PR #53, opened 2015-03-17
Add "--visit-transitive-closure" command line option.

stereotype441:
This option causes the linter to show results for all libraries
reachable from the project via import and export directives.


pq:
For extra credit, adding some tests would be nice.  See here for coverage details:

https://coveralls.io/builds/2136191/source?filename=lib%2Fsrc%2Fanalysis.dart

PR #54, opened 2015-03-17
Don't use "yield" in AnalysisDriver._getAllUnitSources().

stereotype441:
This syntax won't be available until Dart 1.9.


pq:
LGTM!

PR #59, opened 2015-03-18
Fix to trim underlines when errors span lines.

pq:


pq:
@stereotype441 : this one's for you!

Results now look more like this:

```
/Users/me/src/git/clones/linter/lib/src/scratch.dart 10:3 [lint] PREFER to use /// for doc comments
  /**
  ^^^

1 file analyzed, 1 issue found.
-------------------------------
slash_for_doc_comments        1
-------------------------------
```

lib/src/formatter.dart
```diff
@@ -69,9 +69,10 @@ class DetailedReporter extends SimpleFormatter {
     var contents = getLineContents(line, error);
     out.writeln(contents);
 
-    var arrows = '^' * error.length;
     var spaces = column - 1;
-    var result = '${" " * spaces}$arrows';
+    var arrows = min(error.length, contents.length - spaces);
```
stereotype441: Consider changing to:

var arrows = max(1, min(error.length, contents.length - spaces));

just to make sure at least one arrow is always printed.

stereotype441:
LGTM

lib/src/formatter.dart
pq: Excellent.  Great idea!

PR #67, opened 2015-03-23
Exclude getters/setters from one_member_abstracts lint.  Fix #64.

### LINKED ISSUE #64: False positives `Avoid defining a one-member abstract class when a simple function will do
/lib/src/connection/connection.dart 50:16 [lint] Avoid defining a one-member abstract class when a simple function will do.
abstract class Config {
               ^^^^^^
```dart
abstract class Config {
  String get datasetId;
}
```
I think when the class has at least one unimplemented getter/setter/method this rule doesn't apply anymore
I'm not sure what this rule is about exactly but I guess this also only applies when it has only static members.

pq:
@bwilkerson


bwilkerson:
LGTM

PR #70, opened 2015-03-23
Machine output. Fix #69.

### LINKED ISSUE #69: Support machine readable output.
We'll need this to play nice with build integration.  For consistency, we should (minimally) support the same format as the analyzer CLI does.

pq:
@keertip @bwilkerson

### DIFF
--- a/lib/src/formatter.dart
+++ b/lib/src/formatter.dart
@@ -39,25 +39,28 @@ class DetailedReporter extends SimpleFormatter {
   DetailedReporter(
       Iterable<AnalysisErrorInfo> errors, LintFilter filter, IOSink out,
       {int fileCount, String fileRoot, bool showStatistics: false,
-      quiet: false})
+      bool machineOutput: false, quiet: false})
       : super(errors, filter, out,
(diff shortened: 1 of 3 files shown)


lib/src/formatter.dart
```diff
@@ -39,25 +39,28 @@ class DetailedReporter extends SimpleFormatter {
   DetailedReporter(
       Iterable<AnalysisErrorInfo> errors, LintFilter filter, IOSink out,
       {int fileCount, String fileRoot, bool showStatistics: false,
-      quiet: false})
+      bool machineOutput: false, quiet: false})
```
bwilkerson: FWIW, when you get a large number of optional parameters, especially if you need to keep passing them all along (like here), I think it's worth considering defining an "options" object as an alternative. Your style might vary.

bwilkerson:
LGTM

lib/src/formatter.dart
pq: Great advice.  All this param passing sure is ungainly!

PR #73, opened 2015-03-26
API doc tracking; project model WIP.

pq:
Rough but ready for iterating.


pq:
@keertip : baby steps towards a semantic project model.  When this is better fleshed out we can talk about generalizing and possibly pushing down.

PR #79, opened 2015-03-30
Bye by kinds.  Fix #78.

### LINKED ISSUE #78: Consider removing lint kinds (Kind.{DO,DONT,PREFER...})
In the spirit of [Tricorder](http://research.google.com/pubs/pub43322.html)'s approach, I think we should consider removing severities.  I think severities are contentious at best and ultimately subject to override anyway.

EDIT: severities == kinds.

pq:
@keertip 

**TL;DR** removes notion of kinds (DO, DONT, PREFER, etc) in favor of the idea that rules will be either enabled or not.

### DIFF
--- a/lib/src/linter.dart
+++ b/lib/src/linter.dart
@@ -86,22 +86,27 @@ class FileGlobFilter extends LintFilter {
   }
 }
 
-class Group {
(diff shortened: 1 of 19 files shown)


keertip:
lgtm

PR #90, opened 2015-05-10
Error group.

pq:
Catch-all group for lints that catch possible coding errors.

@danschultz @bwilkerson


lib/src/linter.dart
```diff
@@ -108,10 +110,12 @@ class Group implements Comparable<Group> {
 
   factory Group(String name, {String description: '', Hyperlink link}) {
     switch (name.toLowerCase()) {
```
bwilkerson: You might want a Map here rather than a hard-coded list of group names.

bwilkerson:
LGTM

PR #94, opened 2015-05-12
Plugin babysteps.

pq:
@bwilkerson @keertip

### DIFF
--- a/lib/src/plugin/linter_plugin.dart
+++ b/lib/src/plugin/linter_plugin.dart
@@ -26,21 +26,6 @@ final LinterPlugin linterPlugin = new LinterPlugin();
 /// inherently defined by the linter.
 class LinterPlugin implements Plugin {
 
-  /// A subset of rules that we are considering enabled by "default".
-  static final List<LintRule> _rules = [
-    new CamelCaseTypes(),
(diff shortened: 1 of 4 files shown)


keertip:
lgtm!

lib/src/plugin/linter_plugin.dart
```diff
@@ -0,0 +1,78 @@
...
+
+  @override
+  void registerExtensionPoints(RegisterExtensionPoint registerExtensionPoint) {
+    lintRuleExtensionPoint = registerExtensionPoint(
+        LINT_RULE_EXTENSION_POINT, _validateTaskExtension);
+  }
+
+  @override
+  void registerExtensions(RegisterExtension registerExtension) {
+    _rules.forEach((LintRule rule) =>
```
bwilkerson: This is fine, but if you wanted to you could get rid of the '_rules' field by in-lining it here. As it is, you have effectively stored the collection in two places.

bwilkerson:
LGTM

lib/src/plugin/linter_plugin.dart
pq: Good point.  I've done away with `_rules` in favor of inlining.

Thanks!

PR #95, opened 2015-05-12
Const fix.

pq:
@kevmoo

### DIFF
--- a/lib/src/linter.dart
+++ b/lib/src/linter.dart
@@ -101,7 +101,7 @@ class Group implements Comparable<Group> {
           'https://www.dartlang.org/articles/style-guide/'));
 
   /// List of builtin groups in presentation order.
-  static Iterable<Group> get builtin => [errors, style, pub];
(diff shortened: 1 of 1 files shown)


lib/src/linter.dart
```diff
@@ -101,7 +101,7 @@ class Group implements Comparable<Group> {
           'https://www.dartlang.org/articles/style-guide/'));
 
   /// List of builtin groups in presentation order.
-  static Iterable<Group> get builtin => [errors, style, pub];
+  static const Iterable<Group> builtin = const [errors, style, pub];
```
kevmoo: Nice to make this a List – so folks know it's cheap to index into, get length, etc.

pq: Good point.  This is internal though.  Moreover, `List`s can be modified and this should not be.

FWIW: this is a temporary shim anyway.  Odds are good that groups will be defined via an extension point when we have the rest of the plugin story plumbed in.

kevmoo:
LGTM!

PR #97, opened 2015-05-13
migrate tests to use pkg/test

kevmoo:
[Review on Reviewable](https://reviewable.io/reviews/dart-lang/linter/97)


test/project_test.dart
```diff
@@ -9,57 +9,46 @@ library linter.test.project;
...
-import 'package:unittest/unittest.dart';
+import 'package:test/test.dart';
 
-main() {
-  groupSep = ' | ';
-
-  defineTests();
-}
-
-defineTests() {
```
pq: As a general rule, I've adopted the idiom that tests be defined outside of `main`.

test/pub_test.dart
```diff
@@ -149,7 +139,7 @@ dev_dependencies:
 PSDependency findDependency(PSDependencyList deps, {String name}) =>
     deps.firstWhere((dep) => dep.name.text == name, orElse: () => null);
 
-testDepListContains(
+void testDepListContains(
```
pq: *sigh*

I thought this was about updating to `test`?  We've been back and forth about `void` and you know I'm allergic to gratuitous declarations.  Why do you bait me so? :)

test/project_test.dart
kevmoo: For why?

In a world where you want to set up config and aggregate things, it makes sense. In our new world, config should be completely separate, so it seems a tad gratuitous

test/pub_test.dart
kevmoo: if it returns nothing, explicit is gold! The tools love it. Why do you hate our tools so? :stuck_out_tongue_winking_eye:

test/all.dart
```diff
@@ -3,101 +3,22 @@
...
-  config_test.main();
-  project_test.main();
-  integration_test.main();
-}
-
-void useTimingConfig() {
-  unittestConfiguration = new TimingTestConfig();
-}
-
-class TimingTestConfig extends SimpleConfiguration {
```
pq: Why are you removing this?

kevmoo: There is no configuration in the new world.

test/pub_test.dart
pq: meh.  Still not convinced.  For API I'm with you.  For test code that isn't exported what is it buying me?

kevmoo: Clarity for future maintainers.

Since omitting a return annotation implies dynamic, it's super clear that there is *no* expectation that this function could/should return a value.

We spend much more time reading code than writing. Explicit intent FTW! :metal:

test/project_test.dart
pq: Are we in the new world yet?  How is config changed?

Also, you're removing my `groupSep`...

kevmoo: pkg/test == new world
groupSep is also gone.

All that type of visual tweaking will happen in custom runners

test/all.dart
pq: Is there a new way to capture timing info for tests?

pq: Oh and why nuke the `IOSink`?

test/pub_test.dart
pq: This feels a little pedantic.  Do you really think there will be confusion over what method called `testFooBar` returns?

test/all.dart
kevmoo: I think the data is there, it just needs to be surfaced. Likely a bug request for pkg/test

Not sure what you mean by `IOSink`...

test/pub_test.dart
kevmoo: It's 100% our coding style across almost all of our code.

The only time you omit the return type is if you want dynamic implied. Period.

test/project_test.dart
pq: Do we have new runners yet?

kevmoo: Not yet...

test/all.dart
pq: I have a mock IOSink.

``outSink = new MockIOSink();``

That prevents tons of stdout spew in the test results.

test/pub_test.dart
pq: > It's 100% our coding style across almost all of our code.
> The only time you omit the return type is if you want dynamic implied. Period.

Is this in the style guide?

Happy to take the argument to the appropriate place. :)

pq:
Now defunct.

PR #101, opened 2015-06-03
Exit code setting. Fix #98.

### LINKED ISSUE #98: Exit code on lint failures
Should the linter return a non-zero exit code if a source file fails a lint? The linter is always returning 0, which makes it difficult for a script to know if linting fails without having to parse the result.

pq:


pq:
Close to the simplest thing that could possibly work.

@bwilkerson take a look to make sure I'm not missing anything obvious?

Relevant issue #98.

bwilkerson:
LGTM. This is similar to what dartanalyzer does, but it's been long enough since I looked at dartanalyzer that I don't remember whether there are interesting differences.

PR #105, opened 2015-06-16
Analysis options integration.

pq:
This integrates the linter with the analyzer's options mechanism.

### DIFF
--- a/lib/src/config.dart
+++ b/lib/src/config.dart
@@ -4,11 +4,51 @@
 
 library linter.src.config;
 
+import 'package:analyzer/plugin/options.dart';
 import 'package:yaml/yaml.dart';
+
+/// Process the given option [fileContents] and produce a corresponding
+/// [LintConfig].
+LintConfig processAnalysisOptionsFile(String fileContents, {String fileUrl}) {
+  var yaml = loadYamlNode(fileContents, sourceUrl: fileUrl);
+  if (yaml is YamlMap) {
+    return _parseConfig(yaml);
+  }
+  return null;
+}
+
+LintConfig _parseConfig(Map optionsMap) {
+  if (optionsMap != null) {
+    var options = optionsMap['linter'];
+    // Quick check of basic contract.
+    if (options is YamlMap) {
+      return new _LintConfig().._parseYaml(options);
+    }
+  }
+  return null;
+}
+
+/// Processes analysis options files and translates them into [LintConfig]s.
+class AnalysisOptionsProcessor extends OptionsProcessor {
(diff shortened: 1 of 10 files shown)


pq:
Note: there's a bit of refactoring in the works here to cleanup the way that rule defaults are set.  (I'm leaning towards an actual default options file to specify what rules are enabled by default.)  IOW: maybe best to ignore weirdness around defaults.  I'll fix that in a follow-up.

pq:
@bwilkerson PTAL.

lib/src/config.dart
```diff
@@ -4,11 +4,51 @@
...
+  var yaml = loadYamlNode(fileContents, sourceUrl: fileUrl);
+  if (yaml is YamlMap) {
+    return _parseConfig(yaml);
+  }
+  return null;
+}
+
+LintConfig _parseConfig(Map optionsMap) {
+  if (optionsMap != null) {
+    var options = optionsMap['linter'];
```
bwilkerson: We might want to encourage (and use) a convention of fully qualified top-level names, such as 'org.dartlang.linter'.

lib/src/linter.dart
```diff
@@ -154,7 +154,9 @@ class LinterOptions extends DriverOptions {
     }
   }
   void configure(LintConfig config) {
-    enabledLints = ruleRegistry.enabled(config);
+    // TODO(pquitslund): revisit these default-to-on semantics.
+    enabledLints = ruleRegistry.where((LintRule rule) =>
+        !config.ruleConfigs.any((rc) => rc.disables(rule.name)));
```
bwilkerson: This condition was reversed in src/rules.dart. Should it have been changed here as well?

bwilkerson:
LGTM

lib/src/linter.dart
pq: Nope.  That was deliberate.  The current semantics are actually I think **wrong**.  I'll clean that up in a follow-up CL.

lib/src/config.dart
pq: Assuming options processors are associated with plugins and that plugins will be published via pub I was thinking that the pub namespace would solve this for us.  Does that sound right?

bwilkerson: I'm not sure that we should assume that all plugins will be written as packages, or that if they are they will be published on pub. But it does seem reasonable to say that plugins that meet those criteria should use the package name and that others should use a fully qualified name.

bwilkerson: Whatever it is, we should document the convention somewhere.

PR #107, opened 2015-06-16
Move from deprecated to preferred API calls.

pq:


pq:
@bwilkerson PTAL.

bwilkerson:
LGTM

PR #118, opened 2015-08-20
escape regex operator

guillermooo:


pq:
Change looks good to me.  I'm guessing you've signed the CLA already?  Could you confirm and then I'll merge?

Thanks!

guillermooo:
Yep, there's one recorded in my Google profile. At won't point I'll have to sign a new one to get rid of this warning, but now it's more trouble than it's worth for me.

pq:
Cool.  And thanks for the fix!

PR #119, opened 2015-08-21
Update unnecessary brace checking (Fixes #112).

### LINKED ISSUE #112: False positive for "Avoid using braces in interpolation when not needed" rules
Linting fails for the following scenario when ${varname} is required.

    /var/cache/drone/src/github.com/dogma-dart/dogma-data/lib/build.dart 252:11 [lint] Avoid using  braces in interpolation when not needed.
      return '${basename}_convert.dart';
              ^^^^^^^^^^^
    /var/cache/drone/src/github.com/dogma-dart/dogma-data/lib/build.dart 258:24 [lint] Avoid using braces in interpolation when not needed.
      return 'unmodifiable_${basename}_view.dart';
                           ^^^^^^^^^^^

pq:
@keertip PTAL.

Thanks!


keertip:
lgtm!

Issue #113, opened 2015-08-02
Tests on Windows fail due to path differences


donny-dont:
The following tests fail on Windows with a clean pull of linter.

    ERROR: rule | pub | package_names
      Test failed: Caught Unsupported operation: Illegal character in path
  
    FAIL: names | libary_name_prefixes | acme.test.data.a
      Expected: true
        Actual: <false>
  
    FAIL: pubspec | initialization | sourceUrl
      Expected: 'test/_data/p1/_pubspec.yaml'
        Actual: 'test\\_data\\p1\\_pubspec.yaml'
         Which: is different.
      Expected: test/_data/p1/ ...
        Actual: test\\_data\\p ...
                ^
       Differ at offset 4
	   
	FAIL: collecting | files | basic
      Expected: equals [
  	  'test/_data/p1/p1.dart',
 	   'test/_data/p1/_pubspec.yaml',
 	   'test/_data/p1/src/p2.dart'
 	 ] unordered
 	   Actual: MappedListIterable:[
 	   'test/_data/p1\\p1.dart',
 	   'test/_data/p1\\src\\p2.dart',
 	   'test/_data/p1\\_pubspec.yaml'
 	 ]
  	   Which: has no match for 'test/_data/p1/p1.dart' at index 0

The "[lint] Prefix library names with the package name and a dot-separated path." rule also fails on Windows when running the linter.

One potential workaround is using `p.posix._____` in place of `p.____`. As an example this fixes one of the tests.

	String createLibraryNamePrefix(
        {String libraryPath, String projectRoot, String packageName}) {
      var libraryDirectory = p.dirname(libraryPath);
      var path = p.relative(libraryDirectory, from: projectRoot);
      // Drop 'lib/'
      var segments = p.split(path);
      path = p.posix.joinAll((segments[0] == 'lib') ? segments.sublist(1): segments);
      // Replace separators
      path = path.replaceAll('/', '.');
      // Add separator if needed
      if (path.isNotEmpty) {
        path = '.$path';
      }

      return '$packageName$path';
    }

pq:
Thanks for the report.  And sorry for the slow follow-up!

I'll look at getting these tests running on `appveyor` so we have better automated visibility into windows failures.

donny-dont:
No problem. I found in my own stuff using the analyzer that I was better off using the uri for the source and then using toFilePath(windows: false) then using path.posix everywhere.

pq:
[...]

PR #123, opened 2015-08-26
Windows test fixes.

pq:
Baby steps towards #113.

### DIFF
--- a/lib/src/util.dart
+++ b/lib/src/util.dart
@@ -21,16 +21,17 @@ final _pubspec = new RegExp(r'^[_]?pubspec\.yaml$');
 /// current [packageName].
 String createLibraryNamePrefix(
     {String libraryPath, String projectRoot, String packageName}) {
-  var libraryDirectory = p.dirname(libraryPath);
-  var path = p.relative(libraryDirectory, from: projectRoot);
-  // Drop 'lib/'
+  // Use the posix context to canonicalize separators (`\`).
+  var libraryDirectory = p.posix.dirname(libraryPath);
+  var path = p.posix.relative(libraryDirectory, from: projectRoot);
+  // Drop 'lib/'.
   var segments = p.split(path);
   if (segments[0] == 'lib') {
-    path = p.joinAll(segments.sublist(1));
+    path = p.posix.joinAll(segments.sublist(1));
   }
-  // Replace separators
+  // Replace separators.
   path = path.replaceAll('/', '.');
-  // Add separator if needed
+  // Add separator if needed.
   if (path.isNotEmpty) {
     path = '.$path';
   }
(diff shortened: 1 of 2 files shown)


pq:
@keertip PTAL?

keertip:
lgtm

lib/src/util.dart
```diff
@@ -21,16 +21,17 @@ final _pubspec = new RegExp(r'^[_]?pubspec\.yaml$');
 /// current [packageName].
 String createLibraryNamePrefix(
     {String libraryPath, String projectRoot, String packageName}) {
-  var libraryDirectory = p.dirname(libraryPath);
-  var path = p.relative(libraryDirectory, from: projectRoot);
-  // Drop 'lib/'
+  // Use the posix context to canonicalize separators (`\`).
+  var libraryDirectory = p.posix.dirname(libraryPath);
```
donny-dont: I don't think you want to start using p.posix until you do a join.

lib/src/util.dart
```diff
@@ -21,16 +21,17 @@ final _pubspec = new RegExp(r'^[_]?pubspec\.yaml$');
...
+  var path = p.posix.relative(libraryDirectory, from: projectRoot);
+  // Drop 'lib/'.
   var segments = p.split(path);
   if (segments[0] == 'lib') {
-    path = p.joinAll(segments.sublist(1));
+    path = p.posix.joinAll(segments.sublist(1));
   }
-  // Replace separators
+  // Replace separators.
   path = path.replaceAll('/', '.');
```
donny-dont: If it didn't hit the if statement above the paths aren't normalized to posix style.

donny-dont:
Sorry for being late to the party. I did some fixes like this locally and I believe what you want to do is use the os specific path until you do a join. At that point for sure everything is in `/path/to/foo` form.

pq:
Nothing to be sorry about!  Thanks for the catch.  I've not had much time on a windows box lately so this input is invaluable.  Really!

If you feel up for it, maybe through together a Pull Request?  If not, I can throw some changes your way to review but I won't have access to my windows machine until early next week...

Cheers!

donny-dont:
I didn't end up doing a patch because I wasn't sure at the time on how to properly deal with windows/posix differences.

I found that with the analyzer if I used URIs that the problem was simplified as those are all in a posix compatible form. LibraryElement.definingCompilationUnit.source.uri is what I use when I get the library. From there I wrote a wrapper over path that handled things.

Applicable code at
https://github.com/dogma-dart/dogma-codegen/blob/master/lib/path.dart

Now I'm pretty happy with it for my own codegen library and I would probably suggest something similar for the linter. Not sure how that might change things but for me it all fell into place.

Happy to discuss more about it and if you want it done a certain way I can potentially hack on that.

PR #127, opened 2015-08-26
Tracing out to debug windows tests.

pq:
Another tracer bullet.


pq:
Quick merge in the interest of tracing bad paths in windows via appveyor.

PR #130, opened 2015-09-02
Improved pubspec URI handling.

pq:


pq:
@stereotype441 PTAL?

lib/src/linter.dart
```diff
@@ -325,9 +331,11 @@ class SourceLinter implements DartLinter, AnalysisErrorListener {
 
   @override
   Iterable<AnalysisErrorInfo> lintPubspecSource(
-      {String contents, String sourceUrl}) {
+      {String contents, String sourcePath}) {
     var results = <AnalysisErrorInfo>[];
 
+    Uri sourceUrl = sourcePath == null ? null : p.toUri(sourcePath);
```
stereotype441: Nit: maybe call it `sourceUri`?  It's confusing to have its type be `Uri` but to call it a "Url".

stereotype441:
lgtm

lib/src/linter.dart
pq: `url` is consistent with the `yaml` API I eventually call and that was my thinking but I tend to agree!

PR #132, opened 2015-09-09
Fix rule enablement arg setting.

pq:
Corrects a typo: `'enabled:'` => `'enabled'` that breaks enablement via options files.


pq:
@keertip PTAL?

keertip:
lgtm!

PR #219, opened 2016-04-12
Lint rule: Always test type arguments in operator ==(Object other).

alexeieleusis:
LGTMed code review https://codereview.chromium.org/1880163002

### DIFF
--- a/lib/src/rules/equals_test_argument_type.dart
+++ b/lib/src/rules/equals_test_argument_type.dart
@@ -1,0 +1,108 @@
+// Copyright (c) 2016, the Dart project authors.  Please see the AUTHORS file
+
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+library linter.src.rules.equals_test_argument_type;
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:linter/src/linter.dart';
+
+const desc = r'Test type arguments in operator ==(Object other).';
+
+const details = r'''
+**ALWAYS** test type arguments in operator ==(Object other).
+Not doing so might result in Null Pointer Exceptions which will be unexpected
+for consumers of your class.
+
+**GOOD:**
+```
+class Field {
+}
+
+class Good {
+  final Field someField;
+
+  Good(this.someField);
+
+  @override
+  bool operator ==(Object other) {
+    if (identical(this, other)) {
+      return true;
+    }
+    return other is Good &&
+        this.someField == other.someField;
+  }
+
+  @override
+  int get hashCode {
+    return someField.hashCode;
+  }
+}
+```
+
+**BAD:**
+```
+class Field {
+}
+
+class Bad {
+  final Field someField;
+
+  Bad(this.someField);
+
+  @override
+  bool operator ==(Object other) {
+    Bad otherBad = other as Bad; // LINT
+    bool areEqual = otherBad != null && otherBad.someField == someField;
+    return areEqual;
+  }
+
+  @override
+  int get hashCode {
+    return someField.hashCode;
+  }
+}
+```
+''';
+
+class EqualsTestArgumentType extends LintRule {
+  EqualsTestArgumentType()
+      : super(
+            name: 'equals_test_argument_type',
+            description: desc,
+            details: details,
+            group: Group.errors);
+
+  @override
+  AstVisitor getVisitor() => new _Visitor(this);
+}
+
+class _Visitor extends SimpleAstVisitor {
+  final LintRule rule;
+
+  _Visitor(this.rule);
+
+  @override
+  visitAsExpression(AsExpression node) {
+    MethodDeclaration declaration =
+        node.getAncestor((n) => n is MethodDeclaration);
+    if (!_isEqualsOverride(declaration) ||
+        node.expression is! SimpleIdentifier) {
+      return;
+    }
+
+    SimpleIdentifier identifier = node.expression;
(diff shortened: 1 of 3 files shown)


lib/src/rules/equals_test_argument_type.dart
```diff
@@ -0,0 +1,108 @@
...
+  visitAsExpression(AsExpression node) {
+    MethodDeclaration declaration =
+        node.getAncestor((n) => n is MethodDeclaration);
+    if (!_isEqualsOverride(declaration) ||
+        node.expression is! SimpleIdentifier) {
+      return;
+    }
+
+    SimpleIdentifier identifier = node.expression;
+    String parameterName = declaration.parameters.parameterElements.first.name;
```
pq: Method declarations may have `null` parameters in which case I think we want to be safe and guard with `?.`.  Moreover, if `parameterElements` haven't been resolved, `first` may contain a `null`? In any case, I think playing safe wouldn't hurt.

```String parameterName = declaration.parameters?.parameterElements?.first?.name;```

?.yuck. :)

@bwilkerson : thoughts?

File: lib/src/rules/equals_test_argument_type.dart [hunk 2]
```diff
@@ -0,0 +1,108 @@
...
+library linter.src.rules.equals_test_argument_type;
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:linter/src/linter.dart';
+
+const desc = r'"operator ==(Object other)" should test argument type.';
+
+const details = r'''
+**AVOID** "operator ==(Object other)" should test argument type.
```
pq: How about something like 

**ALWAYS** test type arguments in `operator ==(Object other)`.

pq:
Awesome.  Added a few quick questions.  Thanks!

lib/src/rules/equals_test_argument_type.dart
bwilkerson: Good catch. Yes, getters have no parameters, so they return `null`.

lib/src/rules/equals_test_argument_type.dart
alexeieleusis: Done

lib/src/rules/equals_test_argument_type.dart
alexeieleusis: Done, hopefully we never have that kind of error since first we are testing for isOperator, still I didn't realize it was unsafe.

lib/src/rules/equals_test_argument_type.dart
```diff
@@ -0,0 +1,108 @@
...
+library linter.src.rules.equals_test_argument_type;
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:linter/src/linter.dart';
+
+const desc = r'ALWAYS test type arguments in operator ==(Object other).';
+
+const details = r'''
+**AVOID** ALWAYS test type arguments in operator ==(Object other).
```
pq: I think we can drop the ALWAYS, making it:

```
**AVOID** test type arguments in operator ==(Object other).
```

pq:
As for your comment:

> hopefully we never have that kind of error since first we are testing for isOperator, still I didn't realize it was unsafe.

it's a good point.  It's possible that I'm being overly cautious here but it's only based on some painful experiences when the linter has fallen over in the wild when analyzing imperfect sources.

Thanks for all the thoughtful back and forth! :)

alexeieleusis:
Thanks for the review!

Changed rule message.

lib/src/rules/equals_test_argument_type.dart
alexeieleusis: done

lib/src/rules/equals_test_argument_type.dart
```diff
@@ -0,0 +1,108 @@
...
+library linter.src.rules.equals_test_argument_type;
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:linter/src/linter.dart';
+
+const desc = r'test type arguments in operator ==(Object other).';
+
+const details = r'''
+**AVOID** test type arguments in operator ==(Object other).
```
pq: I think you mean

**ALWAYS** test type arguments in operator ==(Object other).

👍

lib/src/rules/equals_test_argument_type.dart
```diff
@@ -0,0 +1,108 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+library linter.src.rules.equals_test_argument_type;
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:linter/src/linter.dart';
+
+const desc = r'test type arguments in operator ==(Object other).';
```
pq: Let's upper-case "test", like so:

```
Test type arguments in operator ==(Object other).
```

(To be consistent with the other lints.)

pq:
Last nits.  I swear!  😄

lib/src/rules/equals_test_argument_type.dart
alexeieleusis: Done

lib/src/rules/equals_test_argument_type.dart
alexeieleusis: apologies, I didn't understand before, thanks for the patience

pq: No worries.  Thanks!  (And sorry if this feels pedantic!)

pq:
LGTM

PR #229, opened 2016-04-26
Conditions should not unconditionally evaluate to "TRUE" or to "FALSE"

alexeieleusis:
I would like to think of this as the first pass implementation (or even first draft) since is quite complex to figure out false things and I am not an expert in either parsing, deducting, logic programming or whatsoever. I am open to discuss the feature and approach and I am aware that might be too complex to be included in the linter. Still I think adds a lot of value since it is very common to break this rule as you can see in the metrics here https://nemo.sonarqube.org/

Ran this rule in the following projects dart-lang/sdk, dart-lang/markdown and dart-lang/reflectable and these were the only occurrences reported, there are some debatable false positives that rely on side effects. SDK at commit 8caa5a6dcb9f35a101516460d6b0abf12affeaf5

```
/Users/.../development/dart/sdk/sdk/lib/_internal/js_runtime/lib/js_mirrors.dart 1967:11 [lint] Conditions should not unconditionally evaluate to "TRUE" or to "FALSE"
/Users/.../development/dart/sdk/sdk/lib/_internal/js_runtime/lib/js_number.dart 249:9 [lint] Conditions should not unconditionally evaluate to "TRUE" or to "FALSE"
/Users/.../development/dart/sdk/sdk/lib/_internal/js_runtime/lib/js_number.dart 288:9 [lint] Conditions should not unconditionally evaluate to "TRUE" or to "FALSE"
/Users/.../development/dart/sdk/sdk/lib/core/duration.dart 247:11 [lint] Conditions should not unconditionally evaluate to "TRUE" or to "FALSE"
/Users/.../development/dart/sdk/sdk/lib/core/uri.dart 2322:15 [lint] Conditions should not unconditionally evaluate to "TRUE" or to "FALSE"
/Users/.../development/dart/sdk/sdk/lib/html/dartium/html_dartium.dart 3108:9 [lint] Conditions should not unconditionally evaluate to "TRUE" or to "FALSE"
/Users/.../development/dart/sdk/sdk/lib/io/secure_socket.dart 868:17 [lint] Conditions should not unconditionally evaluate to "TRUE" or to "FALSE"
```

### DIFF
--- a/lib/src/rules.dart
+++ b/lib/src/rules.dart
@@ -27,10 +27,12 @@ import 'package:linter/src/rules/empty_constructor_bodies.dart';
 import 'package:linter/src/rules/empty_statements.dart';
 import 'package:linter/src/rules/hash_and_equals.dart';
 import 'package:linter/src/rules/implementation_imports.dart';
+import 'package:linter/src/rules/invariant_booleans.dart';
 import 'package:linter/src/rules/iterable_contains_unrelated_type.dart';
 import 'package:linter/src/rules/library_names.dart';
 import 'package:linter/src/rules/library_prefixes.dart';
 import 'package:linter/src/rules/list_remove_unrelated_type.dart';
+import 'package:linter/src/rules/literal_only_boolean_expressions.dart';
 import 'package:linter/src/rules/non_constant_identifier_names.dart';
 import 'package:linter/src/rules/one_member_abstracts.dart';
 import 'package:linter/src/rules/only_throw_errors.dart';
@@ -74,10 +76,12 @@ final Registry ruleRegistry = new Registry()
   ..register(new EmptyStatements())
   ..register(new HashAndEquals())
   ..register(new ImplementationImports())
+  ..register(new InvariantBooleans())
   ..register(new IterableContainsUnrelatedType())
   ..register(new LibraryNames())
   ..register(new LibraryPrefixes())
   ..register(new ListRemoveUnrelatedType())
+  ..register(new LiteralOnlyBooleanExpressions())
   ..register(new NonConstantIdentifierNames())
   ..register(new OneMemberAbstracts())
   ..register(new OnlyThrowErrors())
--- a/lib/src/util/dart_type_utilities.dart
+++ b/lib/src/util/dart_type_utilities.dart
@@ -1,10 +1,10 @@
 // Copyright (c) 2016, the Dart project authors.  Please see the AUTHORS file
-
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
 library linter.src.util.dart_type_utilities;
 
+import 'dart:collection';
 import 'package:analyzer/dart/element/element.dart';
 import 'package:analyzer/dart/element/type.dart';
 import 'package:analyzer/dart/ast/ast.dart';
@@ -66,8 +66,8 @@ class DartTypeUtilities {
 
   /// Builds the list resulting from traversing the node in DFS and does not
   /// include the node itself.
-  static List<AstNode> traverseNodesInDFS(AstNode node) {
-    List<AstNode> nodes = [];
+  static Iterable<AstNode> traverseNodesInDFS(AstNode node) {
+    LinkedHashSet<AstNode> nodes = new LinkedHashSet();
     node.childEntities.where((c) => c is AstNode).forEach((c) {
       nodes.add(c);
       nodes.addAll(traverseNodesInDFS(c));
--- a/lib/src/util/tested_expressions.dart
+++ b/lib/src/util/tested_expressions.dart
@@ -1,0 +1,160 @@
+// Copyright (c) 2016, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+library linter.src.util.tested_expressions;
+
+import 'dart:collection';
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/token.dart';
+import 'package:linter/src/util/boolean_expression_utilities.dart';
+
+typedef void _recurseCallback(Expression expression);
+
+class TestedExpressions {
+  final BinaryExpression testingExpression;
+  final LinkedHashSet<Expression> truths;
+  final LinkedHashSet<Expression> negations;
+  LinkedHashSet<ContradictoryComparisons> _contradictions;
+
+  TestedExpressions(this.testingExpression, this.truths, this.negations);
+
+  LinkedHashSet<ContradictoryComparisons> evaluateInvariant() {
+    if (_contradictions != null) {
+      return _contradictions;
+    }
+
+    _contradictions = _findContradictoryComparisons(
+        new LinkedHashSet.from(
+            [testingExpression.leftOperand, testingExpression.rightOperand]),
+        testingExpression.operator.type);
+
+    if (_contradictions.isEmpty) {
+      HashSet<Expression> set = _extractComparisons(testingExpression)
+        ..addAll(truths)..addAll(negations);
+      // Here and in several places we proceed only for
+      // TokenType.AMPERSAND_AMPERSAND because we then know that all comparisons
+      // must be true.
+      _contradictions.addAll(
+          _findContradictoryComparisons(set, TokenType.AMPERSAND_AMPERSAND));
+    }
+
+    return _contradictions;
+  }
+
+  /// TODO: A truly smart implementation would detect
+  /// (ref.prop && other.otherProp) && (!ref.prop || !other.otherProp)
+  /// assuming properties are pure computations. i.e. dealing with De Morgan's
+  /// laws https://en.wikipedia.org/wiki/De_Morgan%27s_laws
+  LinkedHashSet<ContradictoryComparisons> _findContradictoryComparisons
+      (HashSet<Expression> comparisons,
+      TokenType tokenType) {
+    final Iterable<Expression> binaryExpressions = comparisons
+        .where((e) => e is BinaryExpression).toSet();
+
+    LinkedHashSet<ContradictoryComparisons> contradictions =
+        new LinkedHashSet.identity();
+    binaryExpressions.forEach((Expression ex) {
+      if (contradictions.isNotEmpty) {
+        return;
+      }
+
+      BinaryExpression e = ex as BinaryExpression;
+      final String eLeftOperand = e.leftOperand.toString();
+      final String eRightOperand = e.rightOperand.toString();
+      TokenType eOperatorType = e.operator.type;
+      comparisons
+          .where((c) => c.offset < e.offset && c is BinaryExpression)
+          .forEach((Expression c) {
+        if (contradictions.isNotEmpty) {
+          return;
+        }
+
+        BinaryExpression bc = c;
+        TokenType cOperatorType = negations.contains(c)
+            ? BooleanExpressionUtilities.NEGATIONS[bc.operator.type]
+            : bc.operator.type;
+        final String bcLeftOperand = bc.leftOperand.toString();
+        final String bcRightOperand = bc.rightOperand.toString();
+        final bool sameOperands =
+            eLeftOperand == bcLeftOperand && eRightOperand == bcRightOperand;
+        final bool sameOperandsInverted =
+            eRightOperand == bcLeftOperand && eLeftOperand == bcRightOperand;
+        final bool isNegationOperation =
+            cOperatorType ==
+                BooleanExpressionUtilities.NEGATIONS[eOperatorType] ||
+                BooleanExpressionUtilities.IMPLICATIONS[cOperatorType] ==
+                    BooleanExpressionUtilities.NEGATIONS[eOperatorType];
+        final bool isTrichotomyConjunction =
+            BooleanExpressionUtilities.
+            TRICHOTOMY_OPERATORS.contains(eOperatorType) &&
+                BooleanExpressionUtilities.
+                TRICHOTOMY_OPERATORS.contains(cOperatorType) &&
+                tokenType == TokenType.AMPERSAND_AMPERSAND;
+        if ((isNegationOperation || isTrichotomyConjunction) &&
+            (sameOperands || sameOperandsInverted)) {
+          contradictions.add(new ContradictoryComparisons(c, ex));
+        }
+      });
+    });
+
+    if (contradictions.isEmpty) {
+      binaryExpressions.forEach(_recurseOnChildNodes(contradictions));
+    }
+
+    return contradictions;
+  }
+
+  _recurseCallback _recurseOnChildNodes(
+      LinkedHashSet<ContradictoryComparisons> expressions) =>
+          (Expression e) {
+            BinaryExpression ex = e as BinaryExpression;
+            if (ex.operator.type != TokenType.AMPERSAND_AMPERSAND) {
+              return;
+            }
+
+            LinkedHashSet<ContradictoryComparisons> set =
+            _findContradictoryComparisons(
+                new HashSet.from([ex.leftOperand, ex.rightOperand]),
+                ex.operator.type);
+            expressions.addAll(set);
+          };
+}
+
+class ContradictoryComparisons {
+  final BinaryExpression first;
+  final BinaryExpression second;
+
+  ContradictoryComparisons(this.first, this.second);
+}
+
+HashSet<Expression> _extractComparisons(BinaryExpression node) {
+  final HashSet<Expression> comparisons = new HashSet<Expression>.identity();
+  if (_isComparison(node)) {
+    comparisons.add(node);
+  }
+  if (node.operator.type != TokenType.AMPERSAND_AMPERSAND) {
+    return comparisons;
+  }
+
+  _addNodeComparisons(node.leftOperand, comparisons);
+  _addNodeComparisons(node.rightOperand, comparisons);
+
+  return comparisons;
+}
+
+void _addNodeComparisons(Expression node, HashSet<Expression> comparisons) {
+  if (_isComparison(node)) {
+    comparisons.add(node);
+  } else if (_isBooleanOperation(node)) {
+    comparisons.addAll(_extractComparisons(node));
+  }
+}
+
+bool _isComparison(Expression expression) => expression is BinaryExpression &&
+    BooleanExpressionUtilities.COMPARISONS.contains(expression.operator.type);
+
+bool _isBooleanOperation(Expression expression) =>
+    expression is BinaryExpression &&
+        BooleanExpressionUtilities.BOOLEAN_OPERATIONS.
+        contains(expression.operator.type);
--- a/lib/src/rules/invariant_booleans.dart
+++ b/lib/src/rules/invariant_booleans.dart
@@ -1,0 +1,275 @@
+// Copyright (c) 2016, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+library linter.src.rules.invariant_booleans;
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:linter/src/linter.dart';
+import 'package:linter/src/util/boolean_expression_utilities.dart';
+import 'package:linter/src/util/dart_type_utilities.dart';
+import 'package:linter/src/util/tested_expressions.dart';
+
+const _desc = r'Conditions should not unconditionally evaluate to "TRUE" or to "FALSE"';
+
+const _details = r'''
+
+**DON'T** test for conditions that can be inferred at compile time or test the
+same condition twice.
+Conditional statements using a condition which cannot be anything but FALSE have
+the effect of making blocks of code non-functional. If the condition cannot
+evaluate to anything but TRUE, the conditional statement is completely
+redundant, and makes the code less readable.
+It is quite likely that the code does not match the programmer's intent.
+Either the condition should be removed or it should be updated so that it does
+not always evaluate to TRUE or FALSE and does not perform redundant tests.
+This rule will hint to the test conflicting with the linted one.
+
+**BAD:**
+```
+//foo can't be both equal and not equal to bar in the same expression
+if(foo == bar && something && foo != bar) {...}
+```
+
+**BAD:**
+```
+void compute(int foo) {
+  if (foo == 4) {
+    doSomething();
+    // We know foo is equal to 4 at this point, so the next condition is always false
+    if (foo > 4) {...}
+    ...
+  }
+  ...
+}
+```
+
+**BAD:**
+```
+void compute(boolean foo) {
+  if (foo) {
+    return;
+  }
+  doSomething();
+  // foo is always false here
+  if (foo){...}
+  ...
+}
+```
+
+**GOOD:**
+```
+void nestedOK() {
+  if (foo == bar) {
+    foo = baz;
+    if (foo != bar) {...}
+  }
+}
+```
+
+**GOOD:**
+```
+void nestedOk2() {
+  if (foo == bar) {
+    return;
+  }
+
+  foo = baz;
+  if (foo == bar) {...} // OK
+}
+```
+
+**GOOD:**
+```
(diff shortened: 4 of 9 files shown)


pq:
cc @bwilkerson

bwilkerson:
I looked over the code briefly, but haven't had a chance to look at it in depth. But I'll make a couple of comments.

First, you've stepped on a land-mine :-), by which I mean that this kind of rule is really hard to define without causing lots of discussions / disagreements. We once (foolishly) tried to introduce a much less powerful version of this rule in which we checked for expressions that conformed to the specification of a constant expression, and flagged those cases. The very first case we heard back about was

```
const bool debugging = false;

  if (debugging) {
  }
```

After a few more conversations about why nearly every expression we could think of had some valid use case (and hence shouldn't be flagged), we backed out the rule. Of course, we we foolish enough to try doing this as a hint (we didn't have lints at that point), which at the time couldn't be opted out of, so maybe a lint will have better success.

That said, if you still want to pursue this, I think there's a better approach: write a visitor class that maintains the state you need to track as you traverse the AST structure (such as what you know to be true because you're in the 'then' branch of an 'if' statement). You might also want a separate visitor to visit boolean expressions and return an indication of whether the condition has a known value and what the 'true' and 'false' implications are. You can see an example of this approach in both ExitDetector and ConstantVisitor (used by ConstantEvaluator). It's a little harder to see, but we also do something similar in the resolver when computing the propagated type of variables.

Let me know if you want more detailed feedback at this point in time.

alexeieleusis:
Thanks for the feedback, I'll look in the classes you recommend and come back with questions. In the meantime I am surprised that you recommend writing a visitor class that maintains state, I thought it was a requirement or at least a convention to make the visitor implementors stateless, what I did is build such state with a function and use it to keep track of the things exactly as you mention, the ones I know have to be true or false in order to get to the node under analysis.

The concrete question is, Is there a place (besides the code) where I can read about the design decisions, requirements conventions or similar to write new rules?

mezoni:
I think that it is not so hard to implement the following resolvers:
- Expression evaluators
- Variable assignment trackers

They can be used (at least) for tracking the (re)assignment of the `primitive` types such as the `num`, `bool` and `String`, and of course, the value `null` for all rest of types.
Also they can be useful for the recognition the results of the expressions with type test (`is` and `is!`).
The same approach I used in my `PEG` parser generator.
This is not the same but a similar approach to solving the problem.

mezoni:
Very very simple example of that how it can works and how it can be implemented.
[binary_declarations/lib/src/expression_evaluator/expression_evaluator.dart](https://github.com/mezoni/binary_declarations/blob/master/lib/src/expression_evaluator/expression_evaluator.dart)

Even if it a much more complext then this does not means that is not possible.

bwilkerson:
> I thought it was a requirement or at least a convention to make the visitor implementors stateless ...

It is. I guess I wasn't very clear. What I'm suggesting is that your lint visitor look for top-level functions and methods, then create a separate visitor that can maintain the state as it visits the function body. (You'll have to do some extra work if you want to visit closures in initialization expressions.) The second visitor might even use a third visitor when it gets to the conditions that you're actually testing, but that's mostly for clarity in the code. (You actually can't do what I was suggesting with the current implementation of lint rules because you don't control when the children are visited and hence can't unwind as you leave a node.)

bwilkerson:
The other thing I'll suggest is that you think about whether some of these checks could be smaller and faster when separated out into their own lints. Consider for example the request in https://github.com/dart-lang/sdk/issues/26311, which could be folded into this check, but which also makes a perfectly good stand-alone lint.

Also, having separate lints makes it easier to produce good feedback to the user by being more specific. One of the other lessons I've learned in this area is that at some level of complexity it becomes necessary to explain how you arrived at your conclusion. Seeing code like

    if (x == y) ...

and being told that 'Conditions should not unconditionally evaluate to "TRUE" or to "FALSE" ' doesn't explain why we thought that the condition had a single outcome. Is it because of the types of `x` and `y`? Because we previously tested that `x != y`? Just something else to consider. :-)

alexeieleusis:
Wow! thanks for the valuable feedback, I'll look at all the recommended code and consider splitting this into smaller lints preparing to have this in a better shape :)

pq:
Smaller would be better for sure.  All in all, great stuff!

alexeieleusis:
A comments on the feedback received:

> I think there's a better approach: write a visitor class that maintains the state you need to track as you traverse the AST structure (such as what you know to be true because you're in the 'then' branch of an 'if' statement)

This rule keeps track of that state up to the closest containing function, though in the same visitor, certainly having a more robust engine to evaluate expressions seems to be required.

> Seeing code like `if (x == y) ...` and being told that 'Conditions should not unconditionally evaluate to "TRUE" or to "FALSE" ' doesn't explain why we thought that the condition had a single outcome...

I remember you (@bwilkerson) mentioned in a different issue that making use of an explanatory message would be useful for some other rule, we could use a similar approach here, actually we know what nodes have contradictory values, is just that there is no easy way to communicate back, even linting both nodes would be confusing unless they are on the same line.

> They can be used (at least) for tracking the (re)assignment of the primitive types such as the num, bool and String, and of course, the value null for all rest of types.

The rule keeps track of reassignments and discards such variables like in the test https://github.com/dart-lang/linter/pull/229/commits/317d4556719e3caf9ddd9a6d76cc18c1c525fec6#diff-39221ee11f2f70a30d529b2c35683473R84 though certainly detecting `is` and `is!` should be detected by this kind of rule.

bwilkerson:
> ... making use of an explanatory message ...

I suspect I was referring to the "correction" message, which is intended to explain how to fix the problem, not to provide more information about what's wrong. From the documentation for ErrorCode:

---

Generally, we want to provide messages that consist of three sentences. From the user's perspective these sentences should explain:
1. what is wrong,
2. why is it wrong, and
3. how do I fix it.
However, we combine the first two in the [message] and the last in the [correction].

---

What I'm talking about here is the second part: explaining why it's wrong.

> ... we know what nodes have contradictory values, is just that there is no easy way to communicate back ...

True. What I really want is IDE support for hovering over the highlighted text and having an overlay appear that consists of an arrow from the highlighted text to the earlier expression that contradicts it, with a label on the arrow explaining that this is why we know that this condition has a constant value. Unfortunately, such an interface does not yet exist.

In the meantime, we could potentially craft a message, such as:

This condition will always be false because the condition on line 3 indicates that the opposite condition will be true here.

alexeieleusis:
Finally addressed all comments and worked on performance, currently I detect 3 performance groups in the set of implemented rules, this is in the middle, current implementation is almost an order of magnitude faster than previous implementation.

Ran the rule in the SDK codebase and the only false positives are the ones relying on side effects, which are not in the scope of this rule.

Below some snippets of the addressed comments.

> write a visitor class that maintains the state you need to track as you traverse the AST structure (such as what you know to be true because you're in the 'then' branch of an 'if' statement).

Done

> You might also want a separate visitor to visit boolean expressions and return an indication of whether the condition has a known value and what the 'true' and 'false' implications are.

Done

> They can be used (at least) for tracking the (re)assignment

Done

> What I'm suggesting is that your lint visitor look for top-level functions and methods, then create a separate visitor that can maintain the state as it visits the function body.

Done

> The other thing I'll suggest is that you think about whether some of these checks could be smaller and faster when separated out into their own lints.

Done

> One of the other lessons I've learned in this area is that at some level of complexity it becomes necessary to explain how you arrived at your conclusion.

Done, the rule now prints the conflicting node, which is always before the current node, e.g:

```
void nestedOk4() {
  if (foo <= bar) {
    return;
  }

  if (foo == bar) {} // LINT
}


~/development/dart/linter/test/rules/invariant_boolean.dart 92:7 [lint] Conditions should not unconditionally evaluate to "TRUE" or to "FALSE" verify: foo <= bar.
  if (foo == bar) {} // LINT
      ^^^^^^^^^^
```

lib/src/rules/invariant_boolean.dart
```diff
@@ -0,0 +1,254 @@
...
+  AstVisitor getVisitor() => _visitor;
+}
+
+class _Visitor extends SimpleAstVisitor {
+  final LintRule rule;
+
+  _Visitor(this.rule);
+
+  @override
+  visitBinaryExpression(BinaryExpression node) {
```
bwilkerson: It would be more efficient to only visit binary expressions when they are children of an `if` statement than to test the parent of every binary expression. (That is, don't use a visit method for binary expressions, but invoke a helper method, or maybe just inline the code, when the condition is a binary expression.)

lib/src/rules/invariant_boolean.dart
```diff
@@ -0,0 +1,254 @@
...
+      reportRule.reportLint(e.second);
+    });
+
+    // In dart booleanVariable == true is a valid comparison since the variable
+    // can be null.
+    if (!BooleanExpressionUtilities.EQUALITY_OPERATIONS
+        .contains(node.operator.type) &&
+        (node.leftOperand is BooleanLiteral ||
+            node.rightOperand is BooleanLiteral)) {
+      rule.reportLint(node);
```
bwilkerson: It looks like this rule is actually testing two things: uses of boolean literals and contradictory conditions. Isn't the first covered by the LiteralOnlyBooleanExpression rule? All of the examples in the doc and all of the tests have to do with the second.

lib/src/rules/invariant_boolean.dart
```diff
@@ -0,0 +1,254 @@
...
+  _ContradictionReportRule(ContradictoryComparisons comparisons) : super(
+      name: 'invariant_boolean',
+      description: _desc + ' verify: ${comparisons.first}.',
+      details: _details,
+      group: Group.errors,
+      maturity: Maturity.experimental);
+}
+
+TestedExpressions _findPreviousTestedExpressions(BinaryExpression node) {
+  Block block = node.parent
```
bwilkerson: It doesn't really matter, but I don't think there's any reason to get the `parent` at this point because a binary expression cannot be a direct child of a block.

lib/src/rules/literal_only_boolean_expression.dart
```diff
@@ -0,0 +1,151 @@
...
+  AstVisitor getVisitor() => _visitor;
+}
+
+class _Visitor extends SimpleAstVisitor {
+  final LintRule rule;
+
+  _Visitor(this.rule);
+
+  @override
+  visitBinaryExpression(BinaryExpression node) {
```
bwilkerson: In this case, you already have the helper method written (`_onlyLiterals`), so it would really be more efficient to just invoke it on the conditions in the relevant statements.

lib/src/rules/literal_only_boolean_expression.dart
```diff
@@ -0,0 +1,151 @@
...
+    if (node.bestType.name != 'bool' ||
+        (node.parent is! IfStatement && node.parent is! WhileStatement)) {
+      return;
+    }
+
+    if (_onlyLiterals(node)) {
+      rule.reportLint(node);
+    }
+  }
+
```
bwilkerson: What about do-while statements? What about the condition in a for statement?

lib/src/rules/literal_only_boolean_expression.dart
```diff
@@ -0,0 +1,151 @@
...
+  @override
+  visitIfStatement(IfStatement node) {
+    if (node.condition is BooleanLiteral) {
+      rule.reportLint(node.condition);
+    }
+  }
+}
+
+bool _onlyLiterals(BinaryExpression node) {
```
bwilkerson: This doesn't cover unary expressions, such as "!false", nor parenthesized expressions, such as "(true || false)". There might be other kinds of expressions that could be caught. A simpler approach would be:

```
bool _onlyLiterals(Expression expression) {
  if (expression is Literal) {
    return true;
  } else if (expression is BinaryExpression) {
    return _onlyLiterals(expression.leftOperand) && _onlyLiterals(node.rightOperand);
  } else if (expression is UnaryExpression) {
    return _onlyLiterals(expression.operand);
  } else if (expression is ParenthesizedExpression) {
    return _onlyLiterals(expression.expression) {
  }
  return false;
}
```

lib/src/util/dart_type_utilities.dart
```diff
@@ -61,8 +61,8 @@ class DartTypeUtilities {
 
   /// Builds the list resulting from traversing the node in DFS and does not
   /// include the node itself.
-  static List<AstNode> traverseNodesInDFS(AstNode node) {
-    List<AstNode> nodes = [];
+  static LinkedHashSet<AstNode> traverseNodesInDFS(AstNode node) {
+    LinkedHashSet<AstNode> nodes = new LinkedHashSet();
```
bwilkerson: I don't understand this change. Please add a comment explaining why it's necessary that this be a set rather than a list.

lib/src/rules/invariant_boolean.dart
alexeieleusis: Done.

lib/src/rules/invariant_boolean.dart
alexeieleusis: Removed literals in code.

lib/src/rules/invariant_boolean.dart
alexeieleusis: Fixed.

lib/src/rules/literal_only_boolean_expression.dart
alexeieleusis: Removed check for literals.

lib/src/rules/literal_only_boolean_expression.dart
alexeieleusis: Good catch! added.

lib/src/rules/literal_only_boolean_expression.dart
alexeieleusis: Done.

lib/src/util/dart_type_utilities.dart
alexeieleusis: It was enough for it to be iterable, performance is the reason, I measured something back then and came to the conclusion that the gain was worth the change, the new one is backwards compatible.

lib/src/rules/invariant_boolean.dart
```diff
@@ -0,0 +1,268 @@
...
+    if (!BooleanExpressionUtilities.EQUALITY_OPERATIONS
+        .contains(node.operator.type) &&
+        (node.leftOperand is BooleanLiteral ||
+            node.rightOperand is BooleanLiteral)) {
+      rule.reportLint(node);
+    }
+  }
+
+  @override
+  visitForStatement(ForStatement node) {
```
bwilkerson: We probably also want to visit the condition in a DoStatement.

lib/src/rules/literal_only_boolean_expression.dart
```diff
@@ -0,0 +1,134 @@
...
+  AstVisitor getVisitor() => _visitor;
+}
+
+class _Visitor extends SimpleAstVisitor {
+  final LintRule rule;
+
+  _Visitor(this.rule);
+
+  @override
+  visitWhileStatement(WhileStatement node) {
```
bwilkerson: This should also visit DoStatements.

lib/src/rules/literal_only_boolean_expression.dart
```diff
@@ -0,0 +1,134 @@
...
+
+  @override
+  visitForStatement(ForStatement node) {
+    if (_onlyLiterals(node.condition)) {
+      rule.reportLint(node);
+    }
+  }
+}
+
+bool _onlyLiterals(Expression expression) {
```
bwilkerson: Do we want to special case `??`? For example, `3 ?? m()` might want to return `true`, while `null ?? m()` would return `false`.

lib/src/util/tested_expressions.dart
```diff
@@ -0,0 +1,159 @@
...
+      _contradictions.addAll(
+          _findContradictoryComparisons(set, TokenType.AMPERSAND_AMPERSAND));
+    }
+
+    return _contradictions;
+  }
+
+  /// TODO: A truly smart implementation would detect
+  /// (someComputation() && otherComputation()) &&
+  /// (!someComputation() || !otherComputation())
```
bwilkerson: What if the computations have side-effects that cause them to return different values from one call to the next?

test/rules/invariant_boolean.dart
```diff
@@ -0,0 +1,151 @@
...
+
+void bad() {
+  if ((foo == bar && someComputation()) || (foo != bar && otherComputation())) {} // OK
+  if (foo > bar || foo == bar) {} // OK
+  if (foo > bar && someComputation() || foo == bar) {} // OK
+  if (foo > bar || foo == bar && someComputation()) {} // OK
+  if (foo < bar - baz || foo > bar) {} // OK
+  if (setting == true) {} // OK
+  if (setting != true) {} // OK
+  if (foo == bar || foo != bar) {} // LINT
```
bwilkerson: I notice that all of these tests are based on top-level variables, but I have concerns about false positives when the identifiers reference getters, which can return different values on each invocation. I can't think of a realistic example that would involve the same code patterns, but the spirit of the rule is that the values of the expressions are statically knowable (in some abstract sense) and hence we can reason about them. Getters violate that by being able to return different values on each call. For example, what if `foo` were a getter that returned and incremented a counter on each invocation?

test/rules/invariant_boolean.dart
```diff
@@ -0,0 +1,151 @@
...
+  if (foo == bar || bar != foo) {} // LINT
+  if (foo == bar && foo != bar) {} // LINT
+  if (foo == bar && someComputation() && foo != bar) {} // LINT
```
bwilkerson: Here's another example. What if `someComputation` modifies the value of `foo`? Then the value of `foo` in the first subexpression might not be the same as the value in the second expression.

The fact that arbitrary code is being executed in the condition means that we cannot know anything about the values of any (non-final) variables.

alexeieleusis: Same as above, I would argue that side effects are out of scope. I have seen this kind of rule to warn against residual debugging code, e.g. when forcing a branch of an if statement to reproduce/fix a bug and after fixing it, leaving the invariant condition.

test/rules/invariant_boolean.dart
alexeieleusis: Added getters/fields to the tests. Right now the rule is quite complex, having it deal with side effects is simply too hard.
What about adding a rule that warns about side effects in getters? Is on my todo list now. https://github.com/dart-lang/linter/issues/293

lib/src/util/tested_expressions.dart
alexeieleusis: Fair point, I wanted to reference De Morgan's laws, so I better do it explicitly.

lib/src/rules/literal_only_boolean_expression.dart
alexeieleusis: I really hope this never happens, still added it.

lib/src/rules/literal_only_boolean_expression.dart
alexeieleusis: Done.

lib/src/rules/invariant_boolean.dart
alexeieleusis: Done.

alexeieleusis:
I ran the travis script locally and it passed, also the error is before the tests ran. I don't know how to force it to run again though.

alexeieleusis:
Friendly ping.

lib/src/rules.dart
```diff
@@ -74,10 +76,12 @@ final Registry ruleRegistry = new Registry()
   ..register(new EmptyStatements())
   ..register(new HashAndEquals())
   ..register(new ImplementationImports())
+  ..register(new InvariantBoolean())
```
pq: Nit: in keeping with naming elsewhere, let's rename this to: `InvariantBooleans`.

lib/src/rules/invariant_boolean.dart
```diff
@@ -0,0 +1,275 @@
+// Copyright (c) 2016, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+library linter.src.rules.invariant_boolean;
```
pq: And here: `invariant_booleans` (and the file name).

lib/src/rules/invariant_boolean.dart
```diff
@@ -0,0 +1,275 @@
...
+}
+```
+''';
+
+
+class InvariantBoolean extends LintRule {
+  _Visitor _visitor;
+
+  InvariantBoolean() : super(
+      name: 'invariant_boolean',
```
pq: And here: `invariant_booleans`...

lib/src/rules/literal_only_boolean_expression.dart
```diff
@@ -0,0 +1,148 @@
+// Copyright (c) 2016, the Dart project authors.  Please see the AUTHORS file
+
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+library linter.src.rules.literal_only_boolean_expression;
```
pq: And here: `literal_only_boolean_expressions` (and the file name).

lib/src/rules/literal_only_boolean_expression.dart
```diff
@@ -0,0 +1,148 @@
...
+**BAD:**
+```
+void bad() {
+  if (1 != 0 || 3 < 4 && true) {} // LINT
+}
+```
+
+''';
+
+class LiteralOnlyBooleanExpression extends LintRule {
```
pq: => `LiteralOnlyBooleanExpressions`

test/rules/invariant_boolean.dart
```diff
@@ -0,0 +1,166 @@
+// Copyright (c) 2016, the Dart project authors.  Please see the AUTHORS file
```
pq: And of course=> `invariant_booleans.dart` 👍

test/rules/literal_only_boolean_expression.dart
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) 2016, the Dart project authors.  Please see the AUTHORS file
```
pq: Similarly `literal_only_boolean_expressions.dart`...

pq:
Just a bunch of naming nits from me...  After that, LGTM!

matanlurey:
So this is neat, but just wondering, why isn't this caught by the analyzer's own "unreachable code"-analysis, or what would it take to promote this?

bwilkerson:
@matanlurey We don't want any hints with false positives. Previous experience with this kind of lint suggests that it will have a fair number of false positives. Perhaps we can refine it to the point where it no longer does and then promote it.

matanlurey:
Cool, just checking. SGTM 👍

lib/src/rules.dart
alexeieleusis: done

lib/src/rules.dart
```diff
@@ -74,10 +76,12 @@ final Registry ruleRegistry = new Registry()
...
   ..register(new IterableContainsUnrelatedType())
   ..register(new LibraryNames())
   ..register(new LibraryPrefixes())
   ..register(new ListRemoveUnrelatedType())
+  ..register(new LiteralOnlyBooleanExpression())
```
alexeieleusis: done

lib/src/rules/invariant_boolean.dart
alexeieleusis: done

lib/src/rules/invariant_boolean.dart
alexeieleusis: done

lib/src/rules/literal_only_boolean_expression.dart
alexeieleusis: done

lib/src/rules/literal_only_boolean_expression.dart
alexeieleusis: done

test/rules/invariant_boolean.dart
alexeieleusis: done

test/rules/literal_only_boolean_expression.dart
alexeieleusis: done

pq:
Thanks for all the fixes @alexeieleusis !

alexeieleusis:
Thanks for the review and the patience :)

PR #232, opened 2016-04-30
Equality operator (==) invocation with references of unrelated types

alexeieleusis:
dart-lang/sdk#26311

### DIFF
--- a/lib/src/rules/dont_compare_unrelated_types_for_equality.dart
+++ b/lib/src/rules/dont_compare_unrelated_types_for_equality.dart
@@ -1,0 +1,180 @@
+// Copyright (c) 2016, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+library linter.src.rules.dont_compare_unrelated_types_for_equality;
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/token.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/element.dart';
+import 'package:analyzer/dart/element/type.dart';
+import 'package:linter/src/linter.dart';
+
+const String _desc = r'Equality operator (==) invocation with references of'
+    r' unrelated types.';
+
+const String _details = r'''
+
+**DON'T** Compare references of unrelated types for equality.
+Comparing references of a type where neither is a subtype of the other most
+likely will return false and might not reflect programmer's intent.
+
+**BAD:**
+```
+void someFunction() {
+  var x = '1';
+  if (x == 1) print('someFunction'); // LINT
+}
+```
+
+**BAD:**
+```
+void someFunction1() {
+  String x = '1';
+  if (x == 1) print('someFunction1'); // LINT
+}
+```
+
+**BAD:**
+```
+void someFunction13(DerivedClass2 instance) {
+  var other = new DerivedClass3();
+
+  if (other == instance) print('someFunction13'); // LINT
+}
+
+class ClassBase {}
+
+class DerivedClass1 extends ClassBase {}
+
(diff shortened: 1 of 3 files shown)


lib/src/rules/dont_compare_unrelated_types_for_equality.dart
```diff
@@ -0,0 +1,195 @@
...
+
+  bool _differentDynamicStatic(DartType type, Expression expression) =>
+      type.name == _dynamicClassName && expression is! NullLiteral;
+
+  bool _differentTypes(BinaryExpression node) {
+    final DartType leftStaticType = node.leftOperand.staticType;
+    final DartType rightStaticType = node.rightOperand.staticType;
+    final DartType leftPropagatedType = node.leftOperand.propagatedType;
+    final DartType rightPropagatedType = node.rightOperand.propagatedType;
+    if (leftStaticType.name == _dynamicClassName &&
```
bwilkerson: or leftStaticType.isDynamic

Also, I suspect that you can simplify the logic by using `bestType` rather than either `staticType` or `propagatedType`.

bwilkerson:
Some questions about the semantics of the rule for you to consider.

It seems to me that it's not uncommon to define a class (such as Point) and then define multiple implementations of that class (such as CartesianPoint and PolarPoint) and to define equality to work correctly across all of the subclasses. This rule doesn't support that style. The only approach I can think of that would is to only flag cases where the deepest common ancestor is not Object, but while that would reduce false positives it would likely miss a lot of cases that ought to be flagged. It's a trade-off to think about.

Does the parent of the equality test really have to be an if statement? What about

    if (anInt == aString || itIsTuesday) ...

where the parent will be another binary expression. And what about

    bool mustBeFalse = anInt == aString;

Which seems just as useless. And what about inequality:

    if (anInt != aString) ...

which similarly has a constant value.

alexeieleusis:
Relaxed rule to ignore parent node, so all other mentioned cases are inspected now.

Regarding the classes with a common ancestor, made the rule ignore only the cases where the two classes have the same super class and is not dart.core.Object, hopefully will miss very few cases where it should report while reducing the amount of false positives.

lib/src/rules/dont_compare_unrelated_types_for_equality.dart
alexeieleusis: You suspected the right thing, implementation is much simpler now, thanks!

lib/src/rules/dont_compare_unrelated_types_for_equality.dart
```diff
@@ -0,0 +1,182 @@
...
+    ClassElement leftClass = leftType.element;
+    DartType rightType = node.rightOperand.bestType;
+    ClassElement rightClass = rightType.element;
+    return node.leftOperand is! NullLiteral &&
+        node.rightOperand is! NullLiteral &&
+        !leftType.isMoreSpecificThan(rightType) &&
+        !rightType.isMoreSpecificThan(leftType) &&
+        leftType != rightType &&
+        (leftClass.supertype.name == _objectClassName &&
+            leftClass.library.name == _dartCoreLibraryName ||
```
bwilkerson: Rather than compare by name, you could use `leftClass.supertype.isObject`.

bwilkerson:
lgtm

lib/src/rules/dont_compare_unrelated_types_for_equality.dart
alexeieleusis: done

alexeieleusis:
thank you!

alexeieleusis:
friendly ping, are we merging this? I already have a couple of follow up rules for this, "suggested" in the "Related Types" article mentioned in the feature request :p
@bwilkerson @pq

bwilkerson:
Sorry, I assumed you could merge it in. Let me know if there are other pull requests waiting to be merged.

alexeieleusis:
Thank you!

Issue #234, opened 2016-05-02
Avoid using braces in interpolation false positive


jacehensley-wf:
The `Avoid using braces in interpolation when not needed` reports a lint when it shouldn't. Specifically when the variable starts with a `$`.

``` dart
var $someString = 'Some Value';
var concatenated = 'Value is: ${$someString}'; // Currently lints when it shouldn't.
```

```dart
var $someString = 'Some Value';
var concatenated = 'Value is: $$someString'; // Analyzer Warning
```

alexeieleusis:
https://github.com/dart-lang/linter/pull/235

pq:
Thanks!

Fixed w/ https://github.com/dart-lang/linter/commit/0f6c9caa76d744fc1c4e1544c01ba7819b62a9c2

PR #235, opened 2016-05-03
Don't lint braces on string interpolation if identifier starts with $.

alexeieleusis:
#234

### DIFF
--- a/lib/src/rules/unnecessary_brace_in_string_interp.dart
+++ b/lib/src/rules/unnecessary_brace_in_string_interp.dart
@@ -55,8 +55,10 @@ class Visitor extends SimpleAstVisitor {
     var expressions = node.elements.where((e) => e is InterpolationExpression);
     for (InterpolationExpression expression in expressions) {
       if (expression.expression is SimpleIdentifier) {
+        SimpleIdentifier identifier = expression.expression;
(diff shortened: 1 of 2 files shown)


lib/src/rules/unnecessary_brace_in_string_interp.dart
```diff
@@ -55,8 +55,10 @@ class Visitor extends SimpleAstVisitor {
     var expressions = node.elements.where((e) => e is InterpolationExpression);
     for (InterpolationExpression expression in expressions) {
       if (expression.expression is SimpleIdentifier) {
+        SimpleIdentifier identifier = expression.expression;
         Token bracket = expression.rightBracket;
-        if (bracket != null && !isIdentifierPart(bracket.next)) {
+        if (bracket != null && !isIdentifierPart(bracket.next) &&
+            identifier.name.indexOf(r'$') == -1) {
```
bwilkerson: This doesn't need to be a raw string.

bwilkerson:
lgtm

lib/src/rules/unnecessary_brace_in_string_interp.dart
alexeieleusis: fixed

pq:
LGTM!

Issue #230, opened 2016-04-29
"Do not override fields" for field overridden from `implements SuperClass`


zoechi:
```dart
abstract class BaseLoggingHandler {
  LogRecordTransformer transformer;
  void call(LogRecord logRecord);
}


class LogPrintHandler implements BaseLoggingHandler {
  @override
  LogRecordTransformer transformer;
}
```
complaints about "Do not override fields"

alexeieleusis:
This rule was introduced here https://github.com/dart-lang/linter/pull/217 and requested here https://github.com/dart-lang/sdk/issues/25567, I think this is working as intended, below you can see the docs from the rule. Regardless, do you think the rule is wrong? How would you like it to be changed? It should be noted that the rule will only trigger if it is overriden with a field, if instead you have a getter and setter it won't complain.

http://dart-lang.github.io/linter/lints/overriden_field.html

**DO** Do not override fields.

**BAD:**
```
class Base {
  Object field = 'lorem';

  Object something = 'change';
}

class Bad1 extends Base {
  @override
  final field = 'ipsum'; // LINT
}

class Bad2 extends Base {
  @override
  Object something = 'done'; // LINT
}
```

**GOOD:**
```
class Base {
  Object field = 'lorem';

  Object something = 'change';
}

class Ok extends Base {
  Object newField; // OK

  final Object newFinal = 'ignore'; // OK
}
```

alexeieleusis:
Thinking a bit more about it, I think it is a false positive (which I confirmed) if the implementing class (e.g. LogPrintHandler) has a field from a derived class, which would restrict the values that can be set on instances of such class. Is that what you had in mind?

zoechi:
AFAIR the intention of the rule was that it is inefficient when a subclass overrides a field with a field, because then there are two variables for one value (in superclass and subclass) while only one is actually used.  
That is correct if I extend a class. 
When I **implement** a class I only enforce the interface but don't get any implementation from the interface class, therefore there is still only one field (at least my understanding how `implements` works).

alexeieleusis:
I don't know about that case, but as I said above, when you want to restrict the type in the child class, it makes sense to override, e.g.:

```
abstract class BaseLoggingHandler {
  Base transformer;
}

class LogPrintHandler implements BaseLoggingHandler {
  @override
  Derived transformer; // OK
}
```

I am working on a PR for that case. While in review we can confirm what you mention above.

By the way, thanks for the report!

zoechi:
[...]

PR #236, opened 2016-05-03
False positive fix for overriden_field rule.

alexeieleusis:
When overriding a field, don't lint if the overriden field is inherited through an interface (implements keyword).

dart-lang/linter#230

### DIFF
--- a/lib/src/rules/overriden_field.dart
+++ b/lib/src/rules/overriden_field.dart
@@ -7,7 +7,7 @@ library linter.src.rules.overriden_field;
 import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
 import 'package:analyzer/dart/element/element.dart';
-import 'package:analyzer/src/generated/resolver.dart';
+import 'package:analyzer/dart/element/type.dart';
 import 'package:linter/src/linter.dart';
 
 const desc = r'Do not override fields.';
@@ -50,53 +50,80 @@ class Ok extends Base {
 }
 ```
 
+**GOOD:**
+```
+abstract class BaseLoggingHandler {
+  Base transformer;
+}
+
+class LogPrintHandler implements BaseLoggingHandler {
+  @override
+  Derived transformer; // OK
+}
+```
 ''';
 
 class OverridenField extends LintRule {
+  _Visitor _visitor;
+
   OverridenField()
       : super(
-            name: 'overriden_field',
-            description: desc,
-            details: details,
-            group: Group.style);
+      name: 'overriden_field',
+      description: desc,
+      details: details,
+      group: Group.style,
+      maturity: Maturity.experimental) {
+    _visitor = new _Visitor(this);
+  }
 
   @override
-  AstVisitor getVisitor() => new _Visitor(this);
+  AstVisitor getVisitor() => _visitor;
 }
 
 class _Visitor extends SimpleAstVisitor {
-  InheritanceManager _manager;
-
   final LintRule rule;
-  _Visitor(this.rule);
 
-  @override
-  visitCompilationUnit(CompilationUnit node) {
-    LibraryElement library = node?.element?.library;
-    _manager = library == null ? null : new InheritanceManager(library);
-  }
+  _Visitor(this.rule);
 
   @override
   visitFieldDeclaration(FieldDeclaration node) {
     node.fields.variables.forEach((VariableDeclaration variable) {
-      ExecutableElement member = _getOverriddenMember(variable.element);
-      if (member is PropertyAccessorElement && member.isSynthetic) {
+      PropertyAccessorElement field = _getOverriddenMember(variable.element);
+      if (field != null) {
         rule.reportLint(variable.name);
       }
     });
   }
 
-  ExecutableElement _getOverriddenMember(Element member) {
-    if (member == null || _manager == null) {
-      return null;
-    }
+  PropertyAccessorElement _getOverriddenMember(Element member) {
+    String memberName = member.name;
+    Function isOverriddenMember =
+        (PropertyAccessorElement a) => a.isSynthetic && a.name == memberName;
+    Function containsOverridenMember =
+        (InterfaceType i) => i.accessors.any(isOverriddenMember);
+    ClassElement classElement = member.enclosingElement;
 
-    ClassElement classElement =
-        member.getAncestor((element) => element is ClassElement);
-    if (classElement == null) {
-      return null;
-    }
-
-    return _manager.lookupInheritance(classElement, member.name);
+    Iterable<InterfaceType> interfaces =
+        _findAllSupertypesAndMixins(classElement.type, <InterfaceType>[]);
+    InterfaceType interface =
+        interfaces.firstWhere(containsOverridenMember, orElse: () => null);
+    return interface == null
+        ? null
+        : interface.accessors.firstWhere(isOverriddenMember);
   }
+}
+
+Iterable<InterfaceType> _findAllSupertypesAndMixins(InterfaceType interface, List<InterfaceType> accumulator) {
+  if (interface == null ||
+      interface.isObject ||
+      accumulator.contains(interface)) {
+    return accumulator;
+  }
+
+  accumulator.add(interface);
+  InterfaceType superclass = interface.superclass;
+  Iterable<InterfaceType> interfaces = [superclass]
+    ..addAll(interface.element.mixins)
+    ..addAll(_findAllSupertypesAndMixins(superclass, accumulator));
+  return interfaces.where((i) => i != interface);
 }
--- a/test/rules/overriden_field.dart
+++ b/test/rules/overriden_field.dart
@@ -20,10 +20,97 @@ class Bad2 extends Base {
   Object something = 'done'; // LINT
 }
 
+class Bad3 extends Object with Base {
+  @override
+  Object something = 'done'; // LINT
+}
+
 class Ok extends Base {
   Object newField; // OK
 
   final Object newFinal = 'ignore'; // OK
+}
+
+class OK2 implements Base {
+  @override
+  Object something = 'done'; // OK
+
+  @override
+  Object field;
+}
+
+abstract class OK3 implements Base {
+  @override
+  Object something = 'done'; // OK
+}
+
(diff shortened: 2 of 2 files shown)


bwilkerson:
It would be easier to restrict the search for the overridden member to looking up the superclass chain rather than to search all supertypes and then attempt to weed out items that are not in the superclass chain.

If you don't make that change, then you'll need to fix the way you weed out items that are not in the superclass chain. Consider the following example:

class A {
  int field;
}
class B extends A {
}
class C implements B {
  int field; // Should be OK, but I think the code will currently flag this case.
}

zoechi:
The linked issue is not about the base class being abstract, it's about `extends` vs `implements`. If `extends` is used it overriding the field should be prevented, no matter if the superclass is abstract or not. If `implements` is used overriding should be allowed, no matter if the superclass is abstract or not.

bwilkerson:
Yeah, I was confused by the description as well, but the code is not actually testing for `abstract`, it's trying to remove fields from interfaces.

zoechi:
Ok, didn't look at the code :)

alexeieleusis:
Sorry about the poor description, initially I implemented for abstract classes and then realized my mistake and forgot to change the commit and PR message. Will fix with next patch.

alexeieleusis:
I have fixed the implementation and included mixins, but now I am wondering if that is too much? Thoughs?

zoechi:
From mixins you also inherit the implementation, therefore I think this should be treated like `extends` but lets wait for what @bwilkerson has to say about it.

alexeieleusis:
I think the same, and the difference in implementation (currently) is 1 line of code. I might upload a new patch set and add if needed. Thanks for the feedback!

lib/src/rules/overriden_field.dart
```diff
@@ -79,8 +97,21 @@ class _Visitor extends SimpleAstVisitor {
   @override
   visitFieldDeclaration(FieldDeclaration node) {
     node.fields.variables.forEach((VariableDeclaration variable) {
-      ExecutableElement member = _getOverriddenMember(variable.element);
-      if (member is PropertyAccessorElement && member.isSynthetic) {
+      PropertyAccessorElement field = _getOverriddenMember(variable.element);
+      ClassElement variableClassElement = variable.element.enclosingElement;
```
bwilkerson: Given that most of the time the field should be `null`, the code could be more efficient by enclosing everything from here down in an `if (field != null)` statement. Otherwise the code will create the list of implemented types once for every field.

bwilkerson:
Yes, attempts to override a field defined in a mixin should be flagged.

It would be more efficient to change _getOverriddenMember to only look up the superclass chain. It's a lot less work when looking for an overridden member and it negates the need to compute the list of implemented interfaces. But more importantly, it would produce the correct semantics. Consider:

```
class A {
  int f;
}
class B extends A {}
class C implements A {}
class D extends B implements C {
  int f; // LINT
}
```

I think the code as it stands will fail to lint this case because `A` is a supertype of `C`, and ignores the fact that it's also a superclass of `D`.

lib/src/rules/overriden_field.dart
alexeieleusis: fixed

alexeieleusis:
>It would be more efficient to change _getOverriddenMember to only look up the superclass chain.

Done

Also added test case for provided example.

bwilkerson:
I suspect that this approach doesn't work either. It adds every mixin to the list, then looks up the superclass chain of every thing in the list. But that doesn't match the semantics of mixins. Consider the following

```
class A {
  int f;
}
class B extends A {}
class C extends Object with B {
  int f; // OK
}
```

There should be no lint here. The class `C` has, as it's implementation, all of the members in `C`'s body, all of the members from `B`'s body, and all of the members from `Object`'s body. It effectively _implements_ `A`, but doesn't have any implementation from `A`.

The safest, and most efficient, way to implement this is to walk up the superclass chain, looking at the immediate members of each class and each class' mixins until you reach `Object`. It's much easier to look in all the right places than it is to look everywhere and then try to filter out results from wrong places.

alexeieleusis:
My bad, I always thought that what the mixin was adding to the class was its body and everything it inherited from its supertype, I even added a test here https://github.com/dart-lang/linter/pull/236/commits/ce1e3d58005485c34b8fb930b557cd8aa6a960a9#diff-be264134f6532b540d63f39546c435cdR78 

Thanks for your patience, I will remove that. One more thing, in both examples you provided, I get the following message:

```
[compile time error] The class 'B' cannot be used as a mixin because it extends a class other than Object
class E extends D with B {
                       ^
[compile time error] The class 'B3' cannot be used as a mixin because it extends a class other than Object
class C3 extends Object with B3 {
```

But I think that is going to change soon, does this have any effect on the rule?

alexeieleusis:
A few comments:

* I couldn't get rid of at least iterating the mixins list as per previous comments fields from mixins that are overriden need to be flagged, see last test case.
* I created instances with the proposed examples, removed the field from the more specialized and the field exists, e.g.

```
class A {
  int f;
}
class B extends A {}
class C extends Object with B {
}
```

at runtime when creating an instance of C it has an f field that can be set (and read), what happens in those cases when the field is overriden, also, as I commented before, I get a compile time error when linting (I think it comes from the analyzer, not the linter), but still the code runs and the property works. Very confusing :(

bwilkerson:
> ... in both examples you provided, I get the following message  ... But I think that is going to change soon, does this have any effect on the rule?

Nope. From the perspective of this rule, the semantics of mixins hasn't changed, we've just removed some of the previous restrictions about which classes can be used as mixins, not which members a class gets from a mixin.

> ... at runtime when creating an instance of C it has an f field ...

I don't understand. I thought what you were saying is that if you write code like this

```
main() {
  C c = new C();
  c.f = 3;
  print(c.f);
}
```

that is would run without error and print "3", but when I run it I get exactly what I expected, which is a `NoSuchMethodError` because `C` does not define either a getter named `f` nor a setter named `f=` (to see the exception for the getter, comment out the assignment statement). However, this example shows that analyzer has a bug in that it does not correctly report the absence of these members.

alexeieleusis:
Cannot reproduce, I guess I tried with an example that was actually valid, still confusing:

```
class A {
  int field;
}

class B extends A {}

class C implements B {
  @override
  int field; // OK
}

class D {}

class E extends D with B {
  @override
  int field; // OK

  @override
  String toString() {
    return 'E{field: $field}';
  }
}

main() {
  E e = new E();

  e.field = 5;
  print(e);
}

// ===============

$  dart test/rules/overriden_field.dart 
E{field: 5}
```

It might look like E is overriding field but is not, I ran the linter and also got no error, I guess it is a consequence of the bug you mention in the analyzer. The following throws an error as expected:

```
class A3 {
  int f;
}

class B3 extends A3 {}

class C3 extends D with B3 {
  @override
  String toString() {
    return 'C3{f: $f}';
  }
}

main() {
  C3 c = new C3();
  c.f = 5;
  print(c);
}


// ====================
$ dart test/rules/overriden_field.dart 
Unhandled exception:
Class 'C3' has no instance setter 'f='.

NoSuchMethodError: method not found: 'f='
Receiver: Instance of 'C3'
Arguments: [5]
#0      Object._noSuchMethod (dart:core-patch/object_patch.dart:42)
```

Thanks for your explanation, now I think the rule checks only the cases it should and flags them correctly, finally!

bwilkerson:
Have you uploaded the changes? This looks like the same code I looked at last time, but maybe I'm doing something wrong.

alexeieleusis:
I didn't add a change with my last comment, it was just to confirm that what you said was correct and provide a complete example of what I did and why I thought it was confusing, apologies for the confusion,  this is the last change I uploaded and I think it covers and verifies all the required cases : https://github.com/dart-lang/linter/pull/236/commits/9d6ddc74ac8ec12f7893ee6e243d435a6b70f5cb

I am not certain it is correct, that's why I was pointing to the last test case https://github.com/dart-lang/linter/pull/236/commits/9d6ddc74ac8ec12f7893ee6e243d435a6b70f5cb#diff-be264134f6532b540d63f39546c435cdR106

bwilkerson:
So I think this still isn't sufficient. Try the following test case:

```
class A {
  int field;
}
class B extends Object with A {}
class C extends B {
  int field; // LINT
}
```

If I'm reading the code correctly, it will find the overridden field in `A`, then check to see whether this is a "real" override by looking at the superclasses of `C` (specifically, `B` and `Object`) and any classes mixed in to `C`, but will ignore classes mixed in to `B` (or any other superclass).

The code would be much cleaner if lines 99 - 114 were replaced by

```
FieldElement field = _getOverriddenField(variable.element);
if (field != null) {
  rule.reportLint(variable.name);
}
```

where `_getOverriddenField` doesn't use an `InheritanceManager`, but instead walks up the superclass chain of the variable's class, looking in the current class and all of it's mixins for a non-synthetic field, returning it when it has been found and returning `null` if there is no such field.

alexeieleusis:
I think the latest commit addresses all previous comments and includes tests for all mentioned cases, now given how many cases arose during the review, I am hesitant this implementation is comprehensive at all... should we disable the rule instead? Is there a way to make rules opt in?

bwilkerson:
> I think the latest commit addresses all previous comments and includes tests for all mentioned cases ...

I agree.

> ... now given how many cases arose during the review, I am hesitant this implementation is comprehensive at all ...

Because you've simplified the logic being used, I'm actually quite confident that it's (nearly) correct now.

> ... should we disable the rule instead? Is there a way to make rules opt in?

All lint rules are always opt-in, so there's no need. If there are false positives users can temporarily disable the lint until we can get it fixed.

There's one other change I'd like to see made before this is committed. The method `_findAllSupertypesAndMixins` will work with correct code, but static analysis tools have to be able to work on incorrect code. Consider

```
class A extends B {}
class B extends A {
  int field;
}
```

That's obviously not valid Dart, but is something that we have to be able to deal with. The easiest way to handle this case is to write a method that takes a list as one of its arguments, adds each class and its mixins to the list when it's called, and stops if it ever gets called with a class that's already in the list. (That will also be more efficient because you won't be creating a lot of temporary lists.)

alexeieleusis:
yay! done, thanks :)

bwilkerson:
Thanks for your patience! LGTM

PR #238, opened 2016-05-07
Invocation of Iterable<E>.contains with reference of unrelated type.

alexeieleusis:


lib/src/rules/iterable_contains_unrelated_type.dart
```diff
@@ -0,0 +1,164 @@
...
+    }
+  }
+}
+
+bool _isParameterizedIterableContainsInvocation(MethodInvocation node) =>
+    node.target != null &&
+        node.methodName.name == 'contains' &&
+        _implementsParameterizedIterable(node.target.bestType) &&
+        node.argumentList.arguments.length == 1 &&
+        node.operator.type == TokenType.PERIOD;
```
bwilkerson: Why does the type of the operator matter?

lib/src/rules/iterable_contains_unrelated_type.dart
```diff
@@ -0,0 +1,164 @@
...
+
+bool _implementsParameterizedIterable(DartType type) =>
```
bwilkerson: Does it need to be a class that implements Iterable? What about a direct use of Iterable, such as:

bool takesIterable(Iterable<int> iterable) => iterable.contains('a');

bwilkerson: It would be good to add a test of a class that indirectly implements Iterable:

abstract class A<E> implements Iterable<E> {}
abstract class B<E> extends A<E> {}
bool takesB(B<int> b) => b.contains('a');

alexeieleusis: > `bool takesIterable(Iterable iterable) => iterable.contains('a');`
Nice catch, thanks!

> It would be good to add a test of a class that indirectly implements Iterable:
This one too! Though I expected in that case for Iterable to be in the collection of implemented interfaces, I guess that what happens then is that it contains the interfaces explicitly implemented in the class definition.

lib/src/rules/iterable_contains_unrelated_type.dart
alexeieleusis: It doesn't. I thought it was the way to double check it was a method invocation, but there is also ..

alexeieleusis:
Rebased, ready to continue review. I locally ran the travis script, don't know why it didn't run here as it used to.

alexeieleusis:
friendly ping, travis is green again

lib/src/rules/iterable_contains_unrelated_type.dart
```diff
@@ -0,0 +1,188 @@
...
+import 'package:analyzer/dart/element/type.dart';
+import 'package:linter/src/linter.dart';
+import 'package:linter/src/rules/unrelated_type_equality_checks.dart';
+
+const _desc = r'Invocation of Iterable<E>.contains with references of unrelated'
+    r' types.';
+
+const _details = r'''
+
+**DON'T** Invoke contains on Iterable with an instance of different type than
```
pq: I think some formatting (and a few tweaks) would make this a little easier to read. How about:

**DON'T** Invoke `contains` on an `Iterable` with an instance of a different type than

lib/src/rules/iterable_contains_unrelated_type.dart
```diff
@@ -0,0 +1,188 @@
...
+the parameter type since it will invoke == on its elements and most likely will
```
pq: And here:

the parameter type since it will invoke `==` on its elements and most likely will

pq:
This is AWESOME!

Looks good on my side but I'd love to let @bwilkerson take a look to double check the AST visitor logic.

🍻

bwilkerson:
Note that this won't catch the case where `contain` is being invoked (on `this`) inside a class that implements `Iterable` (see line 132).

I think `_implementsOrExtendsDartCoreIterable` will fail if `Iterable` is implemented indirectly through an interface (that is, I think it will return `false` when asked whether `MyClass` implements `Iterable`):

```
abstract class SomeIterable<E> implements Iterable<E> {}

class MyClass<E> implements SomeIterable<E> {}
```

That's probably not a huge problem in practice, but there should at least be a TODO indicating the issue.

If you had access to a TypeProvider, you could ask the original type whether it `isSubtypeOf` the type `Iterable`. I think there's a convoluted way to get it. When you visit the CompilationUnit, ask it for it's `element`, then ask the `element` for it's `context`. Then you can ask the context for it's `typeProvider`. Ugly, but more correct.

lib/src/rules/iterable_contains_unrelated_type.dart
alexeieleusis: done

lib/src/rules/iterable_contains_unrelated_type.dart
alexeieleusis: done

alexeieleusis:
Found a way without the TypeProvider just in case, also thought that this pattern might be useful in other situations and could be extracted to a utils section, I'll continue thinking on this but in the meantime PTAL.

lib/src/rules/iterable_contains_unrelated_type.dart
```diff
@@ -0,0 +1,198 @@
...
+
+  @override
+  void visitMethodInvocation(MethodInvocation node) {
+    if (!_isParameterizedContainsInvocation(node)) {
+      return;
+    }
+
+    ParameterizedType type = node.target.bestType;
+    Expression argument = node.argumentList.arguments.first;
+    if (unrelatedTypes(argument.bestType, _findIterableTypeArgument(type))) {
```
bwilkerson: The method `_findIterableTypeArgument` can return `null`, but `unrelatedTypes` doesn't test for that case so this will result in null pointer exceptions. There needs to be a test case that catches this case.

lib/src/rules/iterable_contains_unrelated_type.dart
```diff
@@ -0,0 +1,198 @@
...
+      rule.reportLint(node);
+    }
+  }
+}
+
+bool _isParameterizedContainsInvocation(MethodInvocation node) =>
```
bwilkerson: Two problems:
1. The two `is` checks are redundant: every interface type is a parameterized type. I suspect you wanted the `is ParameterizedType` test to tell you whether the class _actually_ has type parameters rather than whether it is a kind of type that _could_ have type parameters, but this test won't do that. Instead, try `node.target.bestType.typeParameters.isNotEmpty`.
2. As noted before, this won't catch uses of contains in classes that implement `Iterable` when the target is implicitly `this`. I think that's fairly minor, but ought to be captured in a TODO.

lib/src/rules/iterable_contains_unrelated_type.dart
alexeieleusis: Added `MyMixedClass extends Object with MyClass` which started throwing the null reference exception, I must say I am impressed by the corner cases you find.

lib/src/rules/iterable_contains_unrelated_type.dart
alexeieleusis: 1. Removed the redundant check and added the suggested in `_findIterableTypeArgument` otherwise the lint in `bool takesB1(B1 b) => b.contains('a'); // OK` was missing.
2. Added `MyClass` `MyIterableMixedClass` `MyDerivedClass` `MyMixedClass` and `MyIterableMixedClass` tetsm hopefully these are enough.

alexeieleusis: Nevermind, I realized the problem with 2. Going back to the editor.

alexeieleusis: Hopefully this is the right test (and implementation) https://github.com/dart-lang/linter/pull/238/commits/d3d290cdddbe786b5c8fc8d7029069d9559d6f3d#diff-73e8bba35c392c1326132397c66682aaR171

bwilkerson:
lgtm

PR #241, opened 2016-05-15
Close instances of dart.core.Sink

alexeieleusis:
First attempt, I am facing a problem I don't know how to solve, which is I can't run the test, I get the following:

```
± |185_close_sinks {2} ✓| → dart test/util/solo_test.dart close_sinks
unittest-suite-wait-for-done
Task failed: GenerateLintsTask for source /home/alexeieleusis/development/dart/linter/test/rules/close_sinks.dart
ERROR: close_sinks
  Test failed: Caught AnalysisException: Cannot compute DART_ERRORS for /home/alexeieleusis/development/dart/linter/test/rules/close_sinks.dart
  Caused by Unexpected exception while performing GenerateLintsTask for source /home/alexeieleusis/development/dart/linter/test/rules/close_sinks.dart
...
```

If I instead run the linter directly, I get the expected lints

```
± |185_close_sinks {2} ✓| → dart ~/development/dart/linter/bin/linter.dart --lints=close_sinks test/rules/close_sinks.dart 
/home/alexeieleusis/development/dart/linter/test/rules/close_sinks.dart 10:10 [lint] Close instances of `dart.core.Sink`.
  IOSink _sinkA;
         ^^^^^^
/home/alexeieleusis/development/dart/linter/test/rules/close_sinks.dart 28:10 [lint] Close instances of `dart.core.Sink`.
  IOSink _sinkF; // LINT
         ^^^^^^

1 file analyzed, 2 issues found, in 2626 ms.
```

dart-lang/linter#185

PR #242, opened 2016-05-15
Cancel subscriptions

alexeieleusis:
In the same spirit that https://github.com/dart-lang/linter/pull/241 perform a check for dart.async.StreamSubscription.

### DIFF
--- a/lib/src/util/leak_detector_visitor.dart
+++ b/lib/src/util/leak_detector_visitor.dart
@@ -1,0 +1,144 @@
+// Copyright (c) 2016, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+library linter.src.util.leak_detector_visitor;
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/type.dart';
+import 'package:linter/src/linter.dart';
+import 'package:linter/src/util/dart_type_utilities.dart';
+
+typedef bool DartTypePredicate(DartType type);
+typedef void _VisitVariableDeclaration(VariableDeclaration node);
+typedef bool _Predicate(AstNode node);
+typedef _Predicate _PredicateBuilder(VariableDeclaration v);
+
+abstract class LeakDetectorVisitor extends SimpleAstVisitor {
+  static List<_PredicateBuilder> _variablePredicateBuilders = [_hasReturn];
+  static List<_PredicateBuilder> _fieldPredicateBuilders =
+    [_hasConstructorFieldInitializers, _hasFieldFormalParameter];
+
+  final LintRule rule;
+
+  LeakDetectorVisitor(this.rule);
+
+  String get methodName;
+
+  DartTypePredicate get predicate;
+
+  @override
+  void visitVariableDeclarationStatement(VariableDeclarationStatement node) {
+    FunctionBody function =
+    node.getAncestor((a) => a is FunctionBody);
+    node.variables.variables.forEach(_buildVariableReporter(methodName,
+        function, _variablePredicateBuilders, predicate, rule));
+  }
+
+  @override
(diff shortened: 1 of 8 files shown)


bwilkerson:
This PR appears to have gotten some extraneous code added to it (there are three rules being added, two of which have already been added).

alexeieleusis:
Apologies, I should have explained better, I uploaded a new patch rebased on master so the PR merged this morning no longer appears as a commit in the PR. The previous commit is #241 because I reused/refactored the code there, and before going all the cycle of review, I wanted to provide visibility on the upcoming refactor just in case it is better to have all in one PR.

bwilkerson:
I like having `LeakDetectorVisitor` abstracted out. I'd probably put it in s separate file (because it seems strange to me to have one rule import another rule), but that's a minor issue.

alexeieleusis:
I was planning a follow up CL to cleanup not only this, but also the functions made available to other rules in recent CLs, I was waiting for you to touch on that and define where should I place all that code.

bwilkerson:
Since you asked... my preference would be to create a 'linter/lib/src/utilities' directory in which this kind of code can live. I'd probably put the visitor class in a separate file. I tend to dislike top-level functions, so I'd create a class like `DartTypeUtilities` in which to define static methods such as `_implementsDartAsyncSubscription` (although I'd shorten the name to something like `isSubscription`).

All the code in 'linter/lib/src/util.dart' could eventually move into that directory as well, but that's a longer term clean-up.

alexeieleusis:
I think similar to you, though I dislike classes with only static methods a bit more than global functions, actually now that I think about it, I don't really dislike global functions without side effects, but I think it is necessary to keep them namespaced, so in the last patch, instead of creating a class, I imported the library with an `as` prefix so the functions are consumed namespaced, that might have the drawback that for the first time reader might look like an instance field. Thoughts?

bwilkerson:
Yeah, there's pro's and cons. I agree that a class with only static methods is a code smell, but the namespace issue is also a factor. Another factor, and probably the one that tips the scale for me, is that by grouping all utility methods related to a single class / hierarchy in one class I (a) know where to put the next utility method I write for that class and (b) can more easily discover the methods that have already been written.

alexeieleusis:
I think most comments are addressed, though given than this PR comes on top of https://github.com/dart-lang/linter/pull/241 I would like to wait until that other one goes through.

lib/src/util/leak_detector_visitor.dart
```diff
@@ -0,0 +1,141 @@
...
+import 'package:linter/src/linter.dart';
+import 'package:linter/src/util/dart_type_utilities.dart';
+
+typedef bool DartTypePredicate(DartType type);
+typedef void _VisitVariableDeclaration(VariableDeclaration node);
+typedef bool _Predicate(AstNode node);
+typedef _Predicate _PredicateBuilder(VariableDeclaration v);
+
+abstract class LeakDetectorVisitor extends SimpleAstVisitor {
+  static _PredicateBuilder _isSinkReturn =
```
bwilkerson: Perhaps generalize the name, given that this is now being used to detect the return of something more than a Sink.

Also, IMO this and the other fields in this class would be more readable if they were static methods rather than static function-valued variables:

```
static _Predicate _isSinkReturn(VariableDeclaration v) => ...
```

bwilkerson:
lgtm

lib/src/util/leak_detector_visitor.dart
alexeieleusis: I thought I Had removed all sink references and made names more general, but I overlooked these last references. Changed now.

Regarding the `_PredicateBuilder`s, certainly they are not doing anything in the class other that being referenced from the lists for fields and variables, still I couldn't figure how to convert them in `_Predicate` since I need a function that will be evaluated in the nodes inside the `container` in the `_buildVariableReporter` function and these functions are different for each `VariableDeclaration`, they could be converted to predicates but at the cost of duplication and I am not sure they would be easier to read, I think they would only be a lower level abstraction but not necessarily more readable.

In spite of all said in the previous paragraph, I moved them out of the class since they are kind of pure functions that happen to be used in the class, but don't really depend on it.

PR #250, opened 2016-05-26
New AvoidImportingForSdk rule

srawlins:
**PROTOTYPE: DO NOT MERGE**

A new AvoidImportingForSdk rule to warn against code like:

```
import 'package:angular2/angular2.dart';

String s = 'just a String';
```

It looks like the analyzer would warn about an UNUSED_IMPORT. However, `angular2.dart` exports `dart:core`, so `String` is seen as being imported by `angular2.dart`. This new rule would catch this problem, helping reduce unnecessary dependencies in large code bases.

This requires a new, unimplemented, `BaseImportsVerifier` class in analyzer, proposed in https://github.com/dart-lang/sdk/issues/26484.


pq:
@srawlins : can we close this PR?

srawlins:
Sure, it's less useful for us since we cleaned up Angular 2 imports/exports.

pq:
Thanks!

PR #262, opened 2016-06-16
Lint that flags usages of implicitly-dynamic values

ochafik:
Designed as a strong-mode-friendly replacement for `always_specify_types` and `always_declare_return_types` lints (complains about implicit dynamics at call sites instead of on declarations).

Accepts explicitly `dynamic` values, and lints against implicit ones used as invocation targets or downcast in argument / assignment positions.

cc/ @leafpetersen I believe you mentioned this kind of checks was on your radar for strong-mode, WDYT?

### DIFF
--- a/lib/src/rules/no_implicit_dynamic.dart
+++ b/lib/src/rules/no_implicit_dynamic.dart
@@ -1,0 +1,202 @@
+// Copyright (c) 2015, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+library linter.src.rules.annotate_types;
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/element.dart';
+import 'package:analyzer/dart/element/type.dart';
+import 'package:linter/src/linter.dart';
+import 'package:linter/src/util.dart';
+
+const desc = 'Implicit use of dynamic.';
+
+const details = '''
+**AVOID** using "implicitly dynamic" values.
+
+Untyped / dynamic invocations may fail or be slower at runtime, but dynamic
+types often creep up unintentionally. Explicitly mark variables or return types
+as `dynamic` (instead of `var`) to express your intent unequivocally.
+
+Note: this works best with the --strong command-line flag and after disabling
+both `always_specify_types` and `always_declare_return_types` lints.
+
+**GOOD:**
+```dart
+String trim(String s) => s.trim();
+
+main() {
+  var s = trim(' a ').toUpperCase();
+
+  dynamic x;
+  x = ... ;
+  x.reallyNotSureThisExists();
+}
+```
+
+**BAD:**
+```dart
+trim(s) => s.trim();
+
+main() {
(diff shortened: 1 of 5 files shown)


alexeieleusis:
I pulled this branch and the intellij dart plugin reports the same warnings than travis.

Also I don't know for sure what the policy is, but I think it would be a good idea to follow more closely https://www.dartlang.org/effective-dart/style/

I know there is a preference to sort methods alphabetically, though some of the code I have contributed does not follow that pattern, but IMO readability would improve if you place the public/inherited api methods before the private ones. I tend to use the java convention where the private methods are placed immediately after they are first used.

alexeieleusis:
I just realized that I might not be following the style guide, so that part might not apply.

leafpetersen:
For strong mode, we are planning to add a `-no-implicit-dynamic` option in the near future.  cc/ @jmesserly @munificent who have both talked about picking this feature up.

pq:
@ochafik : what's the status of this one?  Needless to say, the world has changed a lot since you put this together... 😉   Is there anything worth pursuing?

lib/src/rules/no_implicit_dynamic.dart
```diff
@@ -0,0 +1,202 @@
...
+  @override
+  AstVisitor getVisitor() => new Visitor(this);
+}
+
+// TODO(ochafik): Handle implicit return types of method declarations (vs. overrides).
+class Visitor extends SimpleAstVisitor {
+  final LintRule rule;
+  Visitor(this.rule);
+
+  Element _getBestElement(Expression node) {
```
JPaulsen: Consider to use getCanonicalElementFromIdentifier method from DartTypeUtilities (if applies)

lib/src/rules/no_implicit_dynamic.dart
```diff
@@ -0,0 +1,202 @@
...
+    if (node is SimpleIdentifier) return node.bestElement;
+    if (node is PrefixedIdentifier) return node.bestElement;
+    if (node is PropertyAccess) return node.propertyName.bestElement;
+    return null;
+  }
+
+  bool _isImplicitDynamic(Expression node) {
+    if (node == null) return false;
+    while (node is ParenthesizedExpression) {
+      node = node.expression;
```
JPaulsen: Consider call _isImplicitDynamic(node.expression)

lib/src/rules/no_implicit_dynamic.dart
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    return true;
+  }
+
+  void _checkTarget(Expression target, [token]) {
+    if (_isImplicitDynamic(target)) {
+      // Avoid double taxation (if `x` is dynamic, only lint `x.y.z` once).
+      Expression subTarget;
+      if (target is PropertyAccess) subTarget = target.realTarget;
```
JPaulsen: Perhaps it would be good idea to create a method that return an iterable with all targets, for example when you have something like:

```dart
a.b.foo().c.d.bar()[3].baz();
```

So if you can create a method that could be use in unnecessary_lambdas too, where I had to do a similar check

pq:
@ochafik: this is looking a bit stale (and would need rebasing regardless).  Closing for now; feel to re-open if you want to take another pass.

Thanks!

PR #264, opened 2016-06-22
Only throw instances of dart:core.Error

### LINKED ISSUE #239: Flag throwing non Error/Exception types
```dart
throw 'some string';
```

or `null`, number, etc.

alexeieleusis:
fixes #239

### DIFF
--- a/lib/src/rules/only_throw_error.dart
+++ b/lib/src/rules/only_throw_error.dart
@@ -1,0 +1,84 @@
+// Copyright (c) 2016, the Dart project authors.  Please see the AUTHORS file
+
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+library linter.src.rules.only_throw_error;
+
+import 'dart:collection';
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/type.dart';
+import 'package:linter/src/linter.dart';
+import 'package:linter/src/util/dart_type_utilities.dart';
+
+const _desc = r'Only throw instaces of classes extending either Exception or Error';
+
+const _details = r'''
+
+**DO** throw only instances of classes that extend dart.core.Error or
+dart.core.Exception.
+
+**BAD:**
+```
+void throwString() {
+  throw 'hello world!'; // LINT
+}
+```
+
+**GOOD:**
+```
+void throwArgumentError() {
+  Error error = new ArgumentError('oh!');
+  throw error; // OK
+}
+```
+''';
+
+class OnlyThrowError extends LintRule {
+  _Visitor _visitor;
+
+  OnlyThrowError() : super(
+      name: 'only_throw_error',
+      description: _desc,
+      details: _details,
+      group: Group.style,
+      maturity: Maturity.experimental) {
+    _visitor = new _Visitor(this);
+  }
+
+  @override
+  AstVisitor getVisitor() => _visitor;
+}
+
+class _Visitor extends SimpleAstVisitor {
+  final LintRule rule;
+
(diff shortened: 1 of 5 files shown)


lib/src/rules/only_throw_error.dart
```diff
@@ -0,0 +1,65 @@
...
+
+  @override
+  void visitThrowExpression(ThrowExpression node) {
+    if (node.expression is Literal) {
+      rule.reportLint(node);
+      return;
+    }
+
+    if (!DartTypeUtilities.extendsClass(
+        node.expression.bestType, 'Error', 'dart.core')) {
```
bwilkerson: We need to also allow throwing instances of subclasses of Exception. Then we should try running this rule over some large code bases to see whether there are any false positives.

lib/src/util/dart_type_utilities.dart
```diff
@@ -40,6 +40,12 @@ class DartTypeUtilities {
         element.allSupertypes.any(predicate);
   }
 
+  static bool extendsClass(DartType type, String className, String library) =>
```
bwilkerson: I suspect that we want to allow the `type` to either extend or implement the given class, which means that this has to handle mixins and interfaces as well as superclasses.

test/integration_test.dart
```diff
@@ -238,6 +238,38 @@ defineTests() {
       });
     });
 
+    solo_group('only_throw_error', () {
```
bwilkerson: The ''solo' prefix, both here and on the test, need to be removed.

lib/src/util/dart_type_utilities.dart
alexeieleusis: fixed and included `dynamic`

lib/src/rules/only_throw_error.dart
alexeieleusis: fixed, the only "false" positive I found on the SDK codebase is `throw new DomException.jsInterop`  but when I tried opening the file it didn't exist.

test/integration_test.dart
alexeieleusis: ups, fixed

lib/src/util/dart_type_utilities.dart
```diff
@@ -40,6 +40,12 @@ class DartTypeUtilities {
...
+      type != null && type.name == className &&
+          type.element.library.name == library ||
+          (type is InterfaceType &&
+              extendsClass(type.superclass, className, library));
```
bwilkerson: Admittedly, it's a strange corner case, but should we support using Error and Exception as mixins? For example:

```
class E extends Object with Exception {
  static throw() {
    throw new E();
  }
}
```

lib/src/rules/only_throw_error.dart
```diff
@@ -0,0 +1,78 @@
...
+    }
+  }
+}
+
+const _library = 'dart.core';
+const _errorClassName = 'Error';
+const _exceptionClassName = 'Exception';
+
+bool _isThrowable(DartType type) {
+  return type.isDynamic ||
```
bwilkerson: Perhaps add a comment noting that the performance here could be better if we searched the type hierarchy once and quit once we've found any of a number of classes.

bwilkerson:
Other than minor nits, lgtm.

lib/src/util/dart_type_utilities.dart
alexeieleusis: This was already a handled case, added a test :)

lib/src/rules/only_throw_error.dart
alexeieleusis: Changed implementation, I also thought about this, but I was not sure.

bwilkerson:
lgtm

alexeieleusis:
Thank you!

PR #272, opened 2016-07-11
Linter should catch cases of passing a non-T value to List.remove().

### LINKED ISSUE #271: Linter should catch cases of passing a non-T value to List.remove()
If you have a list of Ts, then there's never a valid reason to be removing an object from the list that isn't of type T, and doing so is an indication of a bug (e.g. you're saying `remove(0)` instead of `removeAt(0)`).

alexeieleusis:
fixes #271


bwilkerson:
lgtm

Hixie:
Thanks!

PR #276, opened 2016-07-13
Do not lint overriding static fields.

### LINKED ISSUE #274: overridden_fields should not warn on static fields
Since these are accessed by prefixing with the class name and it is expected for these values to be different. It might be a design problem, but not a lint IMO.

alexeieleusis:
fixes #274


bwilkerson:
lgtm

PR #290, opened 2016-08-18
Ignore '$' in identifiers; we use in generated code.

matanlurey:
Closes: https://github.com/dart-lang/linter/issues/287

I am not yet _terribly_ familiar with all of the lints, but I think I modified the two lints (and added simple test cases) that we would run into. Both `camel_case_types, `constant_identifier_names`, and `non_constant_identifier_names` should allow the `$` character in between case changes.

i.e., now valid:
- `camel_case_types`: `Foo$Bar`
- `constant_identifier_names` and `non_constant_identifier_names`: `foo$Bar`


matanlurey:
Oops, missed something. I'll fix up and rebase.

bwilkerson:
lgtm

pq:
👍  Awesome!

PR #301, opened 2016-09-02
New rule: Prefer final for variable declaration if reference is not reassigned.

alexeieleusis:
First naive attempt. PTAL.

### DIFF
--- a/lib/src/rules.dart
+++ b/lib/src/rules.dart
@@ -39,6 +39,9 @@ import 'package:linter/src/rules/only_throw_errors.dart';
 import 'package:linter/src/rules/overridden_fields.dart';
 import 'package:linter/src/rules/package_api_docs.dart';
 import 'package:linter/src/rules/package_prefixed_library_names.dart';
+import 'package:linter/src/rules/parameter_assignments.dart';
+import 'package:linter/src/rules/prefer_final_fields.dart';
+import 'package:linter/src/rules/prefer_final_locals.dart';
 import 'package:linter/src/rules/prefer_is_not_empty.dart';
 import 'package:linter/src/rules/pub/package_names.dart';
 import 'package:linter/src/rules/public_member_api_docs.dart';
@@ -88,6 +91,9 @@ final Registry ruleRegistry = new Registry()
   ..register(new OverriddenFields())
   ..register(new PackageApiDocs())
   ..register(new PackagePrefixedLibraryNames())
+  ..register(new ParameterAssignments())
+  ..register(new PreferFinalFields())
+  ..register(new PreferFinalLocals())
   ..register(new PreferIsNotEmpty())
   ..register(new PublicMemberApiDocs())
   ..register(new PubPackageNames())
--- a/lib/src/rules/parameter_assignments.dart
+++ b/lib/src/rules/parameter_assignments.dart
@@ -1,0 +1,87 @@
+// Copyright (c) 2016, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+library linter.src.rules.parameter_assignments;
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:linter/src/linter.dart';
+
+const _desc =
+    r"Don't reassign references to parameters of functions or methods.";
+
+const _details = r'''
+
+**DO NOT** assign new values to parameters of methods or functions.
+
+**BAD:**
+```
+void badFunction(int parameter) { // LINT
+  parameter = 4;
+}
+```
+
+**BAD:**
+```
+class A {
+    void badMethod(int parameter) { // LINT
+    parameter = 4;
+  }
+}
+```
+
+**GOOD:**
+```
+void ok(String parameter) {
+  print(parameter);
+}
+```
+
+**GOOD:**
+```
+class A {
+  void ok(String parameter) {
+    print(parameter);
+  }
+}
+```
+''';
+
+class ParameterAssignments extends LintRule {
+  _Visitor _visitor;
+  ParameterAssignments()
+      : super(
+            name: 'parameter_assignments',
+            description: _desc,
+            details: _details,
+            group: Group.style) {
+    _visitor = new _Visitor(this);
+  }
+
+  @override
+  AstVisitor getVisitor() => _visitor;
+}
+
+class _Visitor extends SimpleAstVisitor {
+  final LintRule rule;
+  _Visitor(this.rule);
+
+  @override
+  void visitFunctionDeclaration(FunctionDeclaration node) {
+    node.functionExpression.parameters.parameters.forEach((e) {
+      if (node.functionExpression.body.isPotentiallyMutatedInScope(e.element)) {
+        rule.reportLint(e);
+      }
+    });
+  }
+
+  @override
+  void visitMethodDeclaration(MethodDeclaration node) {
+    node.parameters.parameters.forEach((e) {
+      if (node.body.isPotentiallyMutatedInScope(e.element)) {
+        rule.reportLint(e);
+      }
+    });
+  }
+}
--- a/lib/src/rules/prefer_final_fields.dart
+++ b/lib/src/rules/prefer_final_fields.dart
@@ -1,0 +1,128 @@
+// Copyright (c) 2016, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+library linter.src.rules.prefer_final_fields;
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/element.dart';
+import 'package:linter/src/linter.dart';
+import 'package:linter/src/util/dart_type_utilities.dart';
+
+const _desc = r'Private field could be final.';
+
+const _details = r'''
+
+**DO** prefer declaring private fields as final if they are not reassigned later
+ in the class.
+
+**BAD:**
+```
+class BadImmutable {
+  var _label = 'hola mundo! BadImmutable'; // LINT
+  var label = 'hola mundo! BadImmutable'; // OK
+}
+```
+
+**BAD:**
+```
+class MultipleMutable {
+  var _label = 'hola mundo! GoodMutable', _offender = 'mumble mumble!'; // LINT
+  var _someOther; // LINT
+
+
+  MultipleMutable() : _someOther = 5;
+
+
+  MultipleMutable(this._someOther);
+
+  void changeLabel() {
+    _label= 'hello world! GoodMutable';
+  }
+}
+```
+
+**GOOD:**
+```
+class GoodImmutable {
+  final label = 'hola mundo! BadImmutable', bla = 5; // OK
+  final _label = 'hola mundo! BadImmutable', _bla = 5; // OK
+}
+```
+
+**GOOD:**
+```
+class GoodMutable {
+  var _label = 'hola mundo! GoodMutable';
+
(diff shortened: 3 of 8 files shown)


zoechi:
Should this include parameters?
I still hope "final by default" makes it into 2.0

bwilkerson:
> Should this include parameters?

I don't know for sure, but my guess is that some users will want to add `final` to local variables without wanting to add it to parameters. I'd recommend having a separate lint for parameters so that users can get the combination they want.

For the same reason we should split this into two lints: one for local variables and one for fields. It's not uncommon to leave a field mutable for use by clients of the class even though there is no code in the class itself that mutates the field. If that style is used heavily in a code base then this lint will produce lots of noise.

lib/src/rules/prefer_final_over_var.dart
```diff
@@ -0,0 +1,121 @@
...
+void badMethod() {
+  var label = 'hola mundo!'; // LINT
+  print(label);
+}
+```
+
+**BAD:**
+```
+class BadImmutable {
+  var label = 'hola mundo!'; // LINT
```
bwilkerson: Should this lint recommend `const` rather than `final` where possible?

lib/src/rules/prefer_final_over_var.dart
```diff
@@ -0,0 +1,121 @@
...
+class GoodMutable {
+  var label = 'hola mundo!';
+
+  void changeLabel() {
+    label = 'hello world!';
+  }
+}
+```
+
+**GOOD:**
```
bwilkerson: This is the same as the example above it.

lib/src/rules/prefer_final_over_var.dart
```diff
@@ -0,0 +1,121 @@
...
+      return;
+    }
+
+    FunctionBody function = node.getAncestor((a) => a is FunctionBody);
+    if (function != null &&
+        !function.isPotentiallyMutatedInScope(node.element)) {
+      rule.reportLint(node);
+    }
+
+    ClassDeclaration classDeclaration =
```
bwilkerson: Note that this method will be invoked for every local variable in every method, and that all of those will be inside a class declaration, which means you'll be traversing the AST for the whole class for every local variable.

You should ignore declarations of fields in this method and handle fields separately (in a separate lint, if I can convince you, but separately none the less). I'd recommend having a visitClassDeclaration method that handles all of the fields in that class. It can then visit all of the assignments in the class and gather a list of the fields that are assigned to in a single pass.

zoechi:
> Should this include parameters?

I also think using `final` for parameters is not very popular while it is for local variables.
For public fields I'd expect most to be false positives.

alexeieleusis:
Agreed on everything, I am realizing now that the rule certainly should not apply for public fields, only private ones. Next patch should have all suggested changes.

lib/src/rules/prefer_final_over_var.dart
alexeieleusis: Still missing the `const` recommendation, coming in a future patch.

lib/src/rules/prefer_final_over_var.dart
alexeieleusis: Fixed.

lib/src/rules/prefer_final_over_var.dart
alexeieleusis: Separated in two rules and added a third one for function and method parameters.

alexeieleusis:
I can't find a way to determine if the initializer expression can be const. Any hints?

bwilkerson:
There probably isn't any. We should eventually add a way and then update the lint, but don't worry about it for now.

alexeieleusis:
Thanks, then the PR is ready to continue review.

test/integration_test.dart
```diff
@@ -165,11 +165,8 @@ defineTests() {
 
       // https://github.com/dart-lang/linter/issues/246
       test('overrides across libraries', () {
-        dartlint.main([
-          'test/_data/overridden_fields',
-          '-c',
-          'test/_data/overridden_fields/lintconfig.yaml'
-        ]);
+        dartlint.main(
+            ['test/_data/overridden_fields', '--rules', 'overridden_fields']);
```
alexeieleusis: Without this change the integration test failed because the new rule was triggering cc @pq

test/rules/dont_assign_param_reference.dart
```diff
@@ -0,0 +1,23 @@
...
+
+class A {
+    void badFunction(int parameter) { // LINT
+    parameter = 4;
+  }
+
+  void ok(String parameter) {
+    print(parameter);
+  }
+}
```
bwilkerson: It would be good to test that it doesn't produce a lint when there is a local variable hiding a parameter:

void ok2(String parameter) {
  if (parameter == null) {
    int parameter = 2;
    parameter = 3;
  }
}

lib/src/rules/prefer_final_over_var_field.dart
```diff
@@ -0,0 +1,120 @@
...
+  AstVisitor getVisitor() => _visitor;
+}
+
+class _Visitor extends SimpleAstVisitor {
+  final LintRule rule;
+  _Visitor(this.rule);
+
+  @override
+  void visitFieldDeclaration(FieldDeclaration node) {
+    if (node.fields.isFinal || node.fields.isConst) {
```
bwilkerson: Extract `node.fields` into a local variable.

lib/src/rules/prefer_final_over_var_field.dart
```diff
@@ -0,0 +1,120 @@
...
+      return;
+    }
+
+    ClassDeclaration classDeclaration =
+        node.getAncestor((a) => a is ClassDeclaration);
+    if (classDeclaration == null) {
+      return;
+    }
+
+    node.fields.childEntities.forEach((/* AstNode | Token */ entity) {
```
bwilkerson: If you use `fields.variables.forEach` you don't need the `is` check on line 98.

lib/src/rules/prefer_final_over_var_field.dart
```diff
@@ -0,0 +1,120 @@
...
+
+      VariableDeclaration variable = entity as VariableDeclaration;
+      if (variable == null || !variable.element.isPrivate) {
+        return;
+      }
+
+      final isMutated = DartTypeUtilities
+          .traverseNodesInDFS(classDeclaration)
+          .any((n) =>
+              n is AssignmentExpression &&
```
bwilkerson: Does this handle operators like `++` and `--`?

lib/src/rules/prefer_final_over_var_field.dart
```diff
@@ -0,0 +1,120 @@
...
+              n.leftHandSide is SimpleIdentifier &&
+              n.leftHandSide.toString() == variable.element.name);
```
bwilkerson: Using names is error prone. Consider

class C {
  int _f = 0;
  void m() {
    String _f;
    _f = '';
  }
}

I believe that this will interpret the assignment to the local variable `f` as being an assignment to the field `f`, resulting in a failure to produce a lint where one is appropriate.

Also, rather than use `any` you could just traverse the whole class and gather the FieldElement's that are assigned to in the class. That way you'd only traverse the class one time rather than traversing half the class N times, where N is the number of private fields. I think that would generally be faster (though it's somewhat questionable).

alexeieleusis: Changed.
Regarding traversing the graph I do `where` to keep only the parent nodes of `PropertyAccessorElement` which at that point should be fast to traverse.

lib/src/rules/prefer_final_over_var_field.dart
alexeieleusis: Now it does.

lib/src/rules/prefer_final_over_var_field.dart
alexeieleusis: Thank you! Changed.

lib/src/rules/prefer_final_over_var_field.dart
alexeieleusis: Done.

test/rules/dont_assign_param_reference.dart
alexeieleusis: Done!

lib/src/rules.dart
```diff
@@ -39,6 +40,8 @@ import 'package:linter/src/rules/only_throw_errors.dart';
 import 'package:linter/src/rules/overridden_fields.dart';
 import 'package:linter/src/rules/package_api_docs.dart';
 import 'package:linter/src/rules/package_prefixed_library_names.dart';
+import 'package:linter/src/rules/prefer_final_over_var_field.dart';
+import 'package:linter/src/rules/prefer_final_over_var_local.dart';
```
pq: Can I propose some shorter names?  How about:

* `prefer_final_fields`
* `prefer_final_locals`

lib/src/rules.dart
```diff
@@ -22,6 +22,7 @@ import 'package:linter/src/rules/close_sinks.dart';
 import 'package:linter/src/rules/comment_references.dart';
 import 'package:linter/src/rules/constant_identifier_names.dart';
 import 'package:linter/src/rules/control_flow_in_finally.dart';
+import 'package:linter/src/rules/dont_assign_param_reference.dart';
```
pq: Wondering how to shorten... What about `parameter_assignments`?

lib/src/rules.dart
alexeieleusis: Done.

lib/src/rules.dart
alexeieleusis: Done.

lib/src/rules/parameter_assignments.dart
```diff
@@ -0,0 +1,87 @@
...
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:linter/src/linter.dart';
+
+const _desc =
+    r"Don't reassign references to parameters of functions or methods.";
+
+const _details = r'''
+
+**DO NOT** Assign new values to parameters of methods or functions.
```
pq: Nit: Assign => assign

lib/src/rules/prefer_final_fields.dart
```diff
@@ -0,0 +1,128 @@
...
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/element.dart';
+import 'package:linter/src/linter.dart';
+import 'package:linter/src/util/dart_type_utilities.dart';
+
+const _desc = r'Private field could be final.';
+
+const _details = r'''
+
+**DO** Prefer declaring private fields as final if they are not reassigned later
```
pq: Nit: Prefer => prefer

lib/src/rules/prefer_final_locals.dart
```diff
@@ -0,0 +1,79 @@
...
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:linter/src/linter.dart';
+
+const _desc =
+    r'Prefer final for variable declaration if reference is not reassigned.';
+
+const _details = r'''
+
+**DO** Prefer declaring variables as final if they are not reassigned later in the code.
```
pq: Nit: Prefer => prefer

pq:
Sorry for all the nits!  Otherwise, LGTM! 👍

lib/src/rules/prefer_final_locals.dart
alexeieleusis: done

lib/src/rules/prefer_final_fields.dart
alexeieleusis: done

lib/src/rules/parameter_assignments.dart
alexeieleusis: done

alexeieleusis:
yay! thanks :)

pq:
Cheers! 🍻

PR #303, opened 2016-09-06
Fix name of test util so it's not run like a test with 'pub run test'

kevmoo:
CC @pq


test/util/solo_test.dart
```diff
@@ -1,18 +0,0 @@
...
-
-import 'package:path/path.dart' as p;
-
-import '../rule_test.dart';
-
-/// Solo test runner.  Handy for debugging.
-main([args]) {
-  var ruleName = args[0];
-  var dir = new Directory(ruleDir).absolute;
-  testRule(ruleName, new File(p.join(dir.path, '$ruleName.dart')), debug: true);
```
pq: How does this bit get invoked in your proposal?

pq:
This is awesome!

I think I'm missing something though...  How do the rule tests get loaded and run?

The trick of the `solo_test` testing jig is how tests are synthesized here:

```
 var ruleName = args[0];
 var dir = new Directory(ruleDir).absolute;
 testRule(ruleName, new File(p.join(dir.path, '$ruleName.dart')), debug: true);
```

where is that happening?

test/util/solo_test.dart
kevmoo: Did you try `pub run test -N xyz` ?

pq: OK, wait.  I think I see now.  Neat.

Am I right that this would stop working if there were another test named `xyz`?

kevmoo: Nope. Both would run.

pq: Neat!

Last reservation: a nice feature of the current `solo_test` jig is that you can easily run it from a debug session in your IDE.  Is there a similar supported flow for `pub run test`?

kevmoo: that I can't help with. 😄 

at a minimum, rename `solo_test.dart` to `test_solo.dart` – it it's not treated like a test when one runs `pub run test`

pq: 👍 

Sounds good.  I'll do that in a follow up (or maybe even `test_debug` which is still more evocative).

Thanks for all the help!

PR #312, opened 2016-09-28
Cover more use cases where '$' may be used

matanlurey:
I missed a couple of use cases with https://github.com/dart-lang/linter/pull/290.

This covers any rule that uses `identifier` or `lowerCamelCase` RegExp checks, including:

- constant_identifier_names
- non_constant_identifier_names

This solves generated code issues in Angular (@ferhatb)

### DIFF
--- a/lib/src/util.dart
+++ b/lib/src/util.dart
@@ -15,9 +15,10 @@ import 'package:analyzer/src/generated/parser.dart' show Parser;
 import 'package:analyzer/src/string_source.dart' show StringSource;
 import 'package:path/path.dart' as p;
 
-final _identifier = new RegExp(r'^([_a-zA-Z]+([_a-zA-Z0-9])*)$');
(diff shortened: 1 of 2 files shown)


alexeieleusis:
IIRC you are doing this to avoid lints on generated code, but we should not be linting that code anyways, Is it because you need to analyze that code and within the same run/configuration you lint it?

matanlurey:
All code, generated or otherwise, has the analyzer run on it.

There is also good prior art for allowing `$` characters, i.e.
https://pub.dartlang.org/packages/charcode

matanlurey:
Ping

matanlurey:
Ping @alexeieleusis @pq

alexeieleusis:
This LGTM, I do think performing the suggested changes can help future maintainers to directly see in the commit message relevant information without needing to go and see issues PRs lists. If you really think it should go as is, is OK to, only a couple of extra clicks will be required to figure out that information.

pq:
LGTM!

alexeieleusis:
Thank you!

PR #313, opened 2016-10-13
New rule for AngularDart2 : Specify template or templateUrl but not both when annotating with @Component.

alexeieleusis:
Sending PR since `git cl upload` stopped working.
@pq @bwilkerson 

previous patches were reviewed at https://codereview.chromium.org/2402943002

### DIFF
--- a/lib/src/rules.dart
+++ b/lib/src/rules.dart
@@ -10,6 +10,7 @@ import 'package:linter/src/config.dart';
 import 'package:linter/src/linter.dart';
 import 'package:linter/src/rules/always_declare_return_types.dart';
 import 'package:linter/src/rules/always_specify_types.dart';
+import 'package:linter/src/rules/angular_template_metadata.dart';
 import 'package:linter/src/rules/annotate_overrides.dart';
 import 'package:linter/src/rules/avoid_as.dart';
 import 'package:linter/src/rules/avoid_empty_else.dart';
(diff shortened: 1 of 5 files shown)


bwilkerson:
I think having angular-specific lints is great, but I don't think the linter package is the right place for this. I have serious concerns about adding a dependency from the linter package to angular2.

Have you talked to the reviewer I added to the previous CL about a different home for this code?

alexeieleusis:
cc @vikerman

pq:
I agree with @bwilkerson that we want to be careful about where these lints go. Adding a dependency on angular to the linter and so by extension onto the Dart SDK which is a show-stopper.

Unfortunately, we really don't (despite lots of effort) have a proper dynamic plugin story for Dart and so things like this are just hard in the general case.  On the bright side, as Brian pointed out, there is a bunch of work going on to solve this problem specifically for angular so there may be hope yet.  @vikerman I expect will have ideas.

vikerman:
Hey - I am resurrecting https://github.com/angular/angular2-dart-analyzer - Updating it for current angular2 library and templates. It would surface errors in the templates in Dart and HTML files, that would be immediately available in the IDE. We are not ready for soft errors/warnings like this particular issue yet and maybe we need a separate Angular 2 linter later - Please file an issue in the other repo and we can track it.

pq:
@alexeieleusis: Can we close this one?

bwilkerson:
This should be part of the angular plugin.

alexeieleusis:
ditto, I will eventually ping the angular team to move this to that plugin.

PR #319, opened 2016-10-19
New rule: cascade_invocations.

alexeieleusis:
Cascade consecutive method invocations on the same reference.

### DIFF
--- a/lib/src/rules/cascade_invocations.dart
+++ b/lib/src/rules/cascade_invocations.dart
@@ -1,0 +1,137 @@
+// Copyright (c) 2016, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+library linter.src.rules.cascade_invocations;
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:linter/src/linter.dart';
+
+const _desc = r'Cascade consecutive method invocations on the same reference.';
+
+const _details = r'''
+
+**DO** Use the cascading style when succesively invoking methods on the same
+ reference.
+
+**BAD:**
+```
+SomeClass someReference = new SomeClass();
+someReference.firstMethod();
+someReference.secondMethod();
+```
+
+**BAD:**
+```
+SomeClass someReference = new SomeClass();
+...
+someReference.firstMethod();
+someReference.aProperty = value;
+someReference.secondMethod();
+```
+
+**GOOD:**
+```
+SomeClass someReference = new SomeClass()
+    ..firstMethod()
+    ..aProperty = value
+    ..secondMethod();
+```
+
+**GOOD:**
+```
+SomeClass someReference = new SomeClass();
+...
+someReference
+    ..firstMethod()
+    ..aProperty = value
+    ..secondMethod();
+```
+
+''';
+
+SimpleIdentifier _findTarget(AstNode node) {
+  Expression expression;
+  if (node is ExpressionStatement) {
+    expression = node.expression;
+  }
+
+  if (expression is MethodInvocation &&
+      expression.realTarget is SimpleIdentifier) {
+    return expression.realTarget;
+  }
+
+  if (expression is PrefixedIdentifier &&
+      expression.prefix is SimpleIdentifier) {
+    return expression.prefix;
+  }
+
+  if (expression is AssignmentExpression &&
+      expression.leftHandSide is PrefixedIdentifier) {
+    return (expression.leftHandSide as PrefixedIdentifier).prefix;
+  }
+
+  return null;
+}
+
+/// Rule to lint consecutive invocations of methods or getters on the same
+/// reference that could be done with the cascade operator.
+class CascadeInvocations extends LintRule {
+  _Visitor _visitor;
+
+  /// Default constructor.
+  CascadeInvocations()
+      : super(
+            name: 'cascade_invocations',
+            description: _desc,
+            details: _details,
+            group: Group.style) {
+    _visitor = new _Visitor(this);
+  }
+
+  @override
+  AstVisitor getVisitor() => _visitor;
+}
+
+class _Visitor extends SimpleAstVisitor {
+  final LintRule rule;
+  _Visitor(this.rule);
+
+  @override
+  void visitExpressionStatement(ExpressionStatement node) {
+    final prefixIdentifier = _findTarget(node);
+    if (prefixIdentifier == null || node.parent is! Block) {
+      return;
+    }
+
+    final Block body = node.parent;
+    final previousNodes = body.statements.takeWhile((n) => n != node);
+    if (previousNodes.isEmpty) {
+      return;
+    }
+
+    final previousNode = previousNodes.last;
+    final SimpleIdentifier previousIdentifier = _findTarget(previousNode);
+    if (previousIdentifier != null &&
+        previousIdentifier.staticElement == prefixIdentifier.staticElement) {
+      rule.reportLint(node);
+    }
+
+    if (previousNode is VariableDeclarationStatement) {
+      _reportIfDeclarationFollowedByMethodInvocation(
+          previousNode, prefixIdentifier, node);
+    }
+  }
+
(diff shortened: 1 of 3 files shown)


alexeieleusis:
ptal @bwilkerson @pq

lib/src/rules/cascade_invocations.dart
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  bool _previousIsMethodInvocation(Iterable previousChilds) =>
+      previousChilds.isNotEmpty &&
+      previousChilds.last is ExpressionStatement &&
+      previousChilds.last.expression is MethodInvocation;
+
+  bool _previousIsSingleVariableDeclaration(
+          Iterable previousChilds, VariableDeclarationStatement variables) =>
+      previousChilds.isNotEmpty &&
```
bwilkerson: I don't understand why we need to test that the collection is not empty. And I don't think that we need to test that `variables` is not null. If we can remove both tests, then this method could be inlined.

lib/src/rules/cascade_invocations.dart
```diff
@@ -0,0 +1,125 @@
...
+  @override
+  AstVisitor getVisitor() => _visitor;
+}
+
+class _Visitor extends SimpleAstVisitor {
+  final LintRule rule;
+  _Visitor(this.rule);
+
+  @override
+  void visitMethodInvocation(MethodInvocation node) {
```
bwilkerson: Consider visiting ExpressionStatement rather than MethodInvocation. That way you won't have to test the parent of every method invocation nested within other expressions.

test/rules/cascade_invocations.dart
```diff
@@ -0,0 +1,27 @@
...
+}
+
+class HasMethodWithNoCascade {
+  List<int> list = [];
+
+  void noCascade() {
+    list.removeWhere((i) => i % 5 == 0);
+    list.clear(); // LINT
+  }
+}
```
bwilkerson: You need to test negative cases, such as:
```
foo.m();
bar.m();
```
or
```
String s = '';
x.m();
```
where it matches the lexical pattern but doesn't refer to the same variable. You should also test that it work correctly when the invocation is nested in a statement or expression:
```
if (x.m() != null) ...
```

lib/src/rules/cascade_invocations.dart
```diff
@@ -0,0 +1,125 @@
...
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:linter/src/linter.dart';
+
+const _desc = r'Cascade consecutive method invocations on the same reference.';
+
+const _details = r'''
+
+**DO** Use the cascading style when succesively invoking methods on the same
+ reference.
```
bwilkerson: Do we care about getters? For example:
```
x.m();
x.g;
x.m();
```

lib/src/rules/cascade_invocations.dart
alexeieleusis: done

lib/src/rules/cascade_invocations.dart
```diff
@@ -0,0 +1,125 @@
...
+  AstVisitor getVisitor() => _visitor;
+}
+
+class _Visitor extends SimpleAstVisitor {
+  final LintRule rule;
+  _Visitor(this.rule);
+
+  @override
+  void visitMethodInvocation(MethodInvocation node) {
+    if (node.parent is! ExpressionStatement ||
```
alexeieleusis: Refactored and applied where relevant.

lib/src/rules/cascade_invocations.dart
```diff
@@ -0,0 +1,125 @@
...
+        node.realTarget is! SimpleIdentifier) {
+      return;
+    }
+
+    final simpleIdentifier = node.realTarget as SimpleIdentifier;
+    Block body = node.parent.parent;
+    Iterable previousChilds =
```
alexeieleusis: Fixed.

lib/src/rules/cascade_invocations.dart
alexeieleusis: IIRC if the collection is empty and we try to access the last element an exception will be thrown. In any case refactored and inlined as suggested.

lib/src/rules/cascade_invocations.dart
```diff
@@ -0,0 +1,125 @@
...
+          Iterable previousChilds, VariableDeclarationStatement variables) =>
+      previousChilds.isNotEmpty &&
+      variables != null &&
+      variables.variables.variables.length == 1;
+
+  void _reportIfConsecutiveInvocationsOnSameTarget(Iterable previousChilds,
+      SimpleIdentifier simpleIdentifier, MethodInvocation node) {
+    ExpressionStatement previousStatement = previousChilds.last;
+    MethodInvocation previousInvocation = previousStatement.expression;
+    if (previousInvocation.realTarget is SimpleIdentifier &&
```
alexeieleusis: Abstracted to variables wherever made sense, hopefully I didn't miss occurrences.

test/rules/cascade_invocations.dart
alexeieleusis: Done

alexeieleusis:
This rule is on the fast side and I think requires no resolution so it could be integrated with tools that have that restriction. cc @matanlurey

lib/src/rules/cascade_invocations.dart
bwilkerson: You recall correctly, but in the only place this method was called we had already accessed the last element, so we knew that it couldn't be empty.

lib/src/rules/cascade_invocations.dart
```diff
@@ -0,0 +1,133 @@
...
+
+  @override
+  void visitExpressionStatement(ExpressionStatement node) {
+    final prefixIdentifier = _findTarget(node);
+    if (prefixIdentifier == null) {
+      return;
+    }
+
+    final Block body = node.getAncestor((n) => n is Block);
+    final previousNodes = body.childEntities.takeWhile((n) => n != node);
```
bwilkerson: If you make the change above, then you can look at the statements in the block rather than the childEntities.

lib/src/rules/cascade_invocations.dart
```diff
@@ -0,0 +1,133 @@
...
+  _Visitor(this.rule);
+
+  @override
+  void visitExpressionStatement(ExpressionStatement node) {
+    final prefixIdentifier = _findTarget(node);
+    if (prefixIdentifier == null) {
+      return;
+    }
+
+    final Block body = node.getAncestor((n) => n is Block);
```
bwilkerson: You don't actually need / want to do this. If the statement isn't the direct child of the block, then all of the rest of this method can be skipped.

lib/src/rules/cascade_invocations.dart
```diff
@@ -0,0 +1,133 @@
...
+    final previousNodes = body.childEntities.takeWhile((n) => n != node);
+    if (previousNodes.isEmpty) {
+      return;
+    }
+
+    final previousNode = previousNodes.last;
+    if (previousNode is! AstNode) {
```
bwilkerson: And if you make the changes above then you don't need to test whether it's an AstNode.

lib/src/rules/cascade_invocations.dart
bwilkerson: Perhaps update the documentation to include getters? Also, do we care about setters:
```
x.m();
x.s = '';
x.m();
```

bwilkerson:
> This rule is on the fast side and I think requires no resolution ...

Unfortunately, it does. If you use staticElement, then it requires resolution.

lib/src/rules/cascade_invocations.dart
alexeieleusis: :D I really disliked this comparison.

lib/src/rules/cascade_invocations.dart
alexeieleusis: Fixed.

lib/src/rules/cascade_invocations.dart
alexeieleusis: Changed for parent.

alexeieleusis:
Friendly ping, I think I addressed open issues.

bwilkerson:
lgtm

Issue #314, opened 2016-10-14
parameter_assignments always fails for optional/named parameters


donny-dont:
I enabled the `parameter_assignments` lint and any code I have that uses optional or named parameters gets called out by it.

Would it be possible for the rule to handle code like the following?

```
void foo({String bar}) {
  bar ??= 'bar';

  ..
}
```

alexeieleusis:
I am not sure I understand the request, so I added the following tests to the code, which are linted where indicated:

```
void badNamedParameter({int parameter}) { // LINT
  parameter = 5;
}

void badNamedParameterNullAware({int parameter}) { // LINT
  parameter ??= 5;
}

void badNamedParameterWithDefault({int parameter: 4}) { // LINT
  parameter = 5;
}
```

You want the second one not to be linted? I guess you have already tried optional parameters with default value (i.e. remove the assignment in the third example). Why is that not good enough for your use case?

donny-dont:
@alexeieleusis sorry if I wasn't clear. Optional or named parameters can only have default values that are `const` so using a default value is really only possible in a limited amount of types.

I've been hitting this lint in a lot of cases but I would like to enable it because its a solid recommendation. Here's a real life example that is hitting it.

```
Uri join(String path, {dynamic base}) {
  // Don't reassign references to parameters of functions or methods.
  base ??= currentPath;

  final basePath = _uriPath(base);
  final baseScheme = _uriScheme(base);
  final toPath = _uriPath(path);
  final joined = p.posix.normalize(p.posix.join(basePath, toPath));

  return new Uri(scheme: baseScheme, path: joined);
}
```

I personally feel that doing the `??=` on optional/named parameters for setting default values should not trigger this lint.

alexeieleusis:
I disagree here :(

`??=` is syntactic sugar for
```
if (base == null) {
    base = currentPath;
}
```

You are indeed assigning a new value to the parameter.

In this case you could either create a new variable `var path = base ?? currentPath` or do the same in every place base is used afterwards. I prefer the former.

Thanks for providing feedback though, I am really glad these rules are being used! Are there any others you would like to see implemented?

donny-dont:
Sorry @alexeieleusis I didn't see this comment.

The `??=` for default parameters is a common enough idiom that I'll probably turn off this particular lint rule if you're against the idea of allowing it for setting a default value. I do agree with the idea of this particular lint as it can cause problems if you aren't careful. Maybe I'll have a change of heart later.

[...]

PR #322, opened 2016-10-31
parameter_assignment : Allow ??= assignation for optional parameters …

alexeieleusis:
…without default value.

 Fixes https://github.com/dart-lang/linter/issues/314

### DIFF
--- a/lib/src/rules/parameter_assignments.dart
+++ b/lib/src/rules/parameter_assignments.dart
@@ -5,8 +5,10 @@
 library linter.src.rules.parameter_assignments;
 
 import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/token.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
 import 'package:linter/src/linter.dart';
+import 'package:linter/src/util/dart_type_utilities.dart';
 
 const _desc =
     r"Don't reassign references to parameters of functions or methods.";
@@ -19,6 +21,20 @@ const _details = r'''
 ```
 void badFunction(int parameter) { // LINT
   parameter = 4;
+}
+```
+
+**BAD:**
+```
+void badFunction(int required, {int optional: 42}) { // LINT
+  optional ??= 8;
+}
+```
+
+**BAD:**
+```
+void badFunctionPositional(int required, [int optional = 42]) { // LINT
+  optional ??= 8;
 }
 ```
 
@@ -40,6 +56,20 @@ void ok(String parameter) {
 
 **GOOD:**
 ```
+void actuallyGood(int required, {int optional}) { // OK
+  optional ??= ...;
+}
+```
+
+**GOOD:**
+```
+void actuallyGoodPositional(int required, [int optional]) { // OK
+  optional ??= ...;
+}
+```
+
+**GOOD:**
+```
 class A {
   void ok(String parameter) {
     print(parameter);
@@ -48,8 +78,31 @@ class A {
 ```
 ''';
 
+bool _isDefaultFormalParameterWithDefaultValue(FormalParameter parameter) =>
+    parameter is DefaultFormalParameter && parameter.defaultValue != null;
+
+bool _isDefaultFormalParameterWithoutDefaultValueReassigned(
+        FormalParameter parameter, AssignmentExpression assignment) =>
+    parameter is DefaultFormalParameter &&
+    parameter.defaultValue == null &&
+    _isFormalParameterReassigned(parameter, assignment);
+
+bool _isFormalParameterReassigned(
+        FormalParameter parameter, AssignmentExpression assignment) =>
+    assignment.leftHandSide is SimpleIdentifier &&
+    (assignment.leftHandSide as SimpleIdentifier).staticElement ==
+        parameter.element;
+
+bool _preOrPostFixExpressionMutation(FormalParameter parameter, AstNode n) {
+  return n is PrefixExpression && n.operand is SimpleIdentifier &&
+          (n.operand as SimpleIdentifier).staticElement == parameter.element ||
+      n is PostfixExpression && n.operand is SimpleIdentifier &&
+          (n.operand as SimpleIdentifier).staticElement == parameter.element;
+}
+
 class ParameterAssignments extends LintRule {
   _Visitor _visitor;
+
   ParameterAssignments()
       : super(
             name: 'parameter_assignments',
@@ -65,6 +118,7 @@ class ParameterAssignments extends LintRule {
 
 class _Visitor extends SimpleAstVisitor {
   final LintRule rule;
+
   _Visitor(this.rule);
 
   @override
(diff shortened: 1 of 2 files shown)


alexeieleusis:
Ran on the SDK, hopefully no NPEs with this change. Also didn't restrict the initialization to be before referencing the parameter in the body since there might be an invocation of other methods or functions that in turn deal with that condition.

lib/src/rules/parameter_assignments.dart
```diff
@@ -88,9 +132,25 @@ class _Visitor extends SimpleAstVisitor {
...
     }
   }
+
+  void _reportIfSimpleParameterOrWithDefaultValue(
+      FormalParameter parameter, AstNode functionOrMethodDeclaration) {
+    final shouldReport = parameter is SimpleFormalParameter ||
+        DartTypeUtilities
+            .traverseNodesInDFS(functionOrMethodDeclaration)
+            .where((n) =>
+                _isDefaultFormalParameterWithDefaultValue(parameter) ||
```
bwilkerson: Can we move at least part of this test outside the enclosing closure? If the parameter isn't a default formal parameter without a default value, then there's no reason to traverse everything in the function body.

test/rules/parameter_assignments.dart
```diff
@@ -45,3 +49,29 @@ void otherBad1(int parameter) { // LINT
...
+  optional = 42;
+}
+
+void onceAgainBad02(int required, {int optional: 42}) { // LINT
+  optional ??= 8;
+}
+
+void onceAgainBad02Positional(int required, [int optional = 42]) { // LINT
+  optional ??= 8;
+}
```
bwilkerson: I think the current implementation will not report a lint for the following:
```
void multipleConditionalAssignments({int optional}) {
  optional ??= 4;
  optional ??= 8;
}
```
Should it?

lib/src/rules/parameter_assignments.dart
alexeieleusis: Nice catch, thank you!

test/rules/parameter_assignments.dart
alexeieleusis: Added.

alexeieleusis:
Will merge later today unless there are any comments.

lib/src/rules/parameter_assignments.dart
```diff
@@ -88,9 +136,33 @@ class _Visitor extends SimpleAstVisitor {
...
       parameterList.parameters.forEach((e) {
         if (node.body.isPotentiallyMutatedInScope(e.element)) {
-          rule.reportLint(e);
+          _reportIfSimpleParameterOrWithDefaultValue(e, node);
         }
       });
     }
   }
+
+  void _reportIfSimpleParameterOrWithDefaultValue(
```
bwilkerson: This could be made more efficient by
(a) testing for simple parameters and default parameters with values before traversing any of the nodes, and
(b) using a for loop to traverse all of the nodes a single time and generating errors as you find them rather than creating a new list.

lib/src/rules/parameter_assignments.dart
```diff
@@ -88,9 +136,33 @@ class _Visitor extends SimpleAstVisitor {
...
+        _isDefaultFormalParameterWithoutDefaultValueIncorrectlyMutated(
+            parameter, n));
+
+    final shouldReport = parameter is SimpleFormalParameter ||
+        _isDefaultFormalParameterWithDefaultValue(parameter) ||
+        nonNullCoalescingAssignments.isNotEmpty ||
+        assignmentsNodes.length > 1;
+
+    if (shouldReport) {
+      rule.reportLint(parameter);
```
bwilkerson: The lint should be reported at the site of the assignment, not on the parameter. Otherwise you're asking the user to try to figure out where the assignment is occurring when you could have told them.

lib/src/rules/parameter_assignments.dart
alexeieleusis: I tested adding print statements to verify what was being invoked and I found that (with new changes):

* `_isDefaultFormalParameterWithoutDefaultValueMutated` in invoked only if `parameter is SimpleFormalParameter ||
        _isDefaultFormalParameterWithDefaultValue(parameter)` is `false`.
* `nonNullCoalescingAssignments` is traversed only if `assignmentsNodes.length <= 1` and since I am invoking toList now, `_isDefaultFormalParameterWithoutDefaultValueMutated` is not invoked again.

And IMO is easier to read :)

bwilkerson: You don't need to traverse the list even once if the parameter is a required parameter or has a default value, and I suspect that those two cases cover a large majority of the parameters in real code. Please at least make that optimization.

test/rules/parameter_assignments.dart
```diff
@@ -45,3 +49,34 @@ void otherBad1(int parameter) { // LINT
   parameter += 3;
```
bwilkerson: Shouldn't the lint be getting reported on line 49 (rather than 48)? Perhaps the location of the "// LINT" comment isn't important, though. (Similarly for some of the other tests.

alexeieleusis: This is because we are directly reporting required parameters before traversing the nodes mutating parameters. We just rely on the fact that isPotentiallyMutatedInScope methods are returning true.

lib/src/rules/parameter_assignments.dart
alexeieleusis: Done

test/rules/parameter_assignments.dart
bwilkerson: Sorry, I guess I wasn't clear. I mean, didn't we decide to report lint violations at the point of the assignment rather than at the point at which the parameter is declared? In which case the lint should be produced for the reference to `parameter` on line 49 rather than at the declaration on line 48.

alexeieleusis: I thought we decided to lint at the mutating node when we know it, of course we could proceed to find it but that will make rule more complex, specially for expressions like `variable++`. I would say that if finding the offending node is hard, is a signal that the method/function is doing too much, but that's how I think about my code, I guess it might be a bit harsh expectation for all users. What do you think?

bwilkerson: I think you know what I think :-)

The principle I would apply here is that errors should be reported as close as possible to the location where the problem occurs. This is a corollary to the more general principle that tools should do work so that people don't have to.

The problem isn't at the point where the parameter is declared, it's at the point where the value of the parameter is changed. Hence, I think we should change the location of the lint error.

alexeieleusis: Will do it then, thank you :)

alexeieleusis:
Lint notifications changed to mutating nodes. Ran on SDK and Flutter code bases with no errors reported.

Apologies for the delay, I've been quite busy, hopefully this is fixed now :)

alexeieleusis:
Friendly ping :)

PR #323, opened 2016-11-01
Add a Dart 1.20 strong-mode fix

srawlins:


lib/src/formatter.dart
```diff
@@ -169,7 +169,7 @@ class SimpleFormatter implements ReportFormatter {
     var codes = stats.keys.toList()..sort();
     var largestCountGuess = 8;
     var longest =
-        codes.fold(0, (int prev, element) => max(prev, element.length));
+        codes.fold(0, (int prev, element) => max/*<num>*/(prev, element.length));
```
alexeieleusis: Why not int?

srawlins: I think it is because `element.length`'s type is unknown.

bwilkerson: That would be because we don't know that `element` is a String, which means we lost type information somewhere. We should identify where and see whether we can do a better job of preserving the information.

srawlins:
sorry, it must have gotten smarter during 1.20 somewhere.

PR #329, opened 2016-11-13
New rule: recursive_getter.

alexeieleusis:
Flag getters recursively returning themselves.

Ran the rule with no problems on SDK and Flutter codebases.

Fixes https://github.com/dart-lang/sdk/issues/27704

### DIFF
--- a/lib/src/rules/recursive_getter.dart
+++ b/lib/src/rules/recursive_getter.dart
@@ -1,0 +1,89 @@
+// Copyright (c) 2016, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+library linter.src.rules.recursive_getter;
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/element.dart';
+import 'package:linter/src/linter.dart';
+import 'package:linter/src/util/dart_type_utilities.dart';
+
+const _desc = r'Property getter recursivlely returns itself.';
+
+const _details = r'''
+
+**DON'T** Return the property itself in a getter body, this can be due to a typo.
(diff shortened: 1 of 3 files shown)


lib/src/rules/recursive_getter.dart
```diff
@@ -0,0 +1,97 @@
...
+      return;
+    }
+
+    SimpleIdentifier expression = node.expression;
+    if (methodDeclaration.element == expression.staticElement) {
+      rule.reportLint(expression);
+    }
+  }
+
+  void _verifyReturnStatement(ReturnStatement node, MethodDeclaration method) {
```
alexeieleusis: done, nice catch thanks!

alexeieleusis:
Extended to top level getters. I still need to perform the extension to ternary expressions, will upload a new commit.

bwilkerson:
Just to be clear, I wasn't recommending extending it to ternary operators. I was asking about references in general. Consider:
```
int get value => plusOne(value);
```
It seems to me that if we find a reference to a getter anywhere inside the getter it's probably a problem. That said, it might not be worth handling at this point; I was just asking what you thought about it.

bwilkerson:
lgtm

alexeieleusis:
I think it simplifies the implementation a lot, at the same time I guess those cases would be extremely rare and my concern is that the probability of false positives will increase.

The current implementation detects silly (typo) mistakes but I would expect those being the ones that will trigger the rule most of the times and therefore where the value of it lies. I think in `invariant_booleans` rule where most of the cases where it is triggered are debugging left overs of the form `if (realCondition || true)`.

I'll revisit tonight testing in the SDK and Flutter codebases, most likely will submit as is and leave the bug open to review later.

bwilkerson:
sgtm

alexeieleusis:
Tested on SDK and Flutter, no errors. Will merge Monday around noon unless there is a concern.

bwilkerson:
No concerns.

PR #333, opened 2016-11-16
invariant_booleans: Fix false positive for nested functions.

alexeieleusis:


lib/src/rules/invariant_booleans.dart
```diff
@@ -182,9 +182,10 @@ class _ContradictionReportRule extends LintRule {
 }
 
 TestedExpressions _findPreviousTestedExpressions(BinaryExpression node) {
-  Block block = node
-      .getAncestor((a) => a is Block && a.parent is BlockFunctionBody);
-  Iterable<AstNode> nodesInDFS = DartTypeUtilities.traverseNodesInDFS(block);
+  Block block =
+      node.getAncestor((a) => a is Block && a.parent is BlockFunctionBody);
```
bwilkerson: I'm guessing that this should just be `FunctionBody`, but I haven't tested it. Consider:
```
someFunction() {
  int a = new DateTime.now().millisecondsSinceEpoch;
  innerFunction() => a > 0 ? false : null;
  if (a > 0) { // OK
    print('bla');
  }
}
```
If that doesn't currently fail, then there's no need to change the test.

alexeieleusis: Added this code to the test and is not linted, but I am guessing that is a missing case for the rule, since it is the opposite bug I won't address it in this CL, also it fits with an improvement I have been thinking for this rule.

alexeieleusis:
I think there are no open issues with this PR, will wait for explicit LGTM before merging.

alexeieleusis:
I forgot to include the issue reference in previous comment https://github.com/dart-lang/linter/issues/336 
merging!

alexeieleusis:
I can't figure exactly what the missing case is nor what the test should be, any hints? I guess the rule is not missing anything, just by coincidence it was ignoring this case.

bwilkerson:
You changed the test case from having an ExpressionFunctionBody (which isn't handled by this version of the code) to having a BlockFunctionBody, which is handled (which is why you didn't see a failure). Hence my comment that if you just change "is BlockFunctionBody" to "is FunctionBody" it would fix the error.

alexeieleusis:
I didn't see that before, definitely is better, changed it and also added the function to the test since is only one more line of code.

Issue #317, opened 2016-10-17
cancel_subscriptions can't tell that I'm closing the subscription


yjbanov:
See [example](https://github.com/flutter/flutter/blob/88e065de861e1930fc306e8e39a98a4377547a6d/packages/flutter_tools/lib/src/commands/drive.dart#L87) in Flutter.

/cc @alexeieleusis

matanlurey:
I honestly think this lint is fundamentally flawed - yes you can whitelist some patterns that will close the subscription, but you'll need almost whole-program analysis to actually tell if something is closed - plus there are some streams (i.e. streamy) that close automatically.

alexeieleusis:
@yjbanov did you have a chance to verify this?

yjbanov:
Nope, but I set a reminder to verify it after the holidays.

Hixie:
Turns out we accidentally had no lint checking at all for the past few weeks, and this isn't actually fixed.

Hixie:
Not sure what @cbracken did there but I'm guessing that wasn't intentional.

cbracken:
Nope! Bizarre, I actually just pushed master to cbracken/flutter master branch. Confused why that would affect this bug at all.

PR #335, opened 2016-11-19
LeakDetectorVisitor: When analyzing places where a node is referenced, include the whole library.

alexeieleusis:
Fixes https://github.com/dart-lang/linter/issues/317

### DIFF
--- a/lib/src/util/leak_detector_visitor.dart
+++ b/lib/src/util/leak_detector_visitor.dart
@@ -59,8 +59,7 @@ _VisitVariableDeclaration _buildVariableReporter(
       validators.add(_findMethodInvocationsWithVariableAsArgument(
           containerNodes, variable));
 
-      // Read this as: validators.forAll((i) => i.isEmpty).
-      if (!validators.any((i) => i.isNotEmpty)) {
+      if (validators.every((i) => i.isEmpty)) {
         rule.reportLint(variable);
       }
     };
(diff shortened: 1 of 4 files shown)


lib/src/util/leak_detector_visitor.dart
```diff
@@ -147,15 +146,15 @@ abstract class LeakDetectorVisitor extends SimpleAstVisitor {
...
+    CompilationUnit unit = node.getAncestor((a) => a is CompilationUnit);
     node.fields.variables.forEach(_buildVariableReporter(
-        classDecl, _fieldPredicateBuilders, rule, predicates));
+        unit, _fieldPredicateBuilders, rule, predicates));
   }
 
   @override
   void visitVariableDeclarationStatement(VariableDeclarationStatement node) {
-    FunctionBody function = node.getAncestor((a) => a is FunctionBody);
+    CompilationUnit unit = node.getAncestor((a) => a is CompilationUnit);
```
bwilkerson: It isn't clear to me what use case this is covering. It seems like VariableDeclarationStatements should only be possible inside function bodies, and the variables cannot be referenced outside the function. What am I forgetting?

If there is a case, let's add a test that covers the case.

alexeieleusis: Reverted, I was trying to refactor both methods but I couldn't since the callback is different and didn't completely reverted back.

PR #342, opened 2016-11-24
Add prefer_const_constructor rule (#129)

dikmax:


lib/src/rules/prefer_const_constructor.dart
```diff
@@ -0,0 +1,57 @@
...
+class A {
+  const A();
+}
+
+void accessA() {
+  A a = const A();
+}
+```
+''';
+
```
alexeieleusis: Add a bad example here.

lib/src/rules/prefer_const_constructor.dart
```diff
@@ -0,0 +1,57 @@
...
+class _Visitor extends SimpleAstVisitor {
+  final LintRule rule;
+
+  _Visitor(this.rule);
+
+  @override
+  void visitInstanceCreationExpression(InstanceCreationExpression node) {
+    if (!node.isConst && node.staticElement != null) {
+      if (node.staticElement.isConst) {
```
alexeieleusis: You could add this as a third condition to the containing if.

dikmax:
@alexeieleusis Could you give some advise how you run rule against sdk and flutter? I think that kind of script will be useful for contributors.

alexeieleusis:
I clone both repos and move to the root folder in the case of Flutter and to sdk folder in the case of tht dart sdk, then I run: `time dart ~/relative/path/linter/bin/linter.dart --rules=rule_name . > linter.txt`, afterwards I look in the `linter.txt` file with something like `cat linter.txt | grep "\[lint\]" | less` and look through the reported cases to discard false positives or errors.

I have a helper script to see results in command line which I paste below:

```
$ cat /.../show-snippet
#!/bin/bash

echo "==============================================================="
head $1 --lines $(($2+5)) | tail
echo "==============================================================="
```

It is invoked with the path of the file to visualize and the line number you want to be at the center, it will display 5 lines above and 5 lines below.

alexeieleusis:
Here is an example of the output I get (this is for the recursive_getter rule):

```
$ show-snippet /.../flutter/bin/cache/dart-sdk/lib/analyzer/lib/src/dart/element/element.dart 2953
===============================================================
  Source get source {
    if (_enclosingElement == null) {
      return null;
    }
    return _enclosingElement.source;
  }

  /**
   * Set whether this element is synthetic.
   */
===============================================================
```

dikmax:
In Dart SDK I've got 317 lint messages, and 416 in flutter.
I've checked some of them (not all, of course), no error or false-positives found.

alexeieleusis:
@dikmax do you want to squash commits here too? github does that automatically.

alexeieleusis:
If not one else does or raise concerns, will squash and merge around 8pm (UTC) tomorrow.

dikmax:
I've done squashing.

eernstg:
Just passing by, I took a look and couldn't verify that this lint rule makes any attempt to detect whether the constructor takes arguments which are not const. For instance, the following forces `new` with a `const` constructor:
```dart
class A {
  final int x;
  const A(this.x);
}

A foo(int x) => new A(x);
```
Will you mark that `new` as a violation of this lint rule? That would outlaw a useful part of the language, namely the ability to use const constructors with non-const arguments. Is that intended?

dikmax:
@eernstg You're right!
I need to rewrite this rule.

a14n:
I don't think it's handled. I just run that rule on the flutter SDK and I get an hint for :

```dart
double bottomPadding = 8.0;
// ...
if (errorText != null) {
  // ....
  if (!config.isDense)
    bottomPadding = 1.0;
  }
EdgeInsets padding = new EdgeInsets.only(top: topPadding, bottom: bottomPadding);
```

In this case, `const` could not be used.

dikmax:
It appeared much harder than I expected, as it requires tighter integration with analyser.
To be specifically we need instantiate `ConstantEvaluationEngine`, which have additional dependencies like TypeProvider and TypeSystem. And all that requires much more knowledge that I have right now.

dikmax:
Here's updated version. Now it checks if constructor cannot be used with const keyword.
Flutter code contains 150 lints and the SDK contains 69.

lib/src/rules/prefer_const_constructor.dart
```diff
@@ -62,7 +75,23 @@ class _Visitor extends SimpleAstVisitor {
     if (!node.isConst &&
         node.staticElement != null &&
         node.staticElement.isConst) {
-      rule.reportLint(node);
+      TypeProvider typeProvider = node.staticElement.context.typeProvider;
+      TypeSystem typeSystem = node.staticElement.context.typeSystem;
+      bool allConst = node.argumentList.arguments.every((Expression argument) {
+        Expression realArgument =
+            argument is NamedExpression ? argument.expression : argument;
+        DeclaredVariables declaredVariables = new DeclaredVariables();
```
bwilkerson: The set of declared variables is also available via the analysis context.

dikmax: Fixed.

dikmax:
I squashed commits once again, so you can merge this pull request if there are no more complaints.

pq:
Awesome.  Late to the part but LGTM.  I'll defer to @bwilkerson though who's been following along more closely. 👍

a14n:
This rule is really great :+1: 

Running this lint on flutter, it looks like some cases are missing (but I don't really understand why the lint fails to report them) see https://github.com/a14n/flutter/commit/d7acae4b78f3987312802a1ad6db78f123ead506 for some examples.

Another point is that the lint report for `new Object()`. I'm not sure this should appear as lint. Flutter use this as unique key. wdyt ?

bwilkerson:
> Another point is that the lint report for new Object(). I'm not sure this should appear as lint. Flutter uses this as unique key. wdyt ?

Yep. Sometimes, even if it would be valid code to create a const instance, you don't want to. This is a good example of such a case. It's clearly a false positive; the question is whether we can figure out that this should be an exception and remove the false positive.

dikmax:
@a14n Looks like other packages are not loaded. Analyzer fails to calculate constants and thus lint is skipped.

dikmax:
> Sometimes, even if it would be valid code to create a const instance, you don't want to. This is a good example of such a case. It's clearly a false positive; the question is whether we can figure out that this should be an exception and remove the false positive.

We can explicitly exclude `new Object` from lints. Of course, it won't fix all such usages but at least it would fix some of them.

zoechi:
What is `new Object()` for? Is this just to avoid `null` for `dynamic` parameters (in strong mode probably rather rare) or are there other use cases?

a14n:
For examples, see :

 - https://github.com/flutter/flutter/blob/45a4f639c9efb31255a0fdf00de87fa25c857616/packages/flutter/lib/src/material/app_bar.dart#L18
- https://github.com/flutter/flutter/blob/45a4f639c9efb31255a0fdf00de87fa25c857616/packages/flutter/lib/src/material/scaffold.dart#L507
- https://github.com/flutter/flutter/blob/45a4f639c9efb31255a0fdf00de87fa25c857616/packages/flutter/lib/src/material/floating_action_button.dart#L20

zoechi:
@a14n Thanks a lot. Looks like it's just used for the object identity. Valid use case of course. And `const` would ruin just that.

dikmax:
I've added an exception for `new Object()`. 

Now it's 4 lints less for Flutter and 1 less for SDK.

pq:
Thanks again for this PR @dikmax!  Sorry also for the slow turn around in integration.  The build is happily green now so things should be speedier from here out. 👍

PR #344, opened 2016-11-26
LeakDetectorVisitor: Include cascade operators when searching for invocations on variable.

alexeieleusis:
Fixes https://github.com/dart-lang/linter/issues/341

This goes on top of https://github.com/dart-lang/linter/pull/343, so this needs that to be merged before merging, and only consider the last commit when reviewing.

### DIFF
--- a/lib/src/util/leak_detector_visitor.dart
+++ b/lib/src/util/leak_detector_visitor.dart
@@ -95,9 +95,11 @@ Iterable<AstNode> _findNodesInvokingMethodOnVariable(
         Map<DartTypePredicate, String> predicates) =>
     classNodes.where((AstNode n) =>
         n is MethodInvocation &&
-        _hasMatch(predicates, variable.element.type, n.methodName.name) &&
-        (_isSimpleIdentifierElementEqualToVariable(n.target, variable) ||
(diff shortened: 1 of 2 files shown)


alexeieleusis:
I think I followed the suggestion... PTAL

lib/src/util/leak_detector_visitor.dart
```diff
@@ -126,6 +128,21 @@ bool _hasMatch(Map<DartTypePredicate, String> predicates, DartType type,
...
+  }
+
+  final identifier = invocation.realTarget as SimpleIdentifier;
+  if (identifier.bestElement is! PropertyAccessorElement) {
+    return false;
+  }
+
+  final element = identifier.bestElement as PropertyAccessorElement;
+  return element.variable == variable.element;
+}
```
bwilkerson: Just something to consider: There is a slightly different coding style that would result in shorter code (fewer lines, fewer casts):
```
final identifier = invocation.realTarget;
if (identifier is SimpleIdentifier) {
  final element = identifier.bestElement;
  if (element is PropertyAccessorElement) {
    return element.variable == variable.element;
  }
}
return false;
```

alexeieleusis:
Performed suggested changes, thank you!

PR #345, opened 2016-11-26
cascade_invocations: Include assignment expressions when looking for cascade candidates.

alexeieleusis:
Fixes https://github.com/dart-lang/linter/issues/340

### DIFF
--- a/lib/src/rules/cascade_invocations.dart
+++ b/lib/src/rules/cascade_invocations.dart
@@ -6,6 +6,7 @@ library linter.src.rules.cascade_invocations;
 
 import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/element.dart';
 import 'package:linter/src/linter.dart';
 
 const _desc = r'Cascade consecutive method invocations on the same reference.';
@@ -50,6 +51,14 @@ someReference
 ```
 
 ''';
+
+Element _findElement(SimpleIdentifier identifier) {
+  if (identifier.bestElement is PropertyAccessorElement) {
+    return (identifier.bestElement as PropertyAccessorElement).variable;
+  }
+
+  return identifier.bestElement;
+}
 
 SimpleIdentifier _findTarget(AstNode node) {
   Expression expression;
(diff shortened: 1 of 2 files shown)


lib/src/rules/cascade_invocations.dart
```diff
@@ -51,6 +52,14 @@ someReference
 
 ''';
 
+Element _findElement(SimpleIdentifier identifier) {
+  if (identifier.bestElement is PropertyAccessorElement) {
```
bwilkerson: style: If you assign "identifier.bestElement" to a local variable, then (a) you won't have to repeat the expression so often and (b) you won't need the cast on the next line.

This probably needs to only return the variable when the accessor is synthetic, otherwise code like the following might start generating lints:
```
Foo get createFoo => new Foo();

void m() {
  createFoo.a();
  createFoo.b();
}
```
Which, as written would create two separate instances of Foo, but if converted to a cascade would create a single instance and invoke both method on it.

alexeieleusis: This is a hard one, it is linted but isSynthetic is not enough to discriminate. The bug is reproducible on master though, I wonder if it should be tracked on a different bug report.

bwilkerson: Can you expain why `isSynthetic` is insufficient?

alexeieleusis: I can't because I don't know the reason, I just ran the test and it failed, so I need to dig deeper into this and gain some knowledge on the structure of the tree.

alexeieleusis:
I think I need to recover the AST node from the element to trigger the rule only if the getter is returning a reference, i.e. skip method/function and constructor invocation. I have tried it invoking `computeNode` method in the `Element` class in `.../analyzer-0.29.1/lib/dart/element/element.dart` but it fails the assertion at `'package:analyzer/src/context/context.dart': Failed assertion: line 719 pos 12: '!driver.isTaskRunning' is not true.`.

If I do this only for synthetic nodes previous tests fail, I will continue exploring to see if there is something else I need to consider.

bwilkerson:
The task that is running is the task to generate lints. I'm not sure how hard it will be to detangle these two.

Can you provide more detail about where you think you need to access the AST node? I'd like to see if there's a way to avoid needing the AST node.

pq:
@alexeieleusis : what's the status on this one?  Do you want to fix the merge conflicts and work towards landing?

alexeieleusis:
I need to rewrite this rule, after lots of annoyances with my team we decided to disable it, so I will close for now and send a new one when ready.

PR #346, opened 2016-11-28
No adjacent strings in list. Fixes #225.

### LINKED ISSUE #225: No Adjacent Strings in lists
It can be easy to forget a comma in String list without error because it's still valid code. In String list we should use `+` operator to prevent this.

**BAD**

``` dart
var list = <String>[
  'a'
  'b',
  'c',
];
```

**GOOD**

``` dart
var list = <String>[
  'a' +
  'b',
  'c',
];
```

dikmax:
Simple rule, but can be useful in some cases.

There're no such lints in SDK and Flutter.

### DIFF
--- a/lib/src/rules/no_adjacent_strings_in_list.dart
+++ b/lib/src/rules/no_adjacent_strings_in_list.dart
@@ -1,0 +1,63 @@
+// Copyright (c) 2016, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+library linter.src.rules.no_adjacent_strings_in_list;
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:linter/src/linter.dart';
+
+const desc = 'Do not use adjacent strings in list';
+
+const details = r'''
+**DO NOT** use adjacent strings in list. This can be
+sign of forgotten comma.
+
+**GOOD:**
+
+```
+List<String> list = <String>[
+  'a' +
+  'b',
+  'c',
+];
+```
+
+**BAD:**
+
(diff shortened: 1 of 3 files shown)


lib/src/rules/no_adjacent_strings_in_list.dart
```diff
@@ -0,0 +1,63 @@
...
+library linter.src.rules.no_adjacent_strings_in_list;
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:linter/src/linter.dart';
+
+const desc = 'Do not use adjacent strings in list';
+
+const details = r'''
+**DO NOT** Do not use adjacent strings in list. This can be
```
alexeieleusis: Remove the second 'Do not'

lib/src/rules/no_adjacent_strings_in_list.dart
```diff
@@ -0,0 +1,63 @@
+// Copyright (c) 2015, the Dart project authors.  Please see the AUTHORS file
```
alexeieleusis: Update the year.

lib/src/rules/no_adjacent_strings_in_list.dart
```diff
@@ -0,0 +1,63 @@
...
+```
+''';
+
+class NoAdjacentStringsInList extends LintRule {
+  NoAdjacentStringsInList()
+      : super(
+            name: 'no_adjacent_strings_in_list',
+            description: desc,
+            details: details,
+            group: Group.style);
```
alexeieleusis: I wonder if the group for this is error.

dikmax:
I've changed group to errors.

alexeieleusis:
We should squash in a single commit before merging, I can do it if needed.

dikmax:
No problem. Done it.

PR #353, opened 2016-12-03
Rule prefer_is_empty.

dikmax:
Prefer using `isEmpty` or `isNotEmpty` over comparison of length with 0.

Description from [Effective Dart](https://www.dartlang.org/guides/language/effective-dart/usage#dont-use-length-to-see-if-a-collection-is-empty).

There are 11 lints in SDK and 1 in Flutter.

### DIFF
--- a/lib/src/rules/prefer_is_empty.dart
+++ b/lib/src/rules/prefer_is_empty.dart
@@ -1,0 +1,226 @@
+// Copyright (c) 2016, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+library linter.src.rules.prefer_is_empty;
+
+import 'package:analyzer/context/declared_variables.dart';
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/token.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/element.dart';
+import 'package:analyzer/dart/element/type.dart';
+import 'package:analyzer/error/listener.dart';
+import 'package:analyzer/src/dart/constant/evaluation.dart';
+import 'package:analyzer/src/dart/constant/value.dart';
+import 'package:analyzer/src/error/codes.dart';
+import 'package:analyzer/src/generated/engine.dart';
+import 'package:analyzer/src/generated/resolver.dart';
+import 'package:analyzer/src/lint/linter.dart';
+
+const desc = 'Use isNotEmpty for Iterables and Maps.';
+
+const useIsNotEmpty = 'Use isNotEmpty instead of length';
+const useIsEmpty = 'Use isEmpty instead of length';
+const alwaysFalse = 'Always false because length is always greater or equal 0.';
+const alwaysTrue = 'Always true because length is always greater or equal 0.';
+
+const details = '''
+**DO NOT** use `.length` to see if a collection is empty.
+
+The `Iterable` contract does not require that a collection know its length or be
+able to provide it in constant time. Calling `.length` just to see if the
+collection contains anything can be painfully slow.
+
+Instead, there are faster and more readable getters: `.isEmpty` and
+`.isNotEmpty`. Use the one that doesn’t require you to negate the result.
+
+**GOOD:**
+```
+if (lunchBox.isEmpty) return 'so hungry...';
+if (words.isNotEmpty) return words.join(' ');
+```
+
+**BAD:**
+```
+if (lunchBox.length == 0) return 'so hungry...';
+if (words.length != 0) return words.join(' ');
+```
+''';
+
+class _LintCode extends LintCode {
+  static final registry = <String, LintCode>{};
+
+  factory _LintCode(String name, String message) => registry.putIfAbsent(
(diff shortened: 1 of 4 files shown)


lib/src/rules/prefer_is_empty.dart
```diff
@@ -0,0 +1,163 @@
...
+    // Should be "length".
+    Element propertyElement = lengthIdentifier.bestElement;
+    if (propertyElement?.name != 'length') {
+      return;
+    }
+
+    // Should have `isNotEmpty` and `isEmpty`.
+    if (type.getGetter('isEmpty') == null ||
+        type.getGetter('isNotEmpty') == null) {
+      return;
```
bwilkerson: This doesn't restrict the lint to invocations on subtypes of `Iterable`. Consider:
```
class BankAccount {
  /// Return true if the balance is zero or negative.
  bool get isEmpty => ...;
  /// Return true if the balance is positive.
  bool get isNotEmpty => ...;
  /// Return the length of time (in days) that this account has been open.
  Duration get length => ...;
}
```
You might want to test whether the `type.isSubtypeOf(typeProvider.iterable)`.

If not, then note that `getGetter` returns the getter defined in the class, but not one that is inherited. So, a class like
```
class MyList extends List {
  void rotateContents() {}
}
```
will return `null` for both getters. If you really want to use duck typing semantics, then you'll need to use `lookUpGetter`.

dikmax:
Updated version.

Now it shows 90 issues in SDK and 30 issues in Flutter.

lib/src/rules/prefer_is_empty.dart
bwilkerson: I don't know, but I don't think so. I'll try to make some time to look into this, but I'm not sure how soon that will be.

lib/src/rules/prefer_is_empty.dart
```diff
@@ -0,0 +1,210 @@
...
+
+    Token operator = binaryExpression.operator;
+    DeclaredVariables declaredVariables = context.declaredVariables;
+
+    // Comparing constants with length.
+
+    DartObjectImpl rightValue = binaryExpression.rightOperand.accept(new ConstantVisitor(
+        new ConstantEvaluationEngine(typeProvider, declaredVariables,
+            typeSystem: typeSystem),
+        rule.reporter));
```
bwilkerson: I missed this the first time, but I think passing in `rule.reporter` will cause any errors produced by the ConstantVisitor to be reported twice. I think you want `AnalysisErrorListener.NULL_LISTENER`.

lib/src/rules/prefer_is_empty.dart
```diff
@@ -0,0 +1,210 @@
...
+      }
+      type = parent.target?.bestType;
+    } else if (parent is PrefixedIdentifier) {
+      lengthIdentifier = parent.identifier;
+      lengthAccess = parent;
+      if (parent.prefix.bestType is! InterfaceType) {
+        return;
+      }
+      type = parent.prefix.bestType;
+    }
```
bwilkerson: I might be missing something, but it looks like this would generate two lints for every violation (though I'm somewhat surprised if the test framework doesn't catch that issue). Given an expression like `list.length`, this method will be called twice: one for `list` and once for `length`, and both times we will exit this `if` with a non-null `lengthIdentifier`.

I think you need to do something like
```
if (parent is PropertyAccess && identifier == parent.propertyName) {
```
In which case you probably don't need `lengthIdentifier` any more. And if you add an `} else { return;` to the `if`, you won't need `lengthAccess` either. (I'm not requesting either of these last two changes, just noting them.)

lib/src/rules/prefer_is_empty.dart
```diff
@@ -0,0 +1,210 @@
...
+
+    if (lengthIdentifier == null) {
+      return;
+    }
+
+    // Should be "length".
+    Element propertyElement = lengthIdentifier.bestElement;
+    if (propertyElement?.name != 'length') {
+      return;
+    }
```
bwilkerson: Not required, but note that this could be replaced by
```
if (lengthIdentifier.name != 'length') {
  return;
}
```
and the test could even be performed before testing the parent. It would probably make this slightly more efficient.

dikmax:
I've fixed all issues with PR.

lib/src/rules/prefer_is_empty.dart
```diff
@@ -0,0 +1,206 @@
...
+```
+
+**BAD:**
+```
+if (lunchBox.length == 0) return 'so hungry...';
+if (words.length != 0) return words.join(' ');
+```
+''';
+
+class _LintCode extends LintCode {
```
bwilkerson: This appears in multiple rules. As a separate PR (by you or someone else) we should move this to a utility library so that we don't have to keep replicating the code. Not something that needs to be addressed now.

dikmax:
I've added a couple more cases. Now it notifies of certain comparisons with `1`.
I.e. `length >= 1` is same as `isNotEmpty` and `length < 1` is same as `isEmpty`.

No such usage cases were found in SDK and Flutter.

lib/src/rules/prefer_is_empty.dart
pq: > This case could be effectively solved by adding optional description parameter to reportLint method.

👍

pq:
Thanks a million for your patience!

PR #355, opened 2016-12-04
Add no_duplicate_cases rule (#354)

dikmax:
This rule has found 4 lints (2 x 2 duplicated cases) in SDK and 0 in Flutter.

Here's the thing with SDK:

KeyCode.WIN_KEY_LEFT and KeyCode.META have same value, so _convertKeyCodeToKeyName will return _KeyName.META instead of _KeyName.WIN for first constant as may be expected.

Check html_dart2js.dart lines 41086 and 41100, and html_dartium.dart line 45611 and 45625.

### DIFF
--- a/lib/src/rules.dart
+++ b/lib/src/rules.dart
@@ -33,6 +33,7 @@ import 'package:linter/src/rules/library_prefixes.dart';
 import 'package:linter/src/rules/list_remove_unrelated_type.dart';
 import 'package:linter/src/rules/literal_only_boolean_expressions.dart';
 import 'package:linter/src/rules/no_adjacent_strings_in_list.dart';
+import 'package:linter/src/rules/no_duplicate_case_values.dart';
 import 'package:linter/src/rules/non_constant_identifier_names.dart';
 import 'package:linter/src/rules/one_member_abstracts.dart';
 import 'package:linter/src/rules/only_throw_errors.dart';
@@ -90,6 +91,7 @@ void registerLintRules() {
     ..register(new ListRemoveUnrelatedType())
     ..register(new LiteralOnlyBooleanExpressions())
     ..register(new NoAdjacentStringsInList())
+    ..register(new NoDuplicateCaseValues())
     ..register(new NonConstantIdentifierNames())
     ..register(new OneMemberAbstracts())
     ..register(new OnlyThrowErrors())
--- a/lib/src/rules/no_duplicate_case_values.dart
+++ b/lib/src/rules/no_duplicate_case_values.dart
@@ -1,0 +1,126 @@
+// Copyright (c) 2016, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+library linter.src.rules.no_duplicate_case_values;
+
+import 'dart:collection';
+import 'package:analyzer/context/declared_variables.dart';
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/error/listener.dart';
+import 'package:analyzer/src/dart/constant/evaluation.dart';
+import 'package:analyzer/src/dart/constant/value.dart';
+import 'package:analyzer/src/error/codes.dart';
+import 'package:analyzer/src/generated/engine.dart';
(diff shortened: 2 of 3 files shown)


lib/src/rules/no_duplicate_cases.dart
```diff
@@ -0,0 +1,108 @@
...
+    });
+
+    for (SwitchMember member in node.members) {
+      if (member is SwitchCase) {
+        Expression expression = member.expression;
+
+        DartObjectImpl result = expression.accept(new ConstantVisitor(
+            new ConstantEvaluationEngine(typeProvider, declaredVariables,
+                typeSystem: typeSystem),
+            rule.reporter));
```
bwilkerson: I suspect that this will cause errors related to constant evaluation to be reported twice (but haven't confirmed that). You probably want to use the null error listener.

lib/src/rules/no_duplicate_cases.dart
```diff
@@ -0,0 +1,108 @@
...
+          cluster.add(member);
+        } else {
+          clusters[result] = <SwitchCase>[member];
+        }
+      }
+    }
+
+    for (List<SwitchCase> cluster in clusters.values) {
+      if (cluster.length > 1) {
+        cluster.forEach(rule.reportLint);
```
bwilkerson: I would probably choose to only generate a lint for every occurrence _after_ the first. That will minimize the amount of feedback without degrading the quality.

It would also allow for a slight simplification in the implementation: instead of keeping a list, you need only remember the first case that evaluated to a given value. If you find another case with the same value, you can report an error immediately without adding it to a list.

test/rules/no_duplicate_cases.dart
```diff
@@ -0,0 +1,33 @@
...
+  String v = 'aa';
+
+  switch (v) {
+    case 'aa':  // LINT
+    case 'bb':  // LINT
+    case A + A: // LINT
+    case 'bb':  // LINT
+    case A + 'b': // OK
+    default:
+  }
```
bwilkerson: It would be good to test an enum type, and also a user defined type with a const constructor. I suspect they will work, but it's good to be complete.

Also, try testing something for which the ConstantVisitor will produce an error.

lib/src/rules/no_duplicate_cases.dart
```diff
@@ -0,0 +1,108 @@
...
+  case A:
+  case 2:
+}
+```
+''';
+
+class NoDuplicateCases extends LintRule {
+  NoDuplicateCases()
+      : super(
+            name: 'no_duplicate_cases',
```
bwilkerson: Consider naming it 'no_duplicate_case_values'. It seems a little less ambiguous to me.

lib/src/rules/no_duplicate_cases.dart
dikmax: I'm not sure about that. If you see a switch with hundred cases and all of them are not so obvious constants, you probably will be glad to see where is duplicate. 
But I think I could add duplicate expression to lint message.

bwilkerson: If you see a switch with a hundred cases I hope you're looking at generated code, not at hand-written code :-)

But yes, I think adding to the error message the lexical representation of the case expression that matches would be helpful.

dikmax:
I'm trying to rebase everything on top of latest master and resolve conflicts, but with no luck.
It uses updated version of analyzer 0.30.0+1, which is nowhere to be found.
The latest version in SDK is 0.30.0+0. I've tried to use it, but API mismatches.

What is the best way to deal with that?

bwilkerson:
Sorry about that! Things are in a bad state at the moment. Basically, you have to checkout the 'sdk' repo, then use a dependency override in the 'linter' package's `pubspec.yaml` file to point 'analyzer' and 'front_end' at the sdk repo. For example, I use
```
dependency_overrides:
  analyzer:
    path: .../sdk/sdk/pkg/analyzer
  front_end:
    path: .../sdk/sdk/pkg/front_end
```
where `...` is the path to the directory containing the checkout of the 'sdk' repo.

dikmax:
I've pushed updated version, but you have to believe me that it passes unit tests, because I have to make environment adjustments to make them actually run.

alexeieleusis:
I pulled the changes and ran the tests, everything passes. Thank you @dikmax

Issue #350, opened 2016-12-02
cascade_invocations false positives on top level functions in prefixed library


alexeieleusis:
```
library_name.function1();
library_name.function2();
```

alexeieleusis:
it also happens with top level functions

PR #358, opened 2016-12-06
cascade_invocations: Exclude top level function invocations in aliased imported library.

alexeieleusis:
Fixes https://github.com/dart-lang/linter/issues/350


alexeieleusis:
Merging since I was able to run tests locally applying the patch suggested by @bwilkerson

Issue #356, opened 2016-12-05
false positive for cascade for `import ... as xxx`


zoechi:
```dart
import 'dart:io' as io;

import 'package:bwu_grinder_tasks/bwu_grinder_tasks.dart' as tasks;
import 'package:path/path.dart' as path;

void main(List<String> args) {
  tasks.getSubProjects = getSubProjectsImpl;
  tasks.main(args);
}
```
![image](https://cloud.githubusercontent.com/assets/405837/20883153/ec4f7292-bae4-11e6-9ff7-745232284c38.png)

but that's not supported on import aliases

```
Dart VM version: 1.21.0-edge.6bb28d748ae7e337de0da62bffd546c2cd393d13 (Fri Dec  2 17:12:48 2016) on "linux_x64"
```

bwilkerson:
@alexeieleusis Did you recently fix this case, or am I misremembering?

alexeieleusis:
I haven't tested, I reported something similar last friday, is on my plate.

zoechi:
A similar example (don't know if it's worth to create another issue)
```dart
class Paths {
  static String packageName;
  statis String outputDirectory;
}

main() {
  Paths.packageName = projectModel.name;
  Paths.outputDirectory = '../fhir_webapp_generated';
}
```
![image](https://cloud.githubusercontent.com/assets/405837/20900577/0fe0897c-bb2f-11e6-8c42-00d147b50334.png)

Changing to cascade leads to

![image](https://cloud.githubusercontent.com/assets/405837/20900637/54a5edf4-bb2f-11e6-8b2f-da6469bffe12.png)

PR #359, opened 2016-12-06
cascade_invocations: Exclude static method invocations.

alexeieleusis:
Fixes https://github.com/dart-lang/linter/issues/356


alexeieleusis:
Also ran tests locally and they pass.

PR #366, opened 2016-12-09
Rule prefer_contains.

### LINKED ISSUE #364: Prefer contains over indexOf() != -1
Prefer `List.contains()` over comparison `List.indexOf(a)` with -1.

dikmax:
Prefer `contains` over `indexOf() != -1` in Lists and Strings.
Closes #364.
Very similar to PR #353.

There's 6 such lints in SDK (2 on lists and 4 on strings) and 0 in Flutter.

### DIFF
--- a/lib/src/rules/prefer_contains.dart
+++ b/lib/src/rules/prefer_contains.dart
@@ -1,0 +1,209 @@
+// Copyright (c) 2016, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+library linter.src.rules.prefer_contains;
+
+import 'package:analyzer/analyzer.dart';
+import 'package:analyzer/context/declared_variables.dart';
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/token.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/element.dart';
+import 'package:analyzer/dart/element/type.dart';
+import 'package:analyzer/error/listener.dart';
+import 'package:linter/src/analyzer.dart';
+import 'package:linter/src/util/dart_type_utilities.dart';
+
+const alwaysFalse =
+    'Always false because indexOf is always greater or equal -1.';
+const alwaysTrue = 'Always true because indexOf is always greater or equal -1.';
+
+const desc = 'Use contains for Lists and Strings.';
+const details = '''
+**DO NOT** use `.indexOf` to see if a collection contains an element.
+
+The `Iterable` contract does not require that a collection know its length or be
+able to provide it in constant time. Calling `.length` just to see if the
+collection contains anything can be painfully slow.
+
+Instead, there are faster and more readable getters: `.isEmpty` and
+`.isNotEmpty`. Use the one that doesn’t require you to negate the result.
+
+**GOOD:**
+```
+if (lunchBox.isEmpty) return 'so hungry...';
+if (words.isNotEmpty) return words.join(' ');
+```
+
+**BAD:**
+```
+if (lunchBox.length == 0) return 'so hungry...';
+if (words.length != 0) return words.join(' ');
+```
+''';
+
+const useContains = 'Use contains instead of indexOf';
+
+class PreferContainsOverIndexOf extends LintRule {
+  PreferContainsOverIndexOf()
+      : super(
+            name: 'prefer_contains',
+            description: desc,
+            details: details,
+            group: Group.style);
+
+  @override
+  AstVisitor getVisitor() => new _Visitor(this);
+
+  void reportLintWithDescription(AstNode node, String description) {
+    if (node != null) {
+      reporter.reportErrorForNode(new _LintCode(name, description), node, []);
+    }
+  }
+}
+
+class _Visitor extends SimpleAstVisitor {
+  final PreferContainsOverIndexOf rule;
+
+  _Visitor(this.rule);
+
+  @override
+  visitSimpleIdentifier(SimpleIdentifier identifier) {
+    // Should be "indexOf".
+    final Element propertyElement = identifier.bestElement;
+    if (propertyElement?.name != 'indexOf') {
+      return;
+    }
+
+    AstNode indexOfAccess;
+    InterfaceType type;
+
+    final AstNode parent = identifier.parent;
+    if (parent is MethodInvocation && identifier == parent.methodName) {
+      indexOfAccess = parent;
+      if (parent.target?.bestType is! InterfaceType) {
+        return;
+      }
+      type = parent.target?.bestType;
+    } else {
+      return;
+    }
+
+    if (!DartTypeUtilities
+        .implementsAnyInterface(type, <InterfaceTypeDefinition>[
+      new InterfaceTypeDefinition('Iterable', 'dart.core'),
+      new InterfaceTypeDefinition('String', 'dart.core'),
+    ])) {
+      return;
+    }
+
+    // Going up in AST structure to find binary comparison operator for this
+    // `indexOf` access. Most of the time it will be a parent, but sometimes
+    // it can be wrapped in parentheses or `as` operator.
+    AstNode search = indexOfAccess;
+    while (
+        search != null && search is Expression && search is! BinaryExpression) {
+      search = search.parent;
+    }
+
+    if (search is! BinaryExpression) {
+      return;
+    }
+
+    final BinaryExpression binaryExpression = search;
+    final Token operator = binaryExpression.operator;
+
+    final AnalysisContext context = propertyElement.context;
+    final TypeProvider typeProvider = context.typeProvider;
+    final TypeSystem typeSystem = context.typeSystem;
+
+    final DeclaredVariables declaredVariables = context.declaredVariables;
+
+    // Comparing constants with result of indexOf.
+
+    final ConstantVisitor visitor = new ConstantVisitor(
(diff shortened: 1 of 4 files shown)


lib/src/rules/prefer_contains.dart
```diff
@@ -0,0 +1,219 @@
...
+        new ErrorReporter(
+            AnalysisErrorListener.NULL_LISTENER, rule.reporter.source));
+
+    DartObjectImpl rightValue = binaryExpression.rightOperand.accept(visitor);
+
+    if (rightValue?.type?.name == "int") {
+      // Constant is on right side of comparison operator
+      int value = rightValue.toIntValue();
+      if (value == -1) {
+        if (operator.type == TokenType.EQ_EQ ||
```
alexeieleusis: Why not creating a set that contains the desired values and use contains instead?

lib/src/rules/prefer_contains.dart
```diff
@@ -0,0 +1,219 @@
...
+      lengthAccess = parent;
+      if (parent.target?.bestType is! InterfaceType) {
+        return;
+      }
+      type = parent.target?.bestType;
+    } else {
+      return;
+    }
+
+    // Should be subtype of Iterable or String.
```
alexeieleusis: We have utility methods to see if a type extends another. If this approach is better, Could you create an issue to refactor the existing one and use this instead?

lib/src/rules/prefer_contains.dart
```diff
@@ -0,0 +1,219 @@
...
+
+    ConstantVisitor visitor = new ConstantVisitor(
+        new ConstantEvaluationEngine(typeProvider, declaredVariables,
+            typeSystem: typeSystem),
+        new ErrorReporter(
+            AnalysisErrorListener.NULL_LISTENER, rule.reporter.source));
+
+    DartObjectImpl rightValue = binaryExpression.rightOperand.accept(visitor);
+
+    if (rightValue?.type?.name == "int") {
```
alexeieleusis: This nesting of if statements looks really hard to understand. Can you extract methods that indicate what you are doing?

lib/src/rules/prefer_contains.dart
dikmax: Do you mean `isSubtypeOf`? I use it for string. But it doesn't work properly with type parameters. For example, `(List<int>).isSubtypeOf(typeProvider.iterableType)` returns `false`.

lib/src/rules/prefer_contains.dart
dikmax: I'm not sure that it will be faster or easier to understand. All conditions are different. But I have an idea how to reduce code size, maybe it will help.

dikmax:
@alexeieleusis It's similar in common structure:

1. Find identifier with specified name.
2. Find target.
3. Check that target supports transformation.
4. List of conditions for constant in comparison operator.

Unfortunately, all parts are differents and it makes not much sense to create an abstract class. But I may be mistaken.

pq:
@alexeieleusis : what's the current status of this?  Besides possibly consolidating some reusable utility functions, does this feel good to you?

kevmoo:
> There's 6 such lints in SDK (2 on lists and 4 on strings) and 0 in Flutter.

@dikmax would be great to have an issue open on this – just copy-paste your results into the SDK bug

lib/src/rules/prefer_contains.dart
alexeieleusis: I was thinking on the util things we have here https://github.com/dart-lang/linter/blob/master/lib/src/util/dart_type_utilities.dart

alexeieleusis:
To be honest I don't remember having reviewed this code in depth, just noticed the similarity I mentioned and thought it was worth mentioning to avoid duplication. I don't think I could review it this week, but I can try.

I think is an important rule to have, thanks @dikmax!

lib/src/rules/prefer_contains.dart
```diff
@@ -0,0 +1,207 @@
...
+    if (leftValue?.type?.name == "int") {
+      // Constants is on left side of comparison operator
+      _checkConstant(binaryExpression, leftValue.toIntValue(),
+          _invertedTokenType(operator.type));
+    }
+  }
+
+  void _checkConstant(Expression expression, int value, TokenType type) {
+    if (value == -1) {
+      if (type == TokenType.EQ_EQ ||
```
alexeieleusis: Consider creating a (private) iterable and test for contains.

lib/src/rules/prefer_contains.dart
```diff
@@ -0,0 +1,207 @@
...
+          type == TokenType.GT_EQ ||
+          type == TokenType.GT) {
+        rule.reportLintWithDescription(expression, alwaysTrue);
+      }
+    }
+  }
+
+  TokenType _invertedTokenType(TokenType type) {
+    switch (type) {
+      case TokenType.LT_EQ:
```
alexeieleusis: Is this meant to represent negation? If so 
* `!(<=)` is `>`.
* `!(>=)` is `<`.

lib/src/rules/prefer_contains.dart
```diff
@@ -0,0 +1,207 @@
+// Copyright (c) 2016, the Dart project authors.  Please see the AUTHORS file
+
```
alexeieleusis: Remove blank line, also consider "Sort members in dart file" (available in the Code menu in intellij if the dart plugin is installed). I don't quite like the implemented order there, but it avoid discussions, is easy and fast to do.

I think there is no requirement/policy around this, so is up to you.

lib/src/rules/prefer_contains.dart
```diff
@@ -0,0 +1,207 @@
...
+library linter.src.rules.prefer_contains;
+
+import 'package:analyzer/context/declared_variables.dart';
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/token.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/element.dart';
+import 'package:analyzer/dart/element/type.dart';
+import 'package:analyzer/error/listener.dart';
+import 'package:analyzer/src/dart/constant/evaluation.dart';
```
alexeieleusis: Don't import libraries under src of a different package.

lib/src/rules/prefer_contains.dart
```diff
@@ -0,0 +1,207 @@
...
+  AstVisitor getVisitor() => new Visitor(this);
+
+  void reportLintWithDescription(AstNode node, String description) {
+    if (node != null) {
+      reporter.reportErrorForNode(new _LintCode(name, description), node, []);
+    }
+  }
+}
+
+class Visitor extends SimpleAstVisitor {
```
alexeieleusis: Even though this file lives under src, make all you can private to avoid inadvertently breaking others, public APIs are forever!

lib/src/rules/prefer_contains.dart
```diff
@@ -0,0 +1,207 @@
...
+```
+
+**BAD:**
+```
+if (lunchBox.length == 0) return 'so hungry...';
+if (words.length != 0) return words.join(' ');
+```
+''';
+
+class _LintCode extends LintCode {
```
pq: This is another candidate to hoist out of here -- with an explanatory name (something like `MultiMessageLintCode` only better).  For now, a `TODO` would be fine too!

lib/src/rules/prefer_contains.dart
```diff
@@ -0,0 +1,207 @@
...
+      return;
+    }
+
+    AstNode search = lengthAccess;
+    while (
+        search != null && search is Expression && search is! BinaryExpression) {
+      search = search.parent;
+    }
+
+    if (search is! BinaryExpression) {
```
alexeieleusis: Is not clear to me what you are trying to do here, why not extract this code into a method? Then exit early if it returns null.

lib/src/rules/prefer_contains.dart
```diff
@@ -0,0 +1,207 @@
...
+class _LintCode extends LintCode {
+  static final registry = <String, LintCode>{};
+
+  factory _LintCode(String name, String message) => registry.putIfAbsent(
+      name + message, () => new _LintCode._(name, message));
+
+  _LintCode._(String name, String message) : super(name, message);
+}
+
+class PreferContains extends LintRule {
```
alexeieleusis: Probably rename the rule to PreferContainsOverIndexOf or something like that conveys a more complete context.

lib/src/rules/prefer_contains.dart
dikmax: No, it meant for different order of operands (`-1 < indexof(...)`).

lib/src/rules/prefer_contains.dart
dikmax: I'm really not sure, that it would improve readability or performance in any way.

lib/src/rules/prefer_contains.dart
dikmax: Here trying to find binary comparison operator for .indexOf access. We should go up in AST because access could be wrapped in parentheses, or `as` expression.
Most of the time, it will only iterate once.

I'm not sure which part you want to extract into method.

alexeieleusis: Fair point, could you add a comment explaining that then?

dikmax:
I fixed almost everything that was noted in this review. Not sure, that rest worth changing, but you can comment, if you think otherwise.

alexeieleusis:
This looks good to me, but as I said, only reviewed superficially, seems to be thoroughly tested though so I have no concerns.

pq:
👍  LGTM and again thanks for the contribution!

PR #377, opened 2017-01-09
Modify linter to make use of analyzer's new "resolution map" interface.

stereotype441:
This will allow analyzer's AST to be decoupled from its element model.


pq:
Late to the party, but LGTM2! 👍

lib/src/ast.dart
```diff
@@ -34,6 +34,7 @@ import 'package:analyzer/dart/ast/ast.dart'
         TypeParameter,
         VariableDeclaration;
 import 'package:analyzer/dart/ast/token.dart';
+import 'package:analyzer/dart/ast/standard_resolution_map.dart';
```
pq: Oh: in the future, let's make sure these imports stay sorted.

PR #379, opened 2017-01-16
Remove invocations of registerDefault method since it no longer exists.

alexeieleusis:


bwilkerson:
I don't understand this change. The method `Registry.registerDefault` is defined in `sdk/pkg/analyzer/lib/src/lint/registry.dart`. What do you mean by "it no longer exists"?

alexeieleusis:
When I ran the linter locally it was crashing and when in intellij it had res wiggles saying the method does not exist.

Will verify.

bwilkerson:
Is it loaded with / compiled against bleeding edge analyzer? That's the only combination that works at the moment.

alexeieleusis:
Discarded, both intellij and command line run flawless without this change now.

Issue #368, opened 2016-12-20
invariant_boolean: false positive with ?? operator


alexeieleusis:
if (values?.isEmpty ?? true) return new Future.value('');

alexeieleusis:
The problem is not invariant_boolean but rather literal_only_boolean_expressions and it is a bug.

alexeieleusis:
I cannot reproduce this at 2129e9b59bffaf678baa1f9f1e8fec9522f6dc87 though I see it in intellij, I assume it is because the linter version shipped with the SDK does not yet have the fix.

alexeieleusis:
I had an old sdk built locally, after rebuilding is no longer linting in intellij.

PR #380, opened 2017-01-16
Verify unreproducible bug

alexeieleusis:
Verifies https://github.com/dart-lang/linter/issues/368 which I could not reproduce even after reviewing the code where the rule was triggering.

Changed maturity type for rule so it gets reported on CI servers (that ignore experimental).

### DIFF
--- a/lib/src/rules/invariant_booleans.dart
+++ b/lib/src/rules/invariant_booleans.dart
@@ -208,7 +208,7 @@ class InvariantBooleans extends LintRule {
             description: _desc,
             details: _details,
             group: Group.errors,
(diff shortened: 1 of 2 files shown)


bwilkerson:
@pq I don't know what the policy is for changing the maturity of lints.

pq:
LGTM.

> @pq I don't know what the policy is for changing the maturity of lints.

No crisp policy but I think this flakiness qualifies it.

Thanks @alexeieleusis!

alexeieleusis:
Assuming implicit LGTM, still I will wait for an explicit one. Thanks!

PR #384, opened 2017-01-26
Make the auto-generated options list more visible

filiph:
Currently, the link to `options/options.html` is hard to find, even if you know what you're looking for. This change:

- Makes the link much more visible
- Clearly explains how to use it.

### DIFF
--- a/tool/doc.dart
+++ b/tool/doc.dart
@@ -34,10 +34,15 @@ rules are considered stable, while others may be marked **experimental**
 to indicate that they are under review.
 
 Rules can be selectively enabled in the analyzer using
-[analysis options](https://pub.dartlang.org/packages/analyzer).  An
-auto-generated list enabling all options is provided
-[here](options/options.html).  As some lints may contradict each other, only
-some lints will be enabled in practice, but this list should provide a
+[analysis options](https://pub.dartlang.org/packages/analyzer)
+or through the
+[.analysis-options file](https://www.dartlang.org/guides/language/analysis-options#the-analysis-options-file). 
+
+* **[An auto-generated list enabling all options is provided
+here.](options/options.html)** 
+
+As some lints may contradict each other, only a subset of these will be
+enabled in practice, but this list should provide a
 convenient jumping-off point.
 
 These rules are under active development.  Feedback is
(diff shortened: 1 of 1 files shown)


pq:
LGTM.  Thanks!

(And sorry for the slow review!)

tool/doc.dart
```diff
@@ -34,10 +34,15 @@ rules are considered stable, while others may be marked **experimental**
 to indicate that they are under review.
 
 Rules can be selectively enabled in the analyzer using
-[analysis options](https://pub.dartlang.org/packages/analyzer).  An
-auto-generated list enabling all options is provided
-[here](options/options.html).  As some lints may contradict each other, only
-some lints will be enabled in practice, but this list should provide a
+[analysis options](https://pub.dartlang.org/packages/analyzer)
+or through the
+[.analysis-options file](https://www.dartlang.org/guides/language/analysis-options#the-analysis-options-file).
```
zoechi: shouldn't that be `analysis_options.yaml`?
From what I got this is now the preferred name.

pq: Good catch.  That's the front-runner but there is a bit of debate still (https://github.com/dart-lang/sdk/issues/28385).  Maybe best to just ref to the dartlang doc and treat that as the source of truth (and trust that it will get updated once the dust settles around the proposed name change).  I'll follow-up with a PR.

Thanks!

pq: https://github.com/dart-lang/linter/pull/396

👍

filiph:
Hi, I noticed this hasn't landed on the official site yet. @pq, can you hit publish on that?

pq:
In general, I try and keep the docs in sync with the latest public release and so have been waiting to ship.  I opened a tracking bug for the next release (#477) and hope we can push that out in the next few days.  Feel free to chime in if you have any thing else you'd like to see landed there.

Thanks!

filiph:
Oh! Got it. I forgot this thing generates the whole documentation in one go, not just the homepage. 

Looking forward to 0.1.30!

pq:
Done!

filiph:
Thanks!

Oh no, the generated site says:

> An auto-generated list enabling all options is **providedhere.**

Not sure why that's the case. `provided` and `here` are on separate lines.

PR #386, opened 2017-02-01
Update to analyzer 0.29.6.

pq:
Fixes: #385 

@bwilkerson


pq:
FWIW: breaking tests to be addressed in #387.

Issue #387, opened 2017-02-01
Fix stale tests left over from engine refactoring.


pq:
A few tests have gone stale since the engine refactoring work.

This issue tracks fixing them. 😸

Issue #391, opened 2017-02-02
Consider moving `plugin_test` to `analyzer` proper.


pq:
This test verifies option processing which is now done in the `analyzer`.  Best to move it there.

pq:
Closing as stale.

Issue #390, opened 2017-02-02
Fix rule tests (by ensuring rule registry is populated).


pq:
Rule tests are failing because the way rules get registered changed in the `analyzer 0.29` stream.  the rub is we need a good place to ensure that `registerLintRules()` has been called before we lookup lints.  We might consider changing `Registry.ruleRegistry` (in analyzer) to a smart getter. Alternatively, we might wrap registry access in the linter in a separate object that handles initialization.  (Other options too.)

bwilkerson:
Is this needed outside of tests? If not, then I think we should just call `registerListRules` in the setUp function.

Analyzer cannot have a dependency on the linter plugin, so I don't think `Registry.ruleRegistry` can be a "smart" getter. And if I understood your idea for linter, it would only change the function that has to be called in setUp, which wouldn't add much value.

pq:
> Is this needed outside of tests? 

At the moment, yes.  It is also used in `bin/linter`.  Of course, if that goes then the point is at least partly moot!

pq:
Fixed w/ #392

PR #392, opened 2017-02-02
Fix tests (#387, #390).

pq:
* remove plugin_test (possibly moving to SDK #391)
* ensure rule registry is initialized

Fixes: #387, #390

@bwilkerson


pq:
screen shot 2017-02-02 at 12 47 54 pm

🎉

PR #396, opened 2017-02-03
Stop mentioning .analysis_options by name (#384).

pq:
In favor of just referencing the dartlang doc as the single source of truth.

Follow-up from: https://github.com/dart-lang/linter/pull/384#pullrequestreview-19956393

@zoechi @filiph @danrubel


pq:
Thanks!

PR #400, opened 2017-02-05
Bugfix: invariant_boolean only triggers on if statements.

alexeieleusis:


lib/src/rules/invariant_booleans.dart
```diff
@@ -100,49 +100,53 @@ void nestedOk5() {
...
   return nodesInDFS
       .where((n) => n == statement.getAncestor((a) => a == n && a != statement))
-      .where((n) => n is IfStatement)
+      .where(_isAnalyzedNode)
       .where(predicate)
       .where(noFurtherAssignmentInvalidatingCondition)
       .fold(<Expression>[], (List<Expression> previous, AstNode statement) {
-    if (statement is IfStatement) {
-      previous.add(statement.condition);
+    if (_isAnalyzedNode(statement)) {
```
bwilkerson: Ditto

lib/src/rules/invariant_booleans.dart
```diff
@@ -100,49 +100,53 @@ void nestedOk5() {
...
     nodesInDFS
+        .where(_isAnalyzedNode)
+        .where(_isConditionalStatementWithReturn(nodesInDFS))
         .takeWhile((n) => n != node.parent)
-        .where(_isIfStatementWithReturn(nodesInDFS))
         .where((_noFurtherAssignmentInvalidatingCondition(node, nodesInDFS)))
         .fold(<Expression>[], (List<Expression> previous, AstNode statement) {
-      if (statement is IfStatement) {
-        previous.add(statement.condition);
+      if (_isAnalyzedNode(statement)) {
```
bwilkerson: Do we need this condition? Isn't the condition already known to be true because of the `where(_isAnalyzedNode)`?

alexeieleusis: Thank you! removed

lib/src/rules/invariant_booleans.dart
alexeieleusis: Same.

PR #401, opened 2017-02-06
Split conjunctions when creating collections with known facts deduced  from the conditions in ancestor nodes.

alexeieleusis:
This allows each fact to be available individually when looking for contradictions.

FIXES: https://github.com/dart-lang/linter/issues/373


lib/src/rules/invariant_booleans.dart
```diff
@@ -224,6 +227,16 @@ AstNodePredicate _noFurtherAssignmentInvalidatingCondition(
       0;
 }
 
+Iterable<Expression> _splitConjunctions(Expression expression) {
+  if (expression is BinaryExpression &&
+      expression.operator.type == TokenType.AMPERSAND_AMPERSAND) {
+    return _splitConjunctions(expression.leftOperand).toList()
```
bwilkerson: If you return `List<Expression>`, then you won't need the `toList` here and can remove some unnecessary list copying operations.

alexeieleusis: Fixed.

Issue #407, opened 2017-02-08
Clean up (or disable) new lints.


pq:
https://github.com/dart-lang/linter/commit/9fa9a20bb0de236c727dd2959fb84ac567ec82c4 flips the bit on a bunch of new lints which flag quite a few new warnings.

These should be vetted first and either fixed or disabled.

alexeieleusis:
I think that what happened is that those were introduced in a different commit that was in the works at the same time, will cleanup some time today.

pq:
Sounds good! 👍

alexeieleusis:
I have a problem, `tool/travis.sh` is linting a file under `test/rules`, even though there is
```yaml
  exclude:
    - test/_data/**
    - test/rules/**
```

Any hint?

pq:
Can you send me to the build output?

alexeieleusis:
I lost it now, but I ended up disabling cascase_invocations which was the triggered rule. I have a rewrite pending for it.

pq:
Ok, cool.  Let me know if there's anything I can do to help!

PR #409, opened 2017-02-08
Cleanup linter warnings.

alexeieleusis:
Disabled `prefer_final_locals` because it has too many occurrences and `cascade_invocations` because it was linting the tests.

Also, my team disabled the rule because the way is implemented is not really useful for us so I might rewrite it to be aligned with usages that look like builder/fluent api.

Some changes are just dartfmt being run.

FIXES https://github.com/dart-lang/linter/issues/407

### DIFF
--- a/lib/src/rules/invariant_booleans.dart
+++ b/lib/src/rules/invariant_booleans.dart
@@ -300,9 +300,10 @@ class _Visitor extends SimpleAstVisitor {
 
     TestedExpressions testedNodes = _findPreviousTestedExpressions(node);
     testedNodes.evaluateInvariant().forEach((ContradictoryComparisons e) {
-      _ContradictionReportRule reportRule = new _ContradictionReportRule(e);
-      reportRule.reporter = rule.reporter;
-      reportRule.reportLint(e.second);
+      final reportRule = new _ContradictionReportRule(e);
+      reportRule
+        ..reporter = rule.reporter
+        ..reportLint(e.second);
     });
 
     // In dart booleanVariable == true is a valid comparison since the variable
--- a/.analysis_options
+++ b/.analysis_options
@@ -16,7 +16,7 @@ linter:
     - await_only_futures
 #    - camel_case_types
     - cancel_subscriptions
-    - cascade_invocations
+#    - cascade_invocations
 #    - close_sinks # https://github.com/dart-lang/linter/issues/268
     - comment_references
 #    - constant_identifier_names
@@ -42,9 +42,9 @@ linter:
     - package_names
     - package_prefixed_library_names
     - parameter_assignments
-    - prefer_const_constructors
+#    - prefer_const_constructors
     - prefer_final_fields
-    - prefer_final_locals
+#    - prefer_final_locals
 #    - prefer_is_empty # not recognized
     - prefer_is_not_empty
 #    - public_member_api_docs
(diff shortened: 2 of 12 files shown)


lib/src/ast.dart
```diff
@@ -36,7 +36,7 @@ bool hasOverrideAnnotation(Declaration node) =>
 /// Returns `true` if this [node] is the child of a private compilation unit
 /// member.
 bool inPrivateMember(AstNode node) {
-  AstNode parent = node.parent;
+  final parent = node.parent;
```
pq: Do you meant to lose types here?

lib/src/rules/invariant_booleans.dart
```diff
@@ -134,8 +134,8 @@ Set<Expression> _findConditionsOfStatementAncestor(
 
 Set<Expression> _findConditionsUnderStatementBranch(Statement statement,
     AstNodePredicate predicate, Iterable<AstNode> nodesInDFS) {
-  Expression condition = _getCondition(statement);
-  AstNodePredicate noFurtherAssignmentInvalidatingCondition =
+  final condition = _getCondition(statement);
```
pq: And here?  I'd probably leave them in (as you do elsewhere).

lib/src/rules/invariant_booleans.dart
```diff
@@ -149,16 +149,16 @@ Set<Expression> _findConditionsUnderStatementBranch(Statement statement,
 }
 
 TestedExpressions _findPreviousTestedExpressions(Expression node) {
-  Block block = node.getAncestor((a) => a is Block && a.parent is FunctionBody);
-  Iterable<AstNode> nodesInDFS = DartTypeUtilities.traverseNodesInDFS(block,
+  final Block block =
+      node.getAncestor((a) => a is Block && a.parent is FunctionBody);
+  final nodesInDFS = DartTypeUtilities.traverseNodesInDFS(block,
```
pq: Similarly, here (and below).  Notice how the `Block` type is intact.  Is that intentional?

bwilkerson:
I do not agree with the lint (I don't think it adds value to mark local variables as final, but adds a lot of visual clutter), so I wouldn't enable it. Mostly I don't want lints enabled unless the code is changed so that no lints are produced.

alexeieleusis:
In general I reviewed types and they were preserved, but I might have missed a few, will revert those instances.

Regarding the noise, I don't think final is more noisy than var or a type annotation if the type is inferred correctly.

pq:
> Regarding the noise, I don't think final is more noisy than var or a type annotation if the type is inferred correctly.

But if the plan is to keep the types, then it adds more right?

    Block block = ...;

becomes

    final Block block = ...;

If the choice is between `var` and `final` on the other hand, I'm all for it.  What extent we decide to depend on type inference is an open question though and one I think the whole ecosystem is still mulling over.

Anyway, back to the present, I did an informal poll around here and the feedback was unanimously dubious, so I propose we hold off on this one.  At least until we're ready to take on a larger conversation around type inference coding conventions.

If you want to chat more, feel free to throw something on my calendar!

alexeieleusis:
Remade the cleanup, I cleaned up cacade_invocations but didn't enable the rule.

pq:
Awesome, thanks.

LGTM!

🍻

PR #410, opened 2017-02-08
Avoid null as operand in if null operator

### LINKED ISSUE #369: flag `xxx ?? null`

a14n:
Fixes #369


pq:
PS: I should have asked if this is ready to land? 😄

a14n:
It is. When not ready I usually prefix the title with _WIP:_

pq:
> It is. When not ready I usually prefix the title with WIP:

Great.  Thanks!

I just updated the contributing docs and added that convention:

https://github.com/dart-lang/linter/pull/418

👍

Issue #337, opened 2016-11-20
invariant_booleans false positive with unary operations.


dikmax:
Consider this code:

```dart
if (offset >= length) {
  return;
}

offset++;
if (offset >= length) { // Got error here.
}
```

With `+=` works fine.

Linter 0.1.29

dikmax:
As mentioned #334, I can be sure that it reproducible in 0.1.27 (version bundled with Dart 1.20.1), not in 0.1.29.

alexeieleusis:
This is indeed a bug, those operators were not considered when looking for mutations, this rule needs some cleanup, I will fix this bug after that. It will take a while, but will be fixed.

PR #412, opened 2017-02-09
invariant_boolean: When searching for references reassignment include PostfixExpression and similar, e.g. variable++.

alexeieleusis:
Fixes https://github.com/dart-lang/linter/issues/337

### DIFF
--- a/lib/src/rules/invariant_booleans.dart
+++ b/lib/src/rules/invariant_booleans.dart
@@ -216,15 +216,29 @@ AstNodePredicate _isConditionalStatementWithReturn(
 AstNodePredicate _noFurtherAssignmentInvalidatingCondition(
     Expression node, Iterable<AstNode> nodesInDFS) {
   Set<Identifier> identifiers = _findStatementIdentifiers(node.parent);
-  return (AstNode statement) =>
-      nodesInDFS
-          .skipWhile((n) => n != statement)
-          .takeWhile((n) => n != node)
-          .where((n) =>
-              n is AssignmentExpression &&
-              !identifiers.contains(n.leftHandSide))
-          .length ==
-      0;
+  return (AstNode statement) {
+    bool isMutation(AstNode n) {
+      if (n is AssignmentExpression) {
+        return !identifiers.contains(n.leftHandSide);
+      } else if (n is PostfixExpression) {
+        TokenType type = n.operator.type;
+        return (type == TokenType.PLUS_PLUS || type == TokenType.MINUS_MINUS) &&
+            !identifiers.contains(n.operand);
+      } else if (n is PrefixExpression) {
+        TokenType type = n.operator.type;
+        return (type == TokenType.PLUS_PLUS || type == TokenType.MINUS_MINUS) &&
+            !identifiers.contains(n.operand);
+      }
+
+      return false;
+    }
+
+    return nodesInDFS
+            .skipWhile((n) => n != statement)
+            .takeWhile((n) => n != node)
+            .where(isMutation)
+            .isEmpty;
+  };
 }
 
 List<Expression> _splitConjunctions(Expression expression) {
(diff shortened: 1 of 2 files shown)


pq:
@bwilkerson ?

lib/src/rules/invariant_booleans.dart
```diff
@@ -216,15 +216,35 @@ AstNodePredicate _isConditionalStatementWithReturn(
...
-      nodesInDFS
-          .skipWhile((n) => n != statement)
-          .takeWhile((n) => n != node)
-          .where((n) =>
-              n is AssignmentExpression &&
-              !identifiers.contains(n.leftHandSide))
-          .length ==
-      0;
+  return (AstNode statement) {
+    bool isMutation(AstNode n) {
```
bwilkerson: This is fairly inefficient because you have to perform multiple `is` checks for every node, even when you've already determined that it isn't necessary. Perhaps re-structure slightly:
```
if (n is AssignmentExpression) {
  return !identifiers.contains(n.leftHandSide);
} else if (n is PostfixExpression) {
  TokenType type = n.operator.type;
  return (type == TokenType.PLUS_PLUS || type == TokenType.MINUS_MINUS) && !identifiers.contains(n.operand);
} else ...
}
```

test/rules/invariant_booleans.dart
```diff
@@ -237,3 +237,13 @@ void bug372(bool foo) {
     // doSomethingElse();
   }
 }
+
+void bug337(int offset, int length) {
+  if (offset >= length) {
+    return;
+  }
+
+  offset++;
```
bwilkerson: You might want to introduce other variables so that you can check other operators.

alexeieleusis: This is a very good idea, but I am not sure which ones to check, can you please provide a few examples?

pq: @bwilkerson : thoughts on what @alexeieleusis should check?

bwilkerson: This checks the postfix increment operator, but the rule checks for all four combinations of (prefix | postfix) (increment | decrement) operators. There are three operators that are not being tested, and I think they should be.

alexeieleusis: Thanks, I was confused, somehow I thought this was the rule's code. Will add those cases to the test.

lib/src/rules/invariant_booleans.dart
alexeieleusis: Fixed

test/rules/invariant_booleans.dart
alexeieleusis: Fixed

alexeieleusis:
Fixes https://github.com/dart-lang/sdk/issues/28967

lib/src/rules/invariant_booleans.dart
pq: @bwilkerson: take another look?

lib/src/rules/invariant_booleans.dart
```diff
@@ -216,15 +216,30 @@ AstNodePredicate _isConditionalStatementWithReturn(
...
+
+      return false;
+    }
+
+    return nodesInDFS
+            .skipWhile((n) => n != statement)
+            .takeWhile((n) => n != node)
+            .where(isMutation)
+            .length ==
+        0;
```
pq: `isEmpty`?

test/rules/invariant_booleans.dart
```diff
@@ -237,3 +237,56 @@ void bug372(bool foo) {
...
+  if (b > 0) {
+    b--;
+    if (b == 0) {
+      return;
+    }
+    if (b > 0) {
+      return;
+    }
+  }
+}
```
pq: Add a newline?

lib/src/rules/invariant_booleans.dart
alexeieleusis: haha! Irony, I should be using the linter :)

test/rules/invariant_booleans.dart
alexeieleusis: Done.

PR #417, opened 2017-02-10
Avoid null in null-aware assignments

a14n:


alexeieleusis:
@a14n you have permissions to merge? I can do it if not.

a14n:
I don't have the permission to merge.

Issue #420, opened 2017-02-16
Implement rule to verify DO place “dart:” imports before other imports.


JPaulsen:
From the [effective dart guide](https://www.dartlang.org/guides/language/effective-dart/style#do-place-dart-imports-before-other-imports)

PR #422, opened 2017-02-17
New rule: dart_imports_go_first

JPaulsen:
Verify that all package:dart imports go first. From https://www.dartlang.org/guides/language/effective-dart/style#do-place-dart-imports-before-other-imports

### DIFF
--- a/lib/src/rules.dart
+++ b/lib/src/rules.dart
@@ -21,6 +21,7 @@ import 'package:linter/src/rules/close_sinks.dart';
 import 'package:linter/src/rules/comment_references.dart';
 import 'package:linter/src/rules/constant_identifier_names.dart';
 import 'package:linter/src/rules/control_flow_in_finally.dart';
+import 'package:linter/src/rules/directives_ordering.dart';
 import 'package:linter/src/rules/empty_catches.dart';
 import 'package:linter/src/rules/empty_constructor_bodies.dart';
 import 'package:linter/src/rules/empty_statements.dart';
@@ -84,6 +85,7 @@ void registerLintRules() {
     ..register(new CommentReferences())
     ..register(new ControlFlowInFinally())
     ..registerDefault(new ConstantIdentifierNames())
+    ..register(new DirectivesOrdering())
     ..register(new EmptyCatches())
     ..registerDefault(new EmptyConstructorBodies())
     ..register(new EmptyStatements())
--- a/lib/src/rules/directives_ordering.dart
+++ b/lib/src/rules/directives_ordering.dart
@@ -1,0 +1,92 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+library linter.src.rules.directives_ordering;
+
+import 'package:analyzer/analyzer.dart';
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:linter/src/analyzer.dart';
+
+const _dartImportGoFirst = r"Place 'dart:' imports before other imports.";
+
+const _desc = r'Adhere to Effective Dart Guide directives sorting conventions.';
+
+const _details = r'''**DO** follow the conventions in the [Effective Dart Guide]
+(https://www.dartlang.org/guides/language/effective-dart/style#ordering)
+
+**BAD:**
+```
+import 'package:bar/bar.dart';
+import 'package:foo/foo.dart';
+
+import 'dart:async';  // LINT
+import 'dart:html';  // LINT
+```
+
+**BAD:**
+```
+import 'dart:html';  // OK
+import 'package:bar/bar.dart';
+
+import 'dart:async';  // LINT
+import 'package:foo/foo.dart';
+```
+
+**GOOD:**
+```
+import 'dart:async';  // OK
+import 'dart:html';  // OK
+
+import 'package:bar/bar.dart';
+import 'package:foo/foo.dart';
+```
+
+''';
+
+bool _isDartImport(Directive node) =>
+    (node as ImportDirective).uriContent.startsWith("dart:");
+
+bool _isImportDirective(Directive node) => node is ImportDirective;
+
+class DirectivesOrdering extends LintRule {
+  _Visitor _visitor;
+
+  DirectivesOrdering()
+      : super(
+            name: 'directives_ordering',
+            description: _desc,
+            details: _details,
+            group: Group.style) {
+    _visitor = new _Visitor(this);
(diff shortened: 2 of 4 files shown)


JPaulsen:
I already signed the CLA

zoechi:
fixes #420

alexeieleusis:
LGTM @pq @bwilkerson we forgot to mention this is ready for review.

lib/src/rules/dart_imports_go_first.dart
```diff
@@ -0,0 +1,80 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+library linter.src.rules.dart_imports_go_first;
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:linter/src/analyzer.dart';
+import 'package:linter/src/util/dart_type_utilities.dart';
+
+const desc = r' ';
```
pq: There should be a short description here.

Probably: 

    Place '*dart:*' imports before other imports.

?

lib/src/rules/dart_imports_go_first.dart
```diff
@@ -0,0 +1,80 @@
...
+
+const details = r'''
+
```
pq: Remove leading whitespace?

test/_data/dart_imports_go_first/bad.dart
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) 2016, the Dart project authors.  Please see the AUTHORS file
```
pq: => 2017

test/_data/dart_imports_go_first/good.dart
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) 2016, the Dart project authors.  Please see the AUTHORS file
```
pq: => 2017

test/_data/dart_imports_go_first/bad.dart
pq: Also curious why we need an integration test here.

Couldn't we just put the contents of `bad.dart` in `test/rules/dart_imports_go_first.dart`?

lib/src/rules/dart_imports_go_first.dart
bwilkerson: I think we want a non-empty description.

lib/src/rules/dart_imports_go_first.dart
```diff
@@ -0,0 +1,80 @@
...
+}
+
+class _Visitor extends SimpleAstVisitor {
+  final LintRule rule;
+  _Visitor(this.rule);
+
+  @override
+  void visitCompilationUnit(CompilationUnit node) {
+    Iterable<ImportDirective> nodes =
+        DartTypeUtilities.traverseNodesInDFS(node).where(_isImportDirective);
```
bwilkerson: It would be much more efficient here for the initializer to be `nodes = node.directives;`.

test/_data/dart_imports_go_first/good.dart
```diff
@@ -0,0 +1,9 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';  // OK
+import 'dummy.dart';
```
pq: Ah, OK.  I see why you want an integration test so that you can define `dummy.dart`.

An alternative would be to simplify things and just ignore the resolution error?

Unless you're really invested in making this an integration test, having tests in the `test/rules` directory is a bit simpler.

I'm not hugely invested either way though! 👍

test/_data/dart_imports_go_first/good.dart
```diff
@@ -0,0 +1,9 @@
...
+
```
pq: If this does stay, maybe remove these trailing lines.

test/_data/dart_imports_go_first/good.dart
JPaulsen: Reply is below

test/_data/dart_imports_go_first/good.dart
JPaulsen: Fixed

lib/src/rules/dart_imports_go_first.dart
JPaulsen: Fixed

lib/src/rules/dart_imports_go_first.dart
JPaulsen: Fixed

test/_data/dart_imports_go_first/good.dart
JPaulsen: We went in the Integration Test way to avoid resolution errors

test/_data/dart_imports_go_first/good.dart
JPaulsen: Fixed

test/_data/dart_imports_go_first/bad.dart
JPaulsen: Fixed

lib/src/rules/dart_imports_go_first.dart
JPaulsen: Fixed

lib/src/rules/dart_imports_go_first.dart
JPaulsen: Fixed

lib/src/rules/dart_imports_go_first.dart
```diff
@@ -0,0 +1,76 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+library linter.src.rules.dart_imports_go_first;
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:linter/src/analyzer.dart';
+import 'package:linter/src/util/dart_type_utilities.dart';
+
+const desc = r"Place '*dart:*' imports before other imports.";
```
bwilkerson: The description is used as the error message, displayed in the IDE and on the console. In neither case will markdown be processed, so it would be better to remove the asterisks.

test/_data/dart_imports_go_first/good.dart
bwilkerson: I'm not sure what that means. What resolution errors were you expecting?

JPaulsen:
We added the good.dart file only to confirm that there is no error, if it is better to delete it we will do it

lib/src/rules/dart_imports_go_first.dart
JPaulsen: Fixed

JPaulsen:
Addressed pending comments.

Renamed the rule to convey and verify the whole ordering section.

PTAL

Issue #423, opened 2017-02-17
Implement rule to verify DO place “package:” imports before relative imports.


JPaulsen:
From https://www.dartlang.org/guides/language/effective-dart/style#do-place-package-imports-before-relative-imports

bwilkerson:
I have to wonder whether we ought to have multiple lints related to the order of import and export directives, or whether it might be better to have a single lint that enforces the one blessed order.

JPaulsen:
Yes, we will add the 5 ordering rules to the lint https://github.com/dart-lang/linter/pull/422 in different pull requests

bwilkerson:
But then you merged https://github.com/dart-lang/linter/pull/425, which added a new lint rule rather than extending the previous rule. I'm not sure we're communicating.

JPaulsen:
I didn't merged it, I discarded it, so we can add to the rule implemented in https://github.com/dart-lang/linter/pull/422 in a subsequent pull request. Master branch is still at https://github.com/dart-lang/linter/commit/47e125bc31ffed31e3482a5d935f3666bcaefe68

bwilkerson:
Oh, sorry, I must have misread something.

PR #425, opened 2017-02-17
New rule: package_imports_before_relative

JPaulsen:
Verify that all "package:" imports go before relative. From https://www.dartlang.org/guides/language/effective-dart/style#do-place-package-imports-before-relative-imports

Fixes: https://github.com/dart-lang/linter/issues/423

### DIFF
--- a/lib/src/rules/package_imports_before_relative.dart
+++ b/lib/src/rules/package_imports_before_relative.dart
@@ -1,0 +1,79 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+library linter.src.rules.package_imports_before_relative;
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:linter/src/analyzer.dart';
+
+const _desc = r"Place '*package:*' imports before relative imports.";
+const _details = r'''**DO** place “*package:*” imports before relative imports.
+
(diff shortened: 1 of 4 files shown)


lib/src/rules/package_imports_before_relative.dart
```diff
@@ -0,0 +1,79 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+library linter.src.rules.package_imports_before_relative;
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:linter/src/analyzer.dart';
+
+const _desc = r"Place '*package:*' imports before relative imports.";
```
bwilkerson: Please remove the asterisks.

lib/src/rules/package_imports_before_relative.dart
```diff
@@ -0,0 +1,79 @@
...
+  final LintRule rule;
+  _Visitor(this.rule);
+
+  @override
+  void visitCompilationUnit(CompilationUnit node) {
+    node.directives
+        .where(_isImportDirective)
+        .where(_isNotDartImport)
+        .skipWhile(_isPackageImport)
+        .where(_isPackageImport)
```
bwilkerson: This doesn't take into account the possibility of errors in the code (that is, just because an import isn't either a `dart:` or `package:` import does not mean that it's a relative import). For example:
```
import `packge:foo/bar.dart`;
```
isn't a `package:` URI because of a typo, but it also isn't a relative URI and `package:` URI's that follow it shouldn't have lints generated for them.

PR #426, opened 2017-02-17
New Rule: Do not pass null arguments where a closure is expected.

### LINKED ISSUE #324: new lint rule for passing null to Iterable's firstWhere's orElse
(original issue is https://github.com/dart-lang/sdk/issues/27729)

Not sure how this might be generalized, but take a look at this snippet:

```dart
void main() {
  List a = ['a', 'b', 'c'];
  print(a.firstWhere((e) => e == 'd', orElse: () => null));
  print(a.firstWhere((e) => e == 'd', orElse: null));
}
```

The first `firstWhere` is a correct implementation, but the second is an easy accident, which results in an exception.

https://dartpad.dartlang.org/a0b43b96664ac071ce84599ed767c4c1

Without great edge-case testing, users may implement an `orElse` as muscle memory for safety, but write `orElse: null` instead of `orElse: () => null`. This then throws when the code is encountered, and I think analyzer should warn about it.

Unfortunately, without non-null types, the code _is valid_, so I'm not sure how to generalize. Since Iterable is dart:core, could we make a special case warning for this? Like prefering `isNotEmpty` over `!isEmpty`?

srawlins:
Fixes #324 

A few things should be addressed before this is accepted:

* [x] Is there enough testing? I think I should have more `// OK` tests.
* [x] is using `name` on a [DartType]
* [x] Do we want to restrict this to SDK methods, with known null functionality? Yes


lib/src/rules/null_closures.dart
```diff
@@ -0,0 +1,77 @@
...
+  @override
+  AstVisitor getVisitor() => new Visitor(this);
+}
+
+class Visitor extends SimpleAstVisitor {
+  final LintRule rule;
+  Visitor(this.rule);
+
+  @override
+  void visitMethodInvocation(MethodInvocation node) {
```
bwilkerson: You might want to visitArgumentList instead because that would cover constructor invocations. Note that neither the current form nor that would cover setters, which show up as assignments in the AST.

lib/src/rules/null_closures.dart
```diff
@@ -0,0 +1,77 @@
...
+    _checkNullArgForClosure(node);
+    return;
+  }
+
+  void _checkNullArgForClosure(MethodInvocation node) {
+    ArgumentList argumentList = node.argumentList;
+    NodeList<Expression> args = argumentList.arguments;
+    List<ParameterElement> params = argumentList.correspondingStaticParameters;
+    if (params == null) {
+      return;}
```
bwilkerson: Please run the formatter.

lib/src/rules/null_closures.dart
```diff
@@ -0,0 +1,77 @@
...
+    for (int i = 0; i < args.length; i++) {
+      var arg = args[i];
+      if (arg is NamedExpression) {
+        arg = arg.expression;
+      }
+      if (arg is NullLiteral) {
+        // [name] is null when the type represents the type of an unnamed
+        // function.
+        // https://www.dartdocs.org/documentation/analyzer/latest/analyzer.dart.element.type/DartType/name.html
+        if (params[i].type.name == null) {
```
bwilkerson: I think `params[i].type is FunctionType` would be a better test (I think it would also cover typedef types).

lib/src/rules/null_closures.dart
```diff
@@ -0,0 +1,77 @@
...
+library linter.src.rules.null_closures;
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:linter/src/analyzer.dart';
+import 'package:linter/src/util/dart_type_utilities.dart';
+
+const _desc = r'Do not pass `null` as an argument where a closure is expected.';
+
+const _details = r'''
```
bwilkerson: You should probably note in the description that this lint only catches explicit uses of null literals. It won't catch
```
var test = null;
[].firstWhere(test);
```
nor even
```
[].firstWhere((null));
```

zoechi:
What's the difference with a closure to any other parameter type for it to get a special lint?

bwilkerson:
Good question. I assume it's based on an argument that many / most developers forget that `null` can be passed in where a closure is expected, but are less likely to forget when an object is being passed in. Although, if that is the argument, then the same "forgetfulness" is also true of enums and types such as `int` or `double` that often are not objects in other languages.

I'll also note that this lint will be obsolete if we get non-nullable by default because the static type checks will actually be stronger than this lint.

srawlins:
This is primarily for when a closure can be passed as an optional argument,
which provides a value. Such as Map.putIfAbsent, or Iterable.firstWhere
(etc) with the orElse arg. Developers often just write `orElse: null` (I've
seen a large codebase with tons of this), and the exception is never seen
because the tests don't trigger the orElse closure. Thus, I'm happy to
restrict the rule to the SDK methods outlined in the script.

bwilkerson:
I wouldn't restrict it to SDK methods; more general rules are usually better. But I'll point out that `putIfAbsent` doesn't have an optional parameter, so I think you're targeting more than you said.

There does need to be more tests, and if you haven't yet you should run it over a large code base looking for false positives.

lib/src/rules/null_closures.dart
srawlins: Oops 😝  done.

lib/src/rules/null_closures.dart
srawlins: Done.

lib/src/rules/null_closures.dart
srawlins: Done.

lib/src/rules/null_closures.dart
srawlins: Done.

srawlins:
OK unfortunately, I ran this on a huge codebase and found a lot of cases where a `null` closure is acceptable, and used. Often it would be a constructor with required closure arguments that get assigned to a class member, and then a null check is always performed before calling.

Now I'm in favor of just using a whitelist, similar to the slow IO methods linter. [This](https://gist.github.com/srawlins/1c674f021f3102004565823934153aac) is a pretty good list.

pq:
> Now I'm in favor of just using a whitelist, similar to the slow IO methods linter. This is a pretty good list.

👍 

In the longer term, this is another case that suggests it'd be nice to support configurable of lints.  I can't find the issue now but the idea here might be something like:

```
rules:
  null_closures:
    allow: 
      - List.map
      - List.reduce
      - List.singleWhere
      - ...
```

Although maybe your list is a bit long for this?

pq:
Still can't find the issue that tracks lint parameters, but there's a similar ask in https://github.com/dart-lang/linter/issues/204.

pq:
@srawlins: do you want to finish this one up or should we close it out?

srawlins:
Eek, sorry. I have some changes I'm still casually working on. I'll upload this week!

pq:
No worries.  Thanks in advance! 🎉

srawlins:
PHEW sorry this is fixed. This lint now has a carefully selected set of SDK functions that take closures that cannot be null. Ready for review.

lib/src/rules/null_closures.dart
```diff
@@ -0,0 +1,201 @@
...
+  new NonNullableFunction('dart.core', 'List', 'lastWhere',
+      positional: [0], named: ['orElse']),
+  new NonNullableFunction('dart.core', 'List', 'map', positional: [0]),
+  new NonNullableFunction('dart.core', 'List', 'reduce', positional: [0]),
+  new NonNullableFunction('dart.core', 'List', 'removeWhere', positional: [0]),
+  new NonNullableFunction('dart.core', 'List', 'retainWhere', positional: [0]),
+  new NonNullableFunction('dart.core', 'List', 'singleWhere', positional: [0]),
+  new NonNullableFunction('dart.core', 'List', 'skipWhile', positional: [0]),
+  new NonNullableFunction('dart.core', 'List', 'takeWhile', positional: [0]),
+  new NonNullableFunction('dart.core', 'List', 'where', positional: [0]),
```
a14n: `Iterable` instead of `List` ?

a14n: Use `DartTypeUtilities.implementsAnyInterface` ?

srawlins: Done! Great suggestion!

example/all.yaml
```diff
@@ -59,6 +59,7 @@ linter:
     - no_adjacent_strings_in_list
     - no_duplicate_case_values
     - non_constant_identifier_names
+    - null_closures
```
bwilkerson: This name doesn't seem consistent, but @pq can probably comment on that better than I.

pq: Unfortunately, we're not hugely self-consistent.  (Try as I did to keep it that way.)

It might be consistent w/ more lints if it were more prescriptive (`no`... or `avoid`...) but at this point I'm sort of wishing we'd never added those since they may get out of sync with content.  For example, if we want to promote avoid advice to no, it's awkward if the lint name has avoid baked in...

How's that for a non answer!

srawlins:
I'll reply inline, but my overall response is: its better to have this rule than not have it.

> To be honest, I have a large concern about defining rules with built-in knowledge like this. The problem is that the rule can quickly become out of sync with the code it's checking. If one of the methods on this list starts accepting null as a valid argument value, this lint can easily not be update and hence continue to trigger (producing a false positive).

That'll be the day. These SDK methods change their signatures very rarely. I'd be happy to write a test into the linter repo that asserts that passing null in each of these locations throws an exception.

> And when the next method is added to the SDK that doesn't allow a null literal, this lint will likely fail to notice (a false negative). Keeping a lint like this consistent with reality is hard.

I don't claim to have found them all. It's a big SDK. The rule provides value if it doesn't know of all things. I originally just wanted it for `Iterable.firstWhere(..., orElse: null)` because I found dozens of examples of this one case.

> It would be much better to add a @notNull annotation and apply it to the SDK in places where null is not a valid argument value. I realize that the probability of adding such an annotation to the SDK is close to zero.

+1. I can put a `TODO(srawlins): delete this rule when @notNull lands.`

> At the very least, we ought to document which methods and constructors are covered and that this isn't a general purpose solution to the problem.

Done.

test/rules/null_closures.dart
```diff
@@ -0,0 +1,33 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// test w/ `pub run test -N null_closures`
+
+import 'dart:async';
+import 'dart:core';
```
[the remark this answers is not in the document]
srawlins: Done.

lib/src/rules/null_closures.dart
```diff
@@ -0,0 +1,200 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
```
pq: Perhaps bump to `2018`?

example/all.yaml
srawlins: Yeah I'm happy to rename or leave it.  I think i named it after something like `empty_statements` and `invariant_booleans`.

Also, I would not be against a wide-reaching rename on all the lints after establishing a policy, and providing aliases after the fact. (So after renames, `empty_statements` would be an alias for `avoid_empty_statements`.

PR #428, opened 2017-02-18
Extend rule: directives_ordering with package: imports before relative. 

JPaulsen:
In this second commit we verify that all package: imports go before relative. 

From https://www.dartlang.org/guides/language/effective-dart/style#do-place-package-imports-before-relative-imports

### DIFF
--- a/lib/src/rules/directives_ordering.dart
+++ b/lib/src/rules/directives_ordering.dart
@@ -9,12 +9,14 @@ import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
 import 'package:linter/src/analyzer.dart';
 
-const _dartImportGoFirst = r"Place 'dart:' imports before other imports.";
-
 const _desc = r'Adhere to Effective Dart Guide directives sorting conventions.';
+const _dartImportGoFirst = r"Place 'dart:' imports before other imports.";
+const _packageImportBeforeRelative =
+    r"Place 'package:' imports before relative imports.";
+const _details =
+    r'''**DO** follow the conventions in the [Effective Dart Guide](https://www.dartlang.org/guides/language/effective-dart/style#ordering)
 
-const _details = r'''**DO** follow the conventions in the [Effective Dart Guide]
-(https://www.dartlang.org/guides/language/effective-dart/style#ordering)
+**DO** place “dart:” imports before other imports.
 
 **BAD:**
 ```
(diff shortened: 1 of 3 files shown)


lib/src/rules/directives_ordering.dart
```diff
@@ -83,10 +124,24 @@ class _Visitor extends SimpleAstVisitor {
...
   }
+
+  void _checkPackageImportBeforeRelative(CompilationUnit node) {
+    node.directives
+        .where(_isImportDirective)
+        .where(_isNotDartImport)
+        .skipWhile(_isAbsoluteImport)
+        .where(_isPackageImport)
+        .forEach(rule._reportLintWithPackageImportBeforeRelativeMessage);
+  }
```
bwilkerson: While the guideline isn't worded this way, I believe the intent is that relative imports should be last. The current implementation is fine, but it might be interesting to consider inverting the logic and run it over the directives in reverse. More specifically, starting at the end, skip over relative imports until you get to a non-relative import, then flag any relative imports you find after that.

I suspect that this would produce fewer lints, and hence provide a better UX.

Of course, this is all moot if you go with my earlier suggestion and write the whole rule as a state machine.

lib/src/rules/directives_ordering.dart
```diff
@@ -83,10 +124,24 @@ class _Visitor extends SimpleAstVisitor {
 
   @override
   void visitCompilationUnit(CompilationUnit node) {
+    _checkDartImportGoFirst(node);
+    _checkPackageImportBeforeRelative(node);
+  }
```
bwilkerson: You might consider taking a different approach here and write the whole rule as a state machine that could make a single pass over the directives. This approach might work fine, but it also might be difficult to prevent generating extraneous lints after you've added logic about how to order `package:` URIs and the sort order logic.

Alternatively, you could take a two pass approach where the first pass collects imports into groups, flagging those that appear in the wrong group, and then a second pass that flags issues related to ordering within each group.

As a bigger question, would the UX be better if we generated a single lint when the imports are out of order that pointed users to the guidelines, rather than producing potentially large numbers of lints? Maybe most users won't see very many lints and it is better to have more specific error messages; I honestly don't know.

Have you run this lint over any existing code bases to see what kind of output it produces? If not, we should do that before enabling the rule in general.

test/_data/directives_ordering/package_imports_before_relative/bad.dart
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:math';
+import 'package:ansicolor/ansicolor.dart';  // OK
+import 'dummy.dart';
+import 'package:async/async.dart';  // LINT
+import 'package:yaml/yaml.dart';  // LINT
```
JPaulsen: Fixed, it didn't produce it

lib/src/rules/directives_ordering.dart
bwilkerson: I'm not a UX expert, but I'll share my intuition.

I think that in general, that kind of behavior tends to be confusing to users. Users, reasonably, tend to assume that if an error appears after they type that it had something to do with what they typed. Consider
```
class Car {}
Car c;
```
If a user types an `s` after `Car` on the last line, they will get an error and will rightly assume that the character they typed caused the error. But if a user moves an import from one place to another in order to fix an error, and then another error appears, they are likely to assume that (a) the change caused the second error and they need to find a different solution or (b) that the error didn't go away (if they don't notice that it really is a different error).

There is an exception if it's clear that the first error was masking the new errors (such as adding a type to a variable to fix one error that causes other errors in other locations to appear). I'm not sure whether the import case would be an exception, but I suspect that it wouldn't because I can't think of the logical dependency that they would have to assume.

But we might be able to get some concrete data to use to resolve the question by running the rule over some existing code. Data tends to be a good way to resolve questions.

JPaulsen: Oh nice, I understood, I'm on it

lib/src/rules/directives_ordering.dart
JPaulsen: It is true, but more generally, is the idea of which rule has more priority than others, and in that way, we think that the order in the guidelines is the better

lib/src/rules/directives_ordering.dart
JPaulsen: We have the messages all together in one block, so we think it will be fine, the user will be able to fix it with just one command, the only problem, is when we implement the rest of the parts, one line can have more than one lint, and in that way, for the next parts we want to implement it with a Set, the idea is with the guideline priority we won't mark a line more than once

bwilkerson: Ok. Then let's proceed and see how it works once you're done.

PR #430, opened 2017-02-19
Expect unawaited futures in test.expect (attempt to fix #419)

### LINKED ISSUE #419: Spurious `unawaited_futures` warnings
**Steps to Reproduce**
  1. `git clone git@github.com:google/file.dart.git`
  2. `cd file.dart`
  3. `dartanalyzer .`

**Expected Results**
No analyzer warnings.  At least, until very recently, there were no warnings.

**Actual Results**
https://gist.github.com/tvolkert/4ae91439e738207d3b566e758e7b7680

What's very strange is that many of the lines it complains about don't even have futures at all, like https://github.com/google/file.dart/blob/master/test/replay_test.dart#L57.  Other lines do have futures and are of the form `expect(futureValue, throwsFoo)`, which have never produced any lint warnings before.

ochafik:
Note: I couldn't find an easy way to test the package origin from the tests themselves, suggestions welcome! Also, not sure if that will work with summaries (do we still have sourcefiles then?), etc... Pointers as to how to test would be welcome!

### DIFF
--- a/lib/src/rules/unawaited_futures.dart
+++ b/lib/src/rules/unawaited_futures.dart
@@ -62,6 +62,7 @@ class Visitor extends SimpleAstVisitor {
     if (type?.isDartAsyncFuture == true) {
       // Ignore a couple of special known cases.
       if (_isFutureDelayedInstanceCreationWithComputation(expr) ||
+          _isTestExpectCall(expr) ||
           _isMapPutIfAbsentInvocation(expr)) {
         return;
       }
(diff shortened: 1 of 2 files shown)


zoechi:
An alternative would be https://github.com/dart-lang/linter/issues/419#issuecomment-280173964 to allow APIs themselves to opt out instead of special-casing in the linter rule.

pq:
> An alternative would be #419 (comment) to allow APIs themselves to opt out instead of special-casing in the linter rule.

@zoechi 👍  --- this is something we're actively discussing.  I think this is a good step in the meantime though.

Thanks @ochafik !

lib/src/rules/unawaited_futures.dart
```diff
@@ -93,6 +105,16 @@ class Visitor extends SimpleAstVisitor {
...
       e is ClassElement && e.name == 'Map' && e.library?.name == 'dart.core';
 
+  bool _isTestExpectFunction(Element e) {
+    if (e is! FunctionElement || e.name != 'expect') return false;
+
+    final libUri = e.library?.source?.uri?.toString();
+    if (libUri == null) return false;
+
+    return libUri.startsWith('package:test/') ||
+        libUri.startsWith('package:unittest/');
```
pq: Do we need the `unittest` check here?

pq:
@kevmoo : is that conversation on an issue somewhere?  If not, could you post status somewhere we can follow?

pq:
Thanks @kevmoo .

I think as of https://github.com/dart-lang/test/pull/546, this becomes moot and @ochafik you can safely close this?

ochafik:
Cool, closing this, thanks!

lib/src/rules/unawaited_futures.dart
ochafik: (not that it matters anymore, but fyi most our tests are still using unittest :-))

PR #433, opened 2017-02-21
Third party package imports before own

JPaulsen:
Extend rule: directives_ordering

In this third commit we verify that all third-party package: imports go before own. From https://www.dartlang.org/guides/language/effective-dart/style#prefer-placing-third-party-package-imports-before-other-imports

Also take care of linting each node at most once.

### DIFF
--- a/lib/src/rules/directives_ordering.dart
+++ b/lib/src/rules/directives_ordering.dart
@@ -9,10 +9,8 @@ import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
 import 'package:linter/src/analyzer.dart';
 
-const _desc = r'Adhere to Effective Dart Guide directives sorting conventions.';
 const _dartImportGoFirst = r"Place 'dart:' imports before other imports.";
-const _packageImportBeforeRelative =
-    r"Place 'package:' imports before relative imports.";
+const _desc = r'Adhere to Effective Dart Guide directives sorting conventions.';
 const _details =
     r'''**DO** follow the conventions in the [Effective Dart Guide](https://www.dartlang.org/guides/language/effective-dart/style#ordering)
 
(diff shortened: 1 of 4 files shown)


lib/src/rules/directives_ordering.dart
```diff
@@ -110,22 +143,46 @@ class DirectivesOrdering extends LintRule {
     if (node == null) {
       return;
     }
+    if (_visitor.lintedNodes.contains(node)) {
+      return;
+    }
+    _visitor.lintedNodes.add(node);
```
bwilkerson: `Set.add` returns `true` if the set was modified (if the argument was not previously an element, so it's more efficient to write
```
if (_visitor.lintedNodes.add(node)) {
  reporter.report...
}
```

lib/src/rules/directives_ordering.dart
```diff
@@ -110,22 +143,46 @@ class DirectivesOrdering extends LintRule {
...
+}
+
+class _PackageBox {
+  String _packageName;
+  _PackageBox(this._packageName);
+
+  bool _isNotOurPackageImport(Directive node) => !_isOurPackageImport(node);
+
+  bool _isOurPackageImport(Directive node) =>
+      (node as ImportDirective).uriContent.startsWith('package:$_packageName/');
```
bwilkerson: If you change the parameter type to ImportDirective you won't need the cast.

JPaulsen: Fixed

lib/src/rules/directives_ordering.dart
JPaulsen: Fixed

lib/src/rules/directives_ordering.dart
```diff
@@ -110,38 +141,74 @@ class DirectivesOrdering extends LintRule {
...
     _reportLintWithDescription(node, _packageImportBeforeRelative);
   }
+
+  void _reportLintWithThirdPartyPackageImportBeforeOwnMessage(AstNode node) {
+    _reportLintWithDescription(node, _thirdPartyPackageImportBeforeOwn);
+  }
+}
+
+class _PackageBox {
+  String _packageName;
```
alexeieleusis: nit: Fix this lint "Private field could be final".

lib/src/rules/directives_ordering.dart
```diff
@@ -110,38 +141,74 @@ class DirectivesOrdering extends LintRule {
...
+        .where(_isPackageImport)
+        .skipWhile(box._isNotOwnPackageImport)
+        .where(box._isNotOwnPackageImport)
+        .forEach(rule._reportLintWithThirdPartyPackageImportBeforeOwnMessage);
+  }
+
+  Iterable<ImportDirective> _getImportDirectives(CompilationUnit node) =>
+      node.directives
+          .where(_isImportDirective)
+          .map((e) => e as ImportDirective);
```
alexeieleusis: This map is not necessary.

JPaulsen: It was, do you remember when tests failed? I changed that and it worked

lib/src/rules/directives_ordering.dart
JPaulsen: Fixed

JPaulsen:
We did some refactoring.

We removed the instance field lintedNodes and made it a local variable, PTAL

Issue #154, opened 2015-11-16
Dart project not getting set when lints are run from within DAS.


pq:
For example, `package_prefixed_library_names` here: https://github.com/dart-lang/sdk/issues/24947.

devoncarew:
Is this still an issue?

devoncarew:
I see it referenced from the `package_prefixed_library_names.dart` file.

pq:
> Is this still an issue?

I believe so.  It looks like we're only initializing project instances here:

https://github.com/dart-lang/sdk/blob/bd3792e7142a6410f3b1fc1289769f14f7ea58cd/pkg/analyzer/lib/src/lint/analysis.dart#L210-L215

```dart
    DartProject project = await DartProject.create(analysisDriver, sources);
    Registry.ruleRegistry.forEach((lint) {
      if (lint is ProjectVisitor) {
        (lint as ProjectVisitor).visit(project);
      }
    });
```

pq:
Also related: https://github.com/dart-lang/linter/issues/1786.

The rub is our notion of "project" hinged on the presence of a pubspec.  That obviously doesn't translate to google3.  We're part of the way there to a better solution where `LinterContext` will surface the workspace `package` which we can use to identify package/project roots.

For that, see: https://dart-review.googlesource.com/c/sdk/+/123341

When we have a new analyzer roll (cc @stereotype441), we can start playing around with that.

pq:
Picked up in #3395.

PR #437, opened 2017-02-23
Extend rule: directives_ordering with directive sections sorted alphabetically

JPaulsen:
In this fifth commit we verify that each directive section must be sorted alphabetically. From https://www.dartlang.org/guides/language/effective-dart/style#do-sort-sections-alphabetically

### DIFF
--- a/lib/src/rules/directives_ordering.dart
+++ b/lib/src/rules/directives_ordering.dart
@@ -89,8 +89,48 @@ import 'package:foo/foo.dart';  // OK
 
 import 'package:myapp/io.dart';
 import 'package:myapp/util.dart';
+```
+
+**DO** specify exports in a separate section after all imports.
+
+**BAD:**
+```
+import 'src/error.dart';
+export 'src/error.dart'; // LINT
+import 'src/string_source.dart';
+```
+
+**GOOD:**
+```
+import 'src/error.dart';
+import 'src/string_source.dart';
+
+export 'src/error.dart'; // OK
+```
+
+**DO** sort sections alphabetically.
+
+**BAD:**
+```
+import 'package:foo/bar.dart'; // OK
+import 'package:bar/bar.dart'; // LINT
+
+import 'a/b.dart'; // OK
+import 'a.dart'; // LINT
+```
+
+**GOOD:**
+```
+import 'package:bar/bar.dart'; // OK
+import 'package:foo/bar.dart'; // OK
+
+import 'a.dart'; // OK
+import 'a/b.dart'; // OK
 
 ''';
+
+const _directiveSectionOrderedAlphabetically =
+    r"Sort directive sections alphabetically.";
 
 const _exportDirectiveAfterImportDirectives =
     r"Specify exports in a separate section after all imports.";
@@ -113,6 +153,8 @@ bool _isNotDartImport(ImportDirective node) => !_isDartImport(node);
 
 bool _isPackageImport(ImportDirective node) =>
     node.uriContent.startsWith("package:");
+
+bool _isRelativeImport(ImportDirective node) => !_isAbsoluteImport(node);
 
 class DirectivesOrdering extends LintRule implements ProjectVisitor {
   _Visitor _visitor;
@@ -146,6 +188,11 @@ class DirectivesOrdering extends LintRule implements ProjectVisitor {
     reporter.reportErrorForNode(new LintCode(name, description), node, []);
   }
 
+  void _reportLintWithDirectiveSectionOrderedAlphabeticallyMessage(
+      AstNode node) {
+    _reportLintWithDescription(node, _directiveSectionOrderedAlphabetically);
+  }
+
   void _reportLintWithExportDirectiveAfterImportDirectiveMessage(AstNode node) {
     _reportLintWithDescription(node, _exportDirectiveAfterImportDirectives);
   }
@@ -164,7 +211,7 @@ class _PackageBox {
   _PackageBox(this._packageName);
 
   bool _isNotOwnPackageImport(ImportDirective node) =>
-      !_isOwnPackageImport(node);
+      _isPackageImport(node) && !_isOwnPackageImport(node);
 
   bool _isOwnPackageImport(ImportDirective node) =>
       node.uriContent.startsWith('package:$_packageName/');
@@ -184,6 +231,7 @@ class _Visitor extends SimpleAstVisitor {
     _checkPackageImportBeforeRelative(lintedNodes, node);
     _checkThirdPartyImportBeforeOwn(lintedNodes, node);
     _checkExportDirectiveAfterImportDirective(lintedNodes, node);
+    _checkDirectiveSectionOrderedAlphabetically(lintedNodes, node);
   }
 
   void _checkDartImportGoFirst(Set<AstNode> lintedNodes, CompilationUnit node) {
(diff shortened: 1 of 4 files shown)


lib/src/rules/directives_ordering.dart
```diff
@@ -199,6 +247,31 @@ class _Visitor extends SimpleAstVisitor {
...
+    final exportDirectives = _getExportDirectives(node);
+
+    final dartImports = importDirectives.where(_isDartImport);
+    final relativeImports = importDirectives.where(_isRelativeImport);
+
+    _checkSectionInOrder(lintedNodes, dartImports);
+    _checkSectionInOrder(lintedNodes, relativeImports);
+    _checkSectionInOrder(lintedNodes, exportDirectives);
+
+    if (project != null) {
```
bwilkerson: Can the project actually ever be null? (I honestly don't know, so this isn't a trick question.) If it can, should we check the order of all `package:` imports, ignoring the distinction between "own" and "third party", or would that sometimes lead to cases where we report lints in code that is actually correctly ordered?

lib/src/rules/directives_ordering.dart
```diff
@@ -229,6 +302,31 @@ class _Visitor extends SimpleAstVisitor {
...
+      }
+
+      return acc;
+    }
+
+    final inOrderNodes = nodes.fold(<NamespaceDirective>[], addIfUpperBound);
+
+    nodes.toSet()
+      ..removeAll(inOrderNodes)
+      ..forEach(reportDirective);
```
bwilkerson: It would be more efficient to reverse the selection choice in `addIfUpperBound` (so that it adds directives that are not in order). That way you wouldn't need to convert `nodes` to a set, nor remove items from the set.

lib/src/rules/directives_ordering.dart
```diff
@@ -249,6 +347,11 @@ class _Visitor extends SimpleAstVisitor {
         .forEach(reportDirective);
   }
 
+  Iterable<ExportDirective> _getExportDirectives(CompilationUnit node) =>
+      node.directives
+          .where(_isExportDirective)
+          .map((e) => e as ExportDirective);
```
bwilkerson: Do we need this `map` operation? If so, do we also need it in `_getImportDirectives`?

test/_data/directives_ordering/sort_directive_sections_alphabetically/bad.dart
```diff
@@ -0,0 +1,28 @@
...
+import 'package:linter/src/rules.dart'; // OK
+
+import 'dummy1.dart'; // OK
+import 'dummy3.dart'; // OK
+import 'dummy2.dart'; // LINT
+import 'dummy4.dart'; // OK
+
+export 'dummy1.dart'; // OK
+export 'dummy3.dart'; // OK
+export 'dummy2.dart'; // LINT
```
bwilkerson: It would be good to test more cases, such as having the out of order directive appear at the beginning of the group or at the end of the group.

lib/src/rules/directives_ordering.dart
JPaulsen: I copied from another lint who has this issue attached https://github.com/dart-lang/linter/issues/154

lib/src/rules/directives_ordering.dart
JPaulsen: You are all right, it would be better, but we couldn't figure it out of how to get the notInOrderNodes with a simple and readable fold.

To know if a directive node is in order, it is necessary to compare it with all nodes that are before it, if its uri content is higher than all of them then it is in order, so we needed to know which node is the maximum node before a node to know if it is in order or not, to do that it was easier to keep all the nodes in order with a ford and compare it just with the last one.

PTAL to the method addIfUpperBound

test/_data/directives_ordering/sort_directive_sections_alphabetically/bad.dart
JPaulsen: Ok, but the rule says that a node is in order if its uriContent is higher alphabetically than each node before it in the group, so it can't be a not-in-order node at the beginning of a group.

Example:

```
import 'z/dart'; // OK because is higher than all nodes before it (there aren't nodes before this)
import 'a/dart'; // LINT because is not higher than z.dart
import 'b/dart'; // LINT because is not higher than z.dart
import 'c/dart'; // LINT because is not higher than z.dart
```

I will test more cases

lib/src/rules/directives_ordering.dart
bwilkerson: Not everything needs to be done with a `fold` :-)

How about a simple loop:
```
NamespaceDirective previousDirective = null;
for (NamespaceDirective directive in nodes) {
  if (previousDirective == null || previousDirective.uriContent.compareTo(directive.uriContent) < 0) {
    previousDirective = directive;
  } else {
    reportDirective(directive);
  }
}
```
It isn't as terse, but it's a lot more performant.

Also, if directive d<sub>i</sub> should be before d<sub>j</sub> when i < j, and the directives being checked are in the order d<sub>1</sub>, d<sub>5</sub>, d<sub>2</sub>, d<sub>3</sub>, d<sub>4</sub>, then this algorithm will produce three lints. But if you rewrite the loop to:
```
NamespaceDirective previousDirective = null;
for (NamespaceDirective directive in nodes) {
  if (previousDirective != null && previousDirective.uriContent.compareTo(directive.uriContent) > 0) {
    reportDirective(directive);
  }
  previousDirective = directive;
}
```
then it will only produce one lint. I'd want to try it on some real data before concluding that this is better, but it's worth considering.

alexeieleusis:
@a14n this false positive is in the code already in master or in the PR?

a14n:
@alexeieleusis it seems to be in master too

JPaulsen:
@a14n fixed

lib/src/rules/directives_ordering.dart
JPaulsen: In that case, the last code would mark d2, and it is more intuitive to mark d5 because it is the one you have to move, it could be done changing the code for this:

```
NamespaceDirective previousDirective = null;
for (NamespaceDirective directive in nodes) {
  if (previousDirective != null && previousDirective.uriContent.compareTo(directive.uriContent) > 0) {
    reportDirective(previousDirective);
  }
  previousDirective = directive;
}
```
but in the case you have d7,d1,d3,d4,d2,d5,d6  the second code would lint d7 and d4, d7 is OK because you have to move it, but d4 not, you should move d2 to fix that, so I think a good way to reduce lints and give a better hint is mark both, with this code.

```
NamespaceDirective previousDirective = null;
for (NamespaceDirective directive in nodes) {
  if (previousDirective != null && previousDirective.uriContent.compareTo(directive.uriContent) > 0) {
    reportDirective(directive);
    reportDirective(previousDirective);
  }
  previousDirective = directive;
}
```
In that case the user would see this:

import 'd7.dart'; // LINT
import 'd1.dart'; // LINT
import 'd3';
import 'd4.dart'; // LINT
import 'd2.dart'; // LINT
import 'd5';
import 'd6';

I think that is a pretty good solution,other one could be  calculate the nodes you have to move with minimal cost of sortering a block moving directives from its line to another then mark them, but it is more sophisticated and hard to implement, what do you think?

lib/src/rules/directives_ordering.dart
JPaulsen: I think so, If I delete it and run $pub run test, A lot of tests start to fail and I got this error message:

```
Caused by type 'WhereIterable<Directive>' is not a subtype of type 'Iterable<ExportDirective>' of 'function result' where
  WhereIterable is from dart:_internal
  Directive is from package:analyzer/dart/ast/ast.dart
  Iterable is from dart:core
  ExportDirective is from package:analyzer/dart/ast/ast.dart
```

test/_data/directives_ordering/sort_directive_sections_alphabetically/bad.dart
JPaulsen: Fixed

lib/src/rules/directives_ordering.dart
JPaulsen: Yes, seeing the issue, I think we would lead to cases where we report lints in code that is actually correctly ordered.

lib/src/rules/directives_ordering.dart
JPaulsen: Fixed, I changed it for the first for you proposed

bwilkerson: > I think that is a pretty good solution,other one could be calculate the nodes you have to move with minimal cost of sortering a block moving directives from its line to another then mark them, but it is more sophisticated and hard to implement, what do you think?

Yeah, I considered more complex algorithms (such as implementing `diff`), but I don't think the complexity of the implementation is warranted by any possible improvement in the UX.

I don't know for sure, but I'm still thinking it might be easier for users to be told once that the imports are not ordered well and then let them use a quick fix (or similar auto sorting technique) to fix the problem.

That said, I'm happy to let you continue as you're going and see what users actually think when they try using it. I could easily be wrong.

test/_data/directives_ordering/sort_directive_sections_alphabetically/bad.dart
JPaulsen: Fixed

PR #445, opened 2017-02-27
New rule: adjacent_strings_to_concatenate_literals

JPaulsen:
In this commit we implement this rule from: https://www.dartlang.org/guides/language/effective-dart/usage#do-use-adjacent-strings-to-concatenate-string-literals

### DIFF
--- a/lib/src/rules.dart
+++ b/lib/src/rules.dart
@@ -66,6 +66,7 @@ import 'package:linter/src/rules/unnecessary_null_aware_assignments.dart';
 import 'package:linter/src/rules/unnecessary_null_in_if_null_operators.dart';
 import 'package:linter/src/rules/unrelated_type_equality_checks.dart';
 import 'package:linter/src/rules/valid_regexps.dart';
+import 'package:linter/src/rules/use_adjacent_strings_to_concatenate_literals.dart';
 
 void registerLintRules() {
   Analyzer.facade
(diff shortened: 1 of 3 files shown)


lib/src/rules/adjacent_strings_to_concatenate_literals.dart
```diff
@@ -0,0 +1,60 @@
...
+    'parts are overrun by martians. Unclear which are which.');
+```
+
+''';
+
+class AdjacentStringsToConcatenateLiterals extends LintRule {
+  _Visitor _visitor;
+  AdjacentStringsToConcatenateLiterals()
+      : super(
+            name: 'adjacent_strings_to_concatenate_literals',
```
bwilkerson: I know the name is already long, but perhaps prefix the name with "use_"?

test/rules/adjacent_strings_to_concatenate_literals.dart
```diff
@@ -0,0 +1,16 @@
...
+
+main() {
+  String string1 = 'hola means' + // LINT
+      ' hello in spanish';
+
+  String string2 = 'hola means' // OK
+      ' hello in spanish';
+
+  List<String> list = ['this is' + // LINT
+    ' not allowed'];
```
bwilkerson: Perhaps add a test (or two) with a non-literal argument:

String prefix = 'this is';
String string3 = prefix + ' perfectly fine';

lib/src/rules/adjacent_strings_to_concatenate_literals.dart
JPaulsen: Fixed

test/rules/adjacent_strings_to_concatenate_literals.dart
JPaulsen: Fixed

PR #446, opened 2017-02-27
Rule name cleanup (pre-publish).

pq:
Some rule name cleanup to stay (more) consistent with existing rules.

@bwilkerson 

FYI @alexeieleusis @JPaulsen

### DIFF
--- a/lib/src/rules.dart
+++ b/lib/src/rules.dart
@@ -50,7 +50,7 @@ import 'package:linter/src/rules/prefer_is_empty.dart';
 import 'package:linter/src/rules/prefer_is_not_empty.dart';
 import 'package:linter/src/rules/pub/package_names.dart';
 import 'package:linter/src/rules/public_member_api_docs.dart';
-import 'package:linter/src/rules/recursive_getter.dart';
+import 'package:linter/src/rules/recursive_getters.dart';
 import 'package:linter/src/rules/slash_for_doc_comments.dart';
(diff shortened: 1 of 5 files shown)


Hixie:
@pq Were these rule changes announced anywhere?

pq:
@Hixie the hope is that these were renamed before anyone got to using them but in general we should have a process in place (a la the conversation around Flutter breaking changes).  Maybe we should create a linter-announce alias for this kind of thing (and generally for releases)?

Hixie:
We use at least one of them! :-) (unnecessary_brace_in_string_interp)

I don't mind if they're not announced or anything. I just want to make sure we keep up to date. I'm not sure I would have noticed except that I happened to be crawling in the list and was lucky.

My ideal scenario would be for there to be a way to run an .analysis_options file through some script and have it tell me all the lints that are omitted and all the listed lints that are unknown (errors), ideally with a way for it to skip omitted lints that are present but commented out or some such. Basically I want to make sure that we aren't using bogus lint names and that we have made an explicit decision on every lint that does exist.

pq:
Thanks @Hixie (and mea culpa!)  I agree that some kind of validation support would be great and as it happens I added some rudimentary support to the server a while back but it never made it into the CLI.   I'll fix that soon though:

https://github.com/dart-lang/sdk/issues/29317

Cheers!

PR #451, opened 2017-02-28
New rule: prefer_interpolation_to_compose_strings

JPaulsen:
In this commit we implement this rule from: https://www.dartlang.org/guides/language/effective-dart/usage#prefer-using-interpolation-to-compose-strings-and-values

Fixes https://github.com/dart-lang/linter/issues/448

### DIFF
--- a/lib/src/rules/prefer_interpolation_to_compose_strings.dart
+++ b/lib/src/rules/prefer_interpolation_to_compose_strings.dart
@@ -1,0 +1,90 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+library linter.src.rules.prefer_interpolation_to_compose_strings;
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/token.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:linter/src/analyzer.dart';
+import 'package:linter/src/util/dart_type_utilities.dart';
+
+const _desc = r'Use interpolation to compose strings and values.';
+
+const _details = r'''
+
+**PREFER** using interpolation to compose strings and values.
+
+**BAD:**
+```
+'Hello, ' + name + '! You are ' + (year - birth) + ' years old.';
+```
+
+**GOOD:**
+```
+'Hello, $name! You are ${year - birth} years old.';
+```
+
+''';
+
+class PreferInterpolationToComposeStrings extends LintRule {
+  _Visitor _visitor;
+  PreferInterpolationToComposeStrings()
+      : super(
+            name: 'prefer_interpolation_to_compose_strings',
+            description: _desc,
+            details: _details,
+            group: Group.style) {
+    _visitor = new _Visitor(this);
+  }
+
+  @override
+  AstVisitor getVisitor() => _visitor;
+}
+
+class _Visitor extends SimpleAstVisitor {
+  final LintRule rule;
+  _Visitor(this.rule);
+
+  @override
+  visitCompilationUnit(CompilationUnit node) {
+    final skippedNodes = new Set<AstNode>();
+    void checkRule(AstNode node) {
+      if (skippedNodes.contains(node)) {
+        return;
+      }
+      if (node is BinaryExpression) {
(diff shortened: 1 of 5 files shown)


test/rules/prefer_interpolation_to_compose_strings.dart
```diff
@@ -0,0 +1,17 @@
...
+void main() {
+  String name = 'Juan'+'Pablo'; // OK // It is not the business of this rule
+  int year = 2017;
+  int birth = 1993;
+  'Hello, $name! You are ${year - birth} years old.';
+  'Hello, ' + // LINT
+      name + // LINT
+      '! You are ' + // LINT
+      (year - birth).toString() + // LINT
+      ' years old.';
```
bwilkerson: Wow! That's a lot of lints for a single expression. Perhaps ignore lints in subexpressions.

test/rules/prefer_interpolation_to_compose_strings.dart
```diff
@@ -0,0 +1,17 @@
...
+}
```
bwilkerson: Should code like the following produce a lint?
```
String pad = '';
for (int i = 0; i < width; i++) {
  pad = pad + ' ';
}
```
It seems like a fairly common idiom, and I don't think `pad = '$pad ';` is easier to read.

What about code like this?
```
string += 'x';
```

test/rules/prefer_interpolation_to_compose_strings.dart
JPaulsen: It is linting the + operator, with Alexei we think it is doing it right

test/rules/prefer_interpolation_to_compose_strings.dart
JPaulsen: FIXED String += 'x';

JPaulsen: In the case:

```
String pad = '';
for (int i = 0; i < width; i++) {
  pad = pad + ' ';
}
```

The programmer should use a StringBuffer, and yes it is linted for this rule (It was added to the test cases)

lib/src/rules/prefer_interpolation_to_compose_strings.dart
```diff
@@ -0,0 +1,71 @@
...
+
+class _Visitor extends SimpleAstVisitor {
+  final LintRule rule;
+  _Visitor(this.rule);
+
+  @override
+  visitAssignmentExpression(AssignmentExpression node) {
+    if (node.operator.type == TokenType.PLUS_EQ &&
+        (node.leftHandSide.bestType.name == 'String' ||
+            node.rightHandSide.bestType.name == 'String')) {
```
bwilkerson: After you've added it, consider using `DartTypeUtilities.isClass(node.bestType, 'String', 'dart.core')` to reduce the chances of false positives.

test/rules/prefer_interpolation_to_compose_strings.dart
bwilkerson: Sorry, but I don't agree. One of the biggest UX issues for users of compilers (and linters) is having multiple errors when a single error would suffice. While analyzer fails in many cases, minimizing the number of unnecessary diagnostics is an important goal. I think this really needs to be changed.

(Because of the way the analysis server works, it also has other implications, such as causing duplicated quick fixes to be displayed in the IDE, causing other UX problems.)

lib/src/rules/prefer_interpolation_to_compose_strings.dart
JPaulsen: Yes!, that is much better! thanks

JPaulsen:
@bwilkerson we have 2 options to implement this, we can visit a CompilationUnit, and then traverse the nodes in DFS to be sure to not lint the children of linted nodes, the other option is use the visitor like now, but in quadratic order with respect to the nodes in a linted expression, if the visitor visits nodes in DFS order we can do an optimization, can we assume that?

test/rules/prefer_interpolation_to_compose_strings.dart
JPaulsen: Fixed, we changed it so it lints the whole expression only once

lib/src/rules/prefer_interpolation_to_compose_strings.dart
JPaulsen: Fixed

PR #456, opened 2017-02-28
Use collection literals when possible

JPaulsen:
Fixes https://github.com/dart-lang/linter/issues/455


lib/src/rules.dart
```diff
@@ -46,6 +46,7 @@ import 'package:linter/src/rules/prefer_const_constructors.dart';
 import 'package:linter/src/rules/prefer_contains.dart';
 import 'package:linter/src/rules/prefer_final_fields.dart';
 import 'package:linter/src/rules/prefer_final_locals.dart';
+import 'package:linter/src/rules/prefer_interpolation_to_compose_strings.dart';
```
bwilkerson: It appears that you've mixed in changes from an earlier PR. Please separate them.

JPaulsen: Fixed

JPaulsen:
The Integration Test was necessary to import 'dart:collection" and use new LinkedHashMap() how the [rule](https://www.dartlang.org/guides/language/effective-dart/usage#do-use-collection-literals-when-possible) says

PR #458, opened 2017-03-01
New rule: avoid_function_literal_in_foreach_method

JPaulsen:
Fixes https://github.com/dart-lang/linter/issues/457

### DIFF
--- a/lib/src/rules/avoid_function_literal_in_foreach_method.dart
+++ b/lib/src/rules/avoid_function_literal_in_foreach_method.dart
@@ -1,0 +1,71 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+library linter.src.rules.avoid_function_literal_in_foreach_method;
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:linter/src/analyzer.dart';
+import 'package:linter/src/util/dart_type_utilities.dart';
+
+const _desc = r'Avoid using Iterable.forEach() with a function literal.';
+
+const _details = r'''
+
+**AVOID** using Iterable.forEach() with a function literal.
+
+**BAD:**
+```
+people.forEach((person) {
+  ...
+});
+```
+
+**GOOD:**
+```
+for (var person in people) {
+  ...
+}
+```
+
+**GOOD:**
+```
+for (var person in people) {
+  ...
+}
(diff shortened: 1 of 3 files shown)


lib/src/rules/avoid_function_literal_in_foreach_method.dart
```diff
@@ -0,0 +1,69 @@
...
+}
+
+class _Visitor extends SimpleAstVisitor {
+  final LintRule rule;
+  _Visitor(this.rule);
+
+  @override
+  visitMethodInvocation(MethodInvocation node) {
+    if (node.target != null &&
+      DartTypeUtilities.implementsInterface(node.target.bestType, 'Iterable', 'dart.core') &&
```
bwilkerson: The type check is the most expensive part of the condition, and hence should be moved to the end.

JPaulsen: Fixed

test/rules/avoid_function_literal_in_foreach_method.dart
```diff
@@ -0,0 +1,18 @@
...
+
+// test w/ `pub run test -N avoid_function_literal_in_foreach_method`
+
+void main() {
+  Iterable<String> people;
+
+  for (var person in people) { // OK
+    print(person);
+  }
+  people.forEach((person) { // LINT
```
alexeieleusis: Should or should it not lint lambdas?

alexeieleusis: Given they are concise, I would say no:
```dart
people.forEach((p) => foo(p, 0)); // OK

void foo(String bar, int baz) {}
```

Turning this into a for loop would turn one line into three. Thoughts @bwilkerson

bwilkerson: While I disagree with the style guide's recommendation (https://www.dartlang.org/guides/language/effective-dart/usage#avoid-using-iterableforeach-with-a-function-literal), the way I read it there is no distinction made on the basis of the form of the function literal. Hence, the lint, as written, conforms to the letter of the guideline (I can't speak to the spirit).

One could defeat the lint by converting the code into a `for` loop,
```
for (var p in people) {
  foo(p, 0);
}
```
by extracting the function into a local variable,
```
var f = (p) => foo(p, 0);
people.forEach(f);
```
or writing a local function
```
f(p) => foo(p, 0);
people.forEach(f);
```
but I agree that all three options seem counterproductive. (But that might be because, as I said, I don't agree with the guideline.)

alexeieleusis: Acknowledged, thanks.

PR #460, opened 2017-03-01
New rule: use_function_declaration_instead_of_variable

JPaulsen:
Fixes https://github.com/dart-lang/linter/issues/459

### DIFF
--- a/lib/src/rules.dart
+++ b/lib/src/rules.dart
@@ -69,6 +69,7 @@ import 'package:linter/src/rules/unnecessary_null_in_if_null_operators.dart';
 import 'package:linter/src/rules/unrelated_type_equality_checks.dart';
 import 'package:linter/src/rules/use_adjacent_strings_to_concatenate_literals.dart';
 import 'package:linter/src/rules/use_collection_literals_when_possible.dart';
+import 'package:linter/src/rules/use_function_declaration_instead_of_variable.dart';
 import 'package:linter/src/rules/valid_regexps.dart';
 
 void registerLintRules() {
@@ -139,5 +140,6 @@ void registerLintRules() {
     ..register(new UnrelatedTypeEqualityChecks())
     ..register(new UseAdjacentStringsToConcatenateLiterals())
     ..register(new UseCollectionLiteralsWhenPossible())
+    ..register(new UseFunctionDeclarationInsteadOfVariable())
     ..register(new ValidRegExps());
 }
--- a/lib/src/rules/use_function_declaration_instead_of_variable.dart
+++ b/lib/src/rules/use_function_declaration_instead_of_variable.dart
@@ -1,0 +1,62 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+library linter.src.rules.use_function_declaration_instead_variable;
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:linter/src/analyzer.dart';
+
+const _desc = r'Use a function declaration to bind a function to a name.';
+
+const _details = r'''
+
+**DO** use a function declaration to bind a function to a name.
+
+**BAD:**
+```
+void main() {
+  var localFunction = () {
+    ...
+  };
+}
+```
+
+**GOOD:**
+```
+void main() {
+  localFunction() {
+    ...
+  }
+}
+```
+
+''';
+
+class UseFunctionDeclarationInsteadOfVariable extends LintRule {
+  _Visitor _visitor;
+  UseFunctionDeclarationInsteadOfVariable()
+      : super(
+            name: 'use_function_declaration_instead_of_variable',
+            description: _desc,
+            details: _details,
+            group: Group.style) {
+    _visitor = new _Visitor(this);
+  }
+
+  @override
+  AstVisitor getVisitor() => _visitor;
+}
+
+class _Visitor extends SimpleAstVisitor {
+  final LintRule rule;
+  _Visitor(this.rule);
+
+  @override
+  visitVariableDeclaration(VariableDeclaration node) {
+    if (node.initializer != null && node.initializer is FunctionExpression) {
+      rule.reportLint(node);
+    }
+  }
+}
--- a/test/rules/use_function_declaration_instead_of_variable.dart
+++ b/test/rules/use_function_declaration_instead_of_variable.dart
@@ -1,0 +1,27 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// test w/ `pub run test -N use_function_declaration_instead_variable`
+
+void main() {
+  thisIsGood() { // OK
+    print('this is good');
+  }
+
+  var bad1 = () { // LINT
+    print('this is bad');
+  };
+
+  var bad2 = bad1; // OK
+
+  bad2 = bad1; // OK
+
+  bad2 = () { // OK
+    print('this is bad');
+  };
+
+  var bad3 = print; // OK
+
+  bad3 = print; // OK
+}


lib/src/rules/use_function_declaration_instead_variable.dart
```diff
@@ -0,0 +1,70 @@
...
+  AstVisitor getVisitor() => _visitor;
+}
+
+class _Visitor extends SimpleAstVisitor {
+  final LintRule rule;
+  _Visitor(this.rule);
+
+  @override
+  visitAssignmentExpression(AssignmentExpression node) {
+    if (node.rightHandSide.bestType is FunctionType) {
```
JPaulsen: @bwilkerson @alexeieleusis do you know if there is a better way to know if something is a function? 
'''
implementsInterface(node.rightHandSide.bestType, 'Function', 'dart.core')
'''
did not work, and
'''
node.rightHandSide.bestType.isDartFunction
'''
Neither

JPaulsen: @bwilkerson @alexeieleusis do you know a better way to know if a type is a function?

Both of these didn't work
```
DartTypeUtilities.implementsInterface(node.rightHandeSide.bestType, 'Function', 'dart.core');
node.rightHandeSide.bestType.isDartFunction;
```

JPaulsen: Also, in a FunctionType, name = null and element.name = ''

lib/src/rules/use_function_declaration_instead_variable.dart
```diff
@@ -0,0 +1,70 @@
...
+  @override
+  AstVisitor getVisitor() => _visitor;
+}
+
+class _Visitor extends SimpleAstVisitor {
+  final LintRule rule;
+  _Visitor(this.rule);
+
+  @override
+  visitAssignmentExpression(AssignmentExpression node) {
```
bwilkerson: Don't produce a lint for assignments. Code to assign a function to a variable can't always be replaced by a function declaration. Consider
```
class Dispatcher {
  Map<String, Function> callbacks = {};

  void on(String event, Function callback) {
    callbacks[event] = callback; // OK
  }
}
```

lib/src/rules/use_function_declaration_instead_variable.dart
```diff
@@ -0,0 +1,70 @@
...
+    if (node.rightHandSide.bestType is FunctionType) {
+      rule.reportLint(node);
+    }
+  }
+
+  @override
+  visitVariableDeclaration(VariableDeclaration node) {
+    if (node.initializer != null && node.initializer.bestType is FunctionType) {
```
bwilkerson: I don't think you want to test the type of the initializer here at all. I think you want to ask `node.initializer is FunctionExpression`.

Otherwise you'll end up creating a lint for something like
```
void doIt(Function f) {
  var x = f; // OK
  x(this);
}
```

lib/src/rules/use_function_declaration_instead_variable.dart
bwilkerson: You don't actually need to do either, but...

In analyzer, just like in Dart, the types are divided into `InterfaceType`s and `FunctionType`s. If a type is a `FunctionType`, then it's a function, but it's also a function if it's an `InterfaceType` that implements the interface `Function` or defined a method named `call`. The first code snippet does part of that work for you. The second, however, tests to see whether the type is exactly the `InterfaceType` representing the class `Function` (and will return false for subtypes of `Function`).

lib/src/rules/use_function_declaration_instead_variable.dart
JPaulsen: Yes, I intentionally created them, can you please comment the test to know exactly what is right and wrong?

test/rules/use_function_declaration_instead_variable.dart
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// test w/ `pub run test -N use_function_declaration_instead_variable`
+
+void main() {
+  thisIsGood() { // OK
```
bwilkerson: Agreed that this is OK.

test/rules/use_function_declaration_instead_variable.dart
```diff
@@ -0,0 +1,27 @@
...
+    print('this is good');
+  }
+
+  var bad1 = () { // LINT
```
bwilkerson: Agreed that this should produce a lint.

test/rules/use_function_declaration_instead_variable.dart
```diff
@@ -0,0 +1,27 @@
...
+    print('this is bad');
+  };
+
+  var bad2 = bad1; // LINT
```
bwilkerson: I don't think this should be a lint. The only "fix" for this is
```
void bad2() {
  bad1();
}
```
which I don't believe the guideline is advocating.

test/rules/use_function_declaration_instead_variable.dart
```diff
@@ -0,0 +1,27 @@
...
+
+  bad2 = bad1; // LINT
```
bwilkerson: This shouldn't be a lint because there's nothing else that the user can write. You can't have
```
var x = ...;
int x(String s) { ... }
```
because then you'd have two definitions of the same name (the local variable declaration and the local function declaration). If the value of `x` needs to change at this point, an assignment is the only option.

test/rules/use_function_declaration_instead_variable.dart
```diff
@@ -0,0 +1,27 @@
...
+
+  bad2 = () { // LINT
```
bwilkerson: This shouldn't be a lint for the same reason. You can't create lints for assignments because there is no alternative with the same semantics.

test/rules/use_function_declaration_instead_variable.dart
```diff
@@ -0,0 +1,27 @@
...
+    print('this is bad');
+  };
+
+  var bad3 = print; // LINT
```
bwilkerson: Again, I don't think preventing this is the intent of the guideline. Consider
```
var write = print;
if (args['writeToFile']) {
  write = (String s) {
    // write `s` to a file
  }
}
doWork(write);
```
If we define a local function named `write`, then we cannot re-assign it before passing it in somewhere. (Which means, by the way, that the lint should also not be produced when the variable is assigned another value even if the initializer contains a function literal.)

The point of the guideline is that if you're going to define a function, it's easier to read it if it looks like a function definition. The point isn't to prevent users from defining and using function-valued variables.

test/rules/use_function_declaration_instead_variable.dart
```diff
@@ -0,0 +1,27 @@
...
+
+  bad3 = print; // LINT
```
bwilkerson: Again, assignments cannot produce lints.

test/rules/use_function_declaration_instead_variable.dart
JPaulsen: Are you sure about line 20?, for example, I can do this:

```
var bad2;
bad2 = () {
 print('this is bad');
}
```

but not this:
```
var bad1 = () {
 print('this is bad');
}
```

bwilkerson: I believe that's reasonable (that the first does not produce a lint while the second does). Again, the point of the guideline (as I understand it) is *not* to prevent assigning a function to a variable, the point is to encourage code that is easier to read. The hypothesis, which I agree with, is that this
```
void bad1() {
  print('this is good');
}
```
is easier to read than
```
var bad1 = () {
 print('this is bad');
}
```
But I don't think we should generate a lint if the variable is re-assigned somewhere later in the function / method, because then the user really *is* trying to create a variable, not just a function.

To put it another way, the guideline is about understanding user intent and helping them accomplish their goal in a better way. In order to do that we have to understand whether they're trying to define a function or a variable.

lib/src/rules/use_function_declaration_instead_variable.dart
JPaulsen: Fixed

lib/src/rules/use_function_declaration_instead_variable.dart
JPaulsen: Fixed

lib/src/rules/use_function_declaration_instead_variable.dart
JPaulsen: Fixed

test/rules/use_function_declaration_instead_variable.dart
JPaulsen: Fixed

test/rules/use_function_declaration_instead_variable.dart
JPaulsen: Fixed

test/rules/use_function_declaration_instead_variable.dart
JPaulsen: Fixed

test/rules/use_function_declaration_instead_variable.dart
JPaulsen: Fixed

test/rules/use_function_declaration_instead_variable.dart
JPaulsen: Fixed

lib/src/rules/use_function_declaration_instead_variable.dart
```diff
@@ -0,0 +1,62 @@
...
+}
+```
+
+''';
+
+class UseFunctionDeclarationInsteadVariable extends LintRule {
+  _Visitor _visitor;
+  UseFunctionDeclarationInsteadVariable()
+      : super(
+            name: 'use_function_declaration_instead_variable',
```
bwilkerson: Perhaps "use_function_declaration_instead_of_variable".

JPaulsen: Fixed

Issue #461, opened 2017-03-02
Implement CONSIDER omitting the types for local variables.


JPaulsen:
From [Effective Dart](https://www.dartlang.org/guides/language/effective-dart/usage#consider-omitting-the-types-for-local-variables)

PR #462, opened 2017-03-02
New rule: omit_local_variable_types

JPaulsen:
Fixes https://github.com/dart-lang/linter/issues/461

### DIFF
--- a/lib/src/util/dart_type_utilities.dart
+++ b/lib/src/util/dart_type_utilities.dart
@@ -49,7 +49,7 @@ class DartTypeUtilities {
   static bool isClass(DartType type, String className, String library) =>
       type != null &&
       type.name == className &&
-      type.element.library.name == library;
+      type.element?.library?.name == library;
 
   /// Builds the list resulting from traversing the node in DFS and does not
   /// include the node itself, it excludes the nodes for which the exclusion
--- a/lib/src/rules.dart
+++ b/lib/src/rules.dart
@@ -37,6 +37,7 @@ import 'package:linter/src/rules/literal_only_boolean_expressions.dart';
 import 'package:linter/src/rules/no_adjacent_strings_in_list.dart';
 import 'package:linter/src/rules/no_duplicate_case_values.dart';
 import 'package:linter/src/rules/non_constant_identifier_names.dart';
+import 'package:linter/src/rules/omit_local_variable_types.dart';
 import 'package:linter/src/rules/one_member_abstracts.dart';
 import 'package:linter/src/rules/only_throw_errors.dart';
 import 'package:linter/src/rules/overridden_fields.dart';
@@ -108,6 +109,7 @@ void registerLintRules() {
     ..register(new NoDuplicateCaseValues())
     ..registerDefault(new NonConstantIdentifierNames())
     ..registerDefault(new OneMemberAbstracts())
+    ..register(new OmitLocalVariableTypes())
     ..register(new OnlyThrowErrors())
     ..register(new OverriddenFields())
     ..register(new PackageApiDocs())
--- a/lib/src/rules/omit_local_variable_types.dart
+++ b/lib/src/rules/omit_local_variable_types.dart
@@ -1,0 +1,96 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+library linter.src.rules.omit_type_for_local_variables;
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/type.dart';
+import 'package:linter/src/analyzer.dart';
+
+const _desc = r'Omit the types for local variables.';
+
+const _details = r'''
+
+**CONSIDER** omitting the types for local variables.
+
+**BAD:**
+```
+Map<int, List<Person>> groupByZip(Iterable<Person> people) {
+  Map<int, List<Person>> peopleByZip = <int, List<Person>>{};
+  for (Person person in people) {
+    peopleByZip.putIfAbsent(person.zip, () => <Person>[]);
+    peopleByZip[person.zip].add(person);
+  }
+  return peopleByZip;
+}
+```
+
+**GOOD:**
+```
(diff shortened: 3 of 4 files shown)


lib/src/rules/omit_type_for_local_variables.dart
```diff
@@ -0,0 +1,101 @@
...
+      return;
+    }
+    if (!(node.iterable.bestType is InterfaceType)) {
+      return;
+    }
+    if (DartTypeUtilities.areFromSameClass(
+        (node.iterable.bestType as InterfaceType).typeArguments[0],
+        staticType.type)) {
+      rule.reportLint(node);
+    }
```
bwilkerson: While the "if (!condition) return" style is often appropriate, there are times when it's better to code it the other way around. In Dart, this is often true of `is` tests because it allows type promotion to work, which removes the need for explicit casts. Consider:
```
DartType iterableType = node.iterable.bestType;
if (iterableType is InterfaceType && iterableType.typeArguments[0] == staticType.type)) {
  rule.reportLint(node);
}
```

lib/src/util/dart_type_utilities.dart
```diff
@@ -13,6 +13,9 @@ import 'package:analyzer/dart/element/type.dart';
 typedef bool AstNodePredicate(AstNode node);
 
 class DartTypeUtilities {
+  static bool areFromSameClass(DartType type1, DartType type2) =>
+      type2 != null && isClass(type1, type2.name, type2.element?.library?.name);
```
bwilkerson: I think for these use cases you really want to know whether whether the types are the same. That is, I don't think you want to suggest dropping the type if it's the same class but with a different type argument:
```
List<num> x = <int>[];
```
If that's true, then you can just use the `==` operator on the types. If that's not the case (and you should add the above as a test case), then you can just ask whether the elements associated with the types are the same.

test/rules/omit_type_for_local_variables.dart
```diff
@@ -0,0 +1,36 @@
...
+  return peopleByZip;
+}
+
+class Person {
+  int zip = 3; // OK
+
+  Person() {
+    Iterable a = [], b = new Iterable.empty(); // OK
+    Iterable c = new Iterable.empty(), d = new Iterable.empty(); // LINT
+  }
```
bwilkerson: You should have at least one test where the variable declaration list has more than one variable in the list.

Should we be creating a lint for a regular `for` loop?
```
/// A node in a linked list.
class LinkedListNode {
  LinkedListNode next;
}
void traverse(LinkedListNode head) {
  for (LinkedListNode node = head; node != null; node = node.next) { ... }
}
```

lib/src/rules/omit_type_for_local_variables.dart
```diff
@@ -0,0 +1,101 @@
...
+}
+```
+
+''';
+
+class OmitTypeForLocalVariables extends LintRule {
+  _Visitor _visitor;
+  OmitTypeForLocalVariables()
+      : super(
+            name: 'omit_type_for_local_variables',
```
bwilkerson: Perhaps "omit_local_variable_types"?

lib/src/rules/omit_type_for_local_variables.dart
```diff
@@ -0,0 +1,101 @@
...
+  @override
+  visitVariableDeclarationList(VariableDeclarationList node) {
+    final staticType = node.type;
+    if (staticType == null) {
+      return;
+    }
+    if (node.parent is FieldDeclaration) {
+      return;
+    }
+    for (var child in node.variables) {
```
zoechi: I don't know what the coding style rules are exactly for this repo, but in the first line you use `final` but here you use `var`
Doesn't look consistent to me.

test/rules/omit_type_for_local_variables.dart
```diff
@@ -0,0 +1,36 @@
...
+  for (Person person in people) { // LINT
+    peopleByZip.putIfAbsent(person.zip, () => <Person>[]);
+    peopleByZip[person.zip].add(person);
+  }
+
+  return peopleByZip;
+}
+
+Map<int, List<Person>> goodGroupByZip(Iterable<Person> people) {
+  var peopleByZip = <int, List<Person>>{}; // OK
```
zoechi: related to my comment above about `final`, but here in both cases `var`

lib/src/rules/omit_type_for_local_variables.dart
JPaulsen: Fixed

lib/src/rules/omit_type_for_local_variables.dart
```diff
@@ -0,0 +1,101 @@
...
+  final LintRule rule;
+  _Visitor(this.rule);
+
+  @override
+  visitForEachStatement(ForEachStatement node) {
+    final staticType = node.loopVariable.type;
+    if (staticType == null) {
+      return;
+    }
+    if (node.parent is FieldDeclaration) {
```
JPaulsen: Fixed

lib/src/rules/omit_type_for_local_variables.dart
JPaulsen: Fixed

lib/src/util/dart_type_utilities.dart

test/rules/omit_type_for_local_variables.dart
JPaulsen: @bwilkerson I tested it in the SDK, and it lints them, the change to avoid the lint is doing this:

```
class LinkedListNode {
  LinkedListNode next;
}
void traverse(LinkedListNode head) {
  for (var node = head; node != null; node = node.next) { ... }
}
```

I think it should be linted, because the type can be inferred statically

JPaulsen: Fixed (tests added), please tell me if you want to change the behavior in the case of a regular for loop.

lib/src/rules/omit_type_for_local_variables.dart
JPaulsen: Fixed

test/rules/omit_type_for_local_variables.dart
JPaulsen: Fixed

This is a test, it is ok to do not follow a coding style rules in order to test the rule you want to test.

But, I only tested the rule with var, so now is one with var and one with final to test better, thanks!

Issue #471, opened 2017-03-06
Implement DO use initializing formals when possible.


JPaulsen:
From [Effective Dart](https://www.dartlang.org/guides/language/effective-dart/usage#do-use-initializing-formals-when-possible)

JPaulsen:
This rule shall not lint private fields in any case

PR #475, opened 2017-03-07
New rule: use_initializing_formals_when_possible

JPaulsen:
Fixes https://github.com/dart-lang/linter/issues/471

### DIFF
--- a/lib/src/rules/use_initializing_formals_when_possible.dart
+++ b/lib/src/rules/use_initializing_formals_when_possible.dart
@@ -1,0 +1,125 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+library linter.src.rules.use_initializing_formals_when_possible;
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/element.dart';
+import 'package:linter/src/analyzer.dart';
+
+const _desc = r'Use initializing formals when possible.';
+
+const _details = r'''
+
+**DO** use initializing formals when possible.
+
+**BAD:**
+```
+class Point {
+  num x, y;
+  Point(num x, num y) {
+    this.x = x;
+    this.y = y;
+  }
+}
+```
+
+**GOOD:**
+```
+class Point {
+  num x, y;
+  Point(this.x, this.y);
+}
+```
+
+''';
+
+Iterable<AssignmentExpression> _getAssignmentExpressionsInConstructorBody(
+    ConstructorDeclaration node) {
+  final body = node.body;
+  final statements =
+      (body is BlockFunctionBody) ? body.block.statements : <Statement>[];
+  return statements
+      .where((e) => e is ExpressionStatement)
+      .map((e) => (e as ExpressionStatement).expression)
+      .where((e) => e is AssignmentExpression)
+      .map((e) => e as AssignmentExpression);
+}
+
+Iterable<ConstructorFieldInitializer>
+    _getConstructorFieldInitializersInInitializers(
+        ConstructorDeclaration node) {
+  return node.initializers
+      .where((e) => e is ConstructorFieldInitializer)
+      .map((e) => (e as ConstructorFieldInitializer));
+}
+
+Element _getLeftElement(AssignmentExpression assignment) {
+  final leftPart = assignment.leftHandSide;
+  return leftPart is SimpleIdentifier
+      ? leftPart.bestElement
+      : leftPart is PropertyAccess ? leftPart.propertyName.bestElement : null;
+}
+
+Iterable<Element> _getParameters(ConstructorDeclaration node) {
+  return node.parameters.parameters.map((e) => e.identifier.bestElement);
+}
+
+Element _getRightElement(AssignmentExpression assignment) {
+  final rightPart = assignment.rightHandSide;
+  return rightPart is SimpleIdentifier ? rightPart.bestElement : null;
+}
+
+class UseInitializingFormalsWhenPossible extends LintRule {
+  _Visitor _visitor;
+  UseInitializingFormalsWhenPossible()
+      : super(
+            name: 'use_initializing_formals_when_possible',
+            description: _desc,
+            details: _details,
+            group: Group.style) {
+    _visitor = new _Visitor(this);
+  }
+
+  @override
+  AstVisitor getVisitor() => _visitor;
+}
+
+class _Visitor extends SimpleAstVisitor {
+  final LintRule rule;
+  _Visitor(this.rule);
+
+  @override
+  visitConstructorDeclaration(ConstructorDeclaration node) {
+    final parameters = _getParameters(node);
+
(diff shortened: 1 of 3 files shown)


lib/src/rules/use_initializing_formals_when_possible.dart
```diff
@@ -0,0 +1,139 @@
...
+Element _getLeftElement(AssignmentExpression assignment) {
+  final leftPart = assignment.leftHandSide;
+  return leftPart is SimpleIdentifier
+      ? leftPart.bestElement
+      : leftPart is PropertyAccess ? leftPart.propertyName.bestElement : null;
+}
+
+Iterable<Element> _getParametersNotPresentInSuperConstructorInvocation(
+    ConstructorDeclaration node) {
+  final parametersNotPresentInInitializers =
```
JPaulsen: I must change the name of this variable

lib/src/rules/use_initializing_formals_when_possible.dart
```diff
@@ -0,0 +1,139 @@
...
+  AstVisitor getVisitor() => _visitor;
+}
+
+class _Visitor extends SimpleAstVisitor {
+  final LintRule rule;
+  _Visitor(this.rule);
+
+  @override
+  visitConstructorDeclaration(ConstructorDeclaration node) {
+    final parametersNotPresentInInitializers =
```
JPaulsen: I must change the name of this variable

test/rules/use_initializing_formals_when_possible.dart
```diff
@@ -0,0 +1,127 @@
...
+  }
+}
+
+class GoodPoint7 {
+  // ignore: unused_field
+  final num _x, _y;
+  GoodPoint7(num x, num y)
+      : this._x = x, // OK // This is the right way to do
+        this._y = y; // OK // This is the right way to do
+}
```
bwilkerson: Given the large number of cases that need to be tested here, it would be good to be a bit more rigorous and to document (via class name or comment) the cases being tested. Think about the various combinations of parameter kinds, ways of initializing fields, kinds of assignments, etc. and see whether you can identify cases you've missed. For example...

There are no tests here involving optional parameters (both positional and named are valid).

There are no tests where the right hand side of an assignment is anything other than a simple identifier, or where the right hand side is a simple identifier that isn't a parameter.

Also, what about some strange code like:
```
Point(x) {
  this.x = x;
  this.y = x;
}
```
You cannot reference field formal parameters in the body of the constructor. We could move these into the initializer list, but that's a different lint.

JPaulsen: In the case of:
```
Point(x) {
  this.x = x;
  this.y = x;
}
```
You should solve the lint doing:

```
Point(this.x) {
  this.y = x;
}
```

I will test both cases

lib/src/rules/use_initializing_formals_when_possible.dart
JPaulsen: Fixed

lib/src/rules/use_initializing_formals_when_possible.dart
JPaulsen: Fixed

test/rules/use_initializing_formals_when_possible.dart
```diff
@@ -0,0 +1,127 @@
...
+  set y(num y) {
+    name = 'My y value is $y';
+  }
+}
+
+class GoodPoint4 extends GoodPoint1 {
+  num r, theta;
+  GoodPoint4(num r, num theta) : super(r * cos(theta), r * sin(theta)) {
+    this.r = r; // OK
+    this.theta = theta; // OK
```
JPaulsen: In the first local version of the linter, we run it in the SDK and we founded some lints with that case happening, and we realized that this:

```
GoodPoint4(this.r, this.theta) : super(r * cos(theta), r * sin(theta));
```

Is not possible to do, so it was a false positive and we updated the rule

JPaulsen: Those both are the erros, with and without this

initializers

test/rules/use_initializing_formals_when_possible.dart
```diff
@@ -0,0 +1,127 @@
...
+  }
+}
+
+class GoodPoint5 extends GoodPoint1 {
+  num r, theta;
+  GoodPoint5.ok(num r, num theta)
+      : this.r = r, // OK
+        this.theta = theta, // OK
+        super(r * cos(theta), r * sin(theta));
```
JPaulsen: Ditto.

test/rules/use_initializing_formals_when_possible.dart
```diff
@@ -0,0 +1,127 @@
...
+}
+
+class GoodPoint6 {
+  // ignore: unused_field
+  num _x, _y;
+  GoodPoint6(num x, num y) {
+    this._x = x; // OK // This should be lint for other rule
+    this._y = y; // OK // This should be lint for other rule
```
JPaulsen: Fixed

test/rules/use_initializing_formals_when_possible.dart
bwilkerson: I just double checked using the example code below, and there are no errors or warnings reported by analyzer.
```
int cos(int x) => 0;
int sin(int x) => 0;
class Point1 {
  int x;
  int y;
  Point1(this.x, this.y);
}
class Point2 extends Point1 {
  int r;
  int theta;
  Point2(this.r, this.theta) : super(r * cos(theta), r * sin(theta));
}
```
What lints were being generated, and on what code, that made you conclude that this isn't possible?

bwilkerson: It's quite possible that DartPad hasn't been updated to the latest version of the analyzer. Allowing references to field formal parameters within the initializer list is a fairly recent change.

test/rules/use_initializing_formals_when_possible.dart
JPaulsen: Fixed

test/rules/use_initializing_formals_when_possible.dart
JPaulsen: With the first version of this rule we run it in sdk and we founded a lint equivalent to this:

```
int cos(int x) => 0;
int sin(int x) => 0;
class Point1 {
  int x;
  int y;
  Point1(this.x, this.y);
}
class Point2 extends Point1 {
  int r;
  int theta;
  Point2(num r, num theta) : super(r * cos(theta), r * sin(theta)) {
    this.r = r; // LINT
    this.theta = theta; // LINT
  }
}
```

And when we tried to fix it, we noticed something similar to the image below, so we concluded that it was a false positive, and we had to consider that in our implementation.

I think we are going to double check with Alexei and give a better answer

JPaulsen: You were totally right.

Fixed

test/rules/use_initializing_formals_when_possible.dart
JPaulsen: Fixed

pq:
@bwilkerson, @JPaulsen : is this good to go?

bwilkerson:
lgtm

JPaulsen:
@pq @bwilkerson yes, I fixed everything was commented and reanalyzed

pq:
Awesome @JPaulsen! 🤘

PR #479, opened 2017-03-09
dartfmt non-test code

kevmoo:


alexeieleusis:
This is nice, did you manually chose them or ran the formatter on all files and these were the ones modified?

kevmoo:
Ran the formatter and realized it broke some tests – for weird reasons

So I reverted the files that caused breaks. At least it moves us forward a bit...

PR #484, opened 2017-03-10
Fix rule: unnecessary_lambda

JPaulsen:
Fixes:

1.- Check that arguments and parameters lengths are equal
2.- Check if there is a ? operator in the target of a method invocation

### DIFF
--- a/lib/src/rules/unnecessary_lambdas.dart
+++ b/lib/src/rules/unnecessary_lambdas.dart
@@ -5,6 +5,7 @@
 library linter.src.rules.unnecessary_lambdas;
 
 import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/token.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
 import 'package:analyzer/dart/element/element.dart';
 import 'package:linter/src/analyzer.dart';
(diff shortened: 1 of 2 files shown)


lib/src/rules/unnecessary_lambdas.dart
```diff
@@ -99,8 +108,15 @@ class _Visitor extends SimpleAstVisitor {
     if (node is FunctionExpressionInvocation) {
       restOfElements = _extractElementsOfSimpleIdentifiers(node.function);
     } else if (node is MethodInvocation && node.target != null) {
+      if (_hasQuestionPeriod(node) ||
+          _hasQuestionPeriod(node.target) ||
+          DartTypeUtilities
+              .traverseNodesInDFS(node.target)
+              .any(_hasQuestionPeriod)) {
+        return;
```
bwilkerson: I think this will catch more than you intend. For example: `(p) a[b?.m1()].m2(p)`. I think it's semantically equivalent to replace this with a tear-off of `a[b?.m1()].m2` (though not necessarily easier to read, so maybe it's better that we don't lint in this case).

If you do want to create a lint in the above case, then I think you want to recursively look at the target of any `PropertyAccess`, `IndexExpression`, etc.

lib/src/rules/unnecessary_lambdas.dart
```diff
@@ -80,16 +88,17 @@ class _Visitor extends SimpleAstVisitor {
 
   void _visitInvocationExpression(
       InvocationExpression node, FunctionExpression nodeToLint) {
-    if (nodeToLint.parameters.length != node.argumentList.length) {
+    if (nodeToLint.parameters.parameters.length !=
+        node.argumentList.arguments.length) {
```
bwilkerson: We need a test that will catch the previous invalid code.

JPaulsen: I tried, but I could not, the perfect test was the one with generics that was in the issue, but I could not reproduce it in the test because the parser could not do it, I will try to create other test without generics that reproduce the bug

bwilkerson: What we were comparing before was the string length of the source code that produced the node, so try adding a space in one place or another to make the lengths different when there ought to be a lint.

JPaulsen: I added a commented test with a TODO statement, I could not reproduce the bug without generics

lib/src/rules/unnecessary_lambdas.dart
JPaulsen: Fixed

Issue #476, opened 2017-03-07
Update `invariant_booleans` docs to warn of perf issues.


pq:
Follow-up from #434.  As per @JPaulsen's suggestion, we should add some docs to warn of performance in general (though we should still strive to do a bit better).

Consider linking to #434 as well.

pq:
Spitballing text, how about something like this:

> **WARNING:** this lint is comparatively expensive as, in general, calculating [boolean satisfiability](https://en.wikipedia.org/wiki/Boolean_satisfiability_problem) is hard. Performance improvements are planned ([#434](https://github.com/dart-lang/linter/issues/434)) but in the meantime, this lint should be sparingly enabled in large projects or when lint performance is a concern.

cc @JPaulsen @bwilkerson @filiph @alexeieleusis for word-smithing.

bwilkerson:
lgtm

filiph:
Ex-post LGTM!

PR #486, opened 2017-03-13
Update invariant_booleans docs (#476).

pq:
Fixes: #476.

@bwilkerson @alexeieleusis @JPaulsen


alexeieleusis:
Thank you! @JPaulsen has a very interesting WIP for unnecessary_this that could be used for this rule, I am confident it will greatly improve performance. I'll touch base with him today and hopefully we can have good news soon.

Issue #473, opened 2017-03-06
False positive for unnecessary_lambdas


a14n:
False positive for https://github.com/flutter/flutter/blob/aac8762a7a16d931eebde40e7179f9913867ca8b/packages/flutter_test/lib/src/controller.dart#L172

```dart
return finder.evaluate().map((Element element) => _stateOf<T>(element, finder));
```

@JPaulsen

JPaulsen:
@a14n are you sure? I could not reproduce the bug, how do you run the linter in flutter?

a14n:
Once you have the flutter repo, you need to change `packages/flutter_tools/pubspec.yaml` to override linter dependency (eithier with a git reference or with a path reference):

```
diff --git a/packages/flutter_tools/pubspec.yaml b/packages/flutter_tools/pubspec.yaml
index feeaac5..5c60da8 100644
--- a/packages/flutter_tools/pubspec.yaml
+++ b/packages/flutter_tools/pubspec.yaml
@@ -52,3 +52,5 @@ dependency_overrides:
     path: ../../bin/cache/dart-sdk/lib/analyzer
   front_end:
     path: ../../bin/cache/dart-sdk/lib/front_end
+  linter:
+    path: ../../../../dart-lang/linter
```

You may have to remove the flutter_tools snapshot to take this change into account with `rm bin/cache/flutter_tools.snapshot`.

Now you can add any linter rule to `.analysis_options_repo` and run `flutter analyze --flutter-repo` to display lints.

I already send a [PR](https://github.com/flutter/flutter/pull/8595/files) to fix some lint but there are actually other false positives. See the Hixie's comments.

a14n:
Basically the lint should apply only when the function content contains only one method call with the same parameter count:

```dart
call((){ a.m1(); }); // LINT
call((p){ a.m2(p); }) // LINT
call((){ a?.m1(); }) // when a is null, not the same as call(a?.m1)
call((){ a.b.m1(); }) // when b is mutable, not the same as call(a.b.m1)
```

alexeieleusis:
https://github.com/flutter/flutter/pull/8595/files#diff-a58f6bf53c00b37f422f3cccc4e906a8R35

a14n:
Actually here's another false positive:

```dart
main() {
  Function makeCallable(void f()) => f;
  
  var f = () => print('a');
  
  // r1 is linted (r2 is the result to remove the lint)
  var r1 = makeCallable((){
    f();
  });
[...]

PR #487, opened 2017-03-13
Fix rule: unnecessary_lambdas

JPaulsen:
Fixes https://github.com/dart-lang/linter/issues/473

Fix: Do not lint when the method invocation in the lambda has any type of function invocation (including constructors and methods) in the target or a non final field or variable, so the lambda has not the same behavior than the tear-off

### DIFF
--- a/lib/src/rules/unnecessary_lambdas.dart
+++ b/lib/src/rules/unnecessary_lambdas.dart
@@ -31,25 +31,36 @@ names.forEach(print);
 
 ''';
 
+bool _containsNullAwareInvocationInChain(AstNode node) =>
+    node != null &&
+    ((node is PropertyAccess &&
+            (node.operator?.type == TokenType.QUESTION_PERIOD ||
+                _containsNullAwareInvocationInChain(node.target))) ||
+        (node is MethodInvocation &&
+            (node.operator?.type == TokenType.QUESTION_PERIOD ||
+                _containsNullAwareInvocationInChain(node.target))) ||
+        (node is IndexExpression &&
+            _containsNullAwareInvocationInChain(node.target)));
+
 Iterable<Element> _extractElementsOfSimpleIdentifiers(AstNode node) =>
     DartTypeUtilities
         .traverseNodesInDFS(node)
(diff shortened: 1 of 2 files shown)


lib/src/rules/unnecessary_lambdas.dart
```diff
@@ -31,25 +31,37 @@ names.forEach(print);
...
+bool _isNonFinalElement(Element element) =>
+    (element is PropertyAccessorElement &&
+        (!element.isSynthetic || !element.variable.isFinal)) ||
+    (element is VariableElement && !element.isFinal);
+
+bool _isNonFinalField(AstNode node) =>
+    (node is PropertyAccess &&
+        _isNonFinalElement(node.propertyName.bestElement)) ||
+    (node is MethodInvocation &&
+        (_isNonFinalElement(node.methodName.bestElement))) ||
```
bwilkerson: When would a method invocation be a non-final field?

JPaulsen: This case
```
var f = () {
   print('hi')
}
f(); // this is a method invocation and its method name (f) is a non-final field
```
This is a case when f is final.
```
var f = () {
   print('hi')
}
f(); // this is a method invocation and its method name (f) is a non-final field
```

lib/src/rules/unnecessary_lambdas.dart
```diff
@@ -31,25 +31,37 @@ names.forEach(print);
...
-              _hasRecursivelyQuestionPeriod(node.target))) ||
-      (node is IndexExpression &&
-              _hasRecursivelyQuestionPeriod(node.target)) ||
-      (node is MethodInvocation &&
-          (node.operator?.type == TokenType.QUESTION_PERIOD ||
-              _hasRecursivelyQuestionPeriod(node.target)));
-}
+bool _isInvocationExpression(AstNode node) => node is InvocationExpression;
+
+// invocation chain contains null aware invocation (?.)
```
JPaulsen: No, Alexis was helping me to chose better function names and I forgot to delete the comment, sorry for that.

Fixed

Issue #477, opened 2017-03-07
0.1.30 Release Plan


pq:
Punch list for a `0.1.30` release.

## Plan items

  * [x] update `invariant_booleans` docs (#476)
  * [x] fix false positive for `unnecessary_lambdas` (#473)
  * [x] land `use_initializing_formals_when_possible` (#475)
  * [x] do some rule renaming for consistency (#489)
  * [x] update changelist, pubspec
  * [x] update `AUTHORS` (#511)
  * [x] land `invariant_booleans` perf fixes (#501)

## Releasing

  * [x] tag release ([tag/0.1.30](https://github.com/dart-lang/linter/releases/tag/0.1.30))
  * [x] publish package (#438)
  * [x] publish docs ( https://github.com/dart-lang/linter/pull/384#issuecomment-284485488 ) 
  * [x] update `DEPS` (#491)

@alexeieleusis, @JPaulsen : are there any changes you have pending that are close to landing?

alexeieleusis:
https://github.com/dart-lang/linter/pull/475 should land soon and we found some false positives for unnecessary_lambdas, should land soon too.

I think we should wait to fix the false positives because it certainly can change the programs in unintended ways. Or at least disable it.

alexeieleusis:
I should have said @a14n found some false positives, not us. Thank you!

alexeieleusis:
https://github.com/dart-lang/linter/issues/473 is the issue to fix.

pq:
Thanks @alexeieleusis!  Updated punch list.

pq:
Given the ongoing conversation on #473, should we maybe just disable  `unnecessary_lambdas` for this release?  We can follow-up with another as soon as we feel good about a proper fix.

@JPaulsen @alexeieleusis ?

JPaulsen:
@pq I have the final fist almost ready, I think I will have it before 12:30 to check (actually I finished implementing it, I am writing more tests to have better test coverage)

pq:
Nice.  Thanks for the update @JPaulsen !

pq:
Updated punchlist with a new task to do some naming normalization.  (I was starting on a changelist and noticed that we have some names that don't play that well with our existing ones.)  I'll follow-up with a PR.

alexeieleusis:
Awesome, thank you @pq .

alexeieleusis:
[...]

PR #489, opened 2017-03-14
Rule name normalization (#477).

pq:
Rule name normalization as per #477.

Not perfect but I think a little closer to consistent; comments welcome!

@alexeieleusis @bwilkerson @JPaulsen


pq:
Thanks @bwilkerson .  As discussed, I'll reach out to folks who might have visibility before we update `DEPS`.

Issue #469, opened 2017-03-06
Implement DON’T use this. when not needed to avoid shadowing.


JPaulsen:
From [Effective Dart](https://www.dartlang.org/guides/language/effective-dart/usage#dont-use-this-when-not-needed-to-avoid-shadowing)

JPaulsen:
Does someone know how to get the variables in scope in a certain part of the code?, It seems like it will be necessary to implement this rule

bwilkerson:
Analyzer doesn't keep that information around, so you'll effectively need to re-compute that data. There is a class (NamespaceBuilder if I recall correctly) that can help. I think we had a utility somewhere that would take an AstNode and return the namespace in force for that node, but I don't remember off-hand where it was. You can probably find it by searching for references to NamespaceBuilder.

Issue #434, opened 2017-02-22
Fix `invariant_booleans` performance.


pq:
From @scheglov :

> Analysis spends 17% of total CPU time in computing lints.
> Or about 30% of total Dart analysis time in LibraryAnalyzer.

![image](https://cloud.githubusercontent.com/assets/67586/23222767/ca96e590-f8dd-11e6-9e68-10008a01c3fb.png)

He suspects that `invariant_booleans.dart` is using most of this lint time.

![image 1](https://cloud.githubusercontent.com/assets/67586/23222768/cc0de202-f8dd-11e6-9767-ee40ee15e5ce.png)

![image 2](https://cloud.githubusercontent.com/assets/67586/23222772/cd7c106e-f8dd-11e6-8d89-b69fea142ace.png)

@bwilkerson points out:

> It seems likely that we can improve this. The visitor doesn't maintain any state, so it appears to re-compute the same state over and over.

👍 

cc @alexeieleusis

pq:
@alexeieleusis: optimistically assigning to you; if you don't think you can look at this (soonish) please un-assign and bump.

Thanks!

pq:
Collecting stats on the linter source base confirms that this lint is comparatively expensive.

    $ dart bin/linter.dart -s .

yields:

```
166 files analyzed, 6917 issues found, in 6947 ms.
-------------------------------------------------------------
AMBIGUOUS_IMPORT                                            2
CAST_TO_NON_TYPE                                           23
CONFLICTING_GETTER_AND_METHOD                               1
CONFLICTING_INSTANCE_METHOD_SETTER                          1
DUPLICATE_CONSTRUCTOR_DEFAULT                               2
DUPLICATE_DEFINITION                                        2
EXPECTED_EXECUTABLE                                         2
EXTENDS_NON_CLASS                                         151
EXTRA_POSITIONAL_ARGUMENTS                                 15
IMPLEMENTS_NON_CLASS                                       17
INVALID_ANNOTATION                                          3
INVALID_OVERRIDE_REQUIRED                                   1
INVOCATION_OF_NON_FUNCTION                                  7
MIXIN_INHERITS_FROM_NOT_OBJECT                              3
MIXIN_OF_NON_CLASS                                          2
MIXIN_WITH_NON_CLASS_SUPERCLASS                             3
NON_ABSTRACT_CLASS_INHERITS_ABSTRACT_MEMBER_FIVE_PLUS       3
[...]

PR #495, opened 2017-03-15
Introducing ScopeVisitors, improve performance of invariant_booleans and create rule unnecessary_this

JPaulsen:
Fixes https://github.com/dart-lang/linter/issues/434
Fixes https://github.com/dart-lang/linter/issues/469

### DIFF
--- a/lib/src/util/dart_type_utilities.dart
+++ b/lib/src/util/dart_type_utilities.dart
@@ -19,9 +19,7 @@ class DartTypeUtilities {
           extendsClass(type.superclass, className, library));
 
   static Element getCanonicalElement(Element element) =>
-      element is PropertyAccessorElement
-          ? element.variable
-          : element;
+      element is PropertyAccessorElement ? element.variable : element;
 
   static bool implementsAnyInterface(
       DartType type, Iterable<InterfaceTypeDefinition> definitions) {
--- a/lib/src/util/environment_visitors.dart
+++ b/lib/src/util/environment_visitors.dart
@@ -1,0 +1,1117 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+library linter.src.util.visitor_with_scope;
+
+import 'dart:collection';
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/element.dart';
+
+class ElementBox implements Nameable {
+  Element element;
+  ElementBox(this.element);
+
+  @override
+  bool hasName(String name) => element.name == name;
+}
+
+abstract class LookUpVisitor<T> extends SimpleAstVisitor<T> {
+  _MainCallVisitor<T, ElementBox> _mainCallVisitor;
+  LookUpVisitor() {
+    _mainCallVisitor = new _MainCallVisitor<T, ElementBox>(
+        new _IdentifiersScopeVisitor<T>(this));
+  }
+  AstVisitor<T> getVisitor() => _mainCallVisitor;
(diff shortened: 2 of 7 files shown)


lib/src/util/environment_visitors.dart
```diff
@@ -0,0 +1,500 @@
...
+  @override
+  visitVariableDeclaration(VariableDeclaration node) {
+    _baseVisitor.visitVariableDeclaration(node);
+    _addElementToEnvironment(node.element);
+    node.visitChildren(this);
+  }
+}
+
+abstract class _EnvironmentVisitor<E> extends _DelegatingAstVisitor {
+  final Queue<Queue<E>> environments = new Queue();
```
bwilkerson: There's a different approach (that we use elsewhere) that might be better. Rather than using a data structure such as a List or Queue, create a linked list of "scope" objects that know how to interact with the "outer" scope. You can see an example by looking at `ScopedVisitor` (which, by the way, is how analyzer keeps name scope information, which might be a good alternative to `_ElementEnvironmentVisitor` because it will automatically be maintained as the language semantics change).

lib/src/util/dart_type_utilities.dart
```diff
@@ -18,6 +18,11 @@ class DartTypeUtilities {
       (type is InterfaceType &&
           extendsClass(type.superclass, className, library));
 
+  static Element getRealElement(Element element) =>
+      element is PropertyAccessorElement
+          ? getRealElement(element.variable)
+          : element;
```

lib/src/util/environment_visitors.dart
JPaulsen: I saw ScopedVisitor and it is wonderful code, I am trying to figure out if I can resolve this query.

I have, for example a 'this.x', and I would like to know what would give the lookUp if a use 'x' instead, do you have an idea of how to do that with ScopedVisitor?

Of course if the element using 'x' and 'this.x' are the same, that this is unnecessary and not needed to avoid shadowing.

bwilkerson: > I saw ScopedVisitor and it is wonderful code ...

I actually think that it needs to be re-factored, but that's a different conversation. :-)

To use it you would create a subclass and implement `visitThisExpression` (similar to what you already have). When you get to the point of trying to resolve an identifier, you'd write something like
```
SimpleIdentifier identifier;
if (parent is PropertyAccess) {
  identifier = parent.propertyName;
} else if (parent is MethodInvocation) {
  identifier = parent.methodName;
} else {
  return null;
}
Element element = nameScope.lookup(identifier, referencingLibrary);
```
where the `referencingLibrary` is the `LibraryElement` associated with the compilation unit being traversed.

JPaulsen: I got it, but I continue having a problem, I do not know how to get the LibraryElement of a CompilationUnit node, do I have to get the LibraryDirective?

JPaulsen: I can get it using node.element.enclosingElement

bwilkerson: Yep.

JPaulsen: Last problem, I need to initialized the ScopedVisitor or do a super call , whatever, the sign of the ScopedVisitor is:

```
ScopedVisitor(this.definingLibrary, Source source, this.typeProvider,
      AnalysisErrorListener errorListener,
      {Scope nameScope})
```
So what about the rest of the parameters?

bwilkerson: Good question. I had to do a bit of digging. Assuming this is inside your lint rule's visitor:
```
CompilationUnit unit = ...;
LibraryElement library = unit.element.library;
new MyScopedVisitorSubclass(library, rule.reporter.source, library.context.typeProvider, AnalysisErrorListener.NULL_LISTENER);
```

JPaulsen:
@bwilkerson I sent unnecessary_this in other pull request, PTAL, I will close this pull request, and send other to refactor the improvement of invariant_booleans.

Issue #493, opened 2017-03-15
False positive for prefer_initializing_formals


a14n:
In Flutter there's 4 lints (every elements in the initialization list) for [the following code](https://github.com/flutter/flutter/blob/master/examples/layers/rendering/src/sector_layout.dart#L20-L24):

```dart
  const SectorConstraints.tight({ double deltaRadius: 0.0, double deltaTheta: 0.0 })
    : minDeltaRadius = deltaRadius,
      maxDeltaRadius = deltaRadius,
      minDeltaTheta = deltaTheta,
      maxDeltaTheta = deltaTheta;
```

alexeieleusis:
I guess the fix consists in not linting if a parameter gets assigned to more than one field. I kind of like it, and definitely prefer it to lose the final modifier on fields.

alexeieleusis:
Nice finding, thanks!

alexeieleusis:
@JPaulsen had already considered this case:

```dart

class SectorConstraints {
  final double minDeltaRadius;
  final double maxDeltaRadius;
  final double minDeltaTheta;
  final double maxDeltaTheta;

  const SectorConstraints.tight(
      { this.minDeltaRadius: 0.0, this.minDeltaTheta: 0.0 })
      : maxDeltaRadius = minDeltaRadius,
        maxDeltaTheta = minDeltaTheta;
}
```

PR #497, opened 2017-03-15
Fix rule: prefer_initializing_formals

JPaulsen:
Fixes https://github.com/dart-lang/linter/issues/493

### DIFF
--- a/lib/src/rules/prefer_initializing_formals.dart
+++ b/lib/src/rules/prefer_initializing_formals.dart
@@ -8,6 +8,7 @@ import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
 import 'package:analyzer/dart/element/element.dart';
 import 'package:linter/src/analyzer.dart';
+import 'package:linter/src/util/dart_type_utilities.dart';
 
 const _desc = r'Use initializing formals when possible.';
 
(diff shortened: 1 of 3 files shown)


lib/src/rules/prefer_initializing_formals.dart
```diff
@@ -94,16 +97,20 @@ class _Visitor extends SimpleAstVisitor {
   @override
   visitConstructorDeclaration(ConstructorDeclaration node) {
     final parameters = _getParameters(node);
+    final parametersUsedOne = new Set<Element>();
+    final parametersUsedMoreThanOne = new Set<Element>();
```
bwilkerson: For the two lines above, "One" --> "Once".

lib/src/rules/prefer_initializing_formals.dart
```diff
@@ -112,9 +119,30 @@ class _Visitor extends SimpleAstVisitor {
...
           expression is SimpleIdentifier &&
-          parameters.contains(expression.bestElement);
+          parameters.contains(expression.bestElement) &&
+          (!parametersUsedMoreThanOne.contains(expression.bestElement) ||
+              constructorFieldInitializer.fieldName.bestElement?.name ==
+                  expression.bestElement.name);
     }
 
+    void processElement(Element element) {
+      if (parametersUsedOne.contains(element)) {
```
bwilkerson: With sets, you can do this more efficiently by adding the element to the set and then using the value returned by `add` to determine whether it was already in the set.

lib/src/util/dart_type_utilities.dart
```diff
@@ -18,6 +18,11 @@ class DartTypeUtilities {
       (type is InterfaceType &&
           extendsClass(type.superclass, className, library));
 
+  static Element getRealElement(Element element) =>
+      element is PropertyAccessorElement
+          ? getRealElement(element.variable)
```
bwilkerson: The variable of a PropertyAccessorElement will never be a PropertyAccessorElement, so the recursive call is not necessary.

lib/src/rules/prefer_initializing_formals.dart
JPaulsen: Fixed

lib/src/rules/prefer_initializing_formals.dart
JPaulsen: Fixed

lib/src/util/dart_type_utilities.dart

PR #501, opened 2017-03-16
Improve performance: invariant_booleans

JPaulsen:
Fixes https://github.com/dart-lang/linter/issues/434

### DIFF
--- a/lib/src/rules/invariant_booleans.dart
+++ b/lib/src/rules/invariant_booleans.dart
@@ -8,8 +8,10 @@ import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/standard_resolution_map.dart';
 import 'package:analyzer/dart/ast/token.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/element.dart';
 import 'package:linter/src/analyzer.dart';
 import 'package:linter/src/util/boolean_expression_utilities.dart';
+import 'package:linter/src/util/condition_scope_visitor.dart';
 import 'package:linter/src/util/dart_type_utilities.dart';
 import 'package:linter/src/util/tested_expressions.dart';
 
@@ -96,166 +98,12 @@ void nestedOk5() {
 }
 ```
 
-_**WARNING:** this lint is comparatively expensive as, in general, calculating 
-[boolean satisfiability](https://en.wikipedia.org/wiki/Boolean_satisfiability_problem) is hard. 
-Performance improvements are planned ([#434](https://github.com/dart-lang/linter/issues/434)) but 
-in the meantime, this lint should be sparingly enabled in large projects or when lint performance 
-is a concern._
 ''';
 
-Set<Expression> _findConditionsCausingReturns(
-        Expression node, Iterable<AstNode> nodesInDFS) =>
-    nodesInDFS
-        .where(_isAnalyzedNode)
-        .where(_isConditionalStatementWithReturn(nodesInDFS))
-        .takeWhile((n) => n != node.parent)
-        .where((_noFurtherAssignmentInvalidatingCondition(node, nodesInDFS)))
-        .fold(<Expression>[], (List<Expression> previous, AstNode statement) {
-      previous.add(_getCondition(statement));
-      return previous;
-    }).toSet();
-
-Set<Expression> _findConditionsOfElseStatementAncestor(
-    Statement statement, Iterable<AstNode> nodesInDFS) {
-  return _findConditionsUnderStatementBranch(
-      statement,
-      (n) =>
-          n is IfStatement &&
-          statement.getAncestor((a) => a == n.elseStatement) != null,
-      nodesInDFS);
-}
-
-Set<Expression> _findConditionsOfStatementAncestor(
-    Statement statement, Iterable<AstNode> nodesInDFS) {
-  return _findConditionsUnderStatementBranch(
-      statement,
-      (n) =>
-          _isAnalyzedNode(n) &&
-          (n is! IfStatement ||
-              statement.getAncestor(
-                      (a) => a == (n as IfStatement).thenStatement) !=
-                  null),
-      nodesInDFS);
-}
-
-Set<Expression> _findConditionsUnderStatementBranch(Statement statement,
-    AstNodePredicate predicate, Iterable<AstNode> nodesInDFS) {
-  Expression condition = _getCondition(statement);
-  AstNodePredicate noFurtherAssignmentInvalidatingCondition =
-      _noFurtherAssignmentInvalidatingCondition(condition, nodesInDFS);
-  return nodesInDFS
-      .where((n) => n == statement.getAncestor((a) => a == n && a != statement))
-      .where(_isAnalyzedNode)
-      .where(predicate)
-      .where(noFurtherAssignmentInvalidatingCondition)
-      .fold(<Expression>[], (List<Expression> previous, AstNode statement) {
-    previous.add(_getCondition(statement));
-    return previous;
-  }).toSet();
-}
-
-TestedExpressions _findPreviousTestedExpressions(Expression node) {
-  Block block = node.getAncestor((a) => a is Block && a.parent is FunctionBody);
-  Iterable<AstNode> nodesInDFS = DartTypeUtilities.traverseNodesInDFS(block,
-      excludeCriteria: (n) => n is FunctionDeclarationStatement);
-  Iterable<Expression> conjunctions =
-      _findConditionsOfStatementAncestor(node.parent, nodesInDFS)
-          .map(_splitConjunctions)
-          .expand((iterable) => iterable)
-          .toSet();
-  Iterable<Expression> negations = (_findConditionsCausingReturns(
-          node, nodesInDFS)
-        ..addAll(
-            _findConditionsOfElseStatementAncestor(node.parent, nodesInDFS)))
-      .toSet();
-  return new TestedExpressions(node, conjunctions, negations);
-}
-
-Set<Identifier> _findStatementIdentifiers(Statement statement) =>
-    DartTypeUtilities
-        .traverseNodesInDFS(statement)
-        .where((n) => n is Identifier)
-        .toSet();
-
-Expression _getCondition(Statement statement) {
-  if (statement is IfStatement) {
-    return statement.condition;
-  }
-
-  if (statement is DoStatement) {
-    return statement.condition;
-  }
-
-  if (statement is ForStatement) {
-    return statement.condition;
-  }
-
-  if (statement is WhileStatement) {
-    return statement.condition;
-  }
-
-  return null;
-}
-
-bool _isAnalyzedNode(AstNode node) =>
-    node is IfStatement ||
-    node is DoStatement ||
-    node is ForStatement ||
-    node is WhileStatement;
-
-AstNodePredicate _isConditionalStatementWithReturn(
-        Iterable<AstNode> blockNodes) =>
-    (AstNode node) {
-      Block block =
-          node.getAncestor((a) => a is Block && a.parent is BlockFunctionBody);
-      Iterable<AstNode> nodesInDFS = DartTypeUtilities.traverseNodesInDFS(node);
-      return nodesInDFS.any((n) => n is ReturnStatement) &&
-          // Ignore nested if statements.
-          !nodesInDFS.any(_isAnalyzedNode) &&
-          node.getAncestor((n) =>
-                  n != node &&
-                  _isAnalyzedNode(n) &&
-                  n.getAncestor((a) => a == block) == block) ==
-              null;
-    };
-
-AstNodePredicate _noFurtherAssignmentInvalidatingCondition(
-    Expression node, Iterable<AstNode> nodesInDFS) {
-  Set<Identifier> identifiers = _findStatementIdentifiers(node.parent);
-  return (AstNode statement) {
-    bool isMutation(AstNode n) {
-      if (n is AssignmentExpression) {
-        return !identifiers.contains(n.leftHandSide);
-      } else if (n is PostfixExpression) {
-        TokenType type = n.operator.type;
-        return (type == TokenType.PLUS_PLUS || type == TokenType.MINUS_MINUS) &&
-            !identifiers.contains(n.operand);
-      } else if (n is PrefixExpression) {
-        TokenType type = n.operator.type;
-        return (type == TokenType.PLUS_PLUS || type == TokenType.MINUS_MINUS) &&
-            !identifiers.contains(n.operand);
-      }
-
-      return false;
-    }
-
-    return nodesInDFS
-        .skipWhile((n) => n != statement)
-        .takeWhile((n) => n != node)
-        .where(isMutation)
-        .isEmpty;
-  };
-}
-
-List<Expression> _splitConjunctions(Expression expression) {
-  if (expression is BinaryExpression &&
-      expression.operator.type == TokenType.AMPERSAND_AMPERSAND) {
-    return _splitConjunctions(expression.leftOperand)
-      ..addAll(_splitConjunctions(expression.rightOperand));
-  }
-
-  return [expression];
-}
+Iterable<Element> _getElementsInExpression(Expression node) => DartTypeUtilities
+    .traverseNodesInDFS(node)
+    .map(DartTypeUtilities.getCanonicalElementFromIdentifier)
+    .where((e) => e != null);
 
 class InvariantBooleans extends LintRule {
   _Visitor _visitor;
@@ -286,36 +134,16 @@ class _ContradictionReportRule extends LintRule {
             maturity: Maturity.stable);
 }
 
-class _Visitor extends SimpleAstVisitor {
+class _InvariantBooleansVisitor extends ConditionScopeVisitor {
   final LintRule rule;
 
-  _Visitor(this.rule);
+  _InvariantBooleansVisitor(this.rule);
 
   @override
-  visitDoStatement(DoStatement node) {
-    _reportExpressionIfConstantValue(node.condition);
-  }
-
-  @override
-  visitForStatement(ForStatement node) {
-    _reportExpressionIfConstantValue(node.condition);
-  }
-
-  @override
-  visitIfStatement(IfStatement node) {
-    _reportExpressionIfConstantValue(node.condition);
-  }
-
-  @override
-  visitWhileStatement(WhileStatement node) {
-    _reportExpressionIfConstantValue(node.condition);
-  }
-
-  _reportExpressionIfConstantValue(Expression node) {
+  visitCondition(Expression node) {
     // Right part discards reporting a subexpression already reported.
     if (node == null ||
-        resolutionMap.bestTypeForExpression(node).name != 'bool' ||
-        !_isAnalyzedNode(node.parent)) {
+        resolutionMap.bestTypeForExpression(node).name != 'bool') {
       return;
     }
 
@@ -339,5 +167,23 @@ class _Visitor extends SimpleAstVisitor {
         binaryExpression.operator.type != TokenType.QUESTION_QUESTION) {
       rule.reportLint(node);
     }
+  }
+
+  TestedExpressions _findPreviousTestedExpressions(Expression node) {
+    final elements = _getElementsInExpression(node);
+    Iterable<Expression> conjunctions = getTrueExpressions(elements).toSet();
+    Iterable<Expression> negations = getFalseExpressions(elements).toSet();
+    return new TestedExpressions(node, conjunctions, negations);
+  }
+}
+
+class _Visitor extends SimpleAstVisitor {
+  final LintRule rule;
+
+  _Visitor(this.rule);
+
+  @override
+  visitCompilationUnit(CompilationUnit node) {
+    new _InvariantBooleansVisitor(rule).visitCompilationUnit(node);
   }
 }
--- a/lib/src/util/condition_scope_visitor.dart
+++ b/lib/src/util/condition_scope_visitor.dart
@@ -1,0 +1,395 @@
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/token.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/element.dart';
+// ignore: implementation_imports
+import 'package:analyzer/src/generated/resolver.dart';
+import 'package:linter/src/util/dart_type_utilities.dart';
+
+Element _getLeftElement(AssignmentExpression assignment) => DartTypeUtilities
+    .getCanonicalElementFromIdentifier(assignment.leftHandSide);
+
+List<Expression> _splitConjunctions(Expression expression) {
+  if (expression is BinaryExpression &&
+      expression.operator.type == TokenType.AMPERSAND_AMPERSAND) {
+    return _splitConjunctions(expression.leftOperand)
+      ..addAll(_splitConjunctions(expression.rightOperand));
+  } else if (expression is ParenthesizedExpression) {
+    return _splitConjunctions(expression.expression);
+  }
+  return [expression];
+}
+
+class BreakScope {
+  var environment = <BreakStatement>[];
+
+  void add(BreakStatement element) {
+    if (element.target != null) {
+      environment.add(element);
+    }
+  }
+
+  void deleteBreaksWithTarget(AstNode node) {
+    environment = environment.where((e) => e.target != node).toList();
+  }
+
+  bool hasBreak(AstNode node) => environment.any((e) => e.target == node);
+}
+
+class ConditionScope {
+  final environment = <_ExpressionBox>[];
+  final ConditionScope outer;
+
+  ConditionScope(this.outer);
+
+  void add(_ExpressionBox e) {
+    if (e != null) {
+      environment.add(e);
+    }
+  }
+
+  void addAll(Iterable<_ExpressionBox> expressions) {
+    environment.addAll(expressions);
+  }
+
+  Iterable<Expression> getExpressions(Iterable<Element> elements, bool value) {
+    final expressions = <Expression>[];
+    _recursiveGetExpressions(expressions, elements, value);
+    return expressions;
+  }
+
+  Iterable<_ExpressionBox> getUndefinedExpressions() =>
+      environment.where((e) => e is _UndefinedExpression);
+
+  void _recursiveGetExpressions(
+      List<Expression> expressions, Iterable<Element> elements, bool value) {
+    for (final element in environment.reversed) {
+      if (element.haveToStop(elements)) {
+        return;
+      }
+      if (element is _ConditionExpression && element.value == value) {
+        expressions.add(element.expression);
+      }
+    }
+    outer?._recursiveGetExpressions(expressions, elements, value);
+  }
+}
+
+/// An AST visitor that keeps the conditions that are currently evaluated
+/// to true or false.
+///
+/// Subclasses that override a visit method must either invoke the overridden
+/// visit method or must explicitly ask the visited node to visit its children.
+/// Failure to do so will cause the children of the visited node to not be
+/// visited.
+///
+/// Subclasses that override a visit method must either invoke the overridden
+/// visit method and keep the scopes behavior consistent to its changes.
+///
+/// When add a new local scope:
+/// - Visiting a non-empty function body: [BlockFunctionBody] [ExpressionFunctionBody].
+/// - Visiting a flow-control statement: [IfStatement] ElseStatement.
+/// - Visiting loop statements: [DoStatement] [WhileStatement] [ForStatement] [ForEachStatement].
+///
+/// When call the abstract method visitCondition(node.condition):
+/// - After visiting a conditional statements: [IfStatement] [DoStatement] [WhileStatement] [ForStatement].
+///
+/// When undefine an element:
+/// - Visiting reassignments of variables: [AssignmentExpression] [PrefixExpression] [PostfixExpression].
+///
+/// When undefine all elements:
+/// - Visiting a non-empty function body: [BlockFunctionBody] [ExpressionFunctionBody].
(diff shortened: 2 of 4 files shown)


lib/src/rules/invariant_booleans.dart
```diff
@@ -96,157 +98,35 @@ void nestedOk5() {
 }
 ```
 
-_**WARNING:** this lint is comparatively expensive as, in general, calculating 
-[boolean satisfiability](https://en.wikipedia.org/wiki/Boolean_satisfiability_problem) is hard. 
-Performance improvements are planned ([#434](https://github.com/dart-lang/linter/issues/434)) but 
-in the meantime, this lint should be sparingly enabled in large projects or when lint performance 
+_**WARNING:** this lint is comparatively expensive as, in general, calculating
```
alexeieleusis: This is no longer needed :)

JPaulsen:
Benchmark in my machine:

```
---------------------------------------------------
Timings                                          ms
---------------------------------------------------
always_declare_return_types                       9
always_specify_types                             10
annotate_overrides                               25
avoid_as                                          3
avoid_empty_else                                  3
avoid_function_literals_in_foreach_calls          4
avoid_init_to_null                                4
avoid_return_types_on_setters                     3
avoid_slow_async_io                               6
await_only_futures                                3
camel_case_types                                  9
cancel_subscriptions                             18
cascade_invocations                               6
close_sinks                                      51
comment_references                                3
constant_identifier_names                         6
control_flow_in_finally                           4
directives_ordering                              16
empty_catches                                     4
empty_constructor_bodies                          3
empty_statements                                  3
hash_and_equals                                   4
implementation_imports                            4
invariant_booleans                               58
iterable_contains_unrelated_type                  5
library_names                                     7
library_prefixes                                  9
list_remove_unrelated_type                        5
literal_only_boolean_expressions                  5
no_adjacent_strings_in_list                       3
no_duplicate_case_values                          7
non_constant_identifier_names                    18
omit_local_variable_types                         7
one_member_abstracts                              4
only_throw_errors                                 4
overridden_fields                                 6
package_api_docs                                 14
package_prefixed_library_names                    8
parameter_assignments                             6
prefer_adjacent_string_concatenation              3
prefer_collection_literals                        4
prefer_const_constructors                         4
prefer_contains                                  25
prefer_expression_function_bodies                 4
prefer_final_fields                              17
prefer_final_locals                               6
prefer_function_declarations_over_variables       3
prefer_initializing_formals                       7
prefer_interpolation_to_compose_strings         164
prefer_is_empty                                  22
prefer_is_not_empty                               9
public_member_api_docs                           92
recursive_getters                                 5
slash_for_doc_comments                            4
sort_constructors_first                           5
sort_unnamed_constructors_first                   4
super_goes_last                                   3
test_types_in_equals                              3
throw_in_finally                                  4
type_annotate_public_apis                         5
type_init_formals                                 3
unawaited_futures                                 4
unnecessary_brace_in_string_interps               4
unnecessary_getters_setters                       7
unnecessary_lambdas                               7
unnecessary_null_aware_assignments                3
unnecessary_null_in_if_null_operators             3
unnecessary_this                                 17
unrelated_type_equality_checks                    6
valid_regexps                                     4
---------------------------------------------------
Total                                           818
---------------------------------------------------
```

lib/src/rules/invariant_booleans.dart
JPaulsen: Fixed

lib/src/rules/invariant_booleans.dart
```diff
@@ -96,157 +98,30 @@ void nestedOk5() {
...
-          .map(_splitConjunctions)
-          .expand((iterable) => iterable)
-          .toSet();
-  Iterable<Expression> negations = (_findConditionsCausingReturns(
-          node, nodesInDFS)
-        ..addAll(
-            _findConditionsOfElseStatementAncestor(node.parent, nodesInDFS)))
-      .toSet();
-  return new TestedExpressions(node, conjunctions, negations);
+Element _getRealElement(AstNode node) {
```
bwilkerson: Can we rename this method to "Canonical" as well?

lib/src/util/condition_scope_visitor.dart
```diff
@@ -0,0 +1,256 @@
...
+import 'package:linter/src/util/dart_type_utilities.dart';
+
+Element _getLeftElement(AssignmentExpression assignment) {
+  final leftPart = assignment.leftHandSide;
+  return leftPart is SimpleIdentifier
+      ? DartTypeUtilities.getCanonicalElement(leftPart.bestElement)
+      : leftPart is PropertyAccess
+          ? DartTypeUtilities
+              .getCanonicalElement(leftPart.propertyName.bestElement)
+          : null;
```
bwilkerson: Maybe move `_getRealElement` from `invariant_booleans.dart` to here and convert this to ` => _getRealElement(assignment.leftHandSide);`?

lib/src/util/condition_scope_visitor.dart
```diff
@@ -0,0 +1,256 @@
...
+}
+
+bool _isBreakStatement(Statement statement) => (statement is BreakStatement ||
+    (statement is Block &&
+        statement.statements.length == 1 &&
+        statement.statements.first is BreakStatement));
```
bwilkerson: Is it worth a recursive call here to handle strange, but possible, code like this?
```
{
  {
    break;
  }
}
```
Probably not, but it's your call.

lib/src/util/condition_scope_visitor.dart
```diff
@@ -0,0 +1,256 @@
...
+            statement.statements.length == 1 &&
+            statement.statements.first is ContinueStatement));
+
+bool _isReturnStatement(Statement statement) => (statement is ReturnStatement ||
+    (statement is Block &&
+        statement.statements.length == 1 &&
+        statement.statements.first is ReturnStatement));
+
+abstract class ConditionScopeVisitor extends RecursiveAstVisitor {
+  final Queue<Queue<_ExpressionBox>> environments = new Queue();
```
bwilkerson: Still seems odd to me to use a `Queue`, especially when we're treating the top-level queue as a stack. :-)

lib/src/util/condition_scope_visitor.dart
```diff
@@ -0,0 +1,256 @@
...
+
+  Iterable<Expression> getFalseExpressions(Iterable<Element> elements) =>
+      _getExpressions(elements, value: false);
+
+  Iterable<Expression> getTrueExpressions(Iterable<Element> elements) =>
+      _getExpressions(elements);
+
+  @override
+  visitAssignmentExpression(AssignmentExpression node) {
+    _addElementToEnvironment(new _UndefinedExpression(_getLeftElement(node)));
```
bwilkerson: We might want to guard against nodes that cannot be resolved by not adding anything to the environment (here and elsewhere):
```
Element element = _getLeftElement(node);
if (element != null) {
  _addElementToEnvironment(new _UndefinedExpression(node);
}
```
We should minimally add a test.

lib/src/util/condition_scope_visitor.dart
```diff
@@ -0,0 +1,256 @@
...
+      environments.first.addFirst(e);
+    }
+  }
+
+  _addExpressionIfStatementIsStopStatement(
+      Statement statement, _ExpressionBox expressionBox) {
+    if (_isReturnStatement(statement) ||
+        _isBreakStatement(statement) ||
+        _isContinueStatement(statement)) {
+      _addElementToEnvironment(expressionBox);
```
bwilkerson: We do something similar for dead code detection. Check out https://github.com/dart-lang/sdk/blob/master/pkg/analyzer/lib/src/generated/resolver.dart#L2201. Our checking is a bit more complete, and you might be able to make use of it here (although we'd need to make it available).

lib/src/util/condition_scope_visitor.dart
```diff
@@ -0,0 +1,256 @@
...
+    }
+  }
+
+  void _addLocalEnvironment() {
```
bwilkerson: Under what conditions do we need to add a local environment? (Adding a comment would be good.) It seems like we might be adding and removing too many, but I'm not sure.

lib/src/util/condition_scope_visitor.dart
JPaulsen: In fact all of them are stacks xd, I am trying to figure out what is exactly what we need in this kind of "scope"

lib/src/util/condition_scope_visitor.dart
JPaulsen: I think so, but I am not sure yet

lib/src/util/condition_scope_visitor.dart
JPaulsen: Fixed

lib/src/util/condition_scope_visitor.dart
JPaulsen: Fixed

lib/src/util/condition_scope_visitor.dart
JPaulsen: I realized that it is necessary to create new local environments, I created an unifying visitor so now it asks which clases just creates local environments, PTAL at the methods visitNode and _needsLocalEnvironment

lib/src/util/condition_scope_visitor.dart
```diff
@@ -0,0 +1,259 @@
...
+    final operand = node.operand;
+    if (operand is SimpleIdentifier) {
+      _addElementToEnvironment(new _UndefinedExpression(operand.bestElement));
+    }
+    node.visitChildren(this);
+  }
+
+  @override
+  visitVariableDeclaration(VariableDeclaration node) {
+    _addElementToEnvironment(new _UndefinedExpression(node.element));
```
bwilkerson: Are there plans to use either the initialization expression (or the right-hand side in an assignment) when the value assigned is known (a constant expression)? For example, will we create a lint here:
```
var x = 0;
// code that doesn't modify x
if (x > 0) {
 ...
}
```
(It's ok if the answer is "no", or to do that as a follow-on to this CL if the answer is "yes".)

lib/src/rules/invariant_booleans.dart
```diff
@@ -286,32 +161,23 @@ class _ContradictionReportRule extends LintRule {
...
-    _reportExpressionIfConstantValue(node.condition);
-  }
-
-  @override
-  visitIfStatement(IfStatement node) {
-    _reportExpressionIfConstantValue(node.condition);
+  TestedExpressions _findPreviousTestedExpressions(Expression node) {
+    final elements = _getRealElementsInExpression(node);
+    Iterable<Expression> conjunctions = getTrueExpressions(elements)
+        .map(_splitConjunctions)
```
bwilkerson: Would it be more efficient to split conjunctions when adding the conditions rather than when finding tested expressions?

lib/src/util/condition_scope_visitor.dart
```diff
@@ -0,0 +1,259 @@
...
+      node is ForEachStatement ||
+      node is ForStatement ||
+      node is WhileStatement;
+
+_needsLocalEnvironment(AstNode node) =>
+    node is ClassDeclaration ||
+    node is CompilationUnit ||
+    node is ConstructorDeclaration ||
+    node is FunctionDeclaration ||
+    node is MethodDeclaration;
```
JPaulsen: It was not necessary, and yes, now they only are FunctionBody and Block

lib/src/util/condition_scope_visitor.dart
```diff
@@ -0,0 +1,259 @@
...
+    visitCondition(node.condition);
+    node.condition?.accept(this);
+    _addElementToEnvironment(new _ConditionExpression(node.condition));
+    node.updaters.accept(this);
+    node.body?.accept(this);
+    _removeLocalEnvironment();
+    bool hasBreak = DartTypeUtilities
+        .traverseNodesInDFS(node, excludeCriteria: _isLoopStatement)
+        .any((e) => e is BreakStatement);
+    if (!hasBreak) {
```
JPaulsen: It works well, if you have a break, what you know is the j < 10 is not necessary false after the ForStatement, but something that we should add is when you find a label, for example outer:, we should undefine all

lib/src/util/condition_scope_visitor.dart
JPaulsen: It is not supported by now, and before this optimization it was not contemplated, but it could be implemented in the future, I would like to do it, what do you think @alexeieleusis ?

lib/src/rules/invariant_booleans.dart
JPaulsen: Fixed

lib/src/util/condition_scope_visitor.dart
JPaulsen: There is not necessary to create new local environments in blocks

bwilkerson: Just so it isn't forgotten, I'll repeat the request:

Please add documentation somewhere explaining when and/or why we need a local environment.

lib/src/util/condition_scope_visitor.dart
```diff
@@ -0,0 +1,259 @@
...
+      _addElementToEnvironment(
+          new _ConditionExpression(node.condition, value: false));
+    }
+    if (_isLastStatementAnExitStatement(node.elseStatement)) {
+      _addElementToEnvironment(new _ConditionExpression(node.condition));
+    }
+  }
+
+  @override
+  visitNode(AstNode node) {
```
JPaulsen: Fixed

lib/src/util/condition_scope_visitor.dart
JPaulsen: Fixed (good idea)

lib/src/util/condition_scope_visitor.dart
JPaulsen: This is one of the few issues that I have not solved, what do you think is a better option here @bwilkerson?

lib/src/util/condition_scope_visitor.dart
JPaulsen: I fixed this creating a factory constructor, so if the parameter is null, the constructor return null and _addElementToEnvironment does not add null objects, so FIXED, but what kind of test could add?

lib/src/util/condition_scope_visitor.dart
bwilkerson: I would start by writing a class like the following:
```
class ConditionScope {
  /**
   * The outer scope, of `null` if this is the outermost scope.
   */
  final Scope outer;

  /**
   * Initialize a newly created scope to have the given [outer] scope.
   */
  ConditionScope(this.outer);
}
```
Then I would add fields and methods to support adding `_ExpressionBox`s and performing lookup (which would replace one of the method in `ConditionScopeVisitor`). Given that you're not really treating the inner queues as queues, I'd probably change to using a list. The `environments` field would be replaced by
```
ConditionScope currentScope;
```
pushes look like
```
currentScope = new ConditionScope(currentScope);
```
and pops look like
```
currentScope = currentScope.outer;
```

lib/src/util/condition_scope_visitor.dart
alexeieleusis: It is definitely nice to have, I wonder if now is the best time. We need to consider the cost/benefit, I lean towards leaving it for later, there are rules that require less effort and provide more value.

lib/src/util/condition_scope_visitor.dart
bwilkerson: In this case, we need an assignment expression whose left-hand-side is not resolved:
```
notDefined = 3;
```
(any name that isn't defined will work). Of course, you'll want to back out your fix to ensure that the test case actually triggers the bug.

lib/src/util/condition_scope_visitor.dart
JPaulsen: I implemented my own Stack the other day to see if that improved the benchmark of my iterative version of traverseNodeInDfs jaja, I got it, I will be focused on this

lib/src/util/condition_scope_visitor.dart
JPaulsen: Yes but, in that case, we would add to the scope something like, null is undefined, and that would not change the behavior of invariant booleans, at least we do something like:

```dart
if (notDefined2 < 2) {
  notDefined = 3; // we undefine null
  if (notDefined2 < 2) { // this won't be linted, because notDefined is null and null was undefined just before this condition is evaluated
  }
}
```
My point is that checking those nulls wont change the behavior in code with not null elements in conditions inside if/while/do/for, so we should take care about conditions with null elements inside too, what do you think?

lib/src/util/condition_scope_visitor.dart
JPaulsen: Fixed, I used a List but I had to reversed it when getting condition expression, PTAL, maybe is better use a Queue internally in the ConditionScope

lib/src/util/condition_scope_visitor.dart
bwilkerson: Given how few undefined names there usually are in code, if it doesn't cause exceptions to be thrown then I don't suppose it's worth doing a lot of work to guard against having a `null` element.

lib/src/util/condition_scope_visitor.dart
```diff
@@ -0,0 +1,307 @@
...
+  @override
+  visitDoStatement(DoStatement node) {
+    _addScope();
+    visitCondition(node.condition);
+    node.visitChildren(this);
+    _propagateUndefinedExpressions(_removeLastScope());
+  }
+
+  @override
+  visitEmptyFunctionBody(EmptyFunctionBody node) {
```
bwilkerson: Given that an empty function body doesn't have any children, this method should be unnecessary.

lib/src/util/condition_scope_visitor.dart
```diff
@@ -0,0 +1,307 @@
...
+    _addTrueCondition(node.condition);
+    node.variables?.accept(this);
+    node.initialization?.accept(this);
+    visitCondition(node.condition);
+    node.condition?.accept(this);
+    _addTrueCondition(node.condition);
+    node.updaters.accept(this);
+    node.body?.accept(this);
+    _propagateUndefinedExpressions(_removeLastScope());
+    bool hasBreak = DartTypeUtilities
```
bwilkerson: It would be good to add a comment here related to the handling of break statements so that we remember the issues.

lib/src/util/condition_scope_visitor.dart
```diff
@@ -0,0 +1,307 @@
...
+        .traverseNodesInDFS(node, excludeCriteria: _isLoopStatement)
```
bwilkerson: Given that we're not handling break statements with labels correctly, I think that excluding loop statements will cause false positives. Consider:
```
loop: for (...) {
  while (...) {
    break loop;
  }
}
```
Also, not excluding function bodies might well cause false negatives. Consider:
```
while (...) {
  void func() {
    while (...) {
      break;
    }
  }
}
```

lib/src/util/condition_scope_visitor.dart
```diff
@@ -0,0 +1,307 @@
...
+    }
+  }
+
+  @override
+  visitIfStatement(IfStatement node) {
+    final elseScope = _visitElseStatement(node.elseStatement, node.condition);
+    _visitIfStatement(node);
+    _propagateUndefinedExpressions(elseScope);
+    if (_isLastStatementAnExitStatement(node.thenStatement)) {
+      _addFalseCondition(node.condition);
```
bwilkerson: What happens with the following (pathological) case:
```
if (x == null) {
  return false;
} else {
  return true;
}
if (x == null) {
  ...
}
```
(and with the sense of the conditions reversed)?

I think we only want to record either true or false conditions only when exactly one block terminates.

lib/src/util/condition_scope_visitor.dart
```diff
@@ -0,0 +1,307 @@
...
+
+  Iterable<Expression> _getExpressions(Iterable<Element> elements,
+      {bool value: true}) {
+    return outerScope.getExpressions(elements, value);
+  }
+
+  bool _isLastStatementAnExitStatement(Statement statement) {
+    if (statement is Block) {
+      return _isLastStatementAnExitStatement(
+          DartTypeUtilities.getLastStatementInBlock(statement));
```
bwilkerson: Does ExitDetector not handle blocks? (I'm wondering whether you need to special case blocks here.)

If you do need to special case blocks, I think you need to search for a statement within the block that can exit in order to handle pathological cases such as
```
if (...) {
  return;
  print("Shouldn't get here");
}
```
Otherwise we'll think that the last statement means that we can fall through.

lib/src/util/condition_scope_visitor.dart
JPaulsen: In that case the last if is dead code, in the environment we would see that (x == null) is true and false at the same time, and probably would lint the last if. But it still correct that if a then statement is return the condition is false after de if/else block the same if the else is a return block. There is only one exception when you have a label after the if/else statement, in that case, when we see a label, we should undefine all expression, what do you think?

lib/src/util/condition_scope_visitor.dart
JPaulsen: Does the analyzer find the dead code there?, I think in that case we would be doing something that is already done

lib/src/util/condition_scope_visitor.dart
bwilkerson: I don't understand the case you're thinking of. Can you give me an example involving a label after an if statement?

That said, I do think that we might want to forget everything we think we know when we hit a label because we don't know what the state of any variable will be when execution reaches the break or continue that will transfer control to that label. Unfortunately, the same is true for any do, for, switch or while statement because they all have an implicit empty label.

lib/src/util/condition_scope_visitor.dart
bwilkerson: Yes, analyzer will find and report the dead code. The question is what we need to do in order to avoid false positives for cases like these.

JPaulsen: An easy fix would be add an UndefinedAllExpressions when we see a return statement :)

lib/src/util/condition_scope_visitor.dart
JPaulsen: Yes, think in something like this:

```dart
if (x == null) {
  return;
}
// x == null is false from here.
myLabel: 
// we do not know anything about x == null from here
```

lib/src/util/condition_scope_visitor.dart
JPaulsen: About this:

```dart
while (...) {
  void func() {
    while (...) {
      break;
    }
  }
}
```

The first while search for breaks excluding other for and while, so in that case, the break won't be see because is inside other while, an option would be something.

```dart
while (...) {
  void func() {
    break;
  }
}
```

But that is not possible, the analyzer would say 'A break statement cannot be used outside of a loop or switch statement'

bwilkerson: Ok, replace the body of the function with a switch statement and stick a break in one case. Then I think we can get to the break when we shouldn't.

lib/src/util/condition_scope_visitor.dart
bwilkerson: A return, a throw, or a rethrow (assuming I'm not forgetting anything).

As long as we've thought hard about possible false positives and worked to prevent them, and document in the code how we've done that, I'm fine with most solutions. (Including undefining everything after an exit from the current function body.)

lib/src/util/condition_scope_visitor.dart
JPaulsen: Yes, that is a bug that I made, I should exclude loop statements and switch statements, and that's it

JPaulsen:
```dart
loop: for (...) {
  while (...) {
    break loop;
  }
}
```

I have been thinking a lot about this, can you help me?

How do you think this would produce a false positive?

lib/src/util/condition_scope_visitor.dart
```diff
@@ -0,0 +1,307 @@
...
+  Iterable<Expression> getExpressions(Iterable<Element> elements, bool value) {
+    final expressions = <Expression>[];
+    _recursiveGetExpressions(this, expressions, elements, value);
+    return expressions;
+  }
+
+  Iterable<_ExpressionBox> getUndefinedExpressions() =>
+      environment.where((e) => e is _UndefinedExpression);
+
+  static void _recursiveGetExpressions(ConditionScope conditionScope,
```
JPaulsen: Fixed

lib/src/util/condition_scope_visitor.dart
JPaulsen: Fixed

lib/src/util/condition_scope_visitor.dart
JPaulsen: Fixed

lib/src/util/condition_scope_visitor.dart
JPaulsen: Fixed, I will work in documentation now, Alexei will help me also.

lib/src/util/condition_scope_visitor.dart
```diff
@@ -0,0 +1,325 @@
...
+  }
+
+  @override
+  visitRethrowExpression(RethrowExpression node) {
+    node.visitChildren(this);
+    _addElementToEnvironment(new _UndefinedAllExpression());
+  }
+
+  @override
+  visitReturnStatement(ReturnStatement node) {
```
JPaulsen: I have to create test for this, and throw and rethrow

lib/src/util/condition_scope_visitor.dart
bwilkerson: My understanding is that `hasBreak` should be `true` (in both the `for` and `while` cases) if there is any way for control to reach the statement following the loop other than for the condition to evaluate to `false`. But if you don't look inside the nested `while` loop (in the example above) then you won't find the `break` statement that applies to the `for` loop and `hasBreak` will be `false` despite the fact that the condition doesn't have to be `false`. Here's a more concrete example:
```
String s = '...';
loop: for (; s != null; s += '.') {
  while (true != false) {
    break loop;
  }
}
// I think there will be a lint on the following line,
// even though the test is reasonable.
if (s != null) {
  print(s);
}
```

JPaulsen: Exactly, in this code there will be a lint there:

```dart
String s = '...';
loop: for (; s != null; s += '.') {
  while (true != false) {
    break loop;
  }
}
if (s != null) { // LINT
  print(s);
}
```

And it is because in that if s != null will be always false

JPaulsen: Woooow!! I see now, I thought a break with label worked like a goto, I am so sorry, my bad, I understood perfectly now, and I will work fixing this

JPaulsen: Fixed!

lib/src/util/condition_scope_visitor.dart
JPaulsen: Forget what I said, in that case x == null is true and false in the environment, so it will be resolved to a lint, but we can easily detect those kind of cases, because they are dead code and do not lint if you want, what do you think is better lint or not to lint?

bwilkerson: It is better to have one diagnostic ("dead code") than to have multiple diagnostics for the same problem. That's why I think we shouldn't produce a lint in this case.

lib/src/util/condition_scope_visitor.dart
JPaulsen: Fixed

lib/src/util/condition_scope_visitor.dart
bwilkerson: Does the overridden method do exactly what this one does? (I think it does, but it's hard for me to check right now.) If so, let's just delete this method.

JPaulsen: This method made this Visitor have the behavior of a RecursiveAstVisitor, I deleted the method and replace the extends UnifyingAstVisitor by extends RecursiveAstVisitor.

lib/src/util/condition_scope_visitor.dart
JPaulsen: Fixed

lib/src/util/condition_scope_visitor.dart
JPaulsen: Fixed

lib/src/util/condition_scope_visitor.dart
JPaulsen: Replicated issue

lib/src/util/condition_scope_visitor.dart
JPaulsen: Fixed, PTAL

lib/src/util/condition_scope_visitor.dart
JPaulsen: @bwilkerson this is my last issue, and I can not figure out a test where the rule do something and without checking adding null elements would do something different. (I mean lints)

pq:
@bwilkerson : could you take a quick look at @JPaulsen's latest?  Looks like the main issues are addressed.  Thanks in advance!

lib/src/util/condition_scope_visitor.dart
bwilkerson: It's quite possible that having an instance of `_UndefinedExpression` whose element is `null` is harmless. It just wasn't clear to me that that would be the case.

JPaulsen: Yes it is harmless, but other things like:

```dart
if (notDefined < 2) {
  if (notDefined < 2) { // there is a lint here
  }
}
```
can decrease the UX level, because there will be a linter and an error in the same place, I fixed this locally but it is necessary to check for types and it is really expensive (the benchmark in the rule moved from ~55 to >140), so I think it is a cost I would pay to have a faster linter as a user.

If you think this is OK the branch is ready to merge, if you want me to fix this even that means have a worse performance just tell me and I will do it.

bwilkerson: I think I'll merge it, and we can re-visit this after the fact.

PR #507, opened 2017-03-20
Fix rule: unnecessary_this

JPaulsen:
Fixes https://github.com/dart-lang/linter/issues/502

### DIFF
--- a/lib/src/rules/unnecessary_this.dart
+++ b/lib/src/rules/unnecessary_this.dart
@@ -85,11 +85,14 @@ class _UnnecessaryThisVisitor extends ScopedVisitor {
       lookUpElement = DartTypeUtilities.getCanonicalElement(
           nameScope.lookup(parent.methodName, definingLibrary));
       localElement = parent.methodName.bestElement;
-    } else {
+    }
+    // Error in code
+    if (localElement == null) {
       return null;
     }
-
-    if (lookUpElement == localElement) {
+    // If localElement was resolved, but lookUpElement was not, that means
+    // the element is defined in an ancestor class.
+    if (lookUpElement == localElement || lookUpElement == null) {
       rule.reportLint(parent);
     }
   }
--- a/test/rules/unnecessary_this.dart
+++ b/test/rules/unnecessary_this.dart
@@ -4,6 +4,34 @@
 
 // test w/ `pub run test -N unnecessary_this`
 
+void bar() {}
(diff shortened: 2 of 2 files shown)


test/rules/unnecessary_this.dart
```diff
@@ -4,6 +4,32 @@
...
+class Parent {
+  int x, y;
+  void bar() {
+
+  }
+}
+
+class Child extends Parent {
+  void foo() {
+    this.bar(); // LINT
```
bwilkerson: This is wrong. Without the `this`, `bar()` will refer to the top-level function, which is clearly a change is semantics.

test/rules/unnecessary_this.dart
```diff
@@ -4,6 +4,32 @@
...
+  }
+  void baz(int a) {
+    this.x = a; // LINT
```
bwilkerson: This is similarly wrong. Without the `this`, `x` will refer to the top-level variable.

test/rules/unnecessary_this.dart
```diff
@@ -4,6 +4,32 @@
...
+  }
+  void assignY(int a) {
+    this.y = a; // LINT
```
bwilkerson: This one, on the other hand, is valid.

JPaulsen:
@bwilkerson you are right, a field shadow a TopLevelVariable, but a TopLevelVariable shadow a super field, what a tricky behavior

JPaulsen:
@bwilkerson I can not comment your comments, I do not know why, but the error was fixed

alexeieleusis:
Friendly ping! No open issues in this PR.

lib/src/rules/unnecessary_this.dart
```diff
@@ -85,11 +85,12 @@ class _UnnecessaryThisVisitor extends ScopedVisitor {
...
       localElement = parent.methodName.bestElement;
-    } else {
+    }
+    // Error in code
+    if (localElement == null) {
       return null;
     }
-
-    if (lookUpElement == localElement) {
+    if (lookUpElement == localElement || lookUpElement == null) {
```
bwilkerson: Please add a comment indicating the conditions under which `lookUpElement` will be `null`. Given that we resolved it to something with the `this` (the `localElement` is not `null`), I can't think of a case where look-up would fail.

PR #508, opened 2017-03-20
Fix rule: prefer_final_fields

JPaulsen:
Fixes https://github.com/dart-lang/linter/issues/506

### DIFF
--- a/lib/src/rules/prefer_final_fields.dart
+++ b/lib/src/rules/prefer_final_fields.dart
@@ -7,7 +7,6 @@ library linter.src.rules.prefer_final_fields;
 import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/standard_resolution_map.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
-import 'package:analyzer/dart/element/element.dart';
 import 'package:linter/src/analyzer.dart';
 import 'package:linter/src/util/dart_type_utilities.dart';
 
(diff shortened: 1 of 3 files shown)


lib/src/rules/prefer_final_fields.dart
```diff
@@ -64,6 +64,27 @@ class GoodMutable {
...
+          identifier.bestElement as PropertyAccessorElement;
+      if (bestElement.variable != variable.element) {
+        return false;
+      }
+
+      return identifier.getAncestor((a) =>
+              a is AssignmentExpression ||
+              a is PrefixExpression ||
+              a is PostfixExpression) !=
+          null;
```
JPaulsen: If a variable can only be reasigned in an AssignmentExpression, PrefixExpression and PostfixExpression, this is totally right, the bug was think that can only be reasigned in an AssignmentExpression or ExpressionStatement.

lib/src/rules/prefer_final_fields.dart
```diff
@@ -64,6 +64,27 @@ class GoodMutable {
...
+        .any((n) {
+      SimpleIdentifier identifier = n as SimpleIdentifier;
+      PropertyAccessorElement bestElement =
+          identifier.bestElement as PropertyAccessorElement;
+      if (bestElement.variable != variable.element) {
+        return false;
+      }
+
+      return identifier.getAncestor((a) =>
+              a is AssignmentExpression ||
```
bwilkerson: Kind of seems like we want to also ensure that the identifier is on the left-hand side of the assignment. For example, `var local; local = _field;` Doesn't mean that `_field` is modified. Similarly `array[_i] = 0;` doesn't mean that `_i` is modified.

lib/src/rules/prefer_final_fields.dart
bwilkerson: Assuming this was in response to my comment on line 82 (in case that assumption is wrong).

The point was to prevent false negatives. Just because a field is referenced somewhere inside an assignment expression doesn't mean that it is being modified.

lib/src/rules/prefer_final_fields.dart
JPaulsen: Fixed and test added

lib/src/rules/prefer_final_fields.dart
JPaulsen: It was not an answer to that comment, it was a previous comment to yours. [This](https://github.com/dart-lang/linter/pull/508#discussion_r107043160) was my answer to your comment in line 82.

lib/src/rules/prefer_final_fields.dart
```diff
@@ -64,6 +63,25 @@ class GoodMutable {
 ```
 ''';
 
+bool _isMutated(
+        VariableDeclaration variable, CompilationUnit compilationUnit) =>
+    DartTypeUtilities
+        .traverseNodesInDFS(compilationUnit)
+        .where((n) =>
```
bwilkerson: Would `any` allow us to short circuit the traversal as soon as we found one matching node?

JPaulsen: Fixed

Issue #518, opened 2017-03-23
Implement DO use rethrow to rethrow a caught exception.


JPaulsen:
From [Effective Dart](https://www.dartlang.org/guides/language/effective-dart/usage#do-use-rethrow-to-rethrow-a-caught-exception)

PR #519, opened 2017-03-23
New rule: use_rethrow_when_possible

JPaulsen:
Fixes https://github.com/dart-lang/linter/issues/518

### DIFF
--- a/test/rules/use_rethrow_when_possible.dart
+++ b/test/rules/use_rethrow_when_possible.dart
@@ -1,0 +1,38 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// test w/ `pub run test -N use_rethrow_when_possible`
+
+void bad1() {
+  try {} catch (e) {
+    throw e; // LINT
+  }
+}
+
+void bad2() {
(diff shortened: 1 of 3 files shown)


test/rules/use_rethrow_when_possible.dart
```diff
@@ -0,0 +1,30 @@
...
+  }
+}
+
+void bad2() {
+  try {} catch (e, stackTrace) {
+    print(stackTrace);
+    throw e; // LINT
+  }
+}
+
```
bwilkerson: Please add this test case:
```
void bad3() {
  try {} catch (e) {
    try {} catch (f) {
      throw e; // LINT
    }
  }
}
```
unless you think this shouldn't produce a lint.

JPaulsen: I think it should not produce a lint, but I can add this test:

```dart
void good3() {
  try {} catch (e) {
    try {} catch (f) {
      throw e; // OK
    }
  }
}
```

bwilkerson: Ah, you're right. I forgot that rethrow implicitly applies to the exception caught by the most closely surrounding catch clause.

JPaulsen: Ok, thanks for the review, let me add the test.

JPaulsen: Fixed (test good3 added)

PR #521, opened 2017-03-23
New rule: do_not_catch_errors

JPaulsen:
Fixes https://github.com/dart-lang/linter/issues/520
Fixes https://github.com/dart-lang/linter/issues/526

### DIFF
--- a/test/rules/do_not_catch_errors.dart
+++ b/test/rules/do_not_catch_errors.dart
@@ -1,0 +1,36 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// test w/ `pub run test -N do_not_catch_errors`
+
+void bad() {
+  try {} on A catch (e) { // LINT
+    // ignore
+  } on B catch (e) { // LINT
+    // ignore
+  } on C { // LINT
+    // ignore
+  } on D { // LINT
+    // ignore
+  } on Error catch (e) { // LINT
+    // ignore
+  } on Exception catch (e) { // OK
+    // ignore
+  } on String catch (e) { // OK
+    // ignore
+  } catch (e) {// OK
+    // ignore
+  }
+}
+
+class A extends B {}
+
+class B implements Error {
+  @override
+  StackTrace get stackTrace => null;
+}
+
+class C extends D {}
+
+class D extends Error {}
--- a/lib/src/rules.dart
+++ b/lib/src/rules.dart
@@ -23,6 +23,7 @@ import 'package:linter/src/rules/comment_references.dart';
 import 'package:linter/src/rules/constant_identifier_names.dart';
 import 'package:linter/src/rules/control_flow_in_finally.dart';
 import 'package:linter/src/rules/directives_ordering.dart';
+import 'package:linter/src/rules/do_not_catch_errors.dart';
 import 'package:linter/src/rules/empty_catches.dart';
 import 'package:linter/src/rules/empty_constructor_bodies.dart';
 import 'package:linter/src/rules/empty_statements.dart';
@@ -98,6 +99,7 @@ void registerLintRules() {
     ..register(new ControlFlowInFinally())
     ..registerDefault(new ConstantIdentifierNames())
     ..register(new DirectivesOrdering())
+    ..register(new DoNotCatchErrors())
     ..register(new EmptyCatches())
     ..registerDefault(new EmptyConstructorBodies())
     ..register(new EmptyStatements())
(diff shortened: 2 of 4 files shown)


JPaulsen:
@bwilkerson @pq I have a problem with this rule, if a run the linter with this rule in the terminal, it works fine, but when I do pub run test -N do_not_catch_errors the test does not pass because it is not linting anything, do you know why is this happening? Do I need to create an integration test? If that is the case, why?

This is the error message:

```
00:01 +0: test/rule_test.dart: rule dart do_not_catch_errors                                       /usr/local/google/home/casanueva/Development/linter/test/rules/do_not_catch_errors.dart 26:20 [compile time error] Classes can only implement other classes
class A implements Error {
                   ^^^^^
/usr/local/google/home/casanueva/Development/linter/test/rules/do_not_catch_errors.dart 31:17 [compile time error] Classes can only extend other classes
class B extends Error {}
                ^^^^^

/usr/local/google/home/casanueva/Development/linter/test/rules/do_not_catch_errors.dart 12:8 [static warning] The name 'Error' is not a type and cannot be used in an on-catch clause

  } on Error catch (e) { // LINT
       ^^^^^
/usr/local/google/home/casanueva/Development/linter/test/rules/do_not_catch_errors.dart 15:6 [static warning] The name 'Exception' is not a type and cannot be used in an on-catch clause
  on Exception catch(e) { // OK
     ^^^^^^^^^
/usr/local/google/home/casanueva/Development/linter/test/rules/do_not_catch_errors.dart 26:20 [static warning] Undefined class 'Error'
class A implements Error {
                   ^^^^^
/usr/local/google/home/casanueva/Development/linter/test/rules/do_not_catch_errors.dart 31:17 [static warning] Undefined class 'Error'
class B extends Error {}
                ^^^^^

null files analyzed, 6 issues found, in null ms.

00:01 +0 -1: test/rule_test.dart: rule dart do_not_catch_errors [E]                                
  Expected: matches [Annotation:<[LINT]: "null" (line: 8) - [null:null]>, Annotation:<[LINT]: "null" (line: 10) - [null:null]>, Annotation:<[LINT]: "null" (line: 12) - [null:null]>] unordered
    Actual: []
     Which: has too few elements (0 < 3)
  
  test/rule_test.dart 445:7  testRule.<fn>
  
00:01 +0 -1: Some tests failed.     
```

JPaulsen:
In the terminal:

```
$dart bin/linter.dart ./test/rules/do_not_catch_errors.dart --rules=do_not_catch_errors

/usr/local/google/home/casanueva/Development/linter/test/rules/do_not_catch_errors.dart 8:10 [lint] Don’t explicitly catch Error or types that implement it.
  try {} on A catch (e) { // LINT
         ^^^^^^^^^^^^^^^^^^^^^^^^
/usr/local/google/home/casanueva/Development/linter/test/rules/do_not_catch_errors.dart 10:5 [lint] Don’t explicitly catch Error or types that implement it.
  } on B catch (e) { // LINT
    ^^^^^^^^^^^^^^^^^^^^^^^^
/usr/local/google/home/casanueva/Development/linter/test/rules/do_not_catch_errors.dart 12:5 [lint] Don’t explicitly catch Error or types that implement it.
  } on Error catch (e) { // LINT
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^

1 file analyzed, 3 issues found, in 2121 ms.
```

pq:
My quick guess is that it has to do with the way we mock SDK classes.  I'm thinking we don't have enough of `Error` there for your lint to run against the mocked SDK?

pq:
In particular, you might poke around in: 

https://github.com/dart-lang/linter/blob/master/test/mock_sdk.dart

test/rules/do_not_catch_errors.dart
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// test w/ `pub run test -N do_not_catch_errors`
+
+void bad() {
+  try {} on A catch (e) { // LINT
+    // ignore
+  } on B catch (e) { // LINT
```
bwilkerson: Add the following clause to this `try` statement:
```
} on C { // LINT
  // ignore
```
and the following declaration:
```
class C extends Error {}
```

JPaulsen:
@pq  it seems to be that, class Error is in dart.core but it is not defined in mock_sdk, what should I do to fix this?, can I add it in the file?

test/rules/do_not_catch_errors.dart
JPaulsen: Fixed, I created 3 cases with classes A, B and C. PTAL

bwilkerson: Not exactly what I suggested: you added some extra text. :-) The point of the test case I gave is that in Dart the `catch` clause is optional if you have an `on` clause, and if the `on` clause specifies a subclass of `Error` then I think you still want to produce a lint.

(Having a test case with a class that indirectly inherits from `Error` is a good enhancement.)

JPaulsen: Oh I got it now, I will fix this, thank you.

JPaulsen: Fixed :)

pq:
> Tell me if it's ok or I have to do something else, thank you for your help.

Nope.  That was it.  Thanks!

Issue #528, opened 2017-03-24
Implement DON’T check for null in custom == operators.


JPaulsen:
From [Effective Dart](https://www.dartlang.org/guides/language/effective-dart/design#dont-check-for-null-in-custom--operators)

PR #529, opened 2017-03-24
New rule: do_not_check_for_null_in_equality_operators

JPaulsen:
Fixes https://github.com/dart-lang/linter/issues/528


JPaulsen:
@bwilkerson before merge this, I am thinking about checking things this should also be linted:

```dart
@override
operator ==(other) => other?.name == name;
```

What do you think?

bwilkerson:
Yes, I think we should lint that too. Also consider any expression of the form `other ?? ...`.

JPaulsen:
@bwilkerson fixed, PTAL

PR #532, opened 2017-03-24
New rule: avoid_positional_boolean_parameters

JPaulsen:
Fixes https://github.com/dart-lang/linter/issues/531

### DIFF
--- a/lib/src/rules/avoid_positional_boolean_parameters.dart
+++ b/lib/src/rules/avoid_positional_boolean_parameters.dart
@@ -1,0 +1,103 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+library linter.src.rules.avoid_positional_boolean_parameters;
+
+import 'package:analyzer/analyzer.dart';
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:linter/src/analyzer.dart';
+import 'package:linter/src/util/dart_type_utilities.dart';
+
+const _desc = r'Avoid positional boolean parameters.';
+
+const _details = r'''
+
+**AVOID** positional boolean parameters.
+
+**BAD:**
+```
+new Task(true);
+new Task(false);
+new ListBox(false, true, true);
+new Button(false);
+```
+
+**GOOD:**
+```
+new Task.oneShot();
+new Task.repeating();
+new ListBox(scroll: true, showScrollbars: true);
+new Button(ButtonState.enabled);
+```
+
+''';
+
+bool _hasInheritedMethod(MethodDeclaration node) {
+  return DartTypeUtilities.lookUpInheritedMethod(node) != null;
+}
+
+bool _isNamedParameter(FormalParameter node) =>
+    node.kind == ParameterKind.NAMED;
+
+class AvoidPositionalBooleanParameters extends LintRule {
+  _Visitor _visitor;
+  AvoidPositionalBooleanParameters()
+      : super(
+            name: 'avoid_positional_boolean_parameters',
+            description: _desc,
+            details: _details,
+            group: Group.style,
+            maturity: Maturity.experimental) {
+    _visitor = new _Visitor(this);
+  }
+
+  @override
+  AstVisitor getVisitor() => _visitor;
+}
+
+class _Visitor extends SimpleAstVisitor {
+  final LintRule rule;
+  _Visitor(this.rule);
+
+  @override
+  visitConstructorDeclaration(ConstructorDeclaration node) {
+    if (!node.element.isPrivate) {
+      final parametersToLint =
+          node.parameters?.parameters?.where(_isFormalParameterToLint);
+      if (parametersToLint?.isNotEmpty == true) {
+        rule.reportLint(parametersToLint.first);
+      }
+    }
+  }
+
+  @override
+  visitFunctionDeclaration(FunctionDeclaration node) {
+    if (!node.element.isPrivate) {
+      final parametersToLint = node.functionExpression.parameters?.parameters
+          ?.where(_isFormalParameterToLint);
+      if (parametersToLint?.isNotEmpty == true) {
+        rule.reportLint(parametersToLint.first);
+      }
+    }
+  }
+
+  @override
+  visitMethodDeclaration(MethodDeclaration node) {
+    if (!node.isSetter &&
+        !node.element.isPrivate &&
+        !_hasInheritedMethod(node)) {
+      final parametersToLint =
+          node.parameters?.parameters?.where(_isFormalParameterToLint);
+      if (parametersToLint?.isNotEmpty == true) {
+        rule.reportLint(parametersToLint.first);
+      }
+    }
+  }
+
+  bool _isFormalParameterToLint(FormalParameter node) =>
+      DartTypeUtilities.implementsInterface(
+          node.identifier.bestType, 'bool', 'dart.core') &&
+      !_isNamedParameter(node);
+}
--- a/test/rules/avoid_positional_boolean_parameters.dart
+++ b/test/rules/avoid_positional_boolean_parameters.dart
@@ -1,0 +1,80 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// test w/ `pub run test -N avoid_positional_boolean_parameters`
+
+class Library {
+  void checkRegularBook(String name) {
+    _checkBook(name, false);
+  }
+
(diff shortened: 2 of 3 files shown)


test/rules/avoid_positional_boolean_parameters.dart
```diff
@@ -0,0 +1,40 @@
...
+  }
+}
+
+class D {
+  bool value;
+  D.good({this.value}); // OK
+
+  D.bad(this.value); // LINT
+
+}
```
bwilkerson: There should be at least one test that uses optional positional parameters
```
void bad([bool value]) {}
```
and at least one test with a setter
```
void set good(bool value) {}
```

JPaulsen: Fixed

JPaulsen:
@bwilkerson I run it in the SDK, a lot of lints with this form:

```
[lint] Avoid positional boolean parameters.
      [bool lowerOpen = false, bool upperOpen = false]) =>
       ^^^^^^^^^^^^^^^^^^^^^^
 [lint] Avoid positional boolean parameters.
      [bool lowerOpen = false, bool upperOpen = false]) =>
```

lib/src/rules/avoid_positional_boolean_parameters.dart
```diff
@@ -0,0 +1,83 @@
...
+new Task.oneShot();
+new Task.repeating();
+new ListBox(scroll: true, showScrollbars: true);
+new Button(ButtonState.enabled);
+```
+
+''';
+
+bool _isNamedParameter(FormalParameter node) =>
+    node is DefaultFormalParameter && node.kind == ParameterKind.NAMED;
```
bwilkerson: You could drop the `is` test here. All formal parameters can return their kind.

bwilkerson:
> I think the lint is working fine, but I think there is one case that I could not be easy to solved, for example using a framework that needs a lambda with a bool parameter in it, so you can not fix that lint easily, what do you think?

I don't know of any reason why the code that specifies the lambda couldn't specify the parameter as a named parameter.

But I think there are a lot of places where boolean-valued parameters are required, and that there is a lot of existing code that doesn't make use of the `@required` annotation (often because it was written before the annotation was introduced) and maybe even good reasons not to make all boolean-valued parameters be named parmaeters. If existing functions/methods are part of the public API of a package, then it's a breaking change to "fix" it. So I'm concerned that the number of false positives make this a rule that won't provide much value to users.

alexeieleusis:
> But I think there are a lot of places where boolean-valued parameters are required, ..., then it's a breaking change to "fix" it. So I'm concerned that the number of false positives make this a rule that won't provide much value to users.

We should try to avoid those false positives, though those breaking changes are for projects that are used as dependencies. The rule itself should apply to application code where is more likely violations will result in a bug.

lib/src/rules/avoid_positional_boolean_parameters.dart
JPaulsen: Fixed

JPaulsen:
@bwilkerson I expressed so bad last time, sorry about that, I was trying to say that sometimes a closure that is passed as an argument of a function does not have an easy lint fix, for example this:

```dart
void foo() {
  final array = <bool>[true, false, true, false];
  array.where((e) => e);
}
```

Actually the rule is linting the ```(e) => e``` and does not have an easy fix.

Notice that this code does not make much sense, but it is totally possible that somebody creates a function that receive a closure with a positional boolean parameter and in that case it is not easy to fix the lint.

lib/src/rules/avoid_positional_boolean_parameters.dart
```diff
@@ -0,0 +1,83 @@
...
+    node.parameters?.parameters?.forEach(_visitFormalParameter);
+  }
+
+  @override
+  visitFunctionExpression(FunctionExpression node) {
+    node.parameters?.parameters?.forEach(_visitFormalParameter);
+  }
+
+  @override
+  visitMethodDeclaration(MethodDeclaration node) {
```
alexeieleusis: Do not lint private functions/methods https://martinfowler.com/bliki/FlagArgument.html

JPaulsen: Fixed

bwilkerson:
> ... I was trying to say that sometimes a closure that is passed as an argument of a function does not have an easy lint fix ...

You're right.

In this case the user has no choice but to write a closure that takes a bool parameter. Perhaps a more realistic case would be
```
Map<Something, bool> propertyMap = ...;
propertyMap.forEach((key, value) { ... });
```
We should not produce a lint when the user has no available alternatives. Closures passed as arguments to methods is one good example. This, unfortunately, gets quite complex. Consider
```
Map<Something, bool> propertyMap = ...;
void mapProcessor(key, value) { ... }
propertyMap.forEach(mapProcessor);
```
I'd argue that the user has no choice but to use a bool parameter because of the function being passed in to a method later on. But what if the user defines `mapProcessor` as a method rather than a local function? Because of it's intended use, they have no choice but to use a positional parameter.

To avoid all false positives we'd need to look at all uses of every function, even though we often won't have access to those uses. We either need to not implement this rule, or we need to choose to remove a reasonable number of false positives. Where do you think we should draw the line?

By the way, overriding an inherited method is another example of a place where there is no choice and shouldn't produce a lint. Are there other places I haven't thought of yet?

JPaulsen:
@alexeieleusis and me think that the best we can do is to lint only public methods, non-anonymous functions and constructors excluding setters and inherited methods.

This does not lint the false positives above, and if you create a method to do something like:

```dart
propertyMap.forEach(mapProcessor);
```

Then mapProcessor should be private. 

We also think that this rule should be marked as experimental.

bwilkerson:
That seems like a reasonable set of choices.

JPaulsen:
Ok, @bwilkerson I already implemented and added the tests, but I do not know how to mark it as experimental, can you help me with that?

bwilkerson:
You include the `maturity:` argument in the super constructor invocation in your rule's constructor. Take a look at `lib/src/rules/literal_only_boolean_expressions.dart` for an example.

JPaulsen:
Ok, all the changes are done, PTAL.

lib/src/rules/avoid_positional_boolean_parameters.dart
```diff
@@ -0,0 +1,95 @@
...
+}
+
+class _Visitor extends SimpleAstVisitor {
+  final LintRule rule;
+  _Visitor(this.rule);
+
+  @override
+  visitConstructorDeclaration(ConstructorDeclaration node) {
+    if (!node.element.isPrivate) {
+      node.parameters?.parameters?.forEach(_visitFormalParameter);
```
JPaulsen: I should not lint more than one parameter in the same function, constructor or method.

JPaulsen: Fixed

lib/src/rules/avoid_positional_boolean_parameters.dart
```diff
@@ -0,0 +1,103 @@
...
+```
+new Task.oneShot();
+new Task.repeating();
+new ListBox(scroll: true, showScrollbars: true);
+new Button(ButtonState.enabled);
+```
+
+''';
+
+bool _hasInheritedMethod(MethodDeclaration node) {
```
bwilkerson: This method appears in another rule (or more). Perhaps in a separate CL move them into `DartTypeUtilities`.

lib/src/rules/avoid_positional_boolean_parameters.dart
```diff
@@ -0,0 +1,103 @@
...
+
+  @override
+  visitConstructorDeclaration(ConstructorDeclaration node) {
+    if (!node.element.isPrivate) {
+      final parametersToLint =
+          node.parameters?.parameters?.where(_isFormalParameterToLint);
+      if (parametersToLint?.isNotEmpty == true) {
+        rule.reportLint(parametersToLint.first);
+      }
+    }
```
bwilkerson: Lines 67-72 appear in three places. Consider a follow-on CL to add a utility method.

test/rules/avoid_positional_boolean_parameters.dart
```diff
@@ -0,0 +1,75 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// test w/ `pub run test -N avoid_positional_boolean_parameters`
+
```
bwilkerson: There aren't any tests of methods that override another method.

JPaulsen: Fixed

lib/src/rules/avoid_positional_boolean_parameters.dart
JPaulsen: Issue created

lib/src/rules/avoid_positional_boolean_parameters.dart
JPaulsen: Issue created

PR #536, opened 2017-03-28
New rule: prefer_constructors_instead_of_static_methods

JPaulsen:
Fixes https://github.com/dart-lang/linter/issues/535

### DIFF
--- a/lib/src/rules/prefer_constructors_instead_of_static_methods.dart
+++ b/lib/src/rules/prefer_constructors_instead_of_static_methods.dart
@@ -1,0 +1,83 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+library linter.src.rules.prefer_constructors_instead_of_static_methods;
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/type.dart';
+import 'package:linter/src/analyzer.dart';
+import 'package:linter/src/util/dart_type_utilities.dart';
+
+const _desc =
+    r'Prefer defining constructors instead of static methods to create instances.';
+
+const _details = r'''
+
+**PREFER** defining constructors instead of static methods to create instances.
+
+**BAD:**
+```
+class Point {
+  num x, y;
+  Point(this.x, this.y);
+  static Point polar(num theta, num radius) {
+    return new Point(radius * math.cos(theta),
+        radius * math.sin(theta));
+  }
+}
+```
+
+**GOOD:**
+```
+class Point {
+  num x, y;
+  Point(this.x, this.y);
+  Point.polar(num theta, num radius)
+      : x = radius * math.cos(theta),
+        y = radius * math.sin(theta);
+}
+```
+
+''';
+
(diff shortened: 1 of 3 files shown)


lib/src/rules/prefer_constructors_instead_of_static_methods.dart
```diff
@@ -0,0 +1,87 @@
...
+  final LintRule rule;
+  _Visitor(this.rule);
+
+  @override
+  visitMethodDeclaration(MethodDeclaration node) {
+    if (!node.isStatic) {
+      return;
+    }
+    final returnType = node.returnType?.type;
+    if (node.isStatic &&
```
bwilkerson: You tested for this on line 76.

lib/src/rules/prefer_constructors_instead_of_static_methods.dart
```diff
@@ -0,0 +1,87 @@
...
+        !node.isGetter &&
```
bwilkerson: Why does this exclude getters?
```
static Point get origin => new Point(0, 0);
```

lib/src/rules/prefer_constructors_instead_of_static_methods.dart
```diff
@@ -0,0 +1,87 @@
...
+        returnType == (node.parent as ClassDeclaration).element.type &&
+        _hasNewInvocation(returnType, node.body)) {
```
bwilkerson: Is this the right criteria?

Does there need to be an instance creation?
```
static Point _origin = new Point(0, 0);
static Point getOrigin() => _origin;
```
Is it OK if the created instance is not returned, or if other values (such as instances of subclasses) are returned?

lib/src/rules/prefer_constructors_instead_of_static_methods.dart
JPaulsen: Fixed

JPaulsen: (And test added)

lib/src/rules/prefer_constructors_instead_of_static_methods.dart
JPaulsen: Do you think we should be lint those cases?, there are a lot of lints like this in the sdk when I delete that check:

```dart
static InternetAddress get LOOPBACK_IP_V4 { 
  return somethingPrecalculated;
}
```

Maybe it could be new InternetAddress.LOOPBACK_IP_V4;

But what do you think about this?:

```dart
static FileSystemEntityType _lookup(int type) => _typeList[type];
```

bwilkerson: > Do you think we should lint those cases?

I don't know. If there are lots of lints, then probably not. It's usually better to start with a narrow lint that only catches cases we know should be caught and then catch more cases based on user feedback, but I was looking for your opinion. Basically I wanted to know whether you'd considered those cases.

Personally, I think the cases you found are fine and shouldn't be linted.

bwilkerson: I do wonder whether we should only lint when the newly created instance is returned, but maybe that case doesn't occur in practice and isn't worth worrying about. What do you think?

JPaulsen: I think is better not to care about that, I know we have to take care about false positives, but in this case it should be so strange, and probably with a deep reflection the user should realize that it is better to use a factory constructor instead a static method.

PR #538, opened 2017-03-28
New rule: declare_const_constructor_when_possible

JPaulsen:
Fixes https://github.com/dart-lang/linter/issues/537

### DIFF
--- a/lib/src/rules/declare_const_constructor_when_possible.dart
+++ b/lib/src/rules/declare_const_constructor_when_possible.dart
@@ -1,0 +1,104 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+library linter.src.rules.declare_const_constructor_when_possible;
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/element.dart';
+import 'package:linter/src/analyzer.dart';
+
+const _desc =
+    r'Consider making your constructor const if the class supports it.';
+
+const _details = r'''
+
+**CONSIDER** making your constructor const if the class supports it.
+
+**BAD:**
+```
+class Bad {
+  final int foo;
+
+  Bad() : foo = 1;
+}
+```
+
+**GOOD:**
+```
+class Good {
+  final int foo;
+
+  const Good() : foo = 1;
+}
+```
+
+''';
+
+ConstructorDeclaration _asConstructorDeclaration(ClassMember node) {
+  if (node is ConstructorDeclaration) {
+    return node;
+  }
+  return null;
+}
+
+bool _doesNotHaveNonConstInitializer(ConstructorDeclaration element) {
+  return element.initializers.every(_isConstInitializer);
+}
+
+bool _hasEmptyBody(ConstructorDeclaration node) {
+  final body = node.body;
+  return body is EmptyFunctionBody ||
+      (body is BlockFunctionBody && body.block.statements.length == 0);
+}
+
+bool _hasNonFinalFields(ClassElement classElement) =>
+    classElement.fields.any((e) => !e.isFinal);
+
+bool _isConstInitializer(ConstructorInitializer element) {
+  if (element is RedirectingConstructorInvocation) {
+    return element.staticElement?.isConst ?? false;
+  } else if (element is SuperConstructorInvocation) {
+    return element.staticElement?.isConst ?? false;
+  }
+  return true;
+}
+
+bool _isNotConst(ConstructorDeclaration element) => !element.element.isConst;
+
+bool _isNotNull(Object o) => o != null;
+
+class DeclareConstConstructorWhenPossible extends LintRule {
+  _Visitor _visitor;
+  DeclareConstConstructorWhenPossible()
+      : super(
+            name: 'declare_const_constructor_when_possible',
+            description: _desc,
+            details: _details,
+            group: Group.style) {
+    _visitor = new _Visitor(this);
+  }
+
+  @override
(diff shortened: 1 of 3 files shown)


lib/src/rules/declare_const_constructor_when_possible.dart
```diff
@@ -0,0 +1,111 @@
...
+class Bad2 {
+  final int foo;
+
+  const Bad2() : foo = 1; // LINT
+}
+```
+
+''';
+
+ConstructorDeclaration _AsConstructorDeclaration(ClassMember node) {
```
alexeieleusis: Change to camel case.

lib/src/rules/declare_const_constructor_when_possible.dart
```diff
@@ -0,0 +1,111 @@
...
+  visitClassDeclaration(ClassDeclaration node) {
+    if (_hasNonFinalFields(node.element)) {
+      return;
+    }
+    node.members
+        .map(_AsConstructorDeclaration)
+        .where(_isNotNull)
+        .where(_isNotConst)
+        .where(_hasEmptyBody)
+        .where(_doesNotHaveNonConstInitializer)
```
alexeieleusis: Following implementation looks like this function can return `null`.

lib/src/rules/declare_const_constructor_when_possible.dart
```diff
@@ -0,0 +1,111 @@
...
+      ? classElement.allSupertypes.first.element
+      : null;
+  return classElement != null &&
+      (classElement.fields.any((e) => !e.isFinal) ||
+          _hasNonFinalFields(superClass));
+}
+
+bool _isConstInitializer(ConstructorInitializer element) {
+  if (element is RedirectingConstructorInvocation) {
+    return element.staticElement?.isConst;
```
alexeieleusis: Do you prefer `bestElement`?

alexeieleusis: Also provide default values to avoid NPEs at runtime.

lib/src/rules/declare_const_constructor_when_possible.dart
```diff
@@ -0,0 +1,111 @@
...
+  } else if (element is SuperConstructorInvocation) {
+    return element.staticElement?.isConst;
+  }
+  return true;
```
bwilkerson: This is incomplete in two ways. First, the spec says

> The superinitializer that appears, explicitly or implicitly, in the initializer list of a constant constructor must specify a constant constructor of the superclass of the immediately enclosing class or a compile-time error occurs.

This is missing the case where there is an implicit superinitializer. Second, the spec says

> Any expression that appears within the initializer list of a constant constructor must be a potentially constant expression ...

So if the initializer is neither of the above cases, we need to determine whether it is a potentially constant expression.

lib/src/rules/declare_const_constructor_when_possible.dart
JPaulsen: Fixed

lib/src/rules/declare_const_constructor_when_possible.dart
JPaulsen: bestElement is not possible in this case (RedirectingConstructorInvocation does not have that getter). When somesthing is null is considered exactly as false right?, that is the behavior that I want, I will add it anyway to make it explicit.

Fixed

lib/src/rules/declare_const_constructor_when_possible.dart
JPaulsen: Yes, it is not the most beautyful in the world, but we know that use ```is``` and ```as``` invocations is expensive, so with that trick I only do one is.

lib/src/rules/declare_const_constructor_when_possible.dart
```diff
@@ -0,0 +1,111 @@
...
+  return body is EmptyFunctionBody ||
+      (body is BlockFunctionBody && body.block.statements.length == 0);
+}
+
+bool _hasNonFinalFields(ClassElement classElement) {
+  if (classElement == null) return false;
+  final superClass = classElement.allSupertypes.isNotEmpty
+      ? classElement.allSupertypes.first.element
+      : null;
+  return classElement != null &&
```
JPaulsen: Fixed

lib/src/rules/declare_const_constructor_when_possible.dart
```diff
@@ -0,0 +1,111 @@
...
+      (classElement.fields.any((e) => !e.isFinal) ||
+          _hasNonFinalFields(superClass));
```
JPaulsen: You are totally right!, I have to test the implicit super calls how you said, but this is fixed now.

lib/src/rules/declare_const_constructor_when_possible.dart
JPaulsen: @bwilkerson is there an easy way to get the implicit superinitializer in a ConstructorDeclaration?

JPaulsen: About the second part, I used the definition of [Effective Dart](https://www.dartlang.org/guides/language/effective-dart/design#consider-making-your-constructor-const-if-the-class-supports-it)

```If you have a class where all the fields are final, and the constructor does nothing but initialize them, you can make that constructor const```

bwilkerson: > is there an easy way to get the implicit superinitializer in a ConstructorDeclaration?

The only way I can think of is to use the element for the enclosing class to get the element for the superclass and then look for a zero-argument unnamed constructor. If it doesn't have one, then the implicit constructor won't be `const`.

> About the second part, I used the definition of Effective Dart ...

Yes, but Effective Dart wasn't trying to be complete. I don't think we want to suggest making constructors `const` if it isn't possible (without other changes).

> Should the initializers of the fields be const too?

Yes, that's also a requirement.

PR #540, opened 2017-03-28
New rule: avoid_classes_with_only_static_members

JPaulsen:
Fixes https://github.com/dart-lang/linter/issues/539

### DIFF
--- a/lib/src/rules.dart
+++ b/lib/src/rules.dart
@@ -10,6 +10,7 @@ import 'package:linter/src/rules/always_specify_types.dart';
 import 'package:linter/src/rules/annotate_overrides.dart';
 import 'package:linter/src/rules/avoid_as.dart';
 import 'package:linter/src/rules/avoid_catches_without_on_clauses.dart';
+import 'package:linter/src/rules/avoid_classes_with_only_static_members.dart';
 import 'package:linter/src/rules/avoid_empty_else.dart';
 import 'package:linter/src/rules/avoid_function_literals_in_foreach_calls.dart';
 import 'package:linter/src/rules/avoid_init_to_null.dart';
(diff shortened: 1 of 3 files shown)


lib/src/rules/avoid_defining_a_class_that_contains_only_static_members.dart
```diff
@@ -0,0 +1,81 @@
...
+    return classMember.isStatic;
+  }
+  return false;
+}
+
+class AvoidDefiningAClassThatContainsOnlyStaticMembers extends LintRule {
+  _Visitor _visitor;
+  AvoidDefiningAClassThatContainsOnlyStaticMembers()
+      : super(
+            name: 'avoid_defining_a_class_that_contains_only_static_members',
```
bwilkerson: Perhaps we can find a shorter name, such as 'avoid_classes_with_only_static_members' (although something even shorter would be nice).

JPaulsen: Fixed

JPaulsen:
@bwilkerson yes, I have run all my pull requests in the SDK since my last bug in do_not_check_for_null_in_equality_operators

bwilkerson:
How many lints were generated?

JPaulsen:
@bwilkerson 96 in the SDK, they are all classes with at least one non-const static field, and only static methods

test/rules/avoid_defining_a_class_that_contains_only_static_members.dart
```diff
@@ -0,0 +1,39 @@
...
+
+// test w/ `pub run test -N avoid_defining_a_class_that_contains_only_static_members`
+
+class Bad { // LINT
+  static int a;
+
+  static foo() {}
+}
+
+class Bad2 extends Good1 { // LINT
```
JPaulsen: Yes, but every class extends Object and it has nonstatic members (like toString), so I thinks the spirit of the rule, is not to create classes that locally only have static members.

Nathan-Heller:
> @bwilkerson 96 in the SDK, they are all classes with at least one non-const static field, and only static methods

Is there a specific reason that the rule isn't violated when a class contains only static const fields? What was the rationale there?

Thanks for your time 😄

Issue #543, opened 2017-03-28
Implement AVOID annotating types on function expressions.


JPaulsen:
From [Effective Dart](https://www.dartlang.org/guides/language/effective-dart/design#avoid-annotating-types-on-function-expressions)

bwilkerson:
I'm not sure how well this advice works with type inference. There might well be times when we need to annotate in order to avoid strong mode errors. (Just something to check out and keep in mind.)

alexeieleusis:
As a user of the language and the effective dart guidelines, I think this rule makes a lot of sense and if annotating is needed (to help type inference) definitely the expression needs to be extracted to a named function or method.

That said I wonder if a similar approach to how things are done in `omit_type_for_local_variables`, though that needs to have the function expression used as a parameter, which might not always be the case.

PR #545, opened 2017-03-28
New rule: avoid_types_on_closure_parameters

JPaulsen:
Fixes https://github.com/dart-lang/linter/issues/543

### DIFF
--- a/lib/src/rules/avoid_types_on_closure_parameters.dart
+++ b/lib/src/rules/avoid_types_on_closure_parameters.dart
@@ -1,0 +1,74 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+library linter.src.rules.avoid_annotating_types_on_function_expressions;
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:linter/src/analyzer.dart';
+
+const _desc = r'Avoid annotating types on function expression parameters.';
+
+const _details = r'''
+
+**AVOID** annotating types on function expressions.
+
+**BAD:**
+```
+var names = people.map((Person person) => person.name);
+```
+
+**GOOD:**
+```
+var names = people.map((person) => person.name);
+```
+
+''';
+
+NormalFormalParameter _getNormalFormalParameter(FormalParameter node) {
+  if (node is DefaultFormalParameter) {
+    return node.parameter;
+  }
+  return node;
+}
+
+bool _hasNonDynamicType(NormalFormalParameter parameter) {
+  if (parameter is SimpleFormalParameter) {
+    return parameter.type != null && parameter.type.name.name != 'dynamic';
+  }
+  return true;
+}
+
+class AvoidTypesOnClosureParameters extends LintRule {
+  _Visitor _visitor;
+  AvoidTypesOnClosureParameters()
+      : super(
+            name: 'avoid_types_on_closure_parameters',
+            description: _desc,
+            details: _details,
+            group: Group.style) {
+    _visitor = new _Visitor(this);
+  }
+
+  @override
+  AstVisitor getVisitor() => _visitor;
+}
+
+class _Visitor extends SimpleAstVisitor {
+  final LintRule rule;
+  _Visitor(this.rule);
+
+  @override
+  visitFunctionExpression(FunctionExpression node) {
+    if (node.parent is FunctionDeclaration) {
+      return;
+    }
+    final nonDynamicParameters = node.parameters.parameters
+        .map(_getNormalFormalParameter)
+        .where(_hasNonDynamicType);
+    if (nonDynamicParameters.isNotEmpty) {
+      rule.reportLint(nonDynamicParameters.first);
+    }
+  }
+}
--- a/lib/src/rules.dart
+++ b/lib/src/rules.dart
@@ -19,6 +19,7 @@ import 'package:linter/src/rules/avoid_return_types_on_setters.dart';
 import 'package:linter/src/rules/avoid_returning_null.dart';
 import 'package:linter/src/rules/avoid_returning_this.dart';
 import 'package:linter/src/rules/avoid_slow_async_io.dart';
+import 'package:linter/src/rules/avoid_types_on_closure_parameters.dart';
 import 'package:linter/src/rules/await_only_futures.dart';
 import 'package:linter/src/rules/camel_case_types.dart';
 import 'package:linter/src/rules/cancel_subscriptions.dart';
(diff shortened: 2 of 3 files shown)


JPaulsen:
131 lints in the SDK, most of them like this:

```dart
map.forEach((Symbol key, value)
```

lib/src/rules/avoid_annotating_types_on_function_expressions.dart
```diff
@@ -0,0 +1,59 @@
...
+class _Visitor extends SimpleAstVisitor {
+  final LintRule rule;
+  _Visitor(this.rule);
+
+  @override
+  visitFunctionExpression(FunctionExpression node) {
+    if (node.parent is FunctionDeclaration) {
+      return;
+    }
+    node.parameters.parameters.where(_isNotDynamic).forEach(rule.reportLint);
```
JPaulsen: @bwilkerson do you think is this ok? or is it better only one lint for all parameters?

JPaulsen: I think a good option would be just lint the first one

lib/src/rules/avoid_annotating_types_on_function_expressions.dart
```diff
@@ -0,0 +1,59 @@
...
+
+bool _isNotDynamic(FormalParameter parameter) {
+  return !parameter.element.type.isDynamic;
+}
+
+class AvoidAnnotatingTypesOnFunctionExpressions extends LintRule {
+  _Visitor _visitor;
+  AvoidAnnotatingTypesOnFunctionExpressions()
+      : super(
+            name: 'avoid_annotating_types_on_function_expressions',
```
bwilkerson: Perhaps "avoid_types_on_closure_parameters"? It isn't much shorter, but might be more expressive. I don't have a strong opinion, so I leave it to you.

lib/src/rules/avoid_annotating_types_on_function_expressions.dart
bwilkerson: I agree, once per closure would be better.

JPaulsen: Fixed

lib/src/rules/avoid_annotating_types_on_function_expressions.dart
JPaulsen: Fixed

JPaulsen:
@bwilkerson where can I see those rules?

JPaulsen:
@bwilkerson what about this rule? (I rebased it)

JPaulsen:
111 lints in the SDK now, no false positives and not linting when using dynamic because other rule does it now, PTAL

lib/src/rules/avoid_types_on_closure_parameters.dart
```diff
@@ -0,0 +1,81 @@
...
+
+**GOOD:**
+```
+var names = people.map((person) => person.name);
+```
+
+''';
+
+bool _doesNotHaveType(SimpleFormalParameter parameter) {
+  return parameter?.type != null && parameter.type.name.name != 'dynamic';
```
JPaulsen: I have to add tests to check this

lib/src/rules/avoid_types_on_closure_parameters.dart
```diff
@@ -0,0 +1,81 @@
...
+}
+
+SimpleFormalParameter _getSimpleFormalParameter(FormalParameter node) {
+  if (node is SimpleFormalParameter) {
+    return node;
+  }
+  final simpleFormalParameters = DartTypeUtilities
+      .traverseNodesInDFS(node)
+      .where(_isSimpleFormalParameter);
+  return simpleFormalParameters.isNotEmpty
```
JPaulsen: firstWhere returns an Exception if does not found, and that happens with FunctionTypedFormalParameter

lib/src/rules/avoid_types_on_closure_parameters.dart
JPaulsen: Fixed

lib/src/rules/avoid_types_on_closure_parameters.dart
bwilkerson: There isn't any reason to traverse the whole AST, and in fact it's wrong when you consider function typed formal parameters. This really needs to be something like:
```
if (node is DefaultFormalParameter) {
  node = node.parameter;
}
if (node is SimpleFormalParameter) {
  return node;
}
return null;
```

lib/src/rules/avoid_types_on_closure_parameters.dart
```diff
@@ -0,0 +1,81 @@
...
+  final LintRule rule;
+  _Visitor(this.rule);
+
+  @override
+  visitFunctionExpression(FunctionExpression node) {
+    if (node.parent is FunctionDeclaration) {
+      return;
+    }
+    final nonDynamicParameters = node.parameters.parameters
+        .map(_getSimpleFormalParameter)
```
bwilkerson: I don't think you want to only get simple formal parameters because I think that misses a case. Consider:
```
(int f(int x)) { ... }
```
The parameter `f` has a declared type.

JPaulsen: Yes!, you read my mind sometimes, fixed and tests added.

lib/src/rules/avoid_types_on_closure_parameters.dart
JPaulsen: Fixed and tests added.

lib/src/rules/avoid_types_on_closure_parameters.dart
```diff
@@ -0,0 +1,74 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+library linter.src.rules.avoid_annotating_types_on_function_expressions;
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:linter/src/analyzer.dart';
+
+const _desc = r'Avoid annotating types on function expression parameters.';
```
JPaulsen: @bwilkerson what do you think about the message?.

The issues were fixed, 114 lints in the SDK no false positives and no more than 1 lint per method.

bwilkerson: It looks fine to me, but it sounds like you have a concern. What are your concerns?

JPaulsen: No, it is just that we are detecting it in FunctionExpression that are not inside of a FunctionDeclaration (an anonymous function, that are known as "lambdas" for me) even the name of the rule changed to: Avoid types on closure parameters, so I feel uncomfortable about not having a name for the concept of an anonymous function, I think if its closure or lambda or whatever it would be better than "function expression".

So what is the correct name for anonymous functions in Dart?

bwilkerson: The specification refers to either a "function expression" (9 times) or "function literal" (16 times) as an unnamed function written as a literal in code. (In other words, both terms are used interchangeably). A function expression is one kind of "closure", a "tear-off" being another kind of closure. Outside the specification, all three terms appear to be in use, pretty much interchangeably. I'd probably leave it as "function expression" because that's what's in the style guide, and it might be better to change the name to be consistent (though I like "closure" because it's shorter).

JPaulsen: Ok, so this rule is ready.

Issue #546, opened 2017-03-29
Implement DO use getters for operations that conceptually access properties.


JPaulsen:
From [Effective Dart](https://www.dartlang.org/guides/language/effective-dart/design#do-use-getters-for-operations-that-conceptually-access-properties)

PR #547, opened 2017-03-29
New rule: use_getters_to_access_properties

JPaulsen:
Fixes https://github.com/dart-lang/linter/issues/546

### DIFF
--- a/lib/src/rules/use_getters_to_access_properties.dart
+++ b/lib/src/rules/use_getters_to_access_properties.dart
@@ -1,0 +1,212 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+library linter.src.rules.use_getters_to_access_properties;
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/token.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:linter/src/analyzer.dart';
+import 'package:linter/src/util/dart_type_utilities.dart';
+
+const _desc =
+    r'Use getters for operations that conceptually access properties.';
+
+const _details = r'''
+
+**DO** use getters for operations that conceptually access properties.
+
+**BAD:**
+```
+rectangle.getWidth()
+collection.isEmpty()
+button.canShow()
+```
+
+**GOOD:**
+```
+rectangle.width
+collection.isEmpty
+button.canShow
+```
+
+''';
+
+bool _beginsWithAsOrTo(String name) {
+  final regExp = new RegExp(r"(to|as|_to|_as)[A-Z]", caseSensitive: true);
+  return regExp.matchAsPrefix(name) != null;
+}
+
+bool _containsReturn(BlockFunctionBody body) {
+  final visitor = new _ContainsReturnVisitor();
+  body.accept(visitor);
+  return visitor.containsReturn;
+}
+
+bool _hasInheritedMethod(MethodDeclaration node) =>
+    DartTypeUtilities.lookUpInheritedMethod(node) != null;
+
+bool _hasSideEffect(FunctionBody node) {
+  final visitor = new _SideEffectVisitor();
+  node.accept(visitor);
+  return visitor.hasSideEffect;
+}
+
+bool _returnsSomething(FunctionBody body) {
+  final visitor = new _ReturnSomethingVisitor();
+  body.accept(visitor);
+  return visitor.returnsSomething;
+}
+
+class UseGettersToAccessProperties extends LintRule {
+  _Visitor _visitor;
+  UseGettersToAccessProperties()
+      : super(
+            name: 'use_getters_to_access_properties',
+            description: _desc,
+            details: _details,
+            group: Group.style) {
+    _visitor = new _Visitor(this);
+  }
+
+  @override
+  AstVisitor getVisitor() => _visitor;
+}
+
+class _ContainsReturnVisitor extends RecursiveAstVisitor {
+  var containsReturn = false;
+
+  @override
+  visitReturnStatement(ReturnStatement node) {
+    if (node.expression != null) {
+      containsReturn = true;
+    }
+  }
+}
+
+class _ReturnSomethingVisitor extends SimpleAstVisitor {
+  var returnsSomething = false;
+  @override
+  visitBlockFunctionBody(BlockFunctionBody node) {
+    returnsSomething = _containsReturn(node);
+  }
+
+  @override
+  visitExpressionFunctionBody(ExpressionFunctionBody node) {
+    returnsSomething = true;
+  }
+}
+
+class _SideEffectVisitor extends UnifyingAstVisitor {
+  var hasSideEffect = false;
+
+  @override
+  visitAssignmentExpression(AssignmentExpression node) {
+    hasSideEffect = true;
+  }
+
+  @override
+  visitBinaryExpression(BinaryExpression node) {
+    if (hasSideEffect) {
+      return;
+    }
+    if (node.operator.type == TokenType.EQ_EQ ||
+        node.operator.type == TokenType.BANG_EQ) {
+      if (DartTypeUtilities.isNullLiteral(node.leftOperand)) {
+        node.rightOperand.accept(this);
+      } else if (DartTypeUtilities.isNullLiteral(node.rightOperand)) {
+        node.leftOperand.accept(this);
+      } else {
+        hasSideEffect = true;
+      }
+    } else if (node.operator.isUserDefinableOperator) {
+      hasSideEffect = true;
+    } else {
+      visitNode(node);
+    }
+  }
+
+  @override
+  visitFunctionExpressionInvocation(FunctionExpressionInvocation node) {
+    hasSideEffect = true;
+  }
+
+  @override
+  visitIndexExpression(IndexExpression node) {
+    hasSideEffect = true;
+  }
+
+  @override
+  visitInstanceCreationExpression(InstanceCreationExpression node) {
+    hasSideEffect = true;
+  }
+
+  @override
+  visitListLiteral(ListLiteral node) {
+    hasSideEffect = true;
+  }
+
+  @override
+  visitMapLiteral(MapLiteral node) {
+    hasSideEffect = true;
+  }
+
+  @override
+  visitMethodInvocation(MethodInvocation node) {
+    hasSideEffect = true;
+  }
+
+  @override
+  visitNativeClause(NativeClause node) {
+    hasSideEffect = true;
+  }
+
+  @override
+  visitNode(AstNode node) {
+    if (!hasSideEffect) {
+      node.visitChildren(this);
+    }
+  }
+
+  @override
+  visitPostfixExpression(PostfixExpression node) {
+    hasSideEffect = true;
+  }
+
+  @override
+  visitPrefixedIdentifier(PrefixedIdentifier node) {
+    hasSideEffect = true;
+  }
+
+  @override
+  visitPrefixExpression(PrefixExpression node) {
+    hasSideEffect = true;
+  }
+
+  @override
+  visitPropertyAccess(PropertyAccess node) {
+    hasSideEffect = true;
+  }
+}
+
+class _Visitor extends SimpleAstVisitor {
+  final LintRule rule;
+  _Visitor(this.rule);
+
+  @override
+  visitMethodDeclaration(MethodDeclaration node) {
+    final body = node.body;
+    if (!node.isGetter &&
+        !node.isSetter &&
+        node.parameters?.parameters?.isEmpty == true &&
+        node.returnType?.type?.name != 'void' &&
+        node.operatorKeyword == null &&
+        !_beginsWithAsOrTo(node.name.name) &&
+        !_hasInheritedMethod(node) &&
+        _returnsSomething(body) &&
+        !_hasSideEffect(body)) {
+      rule.reportLint(node.name);
+    }
+  }
+}
--- a/lib/src/rules.dart
+++ b/lib/src/rules.dart
@@ -89,6 +89,7 @@ import 'package:linter/src/rules/unnecessary_null_aware_assignments.dart';
 import 'package:linter/src/rules/unnecessary_null_in_if_null_operators.dart';
 import 'package:linter/src/rules/unnecessary_this.dart';
 import 'package:linter/src/rules/unrelated_type_equality_checks.dart';
+import 'package:linter/src/rules/use_getters_to_access_properties.dart';
 import 'package:linter/src/rules/use_rethrow_when_possible.dart';
 import 'package:linter/src/rules/use_setters_to_change_a_property.dart';
 import 'package:linter/src/rules/use_to_and_as_if_applicable.dart';
@@ -182,6 +183,7 @@ void registerLintRules() {
     ..register(new UnnecessaryLambdas())
     ..register(new UnnecessaryThis())
     ..register(new UnrelatedTypeEqualityChecks())
+    ..register(new UseGettersToAccessProperties())
     ..register(new UseRethrowWhenPossible())
     ..register(new UseSettersToChangeAProperty())
     ..register(new UseToAndAsIfApplicable())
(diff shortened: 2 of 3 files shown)


lib/src/rules/use_getters_to_access_properties.dart
```diff
@@ -0,0 +1,75 @@
...
+
+class _Visitor extends SimpleAstVisitor {
+  final LintRule rule;
+  _Visitor(this.rule);
+
+  @override
+  visitMethodDeclaration(MethodDeclaration node) {
+    if (!node.isStatic &&
+        !node.isGetter &&
+        !resolutionMap.elementDeclaredByMethodDeclaration(node).isOverride &&
```
JPaulsen: This only check if has @override annotation, I do not like to repeat all the code of annotate_override, but I thinks would give a much better experience for the user, how can we solve this @bwilkerson? One option is copy paste all that code, or other is to merge both rules in one file, or maybe create an abstract Visitor.

JPaulsen:
463 lints in the SDK, a lot of them like this:

```dart
String toString() => _prettyName;
```

lib/src/rules/use_getters_to_access_properties.dart
```diff
@@ -0,0 +1,75 @@
...
+```
+
+''';
+
+bool _hasSideEffect(AstNode node) {
+  return (node is MethodInvocation ||
+      node is FunctionExpressionInvocation ||
+      node is AssignmentExpression ||
+      node is PrefixExpression ||
+      node is PostfixExpression);
```
bwilkerson: We might want to include a few other cases in this list, even though they are admittedly unlikely cases:
 - Factory constructors can do any amount of work, including causing side-effects in existing objects.
 - User definable operators can have side effects (they're just a method invocation). That includes the index operator `[]`.
- Anything that will implicitly invoke `toString()` because it could have a side-effect (such as string interpolation).
I leave it to you to decide whether these are worth guarding against.

lib/src/rules/use_getters_to_access_properties.dart
JPaulsen: Nice!, reduce lints to 332

JPaulsen:
@bwilkerson I do not how much, but there are some patterns

```dart

**Comment by JPaulsen** [comment]
@bwilkerson I do not how much, but there are some patterns

```dart
String _toString() native "JSObject_toString";
Set<E> _newSet() => new _HashSet<E>();
Set<E> toSet() => new Set<E>.from(this);
int getPointer();
int close();
bool _isControlFrame() {
    return _opcode == _WebSocketOpcode.CLOSE ||
        _opcode == _WebSocketOpcode.PING ||
        _opcode == _WebSocketOpcode.PONG;
  }
}
int available() {
    return _status != CONNECTED ? 0
                                : _secureFilter.buffers[READ_PLAINTEXT].length;
  }
```

JPaulsen:
I think the first is private, it should be a getter or a field.

The second one should be an static method

The third one is ok because it is a toSomething method.

int getPointer(); I think this is really well linted, I good rule would be if a method have an EmptyFunctionBody and its name starts with get and and upper case letter it should be linted.

bool _isControlFrame() {} well linted

int available() well linted

bwilkerson:
> `String _toString() native "JSObject_toString";`
>
> I think the first is private, it should be a getter or a field.

We shouldn't create lints for `external` or `native` methods. These are required to match code written in another language.

> `Set<E> _newSet() => new _HashSet<E>();`
>
> The second one should be an static method

Not if it's there, as I expect, so that subclasses can override it to return a different class of set.

> The third one is ok because it is a toSomething method.

So we should probably special case methods starting with `to` followed by an uppercase letter.

> `int getPointer();`
>
> I think this is really well linted, I good rule would be if a method have an EmptyFunctionBody and its name starts with get and and upper case letter it should be linted.

I'd actually say the opposite. If it's an abstract method (has no body) then it's being defined so that subclasses can override it, and we don't know whether the overriding implementations will have a side-effect, so we should not create a lint.

The others seem like they should have a lint.

JPaulsen:
Yes you are right about the last one, I thought you could override a getter, so in that case we should take care about a child class override a method

lib/src/rules/use_getters_to_access_properties.dart
```diff
@@ -0,0 +1,75 @@
...
+
+  @override
+  visitMethodDeclaration(MethodDeclaration node) {
+    if (!node.isStatic &&
+        !node.isGetter &&
+        !resolutionMap.elementDeclaredByMethodDeclaration(node).isOverride &&
+        node.parameters?.parameters?.isEmpty == true &&
+        node.returnType != null &&
+        node.returnType.type.name != 'void' &&
+        !DartTypeUtilities.traverseNodesInDFS(node.body).any(_hasSideEffect)) {
```
JPaulsen: Fixed

JPaulsen:
@bwilkerson I created an extra visitor to find when something has a side-effect, I can remove all the super calls, but I kept them by know to not miss any node that could have side-effect, PTAL and tell me if you agree the nodes that are OK or you should add/remove as side-effect node any of them.

lib/src/rules/use_getters_to_access_properties.dart
```diff
@@ -0,0 +1,83 @@
...
+  visitMethodDeclaration(MethodDeclaration node) {
+    final body = node.body;
+    if (!node.isSetter &&
+        !node.isGetter &&
+        !_hasInheritedMethod(node) &&
+        node.parameters?.parameters?.isEmpty == true &&
+        node.returnType?.type?.name != 'void' &&
+        !(body is BlockFunctionBody && body.block.statements.isEmpty) &&
+        !_beginsWithAsOrTo(node.name.name) &&
+        !_hasSideEffect(node.body)) {
```
bwilkerson: I think this should also exclude methods with an EmptyFunctionBody, as per earlier comments.

Also, `node.body` can be just `body`.

lib/src/util/side_effect_visitor.dart
```diff
@@ -0,0 +1,546 @@
+import 'package:analyzer/analyzer.dart';
+
+class SideEffectVisitor extends RecursiveAstVisitor {
+  var _hasSideEffect = false;
+
+  bool get hasSideEffect => _hasSideEffect;
+  @override
+  visitAdjacentStrings(AdjacentStrings node) {
+    super.visitAdjacentStrings(node);
```
bwilkerson: If you're doing nothing other than invoking the overridden method, then there's no reason to override the method at all.

lib/src/util/side_effect_visitor.dart
```diff
@@ -0,0 +1,546 @@
...
+  }
+
+  @override
+  visitAwaitExpression(AwaitExpression node) {
+    super.visitAwaitExpression(node);
+  }
+
+  @override
+  visitBinaryExpression(BinaryExpression node) {
+    super.visitBinaryExpression(node);
```
bwilkerson: If the operator is user-definable, then we can't guarantee that it won't have a side effect.

lib/src/util/side_effect_visitor.dart
```diff
@@ -0,0 +1,546 @@
...
+  }
+
+  @override
+  visitConstructorDeclaration(ConstructorDeclaration node) {
+    super.visitConstructorDeclaration(node);
+  }
+
+  @override
+  visitConstructorFieldInitializer(ConstructorFieldInitializer node) {
+    super.visitConstructorFieldInitializer(node);
```
bwilkerson: I don't think you should ever visit one of these, but a constructor field initializer *does* have the side effect of assigning a value to a field.

lib/src/util/side_effect_visitor.dart
```diff
@@ -0,0 +1,546 @@
...
+  }
+
+  @override
+  visitFieldDeclaration(FieldDeclaration node) {
+    super.visitFieldDeclaration(node);
+  }
+
+  @override
+  visitFieldFormalParameter(FieldFormalParameter node) {
+    super.visitFieldFormalParameter(node);
```
bwilkerson: I don't think you should ever visit one of these, but a field formal parameter *does* have the side effect of assigning a value to a field.

lib/src/util/side_effect_visitor.dart
```diff
@@ -0,0 +1,546 @@
...
+  }
+
+  @override
+  visitPrefixExpression(PrefixExpression node) {
+    _hasSideEffect = true;
+  }
+
+  @override
+  visitPropertyAccess(PropertyAccess node) {
+    super.visitPropertyAccess(node);
```
bwilkerson: It is possible for a property access to have side effects. Stylistically if shouldn't, but there's nothing in the language that prevents it.

lib/src/util/side_effect_visitor.dart
```diff
@@ -0,0 +1,546 @@
...
+  }
+
+  @override
+  visitPostfixExpression(PostfixExpression node) {
+    _hasSideEffect = true;
+  }
+
+  @override
+  visitPrefixedIdentifier(PrefixedIdentifier node) {
+    super.visitPrefixedIdentifier(node);
```
bwilkerson: It is possible for a property access to have side effects. Stylistically if shouldn't, but there's nothing in the language that prevents it. Unfortunately, we can't always tell lexically when something like "a.b" is a property access, so we often end up creating a prefixed identifier for getter invocations.

lib/src/util/side_effect_visitor.dart
```diff
@@ -0,0 +1,546 @@
...
+  }
+
+  @override
+  visitPropertyAccess(PropertyAccess node) {
+    super.visitPropertyAccess(node);
+  }
+
+  @override
+  visitRedirectingConstructorInvocation(RedirectingConstructorInvocation node) {
+    _hasSideEffect = true;
```
bwilkerson: I don't think you should ever visit one of these, but it isn't clear to me whether this should be considered to have a side-effect. Depending on your reasoning, visitSuperConstructorInvocation might want to be consistent. (But then there's always a super constructor invocation even when it isn't explicit.)

lib/src/util/side_effect_visitor.dart
```diff
@@ -0,0 +1,546 @@
...
+  }
+
+  @override
+  visitThrowExpression(ThrowExpression node) {
+    super.visitThrowExpression(node);
+  }
+
+  @override
+  visitTopLevelVariableDeclaration(TopLevelVariableDeclaration node) {
+    super.visitTopLevelVariableDeclaration(node);
```
bwilkerson: I don't think you'll ever visit one of these, but they can have a side-effect if there is an initializer and the variable is referenced somewhere that could be executed.

Or we could view the reference to the variable as potentially having a side effect if this is the first time the variable is referenced. (But that would mean looking at all identifiers to see whether they refer to a static field or top-level variable.)

lib/src/util/side_effect_visitor.dart
JPaulsen: Exactly, that is why I said that I can remove them all, it is just to check all the types one by one.

lib/src/util/side_effect_visitor.dart
JPaulsen: I got it, for example the operator ==, how can I know which operators are user-definable?

lib/src/util/side_effect_visitor.dart
JPaulsen: Yep, it won't be visited

lib/src/util/side_effect_visitor.dart
JPaulsen: So what do you think is the best, do we assume that the style is correct or we say that in this case _hasSideEffect = true ?

lib/src/util/side_effect_visitor.dart
JPaulsen: Fixed

lib/src/util/side_effect_visitor.dart
JPaulsen: Fixed

lib/src/util/side_effect_visitor.dart
JPaulsen: Fixed

lib/src/util/side_effect_visitor.dart
JPaulsen: Fixed

lib/src/util/side_effect_visitor.dart
JPaulsen: Fixed

lib/src/rules/use_getters_to_access_properties.dart
JPaulsen: Fixed

JPaulsen:
@bwilkerson we did deep changes, now it is linting 8 cases in the sdk, PTAL.

```dart
Map<String, dynamic> getFileInfoMap() {
  return fullValueMap;
}

Future _onCancel() {
  assert(_isCanceled);
  return null;
}

bool _isError() => (_errorCode != null) && (_errorDetail != null);

Uri normalizePath() => this;

DoubleLinkedQueueEntry<E> previousEntry() => _previousLink;

DoubleLinkedQueueEntry<E> nextEntry() => _nextLink;

static getLibrary() {
  return currentLibrary;
}

bool moveNext() => false;
```

lib/src/util/side_effect_visitor.dart
bwilkerson: Section 10.1.1 of the specification (http://www.ecma-international.org/publications/files/ECMA-ST/ECMA-408.pdf) has a list.

JPaulsen: Fixed :)

bwilkerson:
Five of the methods, based on name alone, look like maybe they could have been designed to appear to be properties:
```
getFileInfoMap()
previousEntry()
nextEntry()
getLibrary()
_isError()
```
but I'd like to know where these methods are defined so that I can check a little further.

Three of them definitely don't appear to be property accessors, and I consider them to be false positives:
```
_onCancel()
normalizePath()
moveNext()
```

JPaulsen:
@bwilkerson these are the path of the lints:

```
/usr/local/google/home/casanueva/Development/sdk/sdk/lib/io/io_resource_info.dart 121:24 [lint] Use getters for operations that conceptually access properties.
  Map<String, dynamic> getFileInfoMap() {
                       ^^^^^^^^^^^^^^
/usr/local/google/home/casanueva/Development/sdk/sdk/lib/async/stream_impl.dart 296:10 [lint] Use getters for operations that conceptually access properties.
  Future _onCancel() {
         ^^^^^^^^^
/usr/local/google/home/casanueva/Development/sdk/sdk/lib/developer/extension.dart 88:8 [lint] Use getters for operations that conceptually access properties.
  bool _isError() => (_errorCode != null) && (_errorDetail != null);
       ^^^^^^^^
/usr/local/google/home/casanueva/Development/sdk/sdk/lib/core/uri.dart 4286:7 [lint] Use getters for operations that conceptually access properties.
  Uri normalizePath() => this;
      ^^^^^^^^^^^^^
/usr/local/google/home/casanueva/Development/sdk/sdk/lib/collection/queue.dart 144:29 [lint] Use getters for operations that conceptually access properties.
  DoubleLinkedQueueEntry<E> previousEntry() => _previousLink;
                            ^^^^^^^^^^^^^
/usr/local/google/home/casanueva/Development/sdk/sdk/lib/collection/queue.dart 147:29 [lint] Use getters for operations that conceptually access properties.
  DoubleLinkedQueueEntry<E> nextEntry() => _nextLink;
                            ^^^^^^^^^
/usr/local/google/home/casanueva/Development/sdk/sdk/lib/html/dartium/html_dartium.dart 48117:10 [lint] Use getters for operations that conceptually access properties.
  static getLibrary() {
         ^^^^^^^^^^
/usr/local/google/home/casanueva/Development/sdk/sdk/lib/internal/iterable.dart 750:8 [lint] Use getters for operations that conceptually access properties.
  bool moveNext() => false;
       ^^^^^^^^
```

bwilkerson:
I don't know what else to say. (Although I'd update my earlier list to put `getFileInfoMap()` in the false positive column.)

The goal of the rule is to find methods that "conceptually represent properties". There is no way, using static analysis, to determine what a method conceptually represents. If you're going to try to implement the rule (not implementing it is still an option), that means that you need to have a set of statically analyzable criteria that give you a nearly perfect correlation to what you're really trying to determine. (The number of false positives you can expect is related to how imperfect the correlation is.)

I don't think the criteria we currently have is adequate to bring the number of false positives to be within a reasonable rate, and I'm not convinced that it's possible to get to that point. Personally, I'd drop this rule because I don't think you're going to be able to get it to a point where it will be used, and I think you'll bring more value to users by focusing on other tasks (such as writing more quick fixes).

But, that's a decision that you and Alexei need to make.

PR #549, opened 2017-03-29
New rule: use_setters_to_change_a_property

JPaulsen:
Fixes https://github.com/dart-lang/linter/issues/548

### DIFF
--- a/lib/src/rules/use_setters_to_change_a_property.dart
+++ b/lib/src/rules/use_setters_to_change_a_property.dart
@@ -1,0 +1,88 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+library linter.src.rules.use_setters_to_change_a_property;
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/element.dart';
+import 'package:linter/src/analyzer.dart';
+import 'package:linter/src/util/dart_type_utilities.dart';
+
+const _desc =
+    r'Use a setter for operations that conceptually change a property.';
+
+const _details = r'''
+
+**DO** use a setter for operations that conceptually change a property.
+
+**BAD:**
+```
+rectangle.setWidth(3);
+button.setVisible(false);
+```
+
+**GOOD:**
+```
+rectangle.width = 3;
+button.visible = false;
+```
+
+''';
+
+bool _hasInheritedMethod(MethodDeclaration node) =>
+    DartTypeUtilities.lookUpInheritedMethod(node) != null;
+
+class UseSettersToChangeAProperty extends LintRule {
+  _Visitor _visitor;
+  UseSettersToChangeAProperty()
+      : super(
+            name: 'use_setters_to_change_a_property',
+            description: _desc,
+            details: _details,
+            group: Group.style) {
+    _visitor = new _Visitor(this);
+  }
+
+  @override
+  AstVisitor getVisitor() => _visitor;
+}
+
+class _Visitor extends SimpleAstVisitor {
+  final LintRule rule;
+  _Visitor(this.rule);
+
+  @override
+  visitMethodDeclaration(MethodDeclaration node) {
+    if (node.isSetter ||
+        node.isGetter ||
+        _hasInheritedMethod(node) ||
+        node.parameters?.parameters?.length != 1 ||
+        node.returnType?.type?.name != 'void') {
+      return;
+    }
+    bool _visitExpression(Expression expression) {
+      if (expression is AssignmentExpression) {
+        final leftOperand = DartTypeUtilities
+            .getCanonicalElementFromIdentifier(expression.leftHandSide);
+        final rightOperand = DartTypeUtilities
+            .getCanonicalElementFromIdentifier(expression.rightHandSide);
+        final parameterElement =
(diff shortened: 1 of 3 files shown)


lib/src/rules/use_setters_to_change_a_property.dart
```diff
@@ -0,0 +1,83 @@
...
+
+class _Visitor extends SimpleAstVisitor {
+  final LintRule rule;
+  _Visitor(this.rule);
+
+  @override
+  visitMethodDeclaration(MethodDeclaration node) {
+    if (node.isStatic ||
+        node.isSetter ||
+        resolutionMap.elementDeclaredByMethodDeclaration(node).isOverride ||
```
JPaulsen: I have to change this

JPaulsen:
Only 1 lint reported in SDK

```dart
void onDone(void handleDone()) { _onDone = handleDone; }
```

lib/src/rules/use_setters_to_change_a_property.dart
```diff
@@ -0,0 +1,83 @@
...
+        node.parameters?.parameters?.length != 1 ||
+        (node.returnType != null && node.returnType.type.name != 'void')) {
+      return;
+    }
+    final body = node.body;
+    if (body is BlockFunctionBody && body.block.statements.length == 1) {
```
bwilkerson: Also need to handle expression function bodies.

JPaulsen: But it would be one, with a return type = null or void, should be another lint?

bwilkerson: I don't think so. It's legal to write
```
void setValue(value) => _value = value;
```
and I think you want to create a lint for this.

JPaulsen: Yes, I know it is legal, and it is legal to do this too:

```dart
void setValue(value) {
    return _value = value;
  }
```

And I think that should be a lint of other rule

bwilkerson: I agree that the second case should be a separate diagnostic, though I'd lean toward making it a hint because it's never useful to return a value from a `void` method. But the first is a common short-hand used for methods that will fit on a single line.

But I'm fine with not handling it for now and waiting to l see what users expect.

JPaulsen: Wait, I have been experiment for a while, and I have a question, does a set implicitly return the parameter that receives?

```dart
class A {
  int _x;
  set x(int x) {
    _x = x+1;
  }
}

class B {
  int _x;
  set x(int x) => _x = x+1;
}

main() {
  int a = (new B().x=3);
  int b = (new A().x=3);
  print(a); // print 3
  print(b); // print 3
}
```

JPaulsen: @bwilkerson I think this:

```dart
void setValue(value) {
  return _value = value;
}
```

Only is legal if setValue is a FunctionDeclaration, if it is a MethodDeclaration then doing that is not legal, can you double check please?

lib/src/rules/use_setters_to_change_a_property.dart
JPaulsen: @bwilkerson fixed

lib/src/rules/use_setters_to_change_a_property.dart
bwilkerson: > ... does a setter implicitly return the argument that receives?

No. The value of an assignment expression is the value of the right-hand side. That's what allows chaining such as `x = y = 0;`.

> ... can you double check please?

Try the following:
```
main() {
  try {
    A a = new A();
    a.setValue('a');
    print(a._value);
  } catch (e, s) {
    print(e);
    print(s);
  }
  try {
    setValue('b');
    print(_value);
  } catch (e, s) {
    print(e);
    print(s);
  }
}

var _value;

void setValue(value) {
  return _value = value;
}

class A {
  var _value;
  void setValue(value) {
    return _value = value;
  }
}
```
I get two exceptions printed.

JPaulsen: @bwilkerson in my machine the output is:

```
a
b
```

bwilkerson: I see that when I run in dartpad, which uses dart2js, but when I use the VM I get two exception reports. How are you running it?

JPaulsen: By terminal, I extracted to a file called file.dart, and then I did $dart file.dart in the directory of the file.

bwilkerson: Perhaps we're using different versions of the VM, or there's a platform difference. I'm running:

    Dart VM version: 1.24.0-edge.e3aa3b8ea99e7ec854ad8d227012a512e68ed685

built from bleeding edge this morning, on my Mac.

JPaulsen: So, sometimes is legal, I think we should considerate it legal in this context, do you agree?

JPaulsen: Yes, we are using different versions, I sent you my version by email, so what do we do?

bwilkerson: So for the record, the difference is whether or not the code is being run in checked mode. I think that means that we can merge this as-is and wait for user feedback before making any changes. Any changes you want to make first?

JPaulsen: Nope, I reformatted a test, and rebased to origin/master :D

Issue #550, opened 2017-03-31
Implement AVOID annotating with dynamic when not required.


JPaulsen:
From [Effective Dart](https://www.dartlang.org/guides/language/effective-dart/design#avoid-annotating-with-dynamic-when-not-required)

bwilkerson:
The guide says:

> In most places in Dart, a type annotation can be omitted, in which case the type will automatically be `dynamic`.

Note that this is no longer going to be true. With type inference, omitting a type will sometimes cause a type other than `dynamic` to be inferred. Figuring out whether a type is required (even `dynamic`) will be fairly difficult.

JPaulsen:
Yes also if you do this:

```dart
foo () {
}
```
you will not have any lint, but if you do this:

```dart
dynamic foo() {
}
```

you will get the lint that says: "this function declares a type of 'dynamic' but it does not end in a return statement."

JPaulsen:
@bwilkerson is it a way to use some code of the SDK or analyzer to get the return type to be inferred from a function or method declaration?

bwilkerson:
The top-level inference *is* implemented in analyzer, so technically, yes, there should be a way, but there isn't a reasonable public API for it yet. I don't think we considered the need to ask "what if" questions like the one we need here, so it also wouldn't be very efficient at the moment (I suspect you'd have to fully analyze an edited version of the file, which would probably take too long).

I suspect that the only practical answer here is to ignore for now the question of whether the type is necessary to avoid type inference.

PR #551, opened 2017-03-31
New rule: avoid_annotating_with_dynamic

JPaulsen:
Fixes https://github.com/dart-lang/linter/issues/550

### DIFF
--- a/lib/src/rules.dart
+++ b/lib/src/rules.dart
@@ -8,6 +8,7 @@ import 'package:linter/src/analyzer.dart';
 import 'package:linter/src/rules/always_declare_return_types.dart';
 import 'package:linter/src/rules/always_specify_types.dart';
 import 'package:linter/src/rules/annotate_overrides.dart';
+import 'package:linter/src/rules/avoid_annotating_with_dynamic.dart';
 import 'package:linter/src/rules/avoid_as.dart';
 import 'package:linter/src/rules/avoid_catches_without_on_clauses.dart';
 import 'package:linter/src/rules/avoid_classes_with_only_static_members.dart';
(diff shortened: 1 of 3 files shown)


JPaulsen:
@bwilkerson @pq I am having big problems with this rule, it is getting a lot of false positives when running in the SDK.

SDK lints: 185

Some false positives:

```dart


StringConversionSink startChunkedConversion(Sink<Object> sink) {
^^^^^^^^^^^^^^^^^^^^

Converter<List<int>, T> fuse<T>(Converter<String, T> next) {
                             ^

static Stream<ProcessSignal> _watchSignal(ProcessSignal signal) {
       ^^^^^^^^^^^^^^^^^^^^^
```

bwilkerson:
Perhaps use textual comparison for return types as you do for parameter types. That might remove the false positives. But note the comment added to the issue.

JPaulsen:
@bwilkerson nice!!!, 63 lints in the SDK, no false positives :), also I added some tests to have better test coverage.

JPaulsen:
@bwilkerson I deleted the check of the return type, if you approve these changes, we should open a new issue to extend this rule in the future

lib/src/rules/avoid_annotating_with_dynamic_when_not_required.dart
```diff
@@ -0,0 +1,62 @@
...
+}
+```
+
+''';
+
+class AvoidAnnotatingWithDynamicWhenNotRequired extends LintRule {
+  _Visitor _visitor;
+  AvoidAnnotatingWithDynamicWhenNotRequired()
+      : super(
+            name: 'avoid_annotating_with_dynamic_when_not_required',
```
bwilkerson: Perhaps shorten the name to "avoid_annotating_with_dynamic".

JPaulsen: Fixed

PR #553, opened 2017-03-31
New rule: avoid_returning_this

JPaulsen:
Fixes https://github.com/dart-lang/linter/issues/552

### DIFF
--- a/lib/src/rules/avoid_returning_this.dart
+++ b/lib/src/rules/avoid_returning_this.dart
@@ -1,0 +1,85 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+library linter.src.rules.avoid_returning_this_to_enable_a_fluent_interface;
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:linter/src/analyzer.dart';
+import 'package:linter/src/util/dart_type_utilities.dart';
+
+const _desc =
+    r'Avoid returning this from methods just to enable a fluent interface.';
+
+const _details = r'''
+
+**AVOID** returning this from methods just to enable a fluent interface.
+
+**BAD:**
+```
+var buffer = new StringBuffer()
+  .write('one')
+  .write('two')
+  .write('three');
+```
+
+**GOOD:**
+```
+var buffer = new StringBuffer()
+  ..write('one')
+  ..write('two')
+  ..write('three');
+```
+
+''';
+
+bool _hasInheritedMethod(MethodDeclaration node) =>
+    DartTypeUtilities.lookUpInheritedMethod(node) != null;
+
+bool _isExpressionFunctionBody(AstNode node) => node is ExpressionFunctionBody;
+
+bool _isReturnStatement(AstNode node) => node is ReturnStatement;
+
+bool _returnsThis(AstNode node) {
+  return (node as ReturnStatement).expression is ThisExpression;
+}
+
+class AvoidReturningThis extends LintRule {
+  _Visitor _visitor;
+  AvoidReturningThis()
+      : super(
+            name: 'avoid_returning_this',
+            description: _desc,
+            details: _details,
+            group: Group.style) {
+    _visitor = new _Visitor(this);
+  }
+
+  @override
+  AstVisitor getVisitor() => _visitor;
+}
+
+class _Visitor extends SimpleAstVisitor {
+  final LintRule rule;
+  _Visitor(this.rule);
+
+  @override
+  visitMethodDeclaration(MethodDeclaration node) {
+    if (node.returnType?.type !=
+            (node.parent as ClassDeclaration).element?.type ||
+        _hasInheritedMethod(node)) {
+      return;
+    }
+    final body = node.body;
+    if (body is BlockFunctionBody && body.block.statements.length > 1) {
+      final returnStatements = DartTypeUtilities
+          .traverseNodesInDFS(body.block,
+              excludeCriteria: _isExpressionFunctionBody)
+          .where(_isReturnStatement);
+      if (returnStatements.isNotEmpty && returnStatements.every(_returnsThis)) {
+        rule.reportLint(node.name);
+      }
+    }
+  }
+}
--- a/test/rules/avoid_returning_this.dart
+++ b/test/rules/avoid_returning_this.dart
@@ -1,0 +1,29 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// test w/ `pub run test -N avoid_returning_this`
+
+class A {
+  int x;
+  A badAddOne() { // LINT
+    x++;
+    return this;
+  }
+
+  int goodAddOne() { // OK
+    x++;
+    return this.x;
+  }
+  A getThing() { // OK
+    return this;
+  }
+}
+
+class B extends A{
+  @override
+  badAddOne() { // OK It is ok because it is an inherited method.
+    x++;
+    return this;
+  }
+}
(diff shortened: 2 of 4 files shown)


lib/src/rules/avoid_returning_this_to_enable_a_fluent_interface.dart
```diff
@@ -0,0 +1,73 @@
...
+  AstVisitor getVisitor() => _visitor;
+}
+
+class _Visitor extends SimpleAstVisitor {
+  final LintRule rule;
+  _Visitor(this.rule);
+
+  @override
+  visitMethodDeclaration(MethodDeclaration node) {
+    if (node.returnType?.name?.name == 'void') {
```
JPaulsen: @bwilkerson is there a way to ask if a DartType implements other DartType?

JPaulsen:
4 Lints in the SDK:

```dart
// This appears twice
@DomName('Node.replaceChild')
Node replaceWith(Node otherNode) {
  try {
    final Node parent = this.parentNode;
    parent._replaceChild(otherNode, this);
  } catch (e) {};
  return this;
}

Iterable<E> skip(int count) {
  RangeError.checkNotNegative(count, "count");
  return this;
}

Iterable<E> take(int count) {
  RangeError.checkNotNegative(count, "count");
  return this;
}
```

lib/src/rules/avoid_returning_this_to_enable_a_fluent_interface.dart
bwilkerson: `TypeSystem.isSubtypeOf`

I don't know whether you have access to a type system, but that's what you need.

bwilkerson:
These are all false positives. I'm not sure how to detect that the first one is, other than building in special knowledge about the html libraries (that the interface is defined externally), which I would prefer not to do.

I suspect we can eliminate the last two by noticing that the methods override an inherited method. Any method that overrides another must have the same signature as the overridden method, and it's sometimes reasonable to return `this` in those cases.

lib/src/rules/avoid_returning_this_to_enable_a_fluent_interface.dart
JPaulsen: If a method returns a subtype of my own class, is a better check than ask if it is void or not, but you said we should not lint an "overriden" method, so maybe we just should check if it is exactly the same type than the class that contains it, what do you think?

bwilkerson: I suspect that I misunderstood your original question. Can you give me an example of what you're trying to detect?

JPaulsen: Yes, sorry my english is not good enough yet

```dart
class A {
  A foo () { // this returns an A and its class is A, so I should continue
  }
  int bar () { // this returns an integer but its class is A, so I should not continue
  }
}
```

bwilkerson: Correct. If it returns a different class than the one it's defined in then it isn't worth checking.

My point about overrides is that it's possible for one implementation of a method to return `this` when other implementations of the method don't. The guideline is specifically targeting an implementation style in which the result is *always* `this` and when the reason is so that another method can be called. So I'll extend what I said before: we should not lint any method if there is either a supertype or subtype that implements the same method. Unfortunately, that's not possible (unless we just don't implement the lint at all) because we don't have access to the subtypes. So we'll have to be content with only checking supertypes.

JPaulsen:
@bwilkerson fixed the last two false positives and tests added.

lib/src/rules/avoid_returning_this_to_enable_a_fluent_interface.dart
```diff
@@ -0,0 +1,85 @@
...
+
+bool _returnsThis(AstNode node) {
+  return (node as ReturnStatement).expression is ThisExpression;
+}
+
+class AvoidReturningThisToEnableAFluentInterface extends LintRule {
+  _Visitor _visitor;
+  AvoidReturningThisToEnableAFluentInterface()
+      : super(
+            name: 'avoid_returning_this_to_enable_a_fluent_interface',
```
bwilkerson: Perhaps shorten the name to something like "avoid_returning_this".

lib/src/rules/avoid_returning_this_to_enable_a_fluent_interface.dart
```diff
@@ -0,0 +1,85 @@
...
+      return;
+    }
+    final body = node.body;
+    if (body is BlockFunctionBody && body.block.statements.length > 1) {
+      final returnStatements = DartTypeUtilities
+          .traverseNodesInDFS(body.block,
+              excludeCriteria: _isExpressionFunctionBody)
+          .where(_isReturnStatement);
+      if (returnStatements.every(_returnsThis)) {
+        returnStatements.forEach(rule.reportLint);
```
bwilkerson: I think the lint should be on the method (and that we should have a single lint). This is more a property of the method than of any single return statement, and creating multiple lints isn't going to make it easier to fix the problem. (And in cases where the lint needs to be ignores, users can ignore it in one place rather than many.)

test/rules/avoid_returning_this_to_enable_a_fluent_interface.dart
```diff
@@ -0,0 +1,29 @@
...
+    return this.x; // OK
+  }
+  A getThing() {
+    return this; // OK
+  }
+}
+
+class B extends A{
+  @override
+  badAddOne() { // It is ok because it is an inherited method.
```
bwilkerson: Perhaps add a return type of "B" so we know that the reason there's no lint is because it overrides an inherited method and not because the type is not the same as the declaring class. (And add another test that shows that we don't create a lint if a method has a different return type even when it doesn't override an inherited method.)

lib/src/rules/avoid_returning_this_to_enable_a_fluent_interface.dart
JPaulsen: Fixed

lib/src/rules/avoid_returning_this_to_enable_a_fluent_interface.dart
JPaulsen: Fixed

lib/src/rules/avoid_returning_this.dart
```diff
@@ -0,0 +1,85 @@
...
+      return;
+    }
+    final body = node.body;
+    if (body is BlockFunctionBody && body.block.statements.length > 1) {
+      final returnStatements = DartTypeUtilities
+          .traverseNodesInDFS(body.block,
+              excludeCriteria: _isExpressionFunctionBody)
+          .where(_isReturnStatement);
+      if (returnStatements.every(_returnsThis)) {
+        rule.reportLint(node);
```
bwilkerson: We typically report errors against the *name* of a declaration; it looks better that marking the whole declaration.

JPaulsen: Yes I was thinking the same but I did not say anything xd sorry

JPaulsen: Fixed

test/rules/avoid_returning_this_to_enable_a_fluent_interface.dart
JPaulsen: Fixed

lib/src/rules/avoid_returning_this.dart
JPaulsen: Fixed

lib/src/rules/avoid_returning_this.dart
```diff
@@ -0,0 +1,85 @@
...
+    if (node.returnType?.type !=
+            (node.parent as ClassDeclaration).element?.type ||
+        _hasInheritedMethod(node)) {
+      return;
+    }
+    final body = node.body;
+    if (body is BlockFunctionBody && body.block.statements.length > 1) {
+      final returnStatements = DartTypeUtilities
+          .traverseNodesInDFS(body.block,
+              excludeCriteria: _isExpressionFunctionBody)
```
bwilkerson: Are we trying to avoid nested functions and closures by this? Because if so, they aren't required to have an expression function body, just a function body.

test/rules/avoid_returning_this.dart
```diff
@@ -0,0 +1,29 @@
...
+    return this.x;
+  }
+  A getThing() { // OK
+    return this;
+  }
+}
+
+class B extends A{
+  @override
+  badAddOne() { // OK It is ok because it is an inherited method.
```
bwilkerson: But the code explicitly tests for the wrong return type *before* it tests for an inherited method, which means that this test would pass even if it stopped checking for an overridden method. That defeats the point of the test. You really need to add a return type of `B` here.

JPaulsen: I think this was fixed https://github.com/dart-lang/linter/pull/559

Issue #554, opened 2017-03-31
Implement AVOID annotating with Function.


JPaulsen:
From [Effective Dart](https://www.dartlang.org/guides/language/effective-dart/design#avoid-annotating-with-function)

JPaulsen:
@bwilkerson @alexeieleusis I think this rule can't be implemented, I was thinking about the most basic of the cases, a Function parameter with only one use of Functions.apply inside and no other reference to the function but even in that case there would be false positives.

```dart
void applyWithNullArguments(Function f, int n) {
  final arguments = [];
  while (n --> 0) {
    arguments.add(null);
  }
  Function.apply(f, arguments);
}
```

So I think we should discard this one, what do you think?

alexeieleusis:
Probably not linting those cases?

I would like to know if @munificent has some advice here.

JPaulsen:
@alexeieleusis but I think this is one of the most basic cases, when you have only one call to apply, only case it would be a Function with 0 calls to apply but I do not know if that would have any sense in the code.

Other case could be limit the rule to some variable assignations and declarations, for example a final variable declaration like this:

```dart
final Function a = something;
```

So if we can find the type of ```something``` statically and it is better than Function then we could lint that.

PR #555, opened 2017-03-31
New rule: avoid_annotating_parameters_with_function

JPaulsen:
Fixes https://github.com/dart-lang/linter/issues/554

### DIFF
--- a/lib/src/rules.dart
+++ b/lib/src/rules.dart
@@ -11,6 +11,7 @@ import 'package:linter/src/rules/always_require_non_null_named_parameters.dart';
 import 'package:linter/src/rules/always_specify_types.dart';
 import 'package:linter/src/rules/annotate_overrides.dart';
 import 'package:linter/src/rules/avoid_annotating_with_dynamic.dart';
+import 'package:linter/src/rules/avoid_annotating_with_function.dart';
 import 'package:linter/src/rules/avoid_as.dart';
 import 'package:linter/src/rules/avoid_catches_without_on_clauses.dart';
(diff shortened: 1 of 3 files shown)


lib/src/rules/avoid_annotating_parameters_with_function.dart
```diff
@@ -0,0 +1,59 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+library linter.src.rules.avoid_annotating_parameters_with_function;
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:linter/src/analyzer.dart';
+import 'package:linter/src/util/dart_type_utilities.dart';
+
+const _desc = r'Avoid annotating parameters with Function.';
```
bwilkerson: I don't know for sure, but I suspect (given the discussion of fields) that the guideline applies to more than just parameters.

lib/src/rules/avoid_annotating_parameters_with_function.dart
```diff
@@ -0,0 +1,59 @@
...
+  AstVisitor getVisitor() => _visitor;
+}
+
+class _Visitor extends SimpleAstVisitor {
+  final LintRule rule;
+  _Visitor(this.rule);
+
+  @override
+  visitSimpleFormalParameter(SimpleFormalParameter node) {
+    if (DartTypeUtilities.isClass(node.type.type, 'Function', 'dart.core')) {
```
bwilkerson: This doesn't account for the exception given in the guideline. It's worth thinking about whether we can detect that case.

Issue #142, opened 2015-10-16
Linter features for Flutter code


Hixie:
Flutter adds to the Dart style guide with:

   https://github.com/flutter/engine/blob/master/sky/specs/style-guide.md

Many things in that list could be caught by a linter (or a code reformatter), such as:
- [x] Never use `var`.
- [x] Always type everything explicitly except arguments named `_`, `__`, `___`, etc.
- [x] Never use `as`.
- [x] Constructors always come first.
- [x] Default (unnamed) constructor should come first among the constructors.
- [ ] Whitespace inside `{}`, `[]`, `()` blocks should be symmetric at the start and end. (i.e. If a block starts with a newline, it ends with a newline. If it starts with two, it ends with two. If it starts with a space, it ends with a space. And so on.)
- [ ] Don't call super if you have no arguments to pass up to the superclass.
- [ ] Always put the body of an `if` on its own line.
- [ ] Avoid braces around one-line code blocks (e.g. in `if` statements) unless there's a chain (with `else`) and one of the elements in the chain has a multi-line block.
- [ ] Use a switch when examining an enum, not an `if` chain.
- [ ] Only use `=>` if both sides fit on the same line.

pq:
For reference, the wrapper bits we want to do away with (formerly `skyanalyzer`) are now [here]( https://github.com/flutter/flutter/blob/master/packages/flutter_tools/lib/src/commands/analyze.dart).

pq:
@Hixie : closing for staleness.  Feel free to re-open or harvest into a fresh ticket.

Thanks!

Hixie:
I don't understand. In what sense is it stale?

(I can't reopen.)

pq:
Sorry, too brash.  I was meaning possibly not up to date and maybe superseded by individual issues.  Happy to keep it open. 👍

Hixie:
I've updated the list slightly but it's mostly still the same as before.

pq:
Awesome thanks!

> it's mostly still the same as before.

(Though that is a little sad.)

pq:
See also: https://github.com/flutter/website/issues/1985.

PR #562, opened 2017-04-04
add always_put_control_body_on_new_line

a14n:
This addresses [Flutter style: Separate the 'if' expression from its statement](https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#separate-the-if-expression-from-its-statement) that is present in #142

### DIFF
--- a/lib/src/rules/always_put_control_body_on_new_line.dart
+++ b/lib/src/rules/always_put_control_body_on_new_line.dart
@@ -91,8 +91,12 @@ class Visitor extends SimpleAstVisitor {
     // if (node is Block) return;
 
     final unit = node.root as CompilationUnit;
-    if (unit.lineInfo.getLocation(controlEnd).lineNumber ==
-        unit.lineInfo.getLocation(node.end).lineNumber) {
+    if (node is Block &&
+            unit.lineInfo.getLocation(controlEnd).lineNumber ==
+                unit.lineInfo.getLocation(node.end).lineNumber ||
+        node is! Block &&
+            unit.lineInfo.getLocation(controlEnd).lineNumber ==
+                unit.lineInfo.getLocation(node.offset).lineNumber) {
       rule.reportLintForToken(node.beginToken);
     }
   }
--- a/lib/src/rules/always_put_control_body_on_new_line.dart
+++ b/lib/src/rules/always_put_control_body_on_new_line.dart
@@ -13,9 +13,9 @@ From the [flutter style guide](https://flutter.io/style-guide/):
 
 **DO** Separate the control structre expression from its statement.
 
-Don't put the statement part of an `if`, `for`, `while` on the same line as the
-expression, even if it is short. (Doing so makes it unobvious that there is
-relevant code there. This is especially important for early returns.)
+Don't put the statement part of an `if`, `for`, `while`, `do` on the same line
+as the expression, even if it is short. (Doing so makes it unobvious that there
+is relevant code there. This is especially important for early returns.)
 
 **GOOD:**
 ```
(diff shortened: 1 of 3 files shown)


lib/src/rules/always_put_control_body_on_new_line.dart
```diff
@@ -0,0 +1,110 @@
...
+            group: Group.style);
+
+  @override
+  AstVisitor getVisitor() => new Visitor(this);
+}
+
+class Visitor extends SimpleAstVisitor {
+  final LintRule rule;
+
+  Visitor(this.rule);
```
bwilkerson: Should this also cover the `do` statement?

lib/src/rules/always_put_control_body_on_new_line.dart
```diff
@@ -0,0 +1,110 @@
...
+    if (_getLine(content, controlEnd) == _getLine(content, node.end)) {
+      rule.reportLintForToken(node.beginToken);
+    }
+  }
+}
+
+const int _LF = 10;
+const int _CR = 13;
+
+int _getLine(String content, int offset) {
```
bwilkerson: Each CompilationUnit has a `lineInfo` object that can be used to more efficiently translate offsets to line numbers.

test/rules/always_put_control_body_on_new_line.dart
```diff
@@ -0,0 +1,59 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// test w/ `pub run test -N always_put_control_body_on_new_line`
+
+testIfElse() {
```
bwilkerson: What about something like
```
if (false) object.longMethodInvocation(
  'First argument',
  'Second argument');
```
Should that produce a lint?

lib/src/rules/always_put_control_body_on_new_line.dart
a14n: Yes, added.

lib/src/rules/always_put_control_body_on_new_line.dart
a14n: I hadn't find it :) Thanks

test/rules/always_put_control_body_on_new_line.dart
a14n: Reading the [Flutter rule](https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#separate-the-if-expression-from-its-statement) I think it should.

bwilkerson: I thought it should too, but looking at the code, I don't think it will. Because we compare the line containing the right paren for the condition with the line containing the semicolon, I think it will decide that the whole statement is on a separate line. I think you need to use the first character of the `node` and have special handling for blocks, such as testing the location of the first character of the first statement in the block. (I'd ignore empty blocks, because that seems like something that should be covered by a different lint.) It would be good to add a test of that form.

a14n: Done (except for empty block).

@Hixie should an empty block be linted ?
```dart
while (handleNext()) { }
```

Hixie: No, the `{` can be on the same line as the control flow keyword and the `}` can be on the same line as the `{`.

I have to say, though, I'm not sure that a linter is the right place for this kind of thing. I think the main value of a linter is in finding semantic problems with the code. Syntactic issues are much more driven by aesthetic and it's much harder to apply hard-and-fast rules. The style guide's overriding concern is that the code be readable. If there's some case where having a return on the same line as an if is clearly a win, I wouldn't want to have to then follow it with an `//ignore`, since that would defeat the entire purpose.

a14n: I agree.

I see this lint as a workaround until a flutter formatter lands. However once a formatter will be used you will face the same kind of issue/concern : the same rule for all the code.

But I don't think there will be any `//ignore`. When I run this lint on the Flutter codebase I have no lint anymore (except for a `while (...) { }`) since https://github.com/flutter/flutter/pull/9177

Obviously this lint could be removed when the flutter formatter will land.

a14n: I forgot to mention that there's an open issue for [Linter features for Flutter code](https://github.com/dart-lang/linter/issues/142) what contains this lint.

lib/src/rules/always_put_control_body_on_new_line.dart
```diff
@@ -0,0 +1,103 @@
...
+    _checkNodeOnNextLine(node.body, node.doKeyword.end);
+  }
+
+  void _checkNodeOnNextLine(AstNode node, int controlEnd) {
+    if (node is Block && node.statements.isEmpty) return;
+
+    final unit = node.root as CompilationUnit;
+    if (node is Block &&
+            unit.lineInfo.getLocation(controlEnd).lineNumber ==
+                unit.lineInfo.getLocation(node.end).lineNumber ||
```
bwilkerson: I think this will fail to lint
```
if (cond) { print('should be on next line');
}
```
For a block we might want to check the first character of the first statement, rather than the closing brace. What do you think?

JPaulsen:
@a14n you should add a test with this code:

```dart
if (something) {
}
else if (otherThing) { // OK
}
else {
}
```

Assuming that the second if statement is a control body of the first else statement, probably you rule could lint that and I think it should not do it. It would suggest something like:

```dart
if (something) {
}
else
  if (otherThing) {
  }
```

Also consider adding test like this:

```dart
for (;;) if (something) {
}
```

with while or other things, I do not have a strong opinion about lint or not lint that (I have seen code like this before), but anyway you should add a test to be clear with that.

I hope my comment was helpful.

bwilkerson:
I suspect that the else-if style is Flutter approved, but the others are probably not.

JPaulsen:
I think the same, IntelliJ formats it like @bwilkerson says

a14n:
ptal

PR #566, opened 2017-04-05
Fix lint in rule: use_setters_to_change_a_property

JPaulsen:
Fixes https://github.com/dart-lang/linter/issues/565

### DIFF
--- a/lib/src/rules/use_setters_to_change_a_property.dart
+++ b/lib/src/rules/use_setters_to_change_a_property.dart
@@ -62,7 +62,7 @@ class _Visitor extends SimpleAstVisitor {
         node.returnType?.type?.name != 'void') {
       return;
     }
-    bool _visitExpression(Expression expression) {
+    void _visitExpression(Expression expression) {
       if (expression is AssignmentExpression) {
         final leftOperand = DartTypeUtilities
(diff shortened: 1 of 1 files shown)


lib/src/rules/use_setters_to_change_a_property.dart
```diff
@@ -62,7 +62,7 @@ class _Visitor extends SimpleAstVisitor {
         node.returnType?.type?.name != 'void') {
       return;
     }
-    bool _visitExpression(Expression expression) {
+    _visitExpression(Expression expression) {
```
bwilkerson: If the expectation is that it doesn't return any value, then we should use `void` rather than remove the return type.

JPaulsen: Should we fix the lint also? (It does not lint if there is no return type)

bwilkerson: I don't have a strong opinion, but the absence of a return type means that the return type defaults to `dynamic`, which means that the method can return anything. If the lint should only lint methods that don't return a value, then linting when the return type is implicit seems wrong.

JPaulsen: Fixed

Issue #483, opened 2017-03-10
Lint to use ??= instead of if statement


alexeieleusis:
```dart
  String get fullName {
    if (_fullName == null) {
      _fullName = getFullUserName(this);
    }
    return _fullName;
  }
```
Should be
```dart
  String get fullName {
    return _fullName ??= getFullUserName(this);;
  }
```

JPaulsen:
@bwilkerson @a14n @alexeieleusis, would someone like to propose a name for this rule?

alexeieleusis:
Probably `prefer_conditional_assignation`.

a14n:
`prefer_conditional_assignment` sgtm

bwilkerson:
`prefer_conditional_assignment` sgtm too.

If you're going to provide a quick fix for the lint, though, consider producing
```
String get fullName => _fullName ??= getFullUserName(this);
```

JPaulsen:
Yes, I am thinking about lint and detect this patter in codes:

```dart
if (_fullName == null) {
  _fullName = getFullUserName(this);
}
```

That must be replaced by:

```dart
_fullName ??= getFullUserName(this);
```

But the return and => is not something relevant in this case: maybe we could create another lint to merge some returns with assignment expressions, what do you think about that? to lint this cases:

```dart
String get fullName {
  _fullName ??= getFullUserName(this);
  return _fullName;
}
```

Because if they join them: 

```dart
String get fullName {
  return _fullName ??= getFullUserName(this);;
}
```

Then this [lint](https://github.com/dart-lang/linter/pull/464) will finish to suggest doing what @bwilkerson said:

```dart
String get fullName => _fullName ??= getFullUserName(this);
```

bwilkerson:
Generally speaking, modularity is good, so thinking about splitting this up into a sequence of steps seems reasonable, but we have to weigh that against the extra steps that a user would need to go through.

Assume that we generate a lint that suggests using conditional assignment for the original example:
```
String get fullName {
  if (_fullName == null) {
    _fullName = getFullUserName(this);
  }
  return _fullName;
}
```
It would be much more valuable for the user if we could offer a single fix that would perform all three transformations than if we offered a fix that transformed it into something (conditional assignment followed by a return) that would cause a different lint to be generated where that had a fix that would transform it into something (single return statement) that would cause a different lint to be generated where that had a fix that would finally get rid of the lints (using an expression body).

[...]

PR #567, opened 2017-04-05
New rule: prefer_conditional_assignment

JPaulsen:
Fixes https://github.com/dart-lang/linter/issues/483

### DIFF
--- a/lib/src/rules/prefer_conditional_assignment.dart
+++ b/lib/src/rules/prefer_conditional_assignment.dart
@@ -1,0 +1,109 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+library linter.src.rules.prefer_conditional_assignment;
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/token.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/element.dart';
+import 'package:linter/src/analyzer.dart';
+import 'package:linter/src/util/dart_type_utilities.dart';
+
+const _desc = r'Prefer ??= over testing for null.';
+
+const _details = r'''
+
+**PREFER** ??= over testing for null.
+
+**BAD:**
+```
+String get fullName {
+  if (_fullName == null) {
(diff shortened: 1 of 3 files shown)


lib/src/rules/prefer_conditional_assignment.dart
```diff
@@ -0,0 +1,102 @@
...
+**GOOD:**
+```
+String get fullName {
+  return _fullName ??= getFullUserName(this);;
+}
+```
+
+''';
+
+Element _checkConditionAndGetElement(Expression node) {
```
JPaulsen: I have to split this method in 2 methods

JPaulsen: Fixed

JPaulsen:
178 lints in the SDK, no false positives found.

lib/src/rules/prefer_conditional_assignment.dart
```diff
@@ -0,0 +1,106 @@
...
+  final LintRule rule;
+  _Visitor(this.rule);
+
+  @override
+  visitIfStatement(IfStatement node) {
+    final elementInCondition = _getElementInCondition(node.condition);
+    if (elementInCondition != null &&
+        _checkStatement(node.thenStatement, elementInCondition)) {
+      rule.reportLint(node);
+    }
```
bwilkerson: What about
```
if (x == null) {
  x = 'string';
} else {
  x = '$x$x';
}
```

test/rules/prefer_conditional_assignment.dart
```diff
@@ -0,0 +1,99 @@
...
+}
+
+class Person {
+  int x;
+  String _fullName;
+  String get badWithBlock1 {
+    if (_fullName == null) { // LINT
+      _fullName = getFullUserName(this);
+    }
+    return _fullName;
```
bwilkerson: The `return` statements are uninteresting. Consider removing them.

lib/src/rules/prefer_conditional_assignment.dart
JPaulsen: I had that in mind and I forgot it!, sorry, elseStatement must be null!, thanks.

JPaulsen: Fixed

test/rules/prefer_conditional_assignment.dart
JPaulsen: Fixed

PR #570, opened 2017-04-06
always_use_required_for_non_null_named_parameter

a14n:


lib/src/rules/always_use_required_for_non_null_named_parameter.dart
```diff
@@ -0,0 +1,115 @@
...
+
+bool _isRequired(Element element) =>
+    element is PropertyAccessorElement &&
+    element.name == _REQUIRED_VAR_NAME &&
+    element.library?.name == _META_LIB_NAME;
+
+class AlwaysUseRequiredForNonNullNamedParameter extends LintRule {
+  AlwaysUseRequiredForNonNullNamedParameter()
+      : super(
+            name: 'always_use_required_for_non_null_named_parameter',
```
bwilkerson: The name is a bit long. I'd like to think of something even shorter, but perhaps shorten it to "always_require_non_null_named_parameters".

lib/src/rules/always_use_required_for_non_null_named_parameter.dart
```diff
@@ -0,0 +1,115 @@
...
+  }
+
+  @override
+  visitFormalParameterList(FormalParameterList node) {
+    final namedParametersWithoutDefault = node.parameters
+        .where((p) => p.kind == ParameterKind.NAMED)
+        .map((p) => p as DefaultFormalParameter)
+        .where((p) => p.defaultValue == null);
+    for (final param in namedParametersWithoutDefault) {
+      if (param.metadata.any((a) => _isRequired(a.element))) continue;
```
bwilkerson: I don't care either way, but you could also include this test in the final `where` closure on line 80. I'm not sure which reads better.

lib/src/rules/always_use_required_for_non_null_named_parameter.dart
```diff
@@ -0,0 +1,115 @@
...
+      }
+    }
+  }
+
+  _checkBody(DefaultFormalParameter param, FunctionBody body) {
+    if (body is BlockFunctionBody) {
+      final hasNotNullAssert = body.block.childEntities
+          .skip(1) // the first `{`
+          .takeWhile((e) => e is AssertStatement)
+          .any((e) => _isAssertNotNull(e, param.identifier));
```
bwilkerson: Given that you only want statements from the block, perhaps `block.statements` would work better than `block.childEntities`.

lib/src/rules/always_use_required_for_non_null_named_parameter.dart
```diff
@@ -0,0 +1,115 @@
...
+      if (hasNotNullAssert) {
+        rule.reportLintForToken(param.identifier.beginToken);
+      }
+    }
+  }
+
+  _isAssertNotNull(AssertStatement node, SimpleIdentifier identifier) {
```
bwilkerson: This finds `p != null`, but won't find `null != p`. I don't know whether you care about that, but wanted to point it out.

lib/src/rules/always_use_required_for_non_null_named_parameter.dart
```diff
@@ -0,0 +1,115 @@
...
+  Visitor(this.rule);
+
+  void checkLiteral(TypedLiteral literal) {
+    if (literal.typeArguments == null) {
+      rule.reportLintForToken(literal.beginToken);
+    }
+  }
+
+  @override
+  visitFormalParameterList(FormalParameterList node) {
```
bwilkerson: The performance might be fine as is, but you could make this more efficient by gathering the names of all of the variables whose null-ness is asserted before iterating over the list of required parameters. For methods with more than one named parameter this would reduce the amount of duplicated work.

lib/src/rules/always_use_required_for_non_null_named_parameter.dart
a14n: done

lib/src/rules/always_use_required_for_non_null_named_parameter.dart
a14n: done

lib/src/rules/always_use_required_for_non_null_named_parameter.dart
a14n: done

lib/src/rules/always_use_required_for_non_null_named_parameter.dart
a14n: done

lib/src/rules/always_use_required_for_non_null_named_parameter.dart
a14n: done

a14n:
Could someone merge this PR ? (I'm not allowed to do that)

PR #572, opened 2017-04-07
New rule: do_not_create_setter_without_getter

JPaulsen:
Fixes https://github.com/dart-lang/linter/issues/571

### DIFF
--- a/lib/src/rules.dart
+++ b/lib/src/rules.dart
@@ -7,8 +7,8 @@ library linter.src.rules;
 import 'package:linter/src/analyzer.dart';
 import 'package:linter/src/rules/always_declare_return_types.dart';
 import 'package:linter/src/rules/always_put_control_body_on_new_line.dart';
-import 'package:linter/src/rules/always_specify_types.dart';
 import 'package:linter/src/rules/always_require_non_null_named_parameters.dart';
+import 'package:linter/src/rules/always_specify_types.dart';
 import 'package:linter/src/rules/annotate_overrides.dart';
 import 'package:linter/src/rules/avoid_annotating_with_dynamic.dart';
 import 'package:linter/src/rules/avoid_as.dart';
@@ -34,6 +34,7 @@ import 'package:linter/src/rules/control_flow_in_finally.dart';
 import 'package:linter/src/rules/directives_ordering.dart';
 import 'package:linter/src/rules/do_not_catch_errors.dart';
 import 'package:linter/src/rules/do_not_check_for_null_in_equality_operators.dart';
+import 'package:linter/src/rules/do_not_create_setter_without_getter.dart';
 import 'package:linter/src/rules/empty_catches.dart';
 import 'package:linter/src/rules/empty_constructor_bodies.dart';
 import 'package:linter/src/rules/empty_statements.dart';
@@ -96,8 +97,8 @@ void registerLintRules() {
   Analyzer.facade
     ..register(new AlwaysDeclareReturnTypes())
     ..register(new AlwaysPutControlBodyOnNewLine())
-    ..register(new AlwaysSpecifyTypes())
     ..register(new AlwaysRequireNonNullNamedParameters())
+    ..register(new AlwaysSpecifyTypes())
     ..register(new AnnotateOverrides())
     ..register(new AvoidAnnotatingWithDynamic())
     ..register(new AvoidTypesOnClosureParameters())
@@ -123,6 +124,7 @@ void registerLintRules() {
     ..register(new DirectivesOrdering())
     ..register(new DoNotCatchErrors())
     ..register(new DoNotCheckForNullInEqualityOperators())
+    ..register(new DoNotCreateSetterWithoutGetter())
     ..register(new EmptyCatches())
     ..registerDefault(new EmptyConstructorBodies())
     ..register(new EmptyStatements())
--- a/lib/src/rules/do_not_create_setter_without_getter.dart
+++ b/lib/src/rules/do_not_create_setter_without_getter.dart
@@ -1,0 +1,79 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:linter/src/analyzer.dart';
+import 'package:linter/src/ast.dart';
+import 'package:linter/src/util/dart_type_utilities.dart';
+
+const _desc = r' ';
+
+const _details = r'''
+
+**DON’T** define a setter without a corresponding getter.
+
+**BAD:**
+```
+class Bad {
+  int l, r;
+
+  set length(int newLength) {
+    r = l + newLength;
+  }
+}
+```
+
+**GOOD:**
+```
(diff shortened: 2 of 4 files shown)


lib/src/rules.dart
```diff
@@ -18,6 +18,7 @@ import 'package:linter/src/rules/avoid_init_to_null.dart';
 import 'package:linter/src/rules/avoid_return_types_on_setters.dart';
 import 'package:linter/src/rules/avoid_returning_null.dart';
 import 'package:linter/src/rules/avoid_returning_this.dart';
+import 'package:linter/src/rules/do_not_create_setter_without_getter.dart';
```
bwilkerson: nit: not sorted

lib/src/rules/do_not_create_setter_without_getter.dart
```diff
@@ -0,0 +1,92 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+library linter.src.rules.avoid_setter_without_getter;
```
bwilkerson: nit: the library name didn't get updated. I'm surprised there was no error from importing multiple libraries with the same name. (By the way, library names are not needed, so one way to fix this would be to remove the library directive.)

lib/src/rules/do_not_create_setter_without_getter.dart
```diff
@@ -0,0 +1,92 @@
...
+
+**DON’T** define a setter without a corresponding getter.
+
+**BAD:**
+```
+class Bad {
+  int l, r;
+
+  set length(int newLength) {
+    r = l + length;
```
bwilkerson: "length" --> "newLength"?

lib/src/rules/do_not_create_setter_without_getter.dart
```diff
@@ -0,0 +1,92 @@
...
+
+**GOOD:**
+```
+class Good {
+  int l, r;
+
+  int get length => r - l;
+
+  set length(int newLength) {
+    r = l + length;
```
bwilkerson: This work, but is suspicious looking.

lib/src/rules/do_not_create_setter_without_getter.dart
```diff
@@ -0,0 +1,92 @@
...
+
+''';
+
+bool _hasInheritedGetter(MethodDeclaration node) =>
+    DartTypeUtilities.lookUpInheritedConcreteGetter(node) != null;
+
+bool _hasInheritedSetter(MethodDeclaration node) =>
+    DartTypeUtilities.lookUpInheritedConcreteSetter(node) != null;
+
+class AvoidSetterWithoutGetter extends LintRule {
```
bwilkerson: nit: the class name should match the rule name (DoNotCreateSetterWithoutGetter)

lib/src/rules/do_not_create_setter_without_getter.dart
```diff
@@ -0,0 +1,92 @@
...
+      if (method.isGetter) {
+        getters.add(method.name.toString());
+      } else if (method.isSetter &&
+          !_hasInheritedSetter(method) &&
+          !_hasInheritedGetter(method)) {
+        setters[method.name.toString()] = method;
+      }
+    }
+    void _reportLintFromKey(String id) {
+      rule.reportLint(setters[id]);
```
bwilkerson: The lint should be on the name of the method, not the whole method declaration.

lib/src/rules/do_not_create_setter_without_getter.dart
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    setters.keys.toSet().difference(getters).forEach(_reportLintFromKey);
```
bwilkerson: I believe that there is API for looking up the declaration of a getter in either a class or its superclasses, in which case there's no reason to build up collections. You could just iterate over the setters and ensure that either there is an overridden setter or that there is a getter, and if not report the lint immediately.

test/rules/do_not_create_setter_without_getter.dart
```diff
@@ -0,0 +1,33 @@
...
+  int _x;
+
+  int get x => _x;
+}
+
+class D extends C {
+  set x(int x) { // OK because has inherited getter.
+    this._x = x;
+  }
+}
```
bwilkerson: You should add a test that there is no lint if the getter and setter are in the same class.

lib/src/rules/do_not_create_setter_without_getter.dart
JPaulsen: I missed when changed the name of the rule, thanks for noticing it

lib/src/rules.dart
JPaulsen: Fixed

lib/src/rules/do_not_create_setter_without_getter.dart
JPaulsen: Fixed

lib/src/rules/do_not_create_setter_without_getter.dart
JPaulsen: Fixed (I replaced it by newLength)

lib/src/rules/do_not_create_setter_without_getter.dart
JPaulsen: Fixed

lib/src/rules/do_not_create_setter_without_getter.dart
JPaulsen: Fixed

test/rules/do_not_create_setter_without_getter.dart
JPaulsen: Fixed

lib/src/rules/do_not_create_setter_without_getter.dart
JPaulsen: Fixed

lib/src/rules.dart
```diff
@@ -105,6 +106,7 @@ void registerLintRules() {
     ..register(new AvoidReturningNull())
     ..register(new AvoidReturnTypesOnSetters())
     ..register(new AvoidReturningThis())
+    ..register(new DoNotCreateSetterWithoutGetter())
```
bwilkerson: nit: I think we try to keep these lines in alphabetic order (which it was before the rename).

test/rules/do_not_create_setter_without_getter.dart
```diff
@@ -0,0 +1,41 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+// test w/ `pub run test -N do_not_create_setter_without_getter`
+
+class A {
+  // ignore: unused_field
+  int _x;
+
+  set x(int x) { // LINT
+    this._x = x;
```
bwilkerson: Not that it matters, but you could replace the whole class by
```
class A {
  set x(x) {} // LINT
}
```
which would shorten the test file. You can minimally remove the bodies from all of the members because this rule ignores them.

lib/src/rules.dart
JPaulsen: Fixed

test/rules/do_not_create_setter_without_getter.dart
JPaulsen: Fixed

Issue #573, opened 2017-04-07
Create util functions.


JPaulsen:
From @bwilkerson comments:

https://github.com/dart-lang/linter/pull/532#discussion_r110283095
https://github.com/dart-lang/linter/pull/532#discussion_r110283630

pq:
Assuming stale.

PR #582, opened 2017-04-07
New rule: use_to_and_as_if_applicable

JPaulsen:
Fixes https://github.com/dart-lang/linter/issues/580
Fixes https://github.com/dart-lang/linter/issues/581
And also is relevant to https://github.com/dart-lang/linter/issues/573

### DIFF
--- a/lib/src/rules/use_to_and_as_if_applicable.dart
+++ b/lib/src/rules/use_to_and_as_if_applicable.dart
@@ -1,0 +1,107 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+library linter.src.rules.use_to_and_as_if_applicable;
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:linter/src/analyzer.dart';
+import 'package:linter/src/util/dart_type_utilities.dart';
+
+const _desc =
+    r'Start the name of the method with to/_to or as/_as if applicable.';
+
+const _details = r'''
+
+**PREFER** naming a method to___() if it copies the object’s state to a new object.
+
+**PREFER** naming a method as___() if it returns a different representation backed by the original object.
+
+**BAD:**
+```
+class Bar {
+  Foo myMethod() {
+    return new Foo.from(this);
+  }
+}
+```
+
+**GOOD:**
+```
+class Bar {
+  Foo toFoo() {
+    return new Foo.from(this);
+  }
+}
+```
+
+**GOOD:**
+```
+class Bar {
+  Foo asFoo() {
+    return new Foo.from(this);
+  }
+}
+```
+
+''';
+
+bool _beginsWithAsOrTo(String name) {
+  final regExp = new RegExp(r"(to|as|_to|_as)[A-Z]", caseSensitive: true);
+  return regExp.matchAsPrefix(name) != null;
+}
+
+bool _isVoid(TypeName returnType) =>
+    returnType != null && returnType.name.name == 'void';
+
+class UseToAndAsIfApplicable extends LintRule {
+  _Visitor _visitor;
+  UseToAndAsIfApplicable()
+      : super(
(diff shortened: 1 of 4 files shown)


JPaulsen:
0 lints in the SDK, if I delete the line that checks if already starts with to and as then apperas 7 lints:

```dart
/usr/local/google/home/casanueva/Development/sdk/sdk/lib/_internal/js_runtime/lib/js_array.dart 592:10 [lint] Start the name of the method with to/_to or as/_as if applicable
  Set<E> toSet() => new Set<E>.from(this);
         ^^^^^
/usr/local/google/home/casanueva/Development/sdk/sdk/lib/_internal/js_runtime/lib/js_array.dart 627:15 [lint] Start the name of the method with to/_to or as/_as if applicable
  Map<int, E> asMap() {
              ^^^^^
/usr/local/google/home/casanueva/Development/sdk/sdk/lib/convert/string_conversion.dart 186:22 [lint] Start the name of the method with to/_to or as/_as if applicable
  ClosableStringSink asStringSink() {
                     ^^^^^^^^^^^^
/usr/local/google/home/casanueva/Development/sdk/sdk/lib/async/future_impl.dart 273:13 [lint] Start the name of the method with to/_to or as/_as if applicable
  Stream<T> asStream() => new Stream<T>.fromFuture(this);
            ^^^^^^^^
/usr/local/google/home/casanueva/Development/sdk/sdk/lib/core/iterable.dart 359:10 [lint] Start the name of the method with to/_to or as/_as if applicable
  Set<E> toSet() => new Set<E>.from(this);
         ^^^^^
/usr/local/google/home/casanueva/Development/sdk/sdk/lib/collection/iterable.dart 91:10 [lint] Start the name of the method with to/_to or as/_as if applicable
  Set<E> toSet() => new Set<E>.from(this);
         ^^^^^
/usr/local/google/home/casanueva/Development/sdk/sdk/lib/collection/list.dart 330:15 [lint] Start the name of the method with to/_to or as/_as if applicable
  Map<int, E> asMap() {
              ^^^^^
```

We did not know if the lint was going to work well, but it seems good.

lib/src/rules/use_to_and_as_if_applicable.dart
```diff
@@ -0,0 +1,109 @@
...
+  Foo asFoo() {
+    return new Foo.from(this);
+  }
+}
+```
+
+''';
+
+bool _beginsWithAsOrTo(String name) {
+  final regExp = new RegExp(r"(to|as|_to|_as)[A-Z].*", caseSensitive: true);
```
bwilkerson: Do we need the `.*`, given that we're matching the prefix?

JPaulsen:
@bwilkerson I don't thinks so, we noticed that in general method that uses to and as, only return a new Something(this); but I do not know how to difference them

JPaulsen:
Maybe we are not explaining the rule well, the idea is that if you have a method that does something like method() => new OtherClase(this); and it is not a getter then should be a toOtherClase or an asOtherClase method. If inside the method keeps the reference to the parameter and changes it when changing itself, then should be an as, but that is really hard to decide without false positives.

bwilkerson:
Ah, I thought the rule was checking for two different cases but not differentiating in the error message. If it isn't, then a single message is appropriate.

lib/src/rules/use_to_and_as_if_applicable.dart
```diff
@@ -0,0 +1,109 @@
...
+        return _checkExpression(statement.expression);
+      }
+    }
+    return false;
+  }
+
+  bool _checkExpression(Expression expression) {
+    if (expression is ParenthesizedExpression) {
+      return _checkExpression(expression.expression);
+    }
```
bwilkerson: I was reminded the other day that there's a better way to write this:
```
expression = expression.unParenthesized;
```
Several recent rules could be cleaned up.

lib/src/rules/use_to_and_as_if_applicable.dart
JPaulsen: Fixed

lib/src/rules/use_to_and_as_if_applicable.dart
JPaulsen: Fixed and issue added.

Issue #584, opened 2017-04-09
Use final expression = rawExpression.unParenthesized to unparenthesize expressions.


JPaulsen:
From https://github.com/dart-lang/linter/pull/582#discussion_r110519961

PR #596, opened 2017-04-10
Fix efficient unparenthesized

JPaulsen:
Fixes https://github.com/dart-lang/linter/issues/584

### DIFF
--- a/lib/src/rules/prefer_conditional_assignment.dart
+++ b/lib/src/rules/prefer_conditional_assignment.dart
@@ -52,29 +52,23 @@ bool _checkStatement(Statement statement, Element element) {
   return false;
 }
 
-Element _getElementInCondition(Expression node) {
-  if (node is ParenthesizedExpression) {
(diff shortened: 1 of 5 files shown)


lib/src/rules/prefer_conditional_assignment.dart
```diff
@@ -52,29 +52,23 @@ bool _checkStatement(Statement statement, Element element) {
...
   return null;
 }
 
-bool _isNullLiteral(Expression node) {
-  if (node is ParenthesizedExpression) {
-    return _isNullLiteral(node.expression);
-  }
-  return node is NullLiteral;
-}
+bool _isNullLiteral(Expression node) => node.unParenthesized is NullLiteral;
```
bwilkerson: Note that this appears in two lints, so it might want to be "promoted" to utils.

JPaulsen: I will open a new issue to do that.

Issue #339, opened 2016-11-24
Invalid hint for cascade


zoechi:
```dart
    final resource = (entry.resource as Practitioner);
    resource.identifier = [
      new Identifier()
        ..system = 'urn:system'
        ..value = 'mdc_${resource.id}'
    ];
```
With cascades `'mdc_${resource.id}'` doesn't work

> Dart VM version: 1.21.0-edge.640b5ca80a2739d2db6466834778c60147ad15c0 (Mon Nov 21 12:35:27 2016) on "linux_x64"

zoechi:
It's also not clear to me where I can add

```dart
// ignore: cascade_invocations  
```

to silence this hint. 

The quickfix adds it before the 2nd line but that doesn't silence the hint.

alexeieleusis:
The comment needs to be one line above the lint you want to silence.

I just tried to reproduce this with the following snippet but I didn't lint:

```
class ClassWithStrings {
  String first;
  String second;
  Foo foo;

  void someThing() {
    final variable = new ClassWithStrings()
        ..first = 'constant'
        ..second = '${foo.bar}';
  }
}
```

Do you have a self contained example? I don't know the classes you are using and might be related with the failure.

zoechi:
```dart
class Identifier {
  String value;
  String system;
}

class Resource {
  String id;
  List<Identifier> identifier;
}

class Practitioner extends Resource{
}

class Entry {
  Resource resource;
}

void main() {
  final entry = new Entry(); 
  final resource = (entry.resource as Practitioner);
  // ignore: cascade_invocations
  resource.identifier = [
    new Identifier()
      ..system = 'urn:system'
      ..value = 'mdc_${resource.id}'
  ];
}
```
I stil get the linter warning for `resource.identifier = [...`
```
analyzer:
  strong-mode:
    implicit-casts: false
    implicit-dynamic: false

#  language:
#    enableGenericMethods: true

  exclude:
#    - 'build/**'
#    - '**.g.dart'
    - 'bazel-*'

linter:
  rules:
# Error Rules
    - avoid_empty_else
    - cancel_subscriptions
    - close_sinks
    - comment_references
    - control_flow_in_finally
    - empty_statements
    - hash_and_equals
    - invariant_booleans
    - iterable_contains_unrelated_type
    - list_remove_unrelated_type
    - literal_only_boolean_expressions
    - test_types_in_equals
    - throw_in_finally
    - unrelated_type_equality_checks
    - valid_regexps

[...]

PR #603, opened 2017-04-11
Fix rule: cascade_invocations

JPaulsen:
Fixes https://github.com/dart-lang/linter/issues/339
Fixes https://github.com/dart-lang/linter/issues/340
Fixes https://github.com/dart-lang/linter/issues/606

### DIFF
--- a/lib/src/rules/cascade_invocations.dart
+++ b/lib/src/rules/cascade_invocations.dart
@@ -9,6 +9,7 @@ import 'package:analyzer/dart/ast/token.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
 import 'package:analyzer/dart/element/element.dart';
 import 'package:linter/src/analyzer.dart';
+import 'package:linter/src/util/dart_type_utilities.dart';
 
 const _desc = r'Cascade consecutive method invocations on the same reference.';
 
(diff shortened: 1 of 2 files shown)


JPaulsen:
@bwilkerson it was refactores how we discussed, tests added and issues refreshed.

test/rules/cascade_invocations.dart
```diff
@@ -40,9 +59,10 @@ class Foo {
 void noCascadeWithGetter() {
   final foo = new Foo();
   foo.baz(); // LINT
-  foo.bar; // LINT
-  foo.foo(); // LINT
+  foo.bar; // OK
+  foo.foo(); // OK
   foo.bar = 8; // LINT
```
bwilkerson: What's the rationale for producing a lint on a setter invocation when we don't producing a lint on a getter invocation? (And why do we not lint for a getter invocation?)

test/rules/cascade_invocations.dart
```diff
@@ -17,15 +24,27 @@ int get someThing => 5;
...
+  // toString is not void and it is not a problem.
+}
+
+void cascadeIntermediate() {
+  List<int> list = [];
+  print(list);
+  list.clear();
+  list.clear(); // LINT
+  list..clear(); // LINT
+  list..clear(); // OK
```
JPaulsen: :O I did not know cascades well at that time, you are totally right we must lint those cases also!

test/rules/cascade_invocations.dart
JPaulsen: @alexeieleusis knows the answer to that question

alexeieleusis: The original intention of the change was to lint things that looked like builder invocations, but since I couldn't find the code that I had difficulties with when I tried to enable the lint in my project we agreed on expanding the scope. That said, this now should be linted too.

JPaulsen: Fixed

test/rules/cascade_invocations.dart
JPaulsen: Fixed

lib/src/rules/cascade_invocations.dart
```diff
@@ -53,30 +54,50 @@ someReference
...
-
-  if (expression is MethodInvocation &&
-      expression.realTarget is SimpleIdentifier) {
-    return expression.realTarget;
+Element _getElementFromVariableDeclarationStatement(
+    VariableDeclarationStatement statement) {
+  final variables = statement.variables.variables;
+  if (variables.length == 1) {
+    return DartTypeUtilities.getCanonicalElementFromIdentifier(
+        statement.variables.variables.first.name);
```
bwilkerson: You've already assigned `statement.variables.variables` to a local variable, so minimally this could be cleaned up to be `variables.first.name`.

But it would be better to just replace the whole statement with `return variables.first.element`.

lib/src/rules/cascade_invocations.dart
```diff
@@ -96,57 +117,147 @@ class CascadeInvocations extends LintRule {
   AstVisitor getVisitor() => _visitor;
 }
 
-class _Visitor extends SimpleAstVisitor {
-  final LintRule rule;
-  _Visitor(this.rule);
+class _ExpressionBox {
```
bwilkerson: The name `_ExpressionBox` doesn't really convey any semantic value. It would be good if you could come up with a better name. (One approach that often works for me is to write a documentation comment because it forces me to think about what the object really represents.) Minimally, add a comment to explain how this is being used.

lib/src/rules/cascade_invocations.dart
JPaulsen: Fixed

lib/src/rules/cascade_invocations.dart
JPaulsen: Fixed

JPaulsen:
Friendly ping @bwilkerson, I added some documentation can you check?

lib/src/rules/cascade_invocations.dart
```diff
@@ -96,57 +116,161 @@ class CascadeInvocations extends LintRule {
   AstVisitor getVisitor() => _visitor;
 }
 
-class _Visitor extends SimpleAstVisitor {
-  final LintRule rule;
-  _Visitor(this.rule);
+/// A CascadableExpression is an object that is build from an expression and
```
bwilkerson: 'is build' --> 'is built'

JPaulsen: Fixed

PR #609, opened 2017-04-14
handle else if in always_put_control_body_on_new_line

### LINKED ISSUE #608: always_put_control_body_on_new_line shall not lint else if.
Just like [this](https://github.com/dart-lang/linter/pull/562#issuecomment-292380219)

a14n:
Fixes #608

### DIFF
--- a/test/rules/always_put_control_body_on_new_line.dart
+++ b/test/rules/always_put_control_body_on_new_line.dart
@@ -19,6 +19,13 @@ testIfElse() {
     return; // OK
   else return; // LINT
 
+  if (false) {
+  }
+  else if (false) { // OK
+  }
+  else {
+  }
+
   if (false) { } // OK
   else return; // LINT
 
--- a/lib/src/rules/always_put_control_body_on_new_line.dart
+++ b/lib/src/rules/always_put_control_body_on_new_line.dart
@@ -63,7 +63,7 @@ class Visitor extends SimpleAstVisitor {
   @override
   visitIfStatement(IfStatement node) {
     _checkNodeOnNextLine(node.thenStatement, node.rightParenthesis.end);
-    if (node.elseKeyword != null)
+    if (node.elseKeyword != null && node.elseStatement is! IfStatement)
       _checkNodeOnNextLine(node.elseStatement, node.elseKeyword.end);
   }
 


test/rules/always_put_control_body_on_new_line.dart
```diff
@@ -19,6 +19,13 @@ testIfElse() {
     return; // OK
   else return; // LINT
 
+  if (false) {
+  }
+  else if (false) { // OK
```
bwilkerson: It probably doesn't matter for this lint, but I think the formatter would format this as
```
if (false) {
} else if (false) {
} else {
}
```
We might want to make sure that formatted code is also handled correctly.

a14n: Actually this lint rule does conflict with `dartfmt` because the Flutter style guide disagree on some formatting rules. So this lint can be seen as an helper until the flutter formatter lands. And once landed I guess this lint could be removed.

BTW I'm a little worried about the existence of 2 formatters.

And last thing, I haven't merge rights :)

bwilkerson: I checked with @sethladd, and as far as I can tell there are no plans to have a second formatter. Some of the code in the Flutter SDK won't be formatted using the existing formatter because of those stylistic differences, but Flutter users are encouraged to use `dartfmt`. That means that this shouldn't be viewed as a short-term "helper", but as a long-term aid to maintaining a different formatting style.

In light of that, I think it's important that any lints we write tolerate (as much as possible) the formatter's style as well. I don't think there's actually any conflict in this case. The formatter isn't trying to put the control body on the same line, and the formatting changes I suggested shouldn't cause any lints of this kind to be generated. (And to be clear, I'm suggesting that we include both styles in the test so that we know the lint is robust across non-critical differences.)

sethladd: > I checked with @sethladd, and as far as I can tell there are no plans to have a second formatter.

Allow me to be even more specific: There are no plans to ask or force our end-user developers to use a different formatter. dartfmt is sufficient for our end-user needs, and the Flutter SDK style isn't required for end-user Flutter developers.

I don't know what the plans are from the Flutter SDK authors for their own formatting needs.

a14n: @bwilkerson this rule is not compatible with `dartfmt`. For instance the following code:

```dart
if (animation.value >= 0.5)
  return childSecondHalf;
```

is formatted to the following by `dartfmt` 

```dart
if (animation.value >= 0.5) return childSecondHalf;
```

and this result is not accepted by the current Flutter SDK style.

bwilkerson: That's why I qualified it with "as much as possible". I still think the example I gave should not produce any lints and that it would be good to have a test case to verify that, but it obviously isn't critical (and your changes were already merged, so it's a rather moot point).

PR #611, opened 2017-04-14
Delint the linter

JPaulsen:


.analysis_options
```diff
@@ -6,13 +6,23 @@ analyzer:
 linter:
   rules:
 #    - always_declare_return_types
+#    - always_put_control_body_on_new_line
+    - always_require_non_null_named_parameters
```
bwilkerson: How did we decide which rules to enable and which to not?

bin/linter.dart
```diff
@@ -173,8 +173,9 @@ void runLinter(List<String> args, LinterOptions initialLintOptions) {
         fileRoot: commonRoot,
         showStatistics: stats,
         machineOutput: options['machine'],
-        quiet: options['quiet'])..write();
-  } catch (err, stack) {
+        quiet: options['quiet'])
+      ..write();
+  } on Exception catch (err, stack) {
```
bwilkerson: This is not semantically equivalent, and I think we want to catch everything here. Which means that I think we should ignore the lint here (assuming that we enable it at all).

.analysis_options
JPaulsen: I enable all the new rules except that because it had a recurrent false positive that I reported and @a14n fixed quickly, I will enable it now

bwilkerson: Sorry, I guess I wasn't clear. I'm not questioning why this specific rule *wasn't* enabled, I'm questioning why any of these rules *are* being enabled.

I think it's great that we have rules to support so much of the style guide! I'm very happy for the work you've been doing. The product is better because of it. Thank you!

But the style guide is not universally accepted among the Dart team, and I think there should be a discussion before we start enabling rules that we have not agreed on. There's a reason that lints are disabled by default. :-)

JPaulsen: I got it, I am going to close this branch and wait for the discussions, we are also are implementing quick-fixes so with the delinte we could try to do it automatically

PR #613, opened 2017-04-16
New rule: join_return_with_assignment_when_possible

JPaulsen:


JPaulsen:
11 lints in the sdk, no false positives.

It does not lint variable declarations with return (it could be other rule)

Also it does not lint when has more assignments together, for example:

```dart
int get hashCode {
  int hash = object.hashCode;
  hash <<= 2;
  hash ^= 15; // OK
  return hash;
}
```

lib/src/rules/join_return_with_assignment_when_possible.dart
```diff
@@ -0,0 +1,133 @@
...
+  final visitor = new _ReturnStatementVisitor();
+  node.accept(visitor);
+  return visitor.element;
+}
+
+class JoinReturnWithAssignmentWhenPossible extends LintRule {
+  _Visitor _visitor;
+  JoinReturnWithAssignmentWhenPossible()
+      : super(
+            name: 'join_return_with_assignment_when_possible',
```
bwilkerson: I think we should drop the `_when_possible`.

lib/src/rules/join_return_with_assignment_when_possible.dart
```diff
@@ -0,0 +1,133 @@
...
+  }
+}
+
+class _ReturnStatementVisitor extends SimpleAstVisitor {
+  Element element;
+
+  @override
+  visitReturnStatement(ReturnStatement node) {
+    element =
+        DartTypeUtilities.getCanonicalElementFromIdentifier(node.expression);
```
bwilkerson: Defining a visitor with a single visit method seems unnecessary to me. Given the cost of polymorphic dispatch, it would be more efficient to test the type in the calling function.

lib/src/rules/join_return_with_assignment_when_possible.dart
JPaulsen: Fixed

lib/src/rules/join_return_with_assignment_when_possible.dart
JPaulsen: Fixed

test/rules/join_return_with_assignment.dart
```diff
@@ -0,0 +1,45 @@
...
+  a++; // LINT
+  return a;
+}
+
+int doubleFoo() {
+  int a = 3;
+  a += 5;
+  a += 15; // OK
+  return a;
+}
```
bwilkerson: Do we want to produce a lint if there are multiple assignments but they are not next to each other, as in
```
int a = 3;
a += 5;
print('about to add 15');
a += 15;
return a;
```
This seems inconsistent. (Maybe I just don't understand why the number of assignments matters.)

JPaulsen: It is because there is some code, specially when calculating the hashcode that is like this:

```dart
int get hashCode {
  int hash = object.hashCode;
  hash <<= 2;
  hash &=  object2.hashCode;
  hash *= 32;
  ...
  hash ^= 15; // OK
  return hash;
}
```

So it would be highly probable that the user would like to add a new assignment between the return and the last assignation statement, so if you lint it and you solve the lint, then adding a new one would take a higher cost for the user, that was what I saw running the rule in the sdk

PR #614, opened 2017-04-16
New rule: unnecessary_override

JPaulsen:


JPaulsen:
23 lints in the sdk, no false positives.

lib/src/rules/unnecessary_override.dart
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  @override
+  visitExpressionStatement(ExpressionStatement node) {
+    node.expression.accept(this);
+  }
+
+  @override
+  visitMethodDeclaration(MethodDeclaration node) {
+    if (!node.isGetter && !node.isSetter) {
```
bwilkerson: Why are getters and setters excluded from being linted?

test/rules/unnecessary_override.dart
```diff
@@ -0,0 +1,43 @@
...
+  int getB(Iterable a) { // LINT
+    return super.getB(a);
+  }
+
+  @override
+  int getC(Iterable a) { // OK
+    print("something");
+    return super.getC(a);
+  }
+}
```
bwilkerson: This needs more tests. What about overriding a method with more than one parameter, both with and without changing the order of the arguments?
```
int max(int a, int b) => super.max(a, b); // LINT
int max(int a, int b) => super.max(b, a); // OK
```
Or a method with positional optional parameters, or named parameters (where the order of the named parameters can be changed, both with and without making the override ok)?
```
int m({int a, int b}) => super.m(a: a, b: b); // LINT
int m({int a, int b}) => super.m(b: b, a: a); // LINT
int m({int a, int b}) => super.m(b: a, a: b); // OK
```
Or an operator:
```
int operator +(int other) => super + other; // LINT
```

lib/src/rules/unnecessary_override.dart
JPaulsen: Fixed

test/rules/unnecessary_override.dart
JPaulsen: Fixed

JPaulsen:
I added the case with no targets in getters and setters and also added tests

PR #615, opened 2017-04-17
New rule: unnecessary_statement

JPaulsen:
35 lints in the sdk.

Two of them are these:
```dart
/usr/local/google/home/casanueva/Development/sdk/sdk/lib/_internal/js_runtime/lib/js_mirrors.dart 2362:5 [lint]  
    metadata; // Compute _parameters as a side-effect of extracting metadata.
    ^^^^^^^^^
/usr/local/google/home/casanueva/Development/sdk/sdk/lib/_internal/js_runtime/lib/js_mirrors.dart 2373:5 [lint]  
    metadata; // Compute _returnType as a side-effect of extracting metadata.
    ^^^^^^^^^
```

And 33 lints are in parameters of functions or methods with generics (AST broken).

### DIFF
--- a/lib/src/rules/unnecessary_statement.dart
+++ b/lib/src/rules/unnecessary_statement.dart
@@ -1,0 +1,152 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+library linter.src.rules.unnecessary_statement;
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/token.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:linter/src/analyzer.dart';
+import 'package:linter/src/util/dart_type_utilities.dart';
+
+const _desc =
+    r"Don't write unnecessary statements neither use getters with side-effect.";
+
+const _details = r'''
+
+**DON'T** use unnecessary statements neither use getters with side-effect.
+
+**BAD:**
+```
+void main() {
+  int a;
+  a; // LINT
+}
+```
+
+**GOOD:**
+```
+void main() {
+  int a;
+  foo(a);
+}
+```
+
+''';
+
+class UnnecessaryStatement extends LintRule {
+  _Visitor _visitor;
+  UnnecessaryStatement()
+      : super(
+            name: 'unnecessary_statement',
+            description: _desc,
+            details: _details,
+            group: Group.style) {
+    _visitor = new _Visitor(this);
+  }
+
+  @override
+  AstVisitor getVisitor() => _visitor;
+}
+
+class _UnnecessaryStatementVisitor extends UnifyingAstVisitor {
+  var isUnnecessaryStatement = true;
+
+  @override
+  visitAssignmentExpression(AssignmentExpression node) {
+    isUnnecessaryStatement = false;
+  }
+
+  @override
+  visitBinaryExpression(BinaryExpression node) {
+    if (!isUnnecessaryStatement) {
+      return;
+    }
+    if (node.operator.type == TokenType.EQ_EQ ||
+        node.operator.type == TokenType.BANG_EQ) {
+      if (DartTypeUtilities.isNullLiteral(node.leftOperand)) {
+        node.rightOperand.accept(this);
+      } else if (DartTypeUtilities.isNullLiteral(node.rightOperand)) {
+        node.leftOperand.accept(this);
+      } else {
+        isUnnecessaryStatement = false;
+      }
+    } else if (node.operator.isUserDefinableOperator) {
+      isUnnecessaryStatement = false;
+    } else {
+      visitNode(node);
+    }
+  }
+
+  @override
+  visitFunctionExpressionInvocation(FunctionExpressionInvocation node) {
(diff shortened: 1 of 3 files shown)


lib/src/util/dart_type_utilities.dart
```diff
@@ -83,6 +86,9 @@ class DartTypeUtilities {
   static bool isNullLiteral(Expression expression) =>
       expression?.unParenthesized is NullLiteral;
 
+  static bool isUserDefinedOperator(TokenType type) =>
+      _userDefinedOperatorsSet.contains(type);
```
bwilkerson: Sorry! I should have dug a little further for you. You don't need to define the set yourself, just use `Token.isUserDefinableOperator`.

lib/src/rules/unnecessary_statement.dart
```diff
@@ -0,0 +1,159 @@
...
+    } else {
+      visitNode(node);
+    }
+  }
+
+  @override
+  visitConditionalExpression(ConditionalExpression node) {
+    node.condition.accept(this);
+    node.thenExpression.accept(this);
+    node.elseExpression.accept(this);
```
bwilkerson: I think this is handled by visitNode.

lib/src/rules/unnecessary_statement.dart
```diff
@@ -0,0 +1,159 @@
...
+    isUnnecessaryStatement = false;
+  }
+
+  @override
+  visitBinaryExpression(BinaryExpression node) {
+    if (!isUnnecessaryStatement) {
+      return;
+    }
+    if (node.operator.type == TokenType.EQ_EQ ||
+        node.operator.type == TokenType.BANG_EQ) {
```
bwilkerson: Equality is user-definable, and therefore could have side-effects.

JPaulsen: Yes, but it is not invoked when you compare with null right?, so this part check if it is something like a == null, if it is that, then == it is not invoked and it does not have side-effects

lib/src/util/dart_type_utilities.dart

lib/src/rules/unnecessary_statement.dart
JPaulsen: Fixed

JPaulsen:
@bwilkerson yes, when the parser sees method declarations with generics everything starts to fail, sometimes the bodies are not well-recognized and in this case parameters of these functions where considered statements, so it started to lint them.

zoechi:
AFAIR the rule that getters should not have side effects, is explicitly not part of the Dart style guide @munificent ?
Some people probably still want to enforce this rule.

lib/src/rules/unnecessary_statement.dart
```diff
@@ -0,0 +1,152 @@
...
+            group: Group.style) {
+    _visitor = new _Visitor(this);
+  }
+
+  @override
+  AstVisitor getVisitor() => _visitor;
+}
+
+class _UnnecessaryStatementVisitor extends UnifyingAstVisitor {
+  var isUnnecessaryStatement = true;
```
bwilkerson: I'd like to see this class renamed to be a bit more abstract. What this visitor is really doing, I think, is trying to determine whether we can prove that a given expression cannot have any side-effect. The fact that side-effect free expressions are "unnecessary" when they appear in certain locations is specific to this particular use case, but I'm sure you can imagine other uses cases for determining whether code has a side-effect.

You could leave the field with the current semantics and call it something like `hasNoSideEffect`, but for long term maintenance it would be better to remove the field and allow each `visit` method to return a `bool` indicating whether the visited node is side-effect free, with a fallback for unsupported expressions that returns `false` because that's safe.

bwilkerson:
> ... when the parser sees method declarations with generics everything starts to fail ...

Generic method syntax is not yet enabled by default. You have to enable it explicitly in the analysis options file.

> AFAIR the rule that getters should not have side effects, is explicitly not part of the Dart style guide?
> Some people probably still want to enforce this rule.

I believe that's true, and I have no problem with disallowing getters with side-effects. My concern is that this rule appears to be attempting to do two things: (1) extend the definition of "dead code" to include statements that cannot have any side-effect (but for some reason not do so as part of dead code detection, which is a separate question), and (2) check for places where getters are being used as if they had a side-effect.

The first is an interesting, but error-prone, goal. I've done this kind of analysis for other languages and I know how difficult it can be to strike the right balance between completeness and performance. If nothing else, I'd want this lint to be marked as experimental just for that reason.

But the second is, I think, the wrong thing to do. If you want to discourage getters with side-effects, you don't do so by flagging *uses* of such getters, you do so by flagging *definitions* of such getters, and that kind of checking should be in a separate rule. It isn't safe to assume that no getter will have side-effects, so *this* rule has to assume that getters *can* have side effects and not lint code that invokes a getter (in order to eliminate false positives).

munificent:
+1 to what Brian said.

I agree that most getters should not have side effects, but the language does not *enforce* that rule (and probably can't without some giant language level notion of purity). We could add a lint that assumes that to be true and warns on therefore uses. But the lint is only valid when the assumption is correct, otherwise you get false positives. But by that same token, erroneous code caught by the lint is likely to be rare since... the getter being called doesn't do anything useful.

I'd be surprised if this lint helped many users.

JPaulsen:
According with what @munificent and @bwilkerson said we are going to cancel this PR, thanks for your answers

PR #616, opened 2017-04-17
New rule: use_string_buffer

JPaulsen:
17 lints in the sdk no false positives.

It is linting the left part of assignations that could be replaced by a buffer.write(something);

### DIFF
--- a/lib/src/rules/use_string_buffer.dart
+++ b/lib/src/rules/use_string_buffer.dart
@@ -1,0 +1,191 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+library linter.src.rules.use_string_buffer;
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/token.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/element.dart';
+import 'package:linter/src/analyzer.dart';
+import 'package:linter/src/util/dart_type_utilities.dart';
+
+const _desc = r'Use string buffer to compose strings.';
+
+const _details = r'''
+
+**DO** use string buffer to compose strings.
+
+**BAD:**
+```
+String foo() {
+  final buffer = '';
+  for (int i = 0; i < 10; i++) {
+    buffer += 'a'; // LINT
+  }
+  return buffer;
+}
+```
+
+**GOOD:**
+```
+String foo() {
+  final buffer = new StringBuffer();
+  for (int i = 0; i < 10; i++) {
+    buffer.write('a');
+  }
+  return buffer.toString();
+}
+```
+
+''';
+
+SimpleIdentifier _getSimpleIdentifier(Expression rawExpression) {
+  final expression = rawExpression.unParenthesized;
+  return expression is SimpleIdentifier ? expression : null;
+}
+
+bool _isEmptyInterpolationString(AstNode node) =>
+    node is InterpolationString && node.value == '';
+
+/// The motivation of this rule is performance stuffs, and the explanation is
+/// that if we use N strings additions using the + operator the order of that
+/// algorithm is O(~N^2) and that is because a String is not mutable, so in each
+/// step it creates an auxiliary String that takes O(amount of chars) to be
+/// computed, in otherwise using a StringBuffer the order is reduced to O(~N)
+/// so the bad case is N times slower than the good case.
+class UseStringBuffer extends LintRule {
+  _Visitor _visitor;
+  UseStringBuffer()
+      : super(
+            name: 'use_string_buffer',
+            description: _desc,
+            details: _details,
+            group: Group.style) {
+    _visitor = new _Visitor(this);
+  }
+
+  @override
+  AstVisitor getVisitor() => _visitor;
+}
+
+class _IdentifierIsPrefixVisitor extends SimpleAstVisitor {
+  LintRule rule;
+  SimpleIdentifier identifier;
+  _IdentifierIsPrefixVisitor(this.rule, this.identifier);
+
+  @override
+  visitBinaryExpression(BinaryExpression node) {
+    if (node.operator.type == TokenType.PLUS) {
+      node.leftOperand.accept(this);
+    }
+  }
+
+  @override
+  visitInterpolationExpression(InterpolationExpression node) {
+    node.expression.accept(this);
+  }
+
+  @override
+  visitParenthesizedExpression(ParenthesizedExpression node) {
+    node.unParenthesized.accept(this);
+  }
+
+  @override
+  visitSimpleIdentifier(SimpleIdentifier node) {
+    if (node.bestElement == identifier.bestElement) {
+      rule.reportLint(identifier);
+    }
+  }
+
+  @override
+  visitStringInterpolation(StringInterpolation node) {
+    if (node.elements.length >= 2 &&
+        _isEmptyInterpolationString(node.elements.first)) {
+      node.elements[1].accept(this);
+    }
+  }
+}
+
+class _UseStringBufferVisitor extends SimpleAstVisitor {
+  LintRule rule;
+  final localElements = new Set<Element>();
+  _UseStringBufferVisitor(this.rule);
+
+  @override
+  visitAssignmentExpression(AssignmentExpression node) {
+    if (node.operator.type == TokenType.PLUS_EQ) {
+      final identifier = _getSimpleIdentifier(node.leftHandSide);
+      if (identifier != null) {
+        _visitSimpleIdentifier(identifier);
+      }
+    }
+    if (node.operator.type == TokenType.EQ) {
+      final identifier = _getSimpleIdentifier(node.leftHandSide);
+      if (identifier != null) {
+        final visitor = new _IdentifierIsPrefixVisitor(rule, identifier);
+        node.rightHandSide.accept(visitor);
+      }
+    }
+  }
+
+  @override
+  visitBlock(Block block) {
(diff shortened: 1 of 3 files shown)


lib/src/rules/use_string_buffer.dart
```diff
@@ -0,0 +1,212 @@
...
+  return buffer.toString();
+}
+```
+
+''';
+
+SimpleIdentifier _getSimpleIdentifier(AstNode node) {
+  final visitor = new SimpleIdentifierVisitor();
+  node.accept(visitor);
+  return visitor.identifier;
```
bwilkerson: Creating a visitor in order to ignore parentheses seems like overkill. Why not:
```
SimpleIdentifier _getSimpleIdentifier(Expression node) {
  final unwrappedNode = node.unParenthesized;
  if (unwrappedNode is SimpleIdentifier) {
    return unwrappedNode;
  }
  return null;
}
```

lib/src/rules/use_string_buffer.dart
```diff
@@ -0,0 +1,212 @@
...
+}
+
+bool _isEmptyInterpolationString(AstNode node) {
+  final visitor = new _IsEmptyInterpolationStringVisitor();
+  node.accept(visitor);
+  return visitor._isEmptyInterpolationString;
```
bwilkerson: Similarly, this feels like overkill. Consider:
```
  return node is InterpolationString && (node as InterpolationString).value == '';
```

test/rules/use_string_buffer.dart
```diff
@@ -0,0 +1,48 @@
...
+    buffer = buffer + 'a'; // LINT
+  }
+}
+
+void bar() {
+  String buffer = '';
+  while (buffer.length < 10) {
+    buffer += 'a'; // LINT
+  }
+}
```
bwilkerson: Should this produce a lint?
```
while (buffer.length < 10) {
  buffer = buffer + 'a';
}
```
Is there a reason we only look for statements of this form inside loops? What about
```
String performOperation() {
  String output = '';
  output = output + performStep1();
  output = output + performStep2();
  ...
  output = output + performStepN();
  return output;
}
```

lib/src/rules/use_string_buffer.dart
```diff
@@ -0,0 +1,212 @@
...
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/element.dart';
+import 'package:linter/src/analyzer.dart';
+import 'package:linter/src/util/dart_type_utilities.dart';
+
+const _desc = r'Use string buffer to compose strings.';
+
+const _details = r'''
+
+**DO** use string buffer to compose strings.
```
bwilkerson: I think this rule (rightly) only lints for uses related to local variables (ignores fields and top-level variables). If that's true, it should probably be documented somewhere.

lib/src/rules/use_string_buffer.dart
JPaulsen: node is InterpolationString makes an implicit cast even with && so it would be the same doing:

```dart
return node is InterpolationString && node.value == '';
```

test/rules/use_string_buffer.dart
JPaulsen: That could be made with interpolation:

return $performStep1()$performStep2()...$performStepN();

JPaulsen: dart```
while (buffer.length < 10) {
  buffer = buffer + 'a';
}
```

The test foo3() checks that, I think it should produce the lint, don't you?

File: lib/src/rules/use_string_buffer.dart (hunk 4 above)
Comment (JPaulsen) [inline lib/src/rules/use_string_buffer.dart]: It produces lints to fields too

File: lib/src/rules/use_string_buffer.dart [hunk 5]
```diff
@@ -0,0 +1,212 @@
...
+    buffer += 'a'; // LINT
+  }
+  return buffer;
+}
+```
+
+**GOOD:**
+```
+String foo() {
+	final buffer = new StringBuffer();
```
JPaulsen: Fixed

lib/src/rules/use_string_buffer.dart
JPaulsen: Fixed

lib/src/rules/use_string_buffer.dart
JPaulsen: Fixed

lib/src/rules/use_string_buffer.dart
JPaulsen: Tests added

test/rules/use_string_buffer.dart
JPaulsen: A better transformation for your example is:
```dart
String performOperation() {
  String output = '';
  output = '$output${performStep1()}';
  output = '$output${performStep2()}';
  ...
  output = '$output${performStepN()}';
  return output;
}
```

you can do it without a `StringBuffer` like this
```dart
String performOperation() {
  String output =  '${performStep1()}'
    '${performStep2()}'
  ...
    '${performStepN()}';
  return output;
}
```

bwilkerson: I missed foo3(), thanks!

The point of the other example isn't "what's the best way to code this" but "does or should this rule produce a lint". One of the things I look for is completeness of tests. I'm really asking whether this is a case that should be tested (in the original form, not in some translated form).

test/rules/use_string_buffer.dart
```diff
@@ -0,0 +1,59 @@
...
+class A {
+  String buffer;
+
+  void foo(int n) {
+    int aux = n;
+    while(aux --> 0) {
+      buffer += 'a'; // LINT
+    }
+  }
+}
```
bwilkerson: I don't think we can test fields the same way we do local variables because we can't (in the general case) reason about what might happen to them while computing the right-hand side. In the same class `A`, consider adding the following:
```
void bar(int count) {
  while (count-- > 0) {
    buffer += baz();
  }
}
String baz() {
  buffer = buffer + buffer;
  return buffer;
}
```
If we produce a lint in `bar`, that would be a false positive.

test/rules/use_string_buffer.dart
JPaulsen: Ok sorry, I added the tests.

test/rules/use_string_buffer.dart
JPaulsen: Yes it is linted, this rule in particular is not about what is easier to read, is about performance, in that way I think we should lint any += (or equivalents) inside a loop of a variable declared outside.

I know these cases are not easy to replace with a StringBuffer, but with the spirit of the rule (performance) they should be linted and detected and if the user is doing something like this and like the rule it should make an effort to improve performance with an StringBuffer or add a comment, that is my opinion.

What do you think is better?

test/rules/use_string_buffer.dart
```diff
@@ -0,0 +1,71 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+// test w/ `pub run test -N use_string_buffer`
+
+class A {
+  String buffer;
+
+  void foo(int n) {
+    int aux = n;
+    while(aux --> 0) {
```
bwilkerson: The formatting is strange here. I keep trying to read `-->` as an operator rather than reading the whole expression as `aux-- > 0`.

test/rules/use_string_buffer.dart
bwilkerson: I think it would be better to catch a subset of cases and have zero false positives than to catch all possible cases and have some false positives. Studies (from other languages) have shown that lint rules that produce false positives are disabled, even when comments or other techniques can be used to silence the rule in specific locations, unless the rule is perceived as bringing a very significant value (which is *very* rare).

And when I say "false positives" I don't mean technical false positives, I mean lints produced in places where the user doesn't want a lint to have been produced, however technically accurate the lint might be.

That said, this is ultimately a judgement call. You know my opinion, now it's up to you and Alexei to make the call.

lib/src/rules/use_string_buffer.dart
```diff
@@ -0,0 +1,185 @@
...
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/token.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/element.dart';
+import 'package:linter/src/analyzer.dart';
+import 'package:linter/src/util/dart_type_utilities.dart';
+
+const _desc = r'Use string buffer to compose strings.';
+
+const _details = r'''
```
bwilkerson: Given that this is a rule motivated by performance concerns, it would be good to state that in the details and to provide some concrete numbers (the bad case is N times slower than the good case).

test/rules/use_string_buffer.dart
JPaulsen: Fixed

lib/src/rules/use_string_buffer.dart
JPaulsen: Fixed

JPaulsen:
@bwilkerson @alexeieleusis what do you think about the documentation? It is the only thing that is being review to approve this rule

lib/src/rules/use_string_buffer.dart
```diff
@@ -0,0 +1,191 @@
...
+
+bool _isEmptyInterpolationString(AstNode node) =>
+    node is InterpolationString && node.value == '';
+
+/// The motivation of this rule is performance stuffs, and the explanation is
+/// that if we use N strings additions using the + operator the order of that
+/// algorithm is O(~N^2) and that is because a String is not mutable, so in each
+/// step it creates an auxiliary String that takes O(amount of chars) to be
+/// computed, in otherwise using a StringBuffer the order is reduced to O(~N)
+/// so the bad case is N times slower than the good case.
```
JPaulsen: @alexeieleusis @bwilkerson

PR #618, opened 2017-04-19
Fix rule: omit_local_variable_types

JPaulsen:
Fixes https://github.com/dart-lang/sdk/issues/29384

### DIFF
--- a/lib/src/rules/omit_local_variable_types.dart
+++ b/lib/src/rules/omit_local_variable_types.dart
@@ -62,17 +62,15 @@ class _Visitor extends SimpleAstVisitor {
 
   @override
   visitForEachStatement(ForEachStatement node) {
-    final staticType = node.loopVariable.type;
+    final staticType = node.loopVariable?.type;
     if (staticType == null) {
       return;
     }
-    if (node.iterable.bestType is! InterfaceType) {
-      return;
-    }
-    DartType iterableType = node.iterable.bestType;
+    final iterableType = node.iterable.bestType;
     if (iterableType is InterfaceType &&
-        iterableType.typeArguments.single == staticType.type) {
-      rule.reportLint(node);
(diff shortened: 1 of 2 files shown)


JPaulsen:
@bwilkerson I know this is the bug and this fixes it but I do not know how to reproduce it, I need a ForEachStatement that has an Iterable with an empty typeArgument

JPaulsen:
If I do this:
```dart
void foo() {
  final array = [];
  for (a in array) {
  }
}
```

the variable array is a variable with 1 typeArgument that is dynamic :S so I have not been able to understand what happened and how to reproduce the bug.

bwilkerson:
Perhaps something like this?
```
void printItems(Iterable items) {
  for (var item in items) {
    print(item);
  }
}
```

bwilkerson:
I agree, it is strange. I would have expected that the type would always have a type argument. Do we know that the type is "Iterable"? What if the type is not generic, something like:
```
void printItems(String items) {
  for (var item in items) {
    print(item);
  }
}
```
Perhaps then the type arguments is an empty list because `String` has no type parameters.

JPaulsen:
@bwilkerson in this case I got an error that says ERROR: The type 'String' used in the 'for' loop must implement Iterable. ([linter] test/rules/omit_local_variable_types.dart:14)

JPaulsen:
I also found other bug, if you reuse a variable in a foreach (it was already declared) then loopVariable is null, so I also added that and created a test

bwilkerson:
Maybe something like this?
```
abstract class StringIterator implements Iterator<String> {}
void printItems(StringIterator items) {
  for (var item in items) {
    print(item);
  }
}
```

JPaulsen:
It was that!!!, I fixed and could lint this case :D

PR #620, opened 2017-04-19
Fix rule: omit_type_for_local_variables

JPaulsen:
From https://github.com/dart-lang/linter/pull/618#issuecomment-295479210

### DIFF
--- a/lib/src/util/dart_type_utilities.dart
+++ b/lib/src/util/dart_type_utilities.dart
@@ -34,6 +34,27 @@ class DartTypeUtilities {
     return null;
   }
 
+  static Iterable<InterfaceType> getImplementedInterfaces(InterfaceType type) {
+    void recursiveCall(InterfaceType type, Set<ClassElement> alreadyVisited,
+        List<InterfaceType> interfaceTypes) {
+      if (type == null || !alreadyVisited.add(type.element)) {
+        return;
+      }
+      interfaceTypes.add(type);
+      recursiveCall(type.superclass, alreadyVisited, interfaceTypes);
+      for (final interface in type.interfaces) {
+        recursiveCall(interface, alreadyVisited, interfaceTypes);
+      }
+      for (final mixin in type.mixins) {
+        recursiveCall(mixin, alreadyVisited, interfaceTypes);
+      }
+    }
+
+    final interfaceTypes = <InterfaceType>[];
+    recursiveCall(type, new Set<ClassElement>(), interfaceTypes);
+    return interfaceTypes;
+  }
+
   static Statement getLastStatementInBlock(Block node) {
     if (node.statements.isEmpty) {
       return null;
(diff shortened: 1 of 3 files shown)


lib/src/util/dart_type_utilities.dart
```diff
@@ -47,13 +47,33 @@ class DartTypeUtilities {
...
+    if (type is! InterfaceType) {
+      return null;
+    }
+    bool predicate(InterfaceType i) => isInterface(i, interface, library);
+
+    ClassElement element = type.element;
+    if (predicate(type)) {
+      return type;
+    }
+    if (!element.isSynthetic && type is InterfaceType) {
```
bwilkerson: I don't believe we need the test `type is InterfaceType` because of line 52.

lib/src/util/dart_type_utilities.dart
```diff
@@ -47,13 +47,33 @@ class DartTypeUtilities {
...
+      final interfaces = element.allSupertypes.where(predicate);
```
bwilkerson: This works in the case where the class specifies a non-type parameter as the type of the interface, but I'm concerned that we loose type information if we have something like
```
abstract class SpecialIterator<E> implements Iterator<E> {}
void printItems(SpecialIterator<String> items) ...
```
Asking for the type of `items` will get you `SpecialIterator<String>`, but the element associated with that is `SpecialIterator<E>`. Its supertypes should then include `Iterator<E>`, but `E` won't match `String`.

Unfortunately, there is no `allSupertypes` on `InterfaceType`. Maybe there should be...

lib/src/util/dart_type_utilities.dart

lib/src/util/dart_type_utilities.dart

@bwilkerson I resolved it doing a DFS with the supertypes and propagating the typeArguments with their respective typeParameter to the super classes and implemented interfaces then I took the value of the class Iterable with their propagated types to compare and lint the node, PTAL.

bwilkerson: You've definitely got the right approach. Unfortunately, this duplicates a lot of existing code. In the analyzer's model, an `InterfaceType` already represents the same thing I think you're trying to do with `_ClassElementBox`: the mapping from type parameters to type arguments. So, for example, when you have a type representing `List<int>`, what you really have is an object that knows that the type is defined by the element for `List<E>` (the element keeps the type parameters) and whose type arguments are the type for `int`.

If you ask the element for `List<E>` for its interfaces, it can tell you that it implements `Iterable<E>` (where this `E` is the type parameter associated with `List`). But if you ask the type `List<int>` for it's interfaces it can propagate the substitution and tell you that it implements `Iterable<int>`.

Given that this is all implemented and well tested (and will be maintained) it seems wrong to re-implement it here.

JPaulsen: You are totally right! the problem was that I was using interfaceType.element.allSupertypes instead interfaceType.superclass and interfaceType.interfaces so I lost the propagated argumentList but I can rescue now, thank you!

JPaulsen: Fixed :)

lib/src/util/dart_type_utilities.dart
```diff
@@ -33,6 +33,24 @@ class DartTypeUtilities {
...
   }
 
+  static Iterable<InterfaceType> getImplementedInterfaces(InterfaceType type) {
+    void recursiveCall(InterfaceType type, Set<ClassElement> alreadyVisited,
+        List<InterfaceType> interfaceTypes) {
+      if (type == null || !alreadyVisited.add(type.element)) {
+        return;
+      }
+      interfaceTypes.add(type);
+      recursiveCall(type.superclass, alreadyVisited, interfaceTypes);
```
bwilkerson: This fails to add any `mixins`.

PR #621, opened 2017-04-20
Delete integration test: prefer_collection_literals

JPaulsen:


JPaulsen:
@bwilkerson here I just moved the test to test/rules and added the LinkedHashMap to mock_sdk.dart just like in the SDK project so it will be easier to change the test in the future.

PR #630, opened 2017-04-25
handle asserts in initializer list

a14n:


a14n:
Some changes to `LinterOptions` (in the analzer package) seem to be needed to enable _assert-initializer_ flag and make CI happy.

/cc @bwilkerson

.analysis_options
```diff
@@ -1,4 +1,6 @@
 analyzer:
+  language:
+    enableAssertInitializer: true
```
bwilkerson: Is this necessary? I don't think we want to enable experimental language features that we might later have to remove. Can we enable the feature just for the test? (I'm assuming that's why this is here.)

lib/src/rules/always_require_non_null_named_parameters.dart
```diff
@@ -88,16 +88,34 @@ class Visitor extends SimpleAstVisitor {
...
     }
   }
 
-  _checkParams(List<DefaultFormalParameter> params, FunctionBody body) {
+  void _checkInitializerList(List<DefaultFormalParameter> params,
+      NodeList<ConstructorInitializer> initializers) {
+    final asserts = initializers
+        .where((i) => i is AssertInitializer)
+        .map((e) => e as AssertInitializer)
+        .map((e) => e.condition)
```
bwilkerson: It probably doesn't matter for lists this short, so no action required, but there is a performance difference between two map operations and one:
```
.map((e) => (e as AssertInitializer).condition)
```
For a list with 2000 items, and using one `where`, one or two `map`s and one `toList`, I'm measuring 7 ms for two and 2 ms for one.

lib/src/rules/always_require_non_null_named_parameters.dart
```diff
@@ -106,18 +124,23 @@ class Visitor extends SimpleAstVisitor {
...
 
-    final expression = node.condition.unParenthesized;
-    if (expression is BinaryExpression &&
-        expression.operator.type == TokenType.BANG_EQ) {
-      final operands = [expression.leftOperand, expression.rightOperand];
-      return operands.any(DartTypeUtilities.isNullLiteral) &&
-          operands.any(_hasSameName);
+    final expression = node.unParenthesized;
+    if (expression is BinaryExpression) {
+      if (expression.operator.type == TokenType.AMPERSAND_AMPERSAND) {
```
bwilkerson: Should we also include `||`?

.analysis_options
a14n: That's indeed only for the test. I don't know how to enable this experimental feature only for the test and I think some changes in the analyzer package are needed to enable this experimental feature.

As it's an experimental feature this feature (check asserts in initializer list) could perhaps lives without test until its support improve (dartfmt, `LinterOptions` in analyzer...) wdyt ?

lib/src/rules/always_require_non_null_named_parameters.dart
a14n: Will do

lib/src/rules/always_require_non_null_named_parameters.dart
a14n: No, because `assert(param1 != null || param2 != null)` doesn't mean that both `param1` and `param2` have to be required.

.analysis_options
bwilkerson: @pq I'm hoping you know how to accomplish this.

lib/src/rules/always_require_non_null_named_parameters.dart
bwilkerson: Makes sense.

.analysis_options
pq: Hmmmm.  Can we move this to an integration test?  Look in `test/_data` at the various projects.  Maybe we can define analysis options for a single project there?  Alternatively, we could analyze a single directory using `--assert-initializer` from `integration_test.dart`?

a14n:
@pq updated with an integration test but the problem remains. The flag to enable assert initializer is not seen and the integration test fails with an error about assert not supported at this place.

test/integration_test.dart
```diff
@@ -418,6 +418,34 @@ defineTests() {
...
+      tearDown(() {
+        collectingOut.buffer.clear();
+        outSink = currentOut;
+        exitCode = 0;
+      });
+
+      test('only throw errors', () {
+        dartlint.main([
+          'test/_data/always_require_non_null_named_parameters',
+          '--rules=always_require_non_null_named_parameters'
```
pq: Can we pass the flag in here?  If it doesn't already, can we make linter main pass arguments down to the analyzer?

pq: Alternatively, maybe we should add support to linter main to allow you to specify analysis options (like you can with analyzer).  If that sounds good, I can help with that.

a14n: > Can we pass the flag in here?

It fails with the message:

>Could not find an option named "assert-initializer".

Your last proposal (_add support to linter main to allow you to specify analysis options_) seems to be the only solution. But it's a change that need to be done in the analyzer package not here.

a14n:
rebased and set enableAssertInitializer but still failing. I don't understand why :-/

bwilkerson:
I'm not sure either (and was hoping @pq would know). Are you running against the bleeding edge branch, or are you running against the published 0.29.11 version of 'analyzer' (or the 0.29.10 version, if you didn't add a dependency override and didn't `pub upgrade`)?

a14n:
Setting the `enableAssertInitializer` to true makes the file parsing not failling. However there seems to be a bug in the structure of elements. `parent.initializers` is `[]` for a constructor with 2 asserts in the initializer list. Is it a bug on the analysis server side ?

a14n:
```
$ dart --version
Dart VM version: 1.23.0-dev.11.11 (Wed Apr 19 01:20:09 2017) on "linux_x64"
```

bwilkerson:
> However there seems to be a bug in the structure of elements. `parent.initializers` is `[]` for a constructor with 2 asserts in the initializer list. Is it a bug on the analysis server side ?

I'm assuming that `parent` is a ConstructorDeclaration (which in "analyzer" terms is an AstNode, not an Element :-).

> $ dart --version
> Dart VM version: 1.23.0-dev.11.11 (Wed Apr 19 01:20:09 2017) on "linux_x64"

Actually, I was asking about the version of 'analyzer' being used when running the tests, not the version of the VM.

I think I've figured out the problem. In the 0.29.x versions of 'analyzer', the assert initializers are parsed but discarded. (If I recall correctly, adding them would have required a backward compatibility breakage.) In the bleeding edge version those initializers are added to the list of initializers associated with the constructor. I think this change will need to wait until we've published a 0.30.x version of analyzer in which they are treated correctly.

a14n:
Thanks for your help. I will wait for analyzer 0.30.x

pq:
> Thanks for your help. I will wait for analyzer 0.30.x

👍 

That's being tracked in https://github.com/dart-lang/sdk/issues/29516.

FYI @danrubel @devoncarew

a14n:
rebased with master (and analyzer-0.30.x) : CI fixed \o/

a14n:
This PR is ready to land. Could someone do it?

a14n:
@alexeieleusis thanks :)

PR #635, opened 2017-04-26
New rule: prefer_foreach

JPaulsen:
Fixes https://github.com/dart-lang/linter/issues/634

### DIFF
--- a/lib/src/rules/prefer_foreach.dart
+++ b/lib/src/rules/prefer_foreach.dart
@@ -1,0 +1,129 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+library linter.src.rules.prefer_foreach_call;
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/element.dart';
+import 'package:linter/src/analyzer.dart';
+import 'package:linter/src/util/dart_type_utilities.dart';
+
+const _desc = r'Use forEach call to only apply a function to all the elements.';
+
+const _details = r'''
+
+**DO** use forEach call if you are only going to apply a function or a method to all the elements of an iterable.
+
+**BAD:**
+```
+for (final key in map.keys.toList()) {
+  map.remove(key);
+}
+```
+
+**GOOD:**
+```
+map.keys.toList().forEach(map.remove);
+```
+
+**NOTE:** Replacing a for each statement with a forEach call may change the behavior in the case there are side-effects on the iterable itself.
+```
+for (final v in myList) {
+  foo().f(v); // This code invokes foo() many times.
+}
+
+myList.forEach(foo().f); // But this one invokes foo() just once.
+```
+
+''';
+
+class PreferForeach extends LintRule {
+  _Visitor _visitor;
+  PreferForeach()
+      : super(
+            name: 'prefer_foreach',
+            description: _desc,
+            details: _details,
+            group: Group.style,
+            maturity: Maturity.experimental) {
+    _visitor = new _Visitor(this);
+  }
+
+  @override
+  AstVisitor getVisitor() => _visitor;
+}
+
+class _PreferForEachVisitor extends SimpleAstVisitor {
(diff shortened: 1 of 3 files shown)


JPaulsen:
24 lints in the sdk, no false positives (Except the implementation of the forEach method itself), some examples:

```dart
for (var e in args) {
  list.add(e);
}
// Could be:
args.forEach(list.add);
```

```dart
for (var arg in args) {
  safeForTypedInterop(arg);
}
// Could be:
args.forEach(safeForTypedInterop);
```

```dart
for (E each in objects) {
  add(each);
}
// Could be:
objects.forEach(add);
```

lib/src/rules/prefer_foreach_call.dart
```diff
@@ -0,0 +1,119 @@
...
+map.keys.toList().forEach(map.remove);
+```
+
+''';
+
+class PreferForeachCall extends LintRule {
+  _Visitor _visitor;
+  PreferForeachCall()
+      : super(
+            name: 'prefer_foreach_call',
```
bwilkerson: I would drop the `_call`.

JPaulsen:
@bwilkerson yes, you are totally right, I thought code like that would case errors so even I thought that the rule could have a little improvement in performance because of computing the Iterable just once.

So what do we do? What you said is totally right, I do not even know if we could computate if the change is safe so I think the options are: 

1.- Allow the rule as an experimental one.
2.- Discard it.

alexeieleusis:
I like the rule, marking it as experimental would not fail analysis in CI and we can see how the community likes it, probably add a comment to the doc so besides being it marked as experimental make very explicit in docs that code relying on side effects should not perform this change.

lib/src/rules/prefer_foreach_call.dart
JPaulsen: Fixed

JPaulsen:
@alexeieleusis done

bwilkerson:
> So what do we do?

I suspect you already know my answer :-) I have a very low tolerance for lints with false positives, so I would lean toward dropping the rule. I won't stop you if you want to commit is as experimental, though.

JPaulsen:
Thanks @bwilkerson

lib/src/rules/prefer_foreach.dart
```diff
@@ -0,0 +1,123 @@
...
+for (final key in map.keys.toList()) {
+  map.remove(key);
+}
+```
+
+**GOOD:**
+```
+map.keys.toList().forEach(map.remove);
+```
+
```
alexeieleusis: Add a comment here saying that the rule is experimental and that code that the suggested change alters logic if users rely on side effects over the iterable, provide an example.

JPaulsen: Fixed

JPaulsen:
@bwilkerson I was trying to figure out an example to show in the documentation but it looks like they iterable in the for each statement is being calculated just once so in that case is exactly the same behavior, I share this code to you https://dartpad.dartlang.org/c1c06ea89b842d6c822935f46bfe10a8

bwilkerson:
By 'target' I meant the target of the method being invoked in the loop, not the iterator computed for the loop. Here's a comparison of the results of two methods, one named `process` that I think would produce a lint, and one named `processTransformed` that I think is what this lint would recommend transforming the code into:
```
main() {
  C c = new C();
  c.init();
  c.process(['a', 'b', 'c']);
  print(c.lists);

  c = new C();
  c.init();
  c.processTransformed(['a', 'b', 'c']);
  print(c.lists);
}

class C {
  var index = 0;
  var lists;

  next() => lists[index++];

  void init() {
    lists = [
    	[],
    	[],
    	[],
  	];
  }

  void process(list) {
    for (var element in list) {
      next().add(element);
    }
  }

  void processTransformed(list) {
    list.forEach(next().add);
  }
}
```
Note that the method `next` is invoked exactly once in the transformed code while it is invoked multiple times in the loop-form.

If the lint won't be produced for the loop in `perform` because the target can't be proven to not have any side-effects, then I'm happy with the rule as is. But if the lint *will* be produced, then there's a chance for false positives.

JPaulsen:
@alexeieleusis comment fixed and added

alexeieleusis:
We can avoid those false positives by linting only on local variables or fields, or said differently not linting method invocations and getters.

I will do that change later.

PR #643, opened 2017-05-03
Turns out this attribute is still needed...

pylaligand:


pylaligand:
@zanderso, sorry for the flip-flopping...

Issue #648, opened 2017-05-08
Review `0.1.31` changes.


pq:
I fat-fingered and pushed a commit (https://github.com/dart-lang/linter/commit/609ff7686c9d942b69922e36a4373fe5b2ef7643) that was meant to be a PR.

Please take a quick look.  Happy to revert if needed.

Note there is some rule name normalization.

@bwilkerson 

FYI @a14n @alexeieleusis

pq:
FYI @zoechi

pq:
@bwilkerson: here's the diff

https://github.com/dart-lang/linter/commit/609ff7686c9d942b69922e36a4373fe5b2ef7643

bwilkerson:
A couple of minor comments, but otherwise lgtm.

lib/src/rules.dart

You changed `do_not_create_setter_without_getter` to `avoid_setters_without_getters`. Should we similarly rename `do_not_catch_errors` and `do_not_check_for_null_in_equality_operators`?

We generally try to keep the list of registrations in alphabetic order.

lib/src/rules/avoid_returning_null.dart

Personally, I think we should remove all `library` directives, but it's your call.

pq:
Thanks!

> You changed do_not_create_setter_without_getter to avoid_setters_without_getters. Should we similarly rename do_not_catch_errors and do_not_check_for_null_in_equality_operators?

> We generally try to keep the list of registrations in alphabetic order.

PR on the way.

> Personally, I think we should remove all library directives, but it's your call.

Agreed.  I made a note to myself to remove them all (and turn on that lint) after this lands. 👍

PR #650, opened 2017-05-08
0.1.31 cleanup.

pq:
Follow-up from https://github.com/dart-lang/linter/issues/648#issuecomment-299886186:

cc @bwilkerson


lib/src/rules.dart
```diff
@@ -33,7 +34,6 @@ import 'package:linter/src/rules/comment_references.dart';
 import 'package:linter/src/rules/constant_identifier_names.dart';
 import 'package:linter/src/rules/control_flow_in_finally.dart';
 import 'package:linter/src/rules/directives_ordering.dart';
-import 'package:linter/src/rules/do_not_catch_errors.dart';
 import 'package:linter/src/rules/do_not_check_for_null_in_equality_operators.dart';
```
bwilkerson: Rename the file too?

pq: Quite right.  Thanks!

PR #653, opened 2017-05-08
Example config update and doc tweaks.

pq:
@bwilkerson @filiph


pq:
Thanks!

alexeieleusis:
This is a bit late but, should we also add a config file for the Effective Dart rules? Similarly we have enough performance/correctness rules that might also make it worth having them in its own file/section.

alexeieleusis:
I can do it later since definitely is not an urgent thing.

pq:
@alexeieleusis 👍  yes please, do add an example file that aggregates the effective dart rules.  Thanks!

pq:
(Happy to push out a quick `0.1.32` when you do as well.)

Issue #658, opened 2017-05-09
incorrect "unconditional condition" using ??


MichaelRFairhurst:
I have

```
        if ((view?.templateText ?? '') != '') {
```

and it thinks this is always true...or false?

zoechi:
Is this the same as #576 ?

PR #667, opened 2017-05-13
literal_only_boolean_expressions: Only check literals on left side for ??.ˆ

alexeieleusis:
Fixes https://github.com/dart-lang/linter/issues/658


a14n:
Fixes also #454

alexeieleusis:
Thank you!

Issue #669, opened 2017-05-16
consider benchmarking the lints


devoncarew:
We've had issues where individual lints can have outsized performance impact on the time for analysis. We should set up a benchmarking system, so we can get a rough sense of which lints are performant, slow, or have severe perf issues.

@pq 

I have worked on some similar systems in the past - we may be able to share some code or concepts here.

bwilkerson:
The linter has a --stats flag that will provide timing information for each of the lints run over whatever code base is being linted. That should be a reasonable start for automating something.

pq:
Sounds great.

As for the stats flag, Brian is right it gives us a nice jumping off point.

For context, here are the salient details running all the lints against the linter codebase:

```
---------------------------------------------------
Timings                                          ms
---------------------------------------------------
always_declare_return_types                      18
always_put_control_body_on_new_line              15
always_require_non_null_named_parameters         24
always_specify_types                             19
annotate_overrides                               31
avoid_annotating_with_dynamic                     8
avoid_as                                          8
avoid_catches_without_on_clauses                  8
avoid_catching_errors                             8
avoid_classes_with_only_static_members           10
avoid_empty_else                                  8
avoid_function_literals_in_foreach_calls          9
avoid_init_to_null                                9
avoid_null_checks_in_equality_operators          10
avoid_positional_boolean_parameters              25
avoid_return_types_on_setters                     8
avoid_returning_null                             25
avoid_returning_this                             10
avoid_setters_without_getters                    11
avoid_slow_async_io                              11
avoid_types_on_closure_parameters                10
await_only_futures                                7
camel_case_types                                 12
cancel_subscriptions                             24
cascade_invocations                              18
close_sinks                                      34
comment_references                                8
constant_identifier_names                        10
control_flow_in_finally                          10
directives_ordering                              26
empty_catches                                     7
empty_constructor_bodies                          7
empty_statements                                  6
hash_and_equals                                   8
implementation_imports                            8
invariant_booleans                               88
iterable_contains_unrelated_type                 11
join_return_with_assignment                      11
library_names                                     9
library_prefixes                                  8
list_remove_unrelated_type                        8
literal_only_boolean_expressions                  9
no_adjacent_strings_in_list                       7
no_duplicate_case_values                         11
non_constant_identifier_names                    19
omit_local_variable_types                        12
one_member_abstracts                              8
only_throw_errors                                 8
overridden_fields                                14
package_api_docs                                 27
package_prefixed_library_names                    8
[...]

PR #672, opened 2017-05-19
Add `--benchmark` support.

pq:
New benchmark mode that prints lint stats based on the best of 10 runs.

Follow-up from: https://github.com/dart-lang/linter/issues/669

@devoncarew @bwilkerson

### DIFF
--- a/bin/linter.dart
+++ b/bin/linter.dart
@@ -12,6 +12,7 @@ import 'package:analyzer/src/lint/io.dart';
 import 'package:analyzer/src/lint/linter.dart';
 import 'package:analyzer/src/lint/registry.dart';
 import 'package:args/args.dart';
+import 'package:linter/src/analyzer.dart';
 import 'package:linter/src/formatter.dart';
 import 'package:linter/src/rules.dart';
 
@@ -54,6 +55,8 @@ void runLinter(List<String> args, LinterOptions initialLintOptions) {
         abbr: "h", negatable: false, help: "Show usage information.")
     ..addFlag("stats",
         abbr: "s", negatable: false, help: "Show lint statistics.")
+    ..addFlag("benchmark",
+        abbr: "b", negatable: false, help: "Show lint benchmarks.")
     ..addFlag('visit-transitive-closure',
         help: 'Visit the transitive closure of imported/exported libraries.')
     ..addFlag('quiet', abbr: 'q', help: "Don't show individual lint errors.")
@@ -141,7 +144,8 @@ void runLinter(List<String> args, LinterOptions initialLintOptions) {
   }
 
   var stats = options['stats'];
-  if (stats == true) {
+  var benchmark = options['benchmark'];
+  if (stats == true || benchmark == true) {
     lintOptions.enableTiming = true;
   }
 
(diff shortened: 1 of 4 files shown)


pq:
FYI @alexeieleusis : even with the best of, it looks like `invariant_booleans` is running a bit slow... ?

alexeieleusis:
What time is it taking?

pq:
144 ms (from [here](https://travis-ci.org/dart-lang/linter/jobs/234151797))

alexeieleusis:
I'll look into it once this is merged, at least we have a baseline now.

bin/linter.dart
```diff
@@ -54,6 +55,8 @@ void runLinter(List<String> args, LinterOptions initialLintOptions) {
         abbr: "h", negatable: false, help: "Show usage information.")
     ..addFlag("stats",
         abbr: "s", negatable: false, help: "Show lint statistics.")
+    ..addFlag("benchmark",
```
devoncarew: may not need a shorthand for this flag, given how rarely it'll be used?

bin/linter.dart
```diff
@@ -141,21 +144,27 @@ void runLinter(List<String> args, LinterOptions initialLintOptions) {
   }
 
   var stats = options['stats'];
-  if (stats == true) {
+  var benchmark = options['benchmark'];
+  if (stats == true || benchmark == true) {
```
devoncarew: perhaps `(stats || benchmark)`?

lib/src/formatter.dart
```diff
@@ -291,6 +266,57 @@ class SimpleFormatter implements ReportFormatter {
   }
 }
 
+writeBenchmarks(IOSink out, List<File> filesToLint, LinterOptions lintOptions) {
```
devoncarew: so few return types :(

tool/travis.sh
```diff
@@ -20,8 +20,8 @@ if [ "$LINTER_BOT" = "benchmark" ]; then
   # The actual lints can have errors - we don't want to fail the benchmark bot.
   set +e
 
-  # Run linter with all lints enabled.
-  dart bin/linter.dart -s -q -c example/all.yaml .
+  # Benchmark linter with all lints enabled.
+  dart bin/linter.dart -b -q -c example/all.yaml .
```
devoncarew: `--benchmark` if you remove the shorthand

bin/linter.dart
pq: Sounds good.

bin/linter.dart
pq: Yes!  For some reason I thought I convinced myself that undefined would return `null` but yeah, much better. 👍

lib/src/formatter.dart
pq: Yep.  Consider this bit of code my last stand against gratuitous returns (in non API bits). 😸

tool/travis.sh
pq: Done.

pq:
@devoncarew : take another gander?

devoncarew:
lgtm!

pq:
Thanks!

PR #673, opened 2017-05-19
Add a lint rule for preferring single quote strings

MichaelRFairhurst:
Do not suggest double quotes for
- strings with ' in them naturally
- strings containing other strings ie "foo ${'bar'}"
- strings within other strings (we check 'bar' in the case above!)

Also be careful for the difference between interpolated and non
interpolated strings.

cc @mk13

### DIFF
--- a/lib/src/rules/prefer_single_quotes.dart
+++ b/lib/src/rules/prefer_single_quotes.dart
@@ -17,11 +17,9 @@ apostrophe isn't escaped (note: we don't lint the other way around, ie, a single
 quoted string with an escaped apostrophe is not flagged).
 
 Its also rare, but possible, to have strings within string interpolations. In
-this case, its more or less necessary to use a double quote somewhere (unless
-you want to do single quotes within triple single quotes, which most people
-don't, and then you still have an isue with strings within strings within
-strings). So double quotes are allowed either within, or containing, an
-interpolated string literal.
+this case, its much more readable to use a double quote somewhere. So double
+quotes are allowed either within, or containing, an interpolated string literal.
+Arguably strings within string interpolations should be its own type of lint.
 
 **BAD:**
 ```
--- a/lib/src/rules/prefer_single_quotes.dart
+++ b/lib/src/rules/prefer_single_quotes.dart
@@ -6,28 +6,40 @@ import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
 import 'package:linter/src/analyzer.dart';
 
-const _desc = "Prefer single quotes where it won't require escape sequences";
+const _desc = "Prefer single quotes where they won't require escape sequences";
 
 const _details = '''
 
-**DO** use single quotes where it wouldn't require additional escapes
+**DO** use single quotes where they wouldn't require additional escapes.
+
+That means strings with an apostrophe may use double quotes so that the
+apostrophe isn't escaped (note: we don't lint the other way around, ie, a single
+quoted string with an escaped apostrophe is not flagged).
+
+Its also rare, but possible, to have strings within string interpolations. In
+this case, its more or less necessary to use a double quote somewhere (unless
+you want to do single quotes within triple single quotes, which most people
+don't, and then you still have an isue with strings within strings within
+strings). So double quotes are allowed either within, or containing, an
+interpolated string literal.
 
 **BAD:**
 ```
 useStrings(
-    "should be single",
-    r"should be single",
-    r"""should be single""",
-    "here's ok");
+    "should be single quote",
+    r"should be single quote",
+    r"""should be single quotes""")
 ```
 
 **GOOD:**
 ```
 useStrings(
-    'should be single',
-    r'should be single",
-    r\'''should be single\''',
-    "here's ok");
+    'should be single quote',
+    r'should be single quote",
+    r\'''should be single quotes\''',
+    "here's ok",
+    "nested \${a ? 'strings' : 'can'} be wrapped by a double quote",
+    'and nested \${a ? "strings" : "can be double quoted themselves"});
 ```
 
 ''';
@@ -57,7 +69,7 @@ class _Visitor extends SimpleAstVisitor {
       return;
     }
 
-    // Bail out on "strings ${x ? 'containing' : 'other'} strings"
+    // Bail out on 'strings ${x ? "containing" : "other"} strings'
     if (!isNestedString(string)) {
       rule.reportLintForToken(string.literal);
     }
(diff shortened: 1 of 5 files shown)


bwilkerson:
> strings containing other strings ie `"foo ${'bar'}"`

Given that the inner string is inside a string interpolation, there wouldn't be any lexical issue with `'foo ${'bar'}'`, though it might admittedly be harder to read.

> strings within other strings (we check 'bar' in the case above!)

Again, not a problem from a lexical perspective.

lib/src/rules/prefer_single_quotes.dart
```diff
@@ -0,0 +1,140 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:linter/src/analyzer.dart';
+
+const _desc = "Prefer single quotes where it won't require escape sequences";
```
bwilkerson: "it" --> "they"?

lib/src/rules/prefer_single_quotes.dart
```diff
@@ -0,0 +1,140 @@
...
+
+const _details = '''
+
+**DO** use single quotes where it wouldn't require additional escapes
```
bwilkerson: "where it" --> "for string literals where they"?

Also, if we're allowing the use of double quotes in other places we should describe the allowed uses.

lib/src/rules/prefer_single_quotes.dart
```diff
@@ -0,0 +1,140 @@
...
+class _WithinStringVisitor extends UnifyingAstVisitor {
+  bool withinString = false;
+
+  @override
+  visitNode(AstNode n) {
+    n.parent?.accept(this);
+  }
+
+  @override
+  visitSimpleStringLiteral(SimpleStringLiteral string) {
```
bwilkerson: A simple string literal can never be the parent of another AST node, so this method is unnecessary. Which means this visitor is just looking to see whether any parent of a node is a `StringInterpolation`, so consider using `getAncestor` instead (it should be more performant).

lib/src/rules/prefer_single_quotes.dart
MichaelRFairhurst: sure, sg

lib/src/rules/prefer_single_quotes.dart
MichaelRFairhurst: Sure thing

lib/src/rules/prefer_single_quotes.dart
MichaelRFairhurst: ah that's a great point. Will fix!

MichaelRFairhurst:
Interesting that nested single strings is allowed. I saw it in the linter codebase itself, and tinkering around came to the conclusion that that was required. But you're right, its technically allowed. Just after my commit clarifying comments, look ike I should clarify my comments a bit more.

lib/src/rules/prefer_single_quotes.dart
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    // Bail out on "strings ${x ? 'containing' : 'other'} strings"
+    if (!containsString(string) && !isNestedString(string)) {
+      rule.reportLint(string);
+    }
+  }
+
+  /// Strings can be within interpolations (ie, nested). Check like this.
+  bool isNestedString(AstNode node) {
```
bwilkerson: I should have been more explicit. The body of this method could be replaced by
```
return node.getAncestor((parent) => parent is StringInterpolation) != null;
```
which would be more performant that dispatching to a visitor.

lib/src/rules/prefer_single_quotes.dart
```diff
@@ -0,0 +1,145 @@
...
+    node.parent?.accept(checkWithinString);
+
+    return checkWithinString.withinString;
+  }
+
+  /// Strings interpolations can contain other string nodes. Check like this.
+  bool containsString(StringInterpolation string) {
+    final checkHasString = new _HasStringVisitor();
+    for (final child in string.elements) {
+      child.accept(checkHasString);
```
bwilkerson: It probably won't matter much for performance, but you could short-circuit this loop by checking the value of `checkHasString.hasString` after each element rather than waiting to check until we've visited all of the elements.

MichaelRFairhurst: I considered this, ultimately didn't to reduce clutter. But if everyone sees that upon reading it, that's not a very good attempt at "clean." Will short-circuit it.

lib/src/rules/prefer_single_quotes.dart
MichaelRFairhurst: Ah, yes! Lost that detail in the text.

zoechi:
How does this rule treat

```dart
var x = "don't";
```
AFAIR according to the style guide this is preferred over
```dart
var x = 'don\'t';
```

MichaelRFairhurst:
that is not linted (in either direction -- it doesn't suggest doubles to avoid escapes)

zoechi:
@MichaelRFairhurst just saw it in your initial comment. 
I looked twice but seems I wasn't able to interpret the first rule properly :D
Sorry for the noise.

lib/src/rules/prefer_single_quotes.dart
```diff
@@ -0,0 +1,148 @@
...
+  /// Strings can be within interpolations (ie, nested). Check like this.
+  bool isNestedString(AstNode node) {
+    // careful: node.getAncestor will check the node itself.
+    return node.parent?.getAncestor((p) => p is StringInterpolation) != null;
+  }
+
+  /// Strings interpolations can contain other string nodes. Check like this.
+  bool containsString(StringInterpolation string) {
+    final checkHasString = new _HasStringVisitor();
+    return string.elements.any((child) => child.accept(checkHasString));
```
bwilkerson: This seems really complex, so I might be confused, but...

I think all you're doing here is checking whether `string` has any immediate children that are either a `SimpleStringLiteral` or a `StringInterpolation`. But all of the elements are `InterpolationElement`s, and therefore will never be either a `SimpleStringLiteral` or a `StringInterpolation`. In other words, I think you can replace the body of this method with `return false;`

MichaelRFairhurst: Yeah, it is fairly complex, perhaps too complex.

The visitor, however, is a recursive visitor. For each visited node, it checks if any of the visited nodes is a string, recursively.

Aside from the `GetImmediateChildrenVisitor`, which I'd say more likely belongs in the sdk itself, all its doing is

```
  bool visitNode(AstNode node) => node.children.any((child) => child.accept(this));
```

which is clearer as

```
  bool containsString(AstNode child) => child.accept(this);
  bool visitNode(AstNode node) => node.children.any(containsString);
```

which I considered as well, but ultimately thought the reaction to such code would be "too many methods doing too little."

bwilkerson: FWIW, I don't object to small methods.

MichaelRFairhurst: Alrighty, the newest commit I think is very very nice. Should not have worried about methods being too small, I think the result is really neat/clean. Hopefully lgty

PR #677, opened 2017-05-22
add lint prefer_asserts_in_initializer_list

a14n:


lib/src/rules/prefer_asserts_in_initializer_list.dart
```diff
@@ -0,0 +1,61 @@
...
+
+  Visitor(this.rule);
+
+  @override
+  visitConstructorDeclaration(ConstructorDeclaration node) {
+    final body = node.body;
+    if (body is BlockFunctionBody &&
+        body.block.statements.isNotEmpty &&
+        body.block.statements.every((s) => s is AssertStatement)) {
+      rule.reportLintForToken(body.beginToken);
```
bwilkerson: I think this fails to take into account the restriction that asserts in the initializer list of `const` constructors can only have constant expressions. We shouldn't create a lint if moving the assert isn't valid.

a14n: I don't understand your comment because const constructors can't have a body. So this lint will never apply on a const constructor.

Am I missing something ?

bwilkerson: No, I was. I missed the fact that this case can't arise.

test/_data/prefer_asserts_in_initializer_list/lib.dart
```diff
@@ -0,0 +1,7 @@
+class A {
+  A.c1(a) : assert(a != null); // OK
+  A.c2(a) { // LINT
+    assert(a != null);
+  }
+  A.c3(a) {} // OK
+}
```
bwilkerson: It would be good to have a test showing that a body with non-assert statements also isn't linted.

a14n: added

bwilkerson: So why is line 7 marked with `// LINT`?

a14n: My bad. Actually those comments are not important in integration tests, only the final result is tested https://github.com/dart-lang/linter/pull/677/files#diff-64cac337b62c744b4fa7ff04a8067aa0R444

PR #678, opened 2017-05-22
add lint always_put_required_named_parameters_first

### LINKED ISSUE #676: consider a lint to ensure that @required params are declared before non-required named params
So

```
void foo({
  @required int bar,
  int baz
}) { ...
```

would be fine, but

```
void foo({
  int baz,
  @required int bar
}) { ...
```

would trigger.

a14n:
Fixes #676

### DIFF
--- a/lib/src/rules/always_put_required_named_parameters_first.dart
+++ b/lib/src/rules/always_put_required_named_parameters_first.dart
@@ -1,0 +1,71 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/analyzer.dart';
+import 'package:analyzer/dart/ast/ast.dart' show AstVisitor, TypedLiteral;
+import 'package:analyzer/dart/ast/token.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/element.dart';
+import 'package:linter/src/analyzer.dart';
+import 'package:linter/src/util/dart_type_utilities.dart';
+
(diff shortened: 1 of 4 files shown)


a14n:
Fixes #676

lib/src/rules/always_put_required_named_parameters_first.dart
```diff
@@ -0,0 +1,71 @@
...
+  Visitor(this.rule);
+
+  @override
+  visitFormalParameterList(FormalParameterList node) {
+    bool nonRequiredSeen = false;
+    for (DefaultFormalParameter param
+        in node.parameters.where((p) => p.kind == ParameterKind.NAMED)) {
+      if (param.metadata.any((a) => _isRequired(a.element))) {
+        if (nonRequiredSeen) {
+          rule.reportLintForToken(param.identifier.beginToken);
```
bwilkerson: `beginToken` --> `token` (in case there's a language change, such as allowing metadata on arbitrary identifiers)

test/rules/always_put_required_named_parameters_first.dart
```diff
@@ -0,0 +1,37 @@
...
+  A(
+    a, // OK
+    {
+    b, // OK
+    @required c, // LINT
+    @required d, // LINT
+    e, // OK
+    @required f, // LINT
+  });
+}
```
bwilkerson: It would be good to have an example of a required named parameter for which a lint should *not* be generated (because it's before any non-required parameters).

lib/src/rules/always_put_required_named_parameters_first.dart
a14n: done

test/rules/always_put_required_named_parameters_first.dart
a14n: done

PR #680, opened 2017-05-23
add lint prefer_const_constructors_in_immutables

a14n:


lib/src/rules/prefer_const_constructors_in_immutables.dart
```diff
@@ -0,0 +1,127 @@
...
+    final declaredVariables = node.element.context.declaredVariables;
+
+    final constantVisitor = new ConstantVisitor(
+        new ConstantEvaluationEngine(typeProvider, declaredVariables,
+            typeSystem: typeSystem),
+        new ErrorReporter(
+            AnalysisErrorListener.NULL_LISTENER, rule.reporter.source));
+
+    bool isConstExpression(Expression expression) =>
+        expression.accept(constantVisitor) != null;
```
a14n: @bwilkerson any idea why `isConstExpression(...)` return false for the expression of constructor field initializer in D.c2 ?

```dart
@immutable
class D {
  final _a;
  D.c1(a) : _a = a.toString(); // OK
  D.c2(a) : _a = a; // LINT
  D.c3(bool a) : _a = a && a; // LINT
}
```

a14n: Just found [_ConstantVerifier_validateInitializerExpression](https://github.com/dart-lang/sdk/blob/54bd8a79c48b4346d926ab3d3c3f986e5e691bc3/pkg/analyzer/lib/src/generated/resolver.dart#L10786) but it is not public.

a14n: I found a way to do with `ConstantVerifier`:

```dart
bool _hasOnlyConstExpressionsInIntializerList(ConstructorDeclaration node) {
  final typeProvider = node.element.context.typeProvider;
  final declaredVariables = node.element.context.declaredVariables;

  final listener = new MyAnalysisErrorListener();

  // put a fake const keyword to use ConstantVerifier
  node.constKeyword = new Token(TokenType.KEYWORD, node.offset);
  try {
    final errorReporter = new ErrorReporter(listener, rule.reporter.source);
    node.accept(new ConstantVerifier(errorReporter, node.element.library,
        typeProvider, declaredVariables));
  } finally {
    // restore const keyword
    node.constKeyword = null;
  }

  return !listener.hasConstError;
}
```

Is it safe to temporary change the `constKeyword` of `ConstructorDeclaration` like that ?

a14n:
Ready for review

lib/src/rules/prefer_const_constructors_in_immutables.dart
```diff
@@ -0,0 +1,136 @@
...
+          typeProvider, declaredVariables));
+    } finally {
+      // restore const keyword
+      node.constKeyword = null;
+    }
+
+    return !listener.hasConstError;
+  }
+
+  Iterable<InterfaceType> _getSelfAndInheritedTypes(InterfaceType type) sync* {
```
bwilkerson: The docs for `@immutable` indicate that it applies to a class C if C is directly annotated, or if it extends, mixes in, or inherits from a class that is immutable, but this method only looks at the extends chain. Was that intentional?

lib/src/rules/prefer_const_constructors_in_immutables.dart
```diff
@@ -0,0 +1,136 @@
...
+        _hasImmutableAnnotation(node.element.enclosingElement) &&
+        _hasConstSuperConstructor(node) &&
+        _hasOnlyConstExpressionsInIntializerList(node)) {
+      rule.reportLintForToken(node.firstTokenAfterCommentAndMetadata);
+    }
+  }
+
+  bool _hasImmutableAnnotation(ClassElement clazz) {
+    final inheritedAndSelfTypes =
+        _getSelfAndInheritedTypes(clazz.type).toList();
```
bwilkerson: Is it necessary to use `toList` here?

test/rules/prefer_const_constructors_in_immutables.dart
```diff
@@ -0,0 +1,43 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// test w/ `pub run test -N prefer_const_constructors_in_immutables`
+
+// Hack to work around issues importing `meta.dart` in tests.
+// Ideally, remove:
+library meta;
```
bwilkerson: Does this deal correctly with factory constructors? redirecting constructors?
There aren't any cases where the constructor has a non-empty body, but there should be.

lib/src/rules/prefer_const_constructors_in_immutables.dart
a14n: no, removed

lib/src/rules/prefer_const_constructors_in_immutables.dart
a14n: I excluded mixin because it leads to error _Const constructor can't be declared for a class with a mixin._

test/rules/prefer_const_constructors_in_immutables.dart
a14n: tests added.

a14n:
PTAL

lib/src/rules/prefer_const_constructors_in_immutables.dart
bwilkerson: Ah, of course. Sounds good.

test/rules/prefer_const_constructors_in_immutables.dart
bwilkerson: There's still no tests with redirecting constructors, such as:
```
@immutable
class G {
  G.a();
  G.b() : this.a();
  const G.c();
  G.d() : this.c();
}
```

a14n: Indeed there was an issue with `G.b()`. It's fixed now.

a14n:
PTAL

lib/src/rules/prefer_const_constructors_in_immutables.dart
```diff
@@ -0,0 +1,146 @@
...
+      final errorReporter = new ErrorReporter(listener, rule.reporter.source);
+      node.accept(new ConstantVerifier(errorReporter, node.element.library,
+          typeProvider, declaredVariables));
+    } finally {
+      // restore const keyword
+      node.constKeyword = null;
+    }
+
+    return !listener.hasConstError;
+  }
```
a14n: @bwilkerson I have a false positive with this `_hasOnlyConstExpressionsInIntializerList` function and the following code:

```dart
@immutable
class H {
  final f;
  H(f) : f = f ?? f == null; // OK
}
```

If I put `const` on constructor the analyzer triggers the error _In constant expressions, operands of this operator must be of type 'bool', 'num', 'String' or 'null'_ .

Any idea ?

bwilkerson: Yes. You're missing some constant-expression related errors. Minimally you need to add (to your listener): `CONST_EVAL_TYPE_BOOL`, `CONST_EVAL_TYPE_BOOL_NUM_STRING`, `CONST_EVAL_TYPE_INT`, `CONST_EVAL_TYPE_NUM`, `CONST_EVAL_THROWS_EXCEPTION` and `CONST_EVAL_THROWS_IDBZE`. There might be others that I'm not seeing.

a14n: Thanks for your help! I will send a PR to add those errors.

PR #688, opened 2017-05-31
update prefer_asserts_in_initializer_list to lint individual asserts

a14n:
With this change the lint is on individual `assert` and it checks that the `assert` can be moved into initializer list (no access to field, method, this)

### DIFF
--- a/lib/src/rules/prefer_asserts_in_initializer_list.dart
+++ b/lib/src/rules/prefer_asserts_in_initializer_list.dart
@@ -13,6 +13,9 @@ import 'package:linter/src/analyzer.dart';
 const desc = 'Prefer put asserts in initializer list.';
 
 const details = '''
+**WARNING** Putting asserts in initializer lists is only possible using an
+experimental language feature that might be removed.
+
 **DO** put asserts in initializer list for constructors with only asserts in
 their body.
 
--- a/lib/src/rules/prefer_asserts_in_initializer_list.dart
+++ b/lib/src/rules/prefer_asserts_in_initializer_list.dart
@@ -7,6 +7,7 @@ library linter.src.rules.prefer_const_constructors_in_immutables;
 import 'package:analyzer/analyzer.dart';
 import 'package:analyzer/dart/ast/ast.dart' show AstVisitor;
 import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/element.dart';
 import 'package:linter/src/analyzer.dart';
 
 const desc = 'Prefer put asserts in initializer list.';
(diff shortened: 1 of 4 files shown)


a14n:
PTAL I fixed the CI and added other cases (factory constructor, field formal parameters)

lib/src/rules/prefer_asserts_in_initializer_list.dart
```diff
@@ -68,30 +70,56 @@ class Visitor extends SimpleAstVisitor {
...
+  final ConstructorElement constructorElement;
   bool needInstance = false;
 
+  ClassElement get classElement => constructorElement.enclosingElement;
+
+  bool _isInstanceNeeded(Expression node) {
+    if (node is MethodInvocation) {
+      return _isInstanceNeeded(node.target);
+    } else if (node is PropertyAccess) {
+      return _isInstanceNeeded(node.target);
```
bwilkerson: Given that `visitMethodInvocation` and `visitPropertyAccess` both cause all of their children to be visited, you shouldn't need to explicitly invoke `_isInstanceNeeded` on their children. In which case, the check inside the `node is SimpleIdentifier` branch can be moved into `visitSimpleIdentifier` and `_isInstanceNeeded` can be removed (as well as two of the visit methods).

lib/src/rules/prefer_asserts_in_initializer_list.dart
```diff
@@ -68,30 +70,56 @@ class Visitor extends SimpleAstVisitor {
...
   @override
   visitSimpleIdentifier(SimpleIdentifier node) {
-    final element = node.staticElement;
-    if ((element is PropertyAccessorElement || element is MethodElement) &&
-        element.enclosingElement == classElement) {
+    if (_isInstanceNeeded(node)) {
       needInstance = true;
     }
-
     super.visitSimpleIdentifier(node);
```
bwilkerson: While it causes no harm, the `super.visitSimpleIdentifier(node);` invocation is not necessary because simple identifiers don't have children.

lib/src/rules/prefer_asserts_in_initializer_list.dart
a14n: Thank you really much. It simplified the code a lot :)

a14n:
PTAL

a14n:
>Does it do the right thing when the constructor is a const constructor?

Const constructors can not have a body. So there's no issue :)

>Should we auto-disable the lint if the context isn't configured to support asserts in initializer lists?

Good idea!

>Either way, we should document that putting asserts in initializer lists is only possible using an experimental language feature that might be removed.

Will do.

a14n:
A warning has been added to the rule description.

Opened #689 to track lint auto-disabling.

bwilkerson:
Then if you're ready we can merge it.

a14n:
Let's go for merge.

PR #692, opened 2017-06-05
add unnecessary_required lint

a14n:


a14n:
moved to dart-lang/sdk#29828

PR #705, opened 2017-06-15
fix prefer_const_constructors

a14n:
Fixes #703 
Fixes #704

PR #713, opened 2017-06-22
0.1.33 change notes.

pq:
@bwilkerson


a14n:
There has also been  a fix of `prefer_const_constructors` ( #705 )

pq:
Thanks @a14n !

PR #725, opened 2017-06-28
Improve `prefer_final_fields` performance (#724).

pq:
@bwilkerson 

Fixes: #724.

### DIFF
--- a/lib/src/rules/prefer_final_fields.dart
+++ b/lib/src/rules/prefer_final_fields.dart
@@ -65,19 +65,15 @@ class GoodMutable {
 ''';
 
 class PreferFinalFields extends LintRule {
-  _Visitor _visitor;
-
   PreferFinalFields()
       : super(
             name: 'prefer_final_fields',
             description: _desc,
             details: _details,
-            group: Group.style) {
-    _visitor = new _Visitor(this);
-  }
+            group: Group.style);
 
   @override
-  AstVisitor getVisitor() => _visitor;
+  AstVisitor getVisitor() => new _Visitor(this);
 }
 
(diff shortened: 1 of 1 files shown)


lib/src/rules/prefer_final_fields.dart
```diff
@@ -97,21 +82,38 @@ class PreferFinalFields extends LintRule {
...
+              .getCanonicalElementFromIdentifier(node.operand));
+        } else if (node is PostfixExpression) {
+          _mutatedElements.add(DartTypeUtilities
+              .getCanonicalElementFromIdentifier(node.operand));
+        }
+        node.childEntities.forEach(recurse);
+      }
+    }
+
+    node.childEntities.forEach(recurse);
```
bwilkerson: Or just `recurse(node)`

lib/src/rules/prefer_final_fields.dart
```diff
@@ -97,21 +82,38 @@ class PreferFinalFields extends LintRule {
 
 class _Visitor extends SimpleAstVisitor {
   final LintRule rule;
+  final Set<Element> _mutatedElements = new HashSet<Element>();
```
bwilkerson: It would be good to clear this set after the visit to avoid holding on to memory we no longer need. One possibility would be to not cache the visitor in the LintRule (I think the getter is only called once, so we could re-create the visitor on each call).

pq: Done!

lib/src/rules/prefer_final_fields.dart
pq: 👍

Issue #701, opened 2017-06-12
cascade_invocations conflict with null operator


AdalbertoLacruz:
Linter generates a false error with this code:

error ??= something;
error
   ..val1 = 1
   ..val2 = 2;

dustyholmes-wf:
I ran into this yesterday, we resolved by wrapping the assignment in parenthesis. 

(error ??= something)..val1=1

I'd really prefer not to have to do that though.

AdalbertoLacruz:
Something strange, but it's a solution.

Thanks.

PR #731, opened 2017-07-08
cascade_invocations: Don't lint cases where assignment expression is conditional

alexeieleusis:
Fixes https://github.com/dart-lang/linter/issues/701

### DIFF
--- a/test/rules/cascade_invocations.dart
+++ b/test/rules/cascade_invocations.dart
@@ -172,3 +172,14 @@ class Bug661 {
     parent.practicioner.bar(3);
   }
 }
+
+class Bug701 {
+  int foo;
+  int bar;
+}
+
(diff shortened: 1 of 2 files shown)


test/rules/cascade_invocations.dart
```diff
@@ -172,3 +172,14 @@ class Bug661 {
...
+class Bug701 {
+  int foo;
+  int bar;
+}
+
+void function701({Bug701 something}) {
+  something ??= new Bug701();
+  something.foo = 1; // OK
+  something.bar = 2; // LINT
+}
```
pq: Add a trailing newline?

alexeieleusis: we should have a lint for that! ;)

PR #732, opened 2017-07-11
cascade_invocations: Don't lint static properties.

alexeieleusis:
Fixes https://github.com/dart-lang/linter/issues/694

### DIFF
--- a/lib/src/rules/cascade_invocations.dart
+++ b/lib/src/rules/cascade_invocations.dart
@@ -178,12 +178,12 @@ class _CascadableExpression {
           isCritical: true);
     }
     // setters
-    return new _CascadableExpression._internal(
-        _getPrefixElementFromExpression(node.leftHandSide),
-        [node.rightHandSide],
-        canJoin: true,
-        canReceive: true,
-        canBeCascaded: true);
+    final VariableElement variable =
+        _getPrefixElementFromExpression(node.leftHandSide);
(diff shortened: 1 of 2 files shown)


lib/src/rules/cascade_invocations.dart
```diff
@@ -178,12 +178,12 @@ class _CascadableExpression {
...
     // setters
-    return new _CascadableExpression._internal(
-        _getPrefixElementFromExpression(node.leftHandSide),
-        [node.rightHandSide],
-        canJoin: true,
-        canReceive: true,
-        canBeCascaded: true);
+    final VariableElement variable =
+        _getPrefixElementFromExpression(node.leftHandSide);
+    final canReceive = node.operator.type != TokenType.QUESTION_QUESTION_EQ &&
```
pq: Isn't it possible for `variable` to be `null` here?

alexeieleusis: No because this was not a simple identifier, so being an assignment expression means it is a setter accessed either with this or identifier.

pq: Ah, great.  Thanks for clarifying!

alexeieleusis: I was wrong, I just had an exception where `variable` is null, will send a PR for this.

pq: 👍

Issue #720, opened 2017-06-27
False positive in invariant_booleans


lexaknyazev:
Dart SDK 1.25.0-dev.3.0
```dart
void main() {
  var a = 0;
  var index = 0;
  var length = 5;
  switch(a) {
    case 0:
      while(index < length) {
        ++index;
      }
      break;
    case 1:
      while(index < length) { // <--------
        ++index;
      }
      break;
  }
}
```
gives (line 12):
```
info: Conditions should not unconditionally evaluate to "TRUE" or to "FALSE" verify: index < length. 
```

alexeieleusis:
My guess is that the `while` statement in `case 0:` makes it trigger on `case 1`. Will look into this later, unless there is already a PR for it.

kevmoo:
Just hit this with stageXL on 2.0.0-dev.3.0

srawlins:
We've deprecated `invariant_booleans` and will not be fixing correctness bugs going forward.

PR #733, opened 2017-07-11
invariant_booleans: Fix false positive in loops inside cases of break statements

alexeieleusis:
Fixes https://github.com/dart-lang/linter/issues/720

### DIFF
--- a/lib/src/util/condition_scope_visitor.dart
+++ b/lib/src/util/condition_scope_visitor.dart
@@ -86,24 +86,35 @@ class ConditionScope {
 /// visit method and keep the scopes behavior consistent to its changes.
 ///
 /// When add a new local scope:
-/// - Visiting a non-empty function body: [BlockFunctionBody] [ExpressionFunctionBody].
+/// - Visiting a non-empty function body: [BlockFunctionBody]
+/// [ExpressionFunctionBody].
 /// - Visiting a flow-control statement: [IfStatement] ElseStatement.
-/// - Visiting loop statements: [DoStatement] [WhileStatement] [ForStatement] [ForEachStatement].
+/// - Visiting loop statements: [DoStatement] [WhileStatement] [ForStatement]
+/// [ForEachStatement].
 ///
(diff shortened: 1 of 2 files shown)


lib/src/util/condition_scope_visitor.dart
```diff
@@ -255,6 +267,20 @@ abstract class ConditionScopeVisitor extends RecursiveAstVisitor {
   }
 
   @override
+  visitSwitchCase(SwitchCase node) {
+    node.visitChildren(this);
+    AstNode previousCase = node.parent.childEntities
+        .where((e) => e is AstNode)
+        .lastWhere((n) => n != node, orElse: () => null);
```
bwilkerson: It seems to me that
```
[0, 1, 2, 3, 4].lastWhere((e) => e != i)
```
will return `4` for every value of `i` other than `4`. I don't think this code works at all like you want it to, despite the test passing. (The test is incomplete; you need some cases that don't end in a break statement.)

lib/src/util/condition_scope_visitor.dart
```diff
@@ -255,6 +267,20 @@ abstract class ConditionScopeVisitor extends RecursiveAstVisitor {
...
+    if (previousCase == null ||
+        previousCase.childEntities.last is! BreakStatement) {
```
bwilkerson: I think this also needs to account for return and throw statements, and all or some continue statements.

pq:
Gentle ping: @alexeieleusis, is this one abandoned?

alexeieleusis:
Apologies, I totally missed the previous ping. Right now I have a lot of things on my plate, I don't think I will be able to work on this in the next two weeks. I scheduled some time to look at this the first week of July, if I can't allocate enough time for it, I'll close the PR as abandoned then.

pq:
Awesome and no sweat.  Thanks for the update!

mit-mit:
Closing as assumed abandoned

PR #737, opened 2017-07-14
Update analysis_options.yaml

devoncarew:
- use the `directives_ordering` in package:linter

@pq


analysis_options.yaml
```diff
@@ -23,6 +23,7 @@ linter:
     - comment_references
 #    - constant_identifier_names
     - control_flow_in_finally
+    - directives_ordering
```
kevmoo: One of my favorites!

PR #746, opened 2017-07-19
cascade_invocations: Don't specify type and rather check for it in condition

alexeieleusis:
This was causing a cast exception.


alexeieleusis:
Thanks @a14n I will wait for @pq or @bwilkerson to be safe.

Issue #627, opened 2017-04-22
unnecessary_override should allow tightened return type


a14n:
the following case shouldn't lint

```dart
class A {
  Object m1() => null;
}
class B extends A {
  @override int m1() => super.m1(); // OK
}

**Comment by a14n** [comment]
A workaround that seems to work is:

```dart
class B extends A {
  @override int m1();
}
```

a14n:
However the semantic is not exactly the same. In checked mode a type check is done for the first snippet but not for the workaround snippet.

Hixie:
The workaround snippet is bogus (the B.m1 is entirely ignored, I think). We probably need a lint that prevents that from happening at all.
See also https://github.com/dart-lang/sdk/issues/28250.

natebosch:
Fixed in https://github.com/dart-lang/linter/pull/748

PR #748, opened 2017-07-19
improve unnecessary_overrides

### LINKED ISSUE #626: unnecessary_override should special case noSuchMethod method
`noSuchMethod` has to be override in some cases:

```dart
abstract class A {
  m();
}
class B implements A {
  @override
  noSuchMethod(Invocation invocation) => super.noSuchMethod(invocation);
}
```

Without the override I get a warning:

>Missing concrete implementation of 'A.m'

a14n:
Fixes #626, #627 and #747.

### DIFF
--- a/lib/src/rules/unnecessary_overrides.dart
+++ b/lib/src/rules/unnecessary_overrides.dart
@@ -124,12 +124,16 @@ abstract class _AbstractUnnecessaryOverrideVisitor extends SimpleAstVisitor {
   bool _haveSameDeclaration() {
     if (declaration.element.returnType != inheritedMethod.returnType)
       return false;
+    if (declaration.element.parameters.length !=
+        inheritedMethod.parameters.length) return false;
     for (var i = 0; i < inheritedMethod.parameters.length; i++) {
       final superParam = inheritedMethod.parameters[i];
       final param = declaration.element.parameters[i];
       if (param.type != superParam.type) return false;
       if (param.name != superParam.name) return false;
       if (param.isCovariant != superParam.isCovariant) return false;
+      if (param.parameterKind != superParam.parameterKind) return false;
+      if (param.defaultValueCode != superParam.defaultValueCode) return false;
     }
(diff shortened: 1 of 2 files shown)


lib/src/rules/unnecessary_overrides.dart
```diff
@@ -106,14 +120,27 @@ abstract class _AbstractUnnecessaryOverrideVisitor extends SimpleAstVisitor {
...
+
+  bool _haveSameDeclaration() {
+    if (declaration.element.returnType != inheritedMethod.returnType)
+      return false;
+    for (var i = 0; i < inheritedMethod.parameters.length; i++) {
+      final superParam = inheritedMethod.parameters[i];
+      final param = declaration.element.parameters[i];
+      if (param.type != superParam.type) return false;
+      if (param.name != superParam.name) return false;
+      if (param.isCovariant != superParam.isCovariant) return false;
```
bwilkerson: We should check the optionality of the parameters. Should we also check the default value for option parameters?

lib/src/rules/unnecessary_overrides.dart
```diff
@@ -54,12 +62,12 @@ class UnnecessaryOverrides extends LintRule {
 
 abstract class _AbstractUnnecessaryOverrideVisitor extends SimpleAstVisitor {
   LintRule rule;
-  Element inheritedMethod;
+  ExecutableElement inheritedMethod;
```
bwilkerson: Consider making `inheritedMethod` and `declaration` be local variables in `visitMethodDeclaration` and parameters to `_haveSameDeclaration`.

a14n: Those variables are used at several places and not only there.

lib/src/rules/unnecessary_overrides.dart
a14n: done

a14n:
PTAL

PR #755, opened 2017-07-25
New rule: Prefer spaces before EOL in string literals on the left

MichaelRFairhurst:
Ie, bad:

```
'Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do '
'eiusmod tempor incididunt ut labore et dolore magna aliqua.'
```

good:

```
'Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do'
' eiusmod tempor incididunt ut labore et dolore magna aliqua.'
```

Now, this *could do the opposite if that's preferred*. In our example
codebase, I found a 2:1 preference for the former and not the latter.
Why did I then make the former bad? Because what I ultimately want to
catch is missing or doubled spaces. And to catch that, only human
review has a decent chance at it.

My hope is that by turning this on, humans become better at seeing when
the spaces between words in wrapped strings are correct, because the
tool enforces one, which is more obvious, and humans come to expect the
obvious.

### DIFF
--- a/lib/src/rules/prefer_string_spaces_at_beginning.dart
+++ b/lib/src/rules/prefer_string_spaces_at_beginning.dart
@@ -1,0 +1,153 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:collection';
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:linter/src/analyzer.dart';
+
+const _desc = "Put spaces in wrapped strings on the beginning of new lines.";
+
+const _details = '''
+
+**DO** put your spaces between words in wrapped strings at the beginning of new
+lines.
+
+That means you should prefer '... foo' \\n ' bar' over '... foo ' \\n 'bar'.
+This provides a few advantages:
+
+- with this lint on, you can't accidentally do *both*.
+- with this lint on, your project will at least be consistent in either way
+- by preferring spaces at the beginning, it is easier for code reviewers to
+  catch when they are missing entirely
+
+Pretty much every exception to this is in raw strings, which is a solid
+indicator of something that is not human text and does not conform to any
+predictable spacing pattern.
+
+**BAD:**
+```
+ 'some amount of words that wrap around where the space is very '
+ 'easily missed'
+```
+
+**GOOD:**
+```
+ 'some amount of words that wrap around where the space is very'
+ ' easily recognized'
+```
+
+''';
+
+class PreferStringSpacesAtBeginning extends LintRule {
+  _Visitor _visitor;
+
+  PreferStringSpacesAtBeginning()
+      : super(
+            name: 'prefer_string_spaces_at_beginning',
+            description: _desc,
+            details: _details,
+            group: Group.style) {
+    _visitor = new _Visitor(this);
+  }
+
+  @override
+  AstVisitor getVisitor() => _visitor;
+}
+
+class _Visitor extends SimpleAstVisitor<bool> {
+  final LintRule rule;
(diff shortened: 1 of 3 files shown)


lib/src/rules/prefer_string_spaces_at_beginning.dart
```diff
@@ -0,0 +1,153 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:collection';
```
a14n: nit: empty line after

lib/src/rules/prefer_string_spaces_at_beginning.dart
```diff
@@ -0,0 +1,153 @@
...
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:linter/src/analyzer.dart';
+
+const _desc = "Put spaces in wrapped strings on the beginning of new lines.";
```
a14n: nit: single quote

lib/src/rules/prefer_string_spaces_at_beginning.dart
```diff
@@ -0,0 +1,153 @@
...
+
+const _details = '''
+
+**DO** put your spaces between words in wrapped strings at the beginning of new
+lines.
+
+That means you should prefer '... foo' \\n ' bar' over '... foo ' \\n 'bar'.
```
a14n: Unnecessary `\\n`

lib/src/rules/prefer_string_spaces_at_beginning.dart
```diff
@@ -0,0 +1,153 @@
...
+      return true;
+    }
+
+    return false;
+  }
+}
+
+class _ReportIfEndsInSpace extends SimpleAstVisitor {
+  final _Visitor owningVisitor;
+  _ReportIfEndsInSpace(this.owningVisitor);
```
a14n: You can pass `rule` instead of `owningVisitor`

lib/src/rules/prefer_string_spaces_at_beginning.dart
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+  @override
+  visitAdjacentStrings(AdjacentStrings strings) {
+    strings.strings.last.accept(this);
+  }
+
+  @override
+  visitSimpleStringLiteral(SimpleStringLiteral string) {
+    if (!string.isRaw && string.value.endsWith(' ')) {
```
a14n: Not sure raw that String should be exclude.

lib/src/rules/prefer_string_spaces_at_beginning.dart
```diff
@@ -0,0 +1,153 @@
...
+    _visitor = new _Visitor(this);
+  }
+
+  @override
+  AstVisitor getVisitor() => _visitor;
+}
+
+class _Visitor extends SimpleAstVisitor<bool> {
+  final LintRule rule;
+  final Set<AstNode> visited = new HashSet<AstNode>();
```
bwilkerson: Not sure why this set is necessary. Please add documentation (assuming you keep it).

lib/src/rules/prefer_string_spaces_at_beginning.dart
```diff
@@ -0,0 +1,153 @@
...
+  _Visitor(this.rule);
+
+  bool isPlusOp(AstNode node) =>
+      node is BinaryExpression && node.operator.lexeme == '+';
```
bwilkerson: This test will work, but we also have a TokenType that you could use, and I think that makes the code cleaner. Your call.

bwilkerson:
@MichaelRFairhurst Have we abandoned this PR?

pq:
@MichaelRFairhurst: is this one abandoned?

mit-mit:
Appears abandoned, closing

PR #756, opened 2017-07-26
New rule to detect (usually) unnecessary statements (no clear effects)

MichaelRFairhurst:
Allow expression statements which clearly have side effects, and allow
expression statements being used as control flow for expressions that
have side effects.

### DIFF
--- a/lib/src/rules/unnecessary_statements.dart
+++ b/lib/src/rules/unnecessary_statements.dart
@@ -1,4 +1,4 @@
-// Copyright (c) 2016, the Dart project authors.  Please see the AUTHORS file
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
@@ -6,10 +6,10 @@ import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
 import 'package:linter/src/analyzer.dart';
 
-const desc = r'Statements should have a clear effect.';
+const desc = 'Statements should have a clear effect.';
 
 const details = r'''
-**AVOID** unnecessary statments.
+**AVOID** unnecessary statements.
 
 Statements which have no clear effect are usually unnecessary, or should be
 broken up.
@@ -35,6 +35,7 @@ instance, the getter was likely supposed to be a function call.
 **GOOD:**
 ```
 some.method();
+new SomeClass();
 methodOne();
 methodTwo();
 foo ? bar() : baz();
@@ -64,7 +65,7 @@ class _Visitor extends SimpleAstVisitor {
     if (node.parent is FunctionBody) {
       return;
     }
-    node.accept(reportNoClearEffect);
+    node.expression.accept(reportNoClearEffect);
   }
 
   @override
@@ -76,12 +77,22 @@ class _Visitor extends SimpleAstVisitor {
   }
 }
 
-class _ReportNoClearEffectVisitor extends GeneralizingAstVisitor {
+class _ReportNoClearEffectVisitor extends UnifyingAstVisitor {
   final LintRule rule;
   _ReportNoClearEffectVisitor(this.rule);
 
   @override
-  visitInvocationExpression(InvocationExpression node) {
+  visitMethodInvocation(MethodInvocation node) {
+    // Has a clear effect
+  }
+
+  @override
+  visitSuperConstructorInvocation(SuperConstructorInvocation node) {
+    // Has a clear effect
+  }
+
+  @override
+  visitFunctionExpressionInvocation(FunctionExpressionInvocation node) {
     // Has a clear effect
   }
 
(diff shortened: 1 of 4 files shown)


lib/src/rules/unnecessary_statements.dart
```diff
@@ -0,0 +1,151 @@
+// Copyright (c) 2016, the Dart project authors.  Please see the AUTHORS file
```
a14n: 2017

lib/src/rules/unnecessary_statements.dart
```diff
@@ -0,0 +1,151 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:linter/src/analyzer.dart';
+
+const desc = r'Statements should have a clear effect.';
```
a14n: unnecessary raw marker

lib/src/rules/unnecessary_statements.dart
```diff
@@ -0,0 +1,151 @@
...
+
+const details = r'''
+**AVOID** unnecessary statments.
```
a14n: typo: statements

lib/src/rules/unnecessary_statements.dart
```diff
@@ -0,0 +1,151 @@
...
+  @override
+  visitForStatement(ForStatement node) {
+    node.initialization?.accept(reportNoClearEffect);
+    node.updaters?.forEach((u) {
+      u.accept(reportNoClearEffect);
+    });
+  }
+}
+
+class _ReportNoClearEffectVisitor extends GeneralizingAstVisitor {
```
bwilkerson: Can this subclass UnifyingAstVisitor instead?

MichaelRFairhurst:
Hey Brian,

what's the simplest way to try this out on a project? I don't actually know how to build a custom analysis server with a custom linter version...

bwilkerson:
You can build an analysis server by running something like `./tools/build.py --mode release --arch x64 create_sdk` (your architecture might be different than my Mac).

The easiest way to use a local copy of the linter package is to manually update the `.packages` file (in the root of the repo) to point to the local copy (prior to doing the build above). Don't forget to revert the `.packages` file when you're done. It's checked in so there's a risk of overwriting it with something invalid.

lib/src/rules/unnecessary_statements.dart
MichaelRFairhurst: Done

lib/src/rules/unnecessary_statements.dart
MichaelRFairhurst: Done

lib/src/rules/unnecessary_statements.dart
MichaelRFairhurst: Done

MichaelRFairhurst:
OK, ctor support has been added and other advice has been taken!

a14n:
Rerunning on flutter: 0 issue.

Regarding [my previous run](https://github.com/dart-lang/sdk/issues/30793#issuecomment-330452505) I'd like to have keeped the lint for https://github.com/flutter/flutter/blob/8e5ec0b153e008a5af139689292f214afeaad8f5/dev/bots/test.dart#L266

MichaelRFairhurst:
Interesting, and agreed! I thought my tests covered that. Maybe the type
parameters make it exceptional, I'll see if I can't get that line to lint
with a few more minutes' work.

MichaelRFairhurst:
> As written it produces a lint for each subexpression that doesn't have a clear effect.

I don't think this should happen; I overrode `visitNode` to not recurse inside `_ReportNoClearEffectVisitor`. The only cases where it does fan out is the conditional expression, since it checks that both the left and right sides have a clear effect....so there is a pathological case for something like:

```
  a ?
   (b ?
       c // LINT
       : d) // LINT
   (e ?
       f // LINT
       : g) // LINT
```

But in general, its tries to only check the direct expression of a subexpression, not its children.

Is there a way I can express this in the tests? Maybe

```
  1 + 2 + 3 + 4; // 1 LINT
```

MichaelRFairhurst:
I'm also trying to reproduce the flutter <String>[] lint. From what I can tell, it should be happening...

However, running into issues trying to build the sdk with the linter https://github.com/dart-lang/sdk/issues/30830

bwilkerson:
> Is there a way I can express this in the tests?

Not that I'm aware of. @pq Do you know of a way?

a14n:
My bad :-( I'd made a `git reset --hard` at a wrong place and I didnt use your version of linter.

a14n:
Actually, it looks good :)

[lint] Statements should have a clear effect. [(dev/bots/test.dart, line 266, col 3)](https://github.com/flutter/flutter/blob/master/dev/bots/test.dart#L266)
[lint] Statements should have a clear effect. [(packages/flutter/test/animation/animations_test.dart, line 162, col 17)](https://github.com/flutter/flutter/blob/master/packages/flutter/test/animation/animations_test.dart#L162)
[lint] Statements should have a clear effect. [(packages/flutter/lib/src/material/text_field.dart, line 216, col 7)](https://github.com/flutter/flutter/blob/master/packages/flutter/lib/src/material/text_field.dart#L216)
[lint] Statements should have a clear effect. [(packages/flutter/lib/src/rendering/image.dart, line 131, col 5)](https://github.com/flutter/flutter/blob/master/packages/flutter/lib/src/rendering/image.dart#L131)

MichaelRFairhurst:
That third one is definitely a bug!

The other three are a mix of exceptional and unnecessary....And I think only animation_test is exceptional.

MichaelRFairhurst:
a14n, anything else you want to see before approval?

MichaelRFairhurst:
I also can't merge this myself. So feel free, either one of you, to merge it or request changes/tests/etc

MichaelRFairhurst:
Thanks!

PR #761, opened 2017-07-31
Linter delinting.

pq:
Updated enabled lints and did a bunch of cleanup.

@bwilkerson

### DIFF
--- a/lib/src/util/tested_expressions.dart
+++ b/lib/src/util/tested_expressions.dart
@@ -65,6 +65,7 @@ bool _sameOperands(String eLeftOperand, String bcLeftOperand,
   return sameOperandsSameOrder || sameOperandsInverted;
 }
 
+// ignore: camel_case_types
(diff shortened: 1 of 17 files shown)


lib/src/util/tested_expressions.dart
```diff
@@ -65,6 +65,7 @@ bool _sameOperands(String eLeftOperand, String bcLeftOperand,
   return sameOperandsSameOrder || sameOperandsInverted;
 }
 
+// ignore: camel_case_types
```
bwilkerson: Is there a reason to not just fix the name rather than ignoring the lint (here and elsewhere)?

pq: Not a good one.  This case is what got me started down this road:

    class MockLineInfo_Location extends Mock implements LineInfo_Location {}

which makes sense to ignore IMO.

I've fixed the other ones though.

👍

PR #762, opened 2017-08-01
Cleanup and dogfood.

pq:
General cleanup and rule enablement:

  * avoid_catches_without_on_clauses
  * avoid_positional_boolean_parameters
  * prefer_expression_function_bodies
  * prefer_foreach

@bwilkerson 

FYI: @alexeieleusis @JPaulsen

### DIFF
--- a/bin/linter.dart
+++ b/bin/linter.dart
@@ -185,8 +185,8 @@ Future runLinter(List<String> args, LinterOptions initialLintOptions) async {
         machineOutput: options['machine'],
         quiet: options['quiet'])
       ..write();
-    // ignore: avoid_catching_errors
-  } on Error catch (err, stack) {
+    // ignore: avoid_catches_without_on_clauses
+  } catch (err, stack) {
     errorSink.writeln('''An error occurred while linting
   Please report it at: github.com/dart-lang/linter/issues
 $err
--- a/bin/linter.dart
+++ b/bin/linter.dart
@@ -185,7 +185,8 @@ Future runLinter(List<String> args, LinterOptions initialLintOptions) async {
         machineOutput: options['machine'],
         quiet: options['quiet'])
       ..write();
-  } catch (err, stack) {
+    // ignore: avoid_catching_errors
+  } on Error catch (err, stack) {
     errorSink.writeln('''An error occurred while linting
   Please report it at: github.com/dart-lang/linter/issues
 $err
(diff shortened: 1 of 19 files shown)


bin/linter.dart
```diff
@@ -185,7 +185,8 @@ Future runLinter(List<String> args, LinterOptions initialLintOptions) async {
         machineOutput: options['machine'],
         quiet: options['quiet'])
       ..write();
-  } catch (err, stack) {
+    // ignore: avoid_catching_errors
+  } on Error catch (err, stack) {
```
bwilkerson: I think it is actually appropriate to catch anything that might be thrown in a top-level method like this. I'd remove the `on` clause and ignore the lint.

lib/src/formatter.dart
```diff
@@ -173,9 +173,7 @@ class SimpleFormatter implements ReportFormatter {
     var tableWidth = max(_summaryLength, longest + largestCountGuess);
     var pad = tableWidth - longest;
     var line = ''.padLeft(tableWidth, '-');
-    out.writeln(line);
-    out.writeln('Counts');
-    out.writeln(line);
+    out..writeln(line)..writeln('Counts')..writeln(line);
```
bwilkerson: Your call, but I find this harder to read (and disagree with the lint in general). Having each line of output on a different line in the code makes it more clear for me.

test/rule_test.dart
```diff
@@ -436,7 +436,8 @@ testRule(String ruleName, File file, {bool debug: false}) {
     });
     try {
       expect(actual, unorderedMatches(expected));
-    } catch (e) {
+      // ignore: avoid_catching_errors
+    } on Error {
```
bwilkerson: This probably also wants to catch everything.

bin/linter.dart
pq: 👍

test/rule_test.dart
pq: 👍

pq:
Thanks!

lib/src/formatter.dart
a14n: It looks like an issue with `dartfmt` rather than with the cascade lint.

If `dartfmt` put cascades on new lines it will be simpler to read.

bwilkerson: I don't know that it's always useful to have cascades on separate lines, but yes, if they were formatted that way I wouldn't have a problem with using cascades in this case. :-)

PR #767, opened 2017-08-02
Improve `use_string_buffer` test coverage.

pq:
@bwilkerson 

As discussed, I'm exploring (and may open an issue to track) why `s = s + 'foo'` is OK in the foreach case.  In the meantime, this at least gets the logic better under coverage.


pq:
image

🎉

PR #774, opened 2017-08-10
add prefer_bool_in_asserts

a14n:
This lint will help for dart-lang/sdk#30326

### DIFF
--- a/lib/src/rules/prefer_bool_in_asserts.dart
+++ b/lib/src/rules/prefer_bool_in_asserts.dart
@@ -1,0 +1,50 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/analyzer.dart';
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:linter/src/analyzer.dart';
+import 'package:linter/src/util/dart_type_utilities.dart';
+
+const desc = 'Prefer bool as assert condition.';
+
+const details = '''
+**DO** use bool as assert condition.
+
+**BAD:**
+```
+assert(() {return f(););
+```
+
+**GOOD:**
+```
+assert(() {return f();)();
+```
+''';
+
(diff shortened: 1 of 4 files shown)


lib/src/rules/prefer_bool_in_asserts.dart
```diff
@@ -0,0 +1,50 @@
...
+import 'package:linter/src/util/dart_type_utilities.dart';
+
+const desc = 'Prefer bool as assert condition.';
+
+const details = '''
+**DO** use bool as assert condition.
+
+**BAD:**
+```
+assert(() {return f(););
```
bwilkerson: Perhaps
```
assert(() => f());
```

lib/src/rules/prefer_bool_in_asserts.dart
```diff
@@ -0,0 +1,50 @@
...
+```
+
+**GOOD:**
+```
+assert(() {return f();)();
```
bwilkerson: Perhaps
```
assert(f());
```

lib/src/rules/prefer_bool_in_asserts.dart
```diff
@@ -0,0 +1,50 @@
...
+}
+
+class Visitor extends SimpleAstVisitor {
+  final LintRule rule;
+  Visitor(this.rule);
+
+  @override
+  visitAssertStatement(AssertStatement node) {
+    if (!DartTypeUtilities.isClass(
+        node.condition.staticType, 'bool', 'dart.core')) {
```
bwilkerson: We might want to use `bestType` here in case analyzer was able to figure out a better type that static type analysis provided (mostly for the non-strong mode case).

bwilkerson:
Let me know when you're ready to merge.

lib/src/rules/prefer_bool_in_asserts.dart
a14n: Change for longer body.

lib/src/rules/prefer_bool_in_asserts.dart
a14n: Change for longer body.

lib/src/rules/prefer_bool_in_asserts.dart
a14n: done

PR #775, opened 2017-08-17
New rule: Prefer type annotating uninitialized variables and fields.

alexeieleusis:
Useful for my project as we are migrating to DDC.

Work done with @bencalabrese

### DIFF
--- a/lib/src/rules/prefer_typing_uninitialized_variables.dart
+++ b/lib/src/rules/prefer_typing_uninitialized_variables.dart
@@ -1,0 +1,86 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:linter/src/analyzer.dart';
+
+const _desc = r'Prefer typing uninitialized variables and fields.';
+
+const _details = r'''
+**PREFER** specifying a type annotation for uninitialized variables
+ and fields.
+
+**BAD:**
+```
+class BadClass {
+  static var bar; // LINT
+  var foo; // LINT
+
+  void method() {
+    var bar; // LINT
+    bar = 5;
+    print(bar);
+  }
+}
+```
+
+**BAD:**
+```
+void aFunction() {
+  var bar; // LINT
+  bar = 5;
+  ...
+}
+```
+
+**GOOD:**
+```
+class GoodClass {
+  static var bar = 7;
+  var foo = 42;
+  int baz; // OK
+
+  void method() {
+    int baz;
+    var bar = 5;
+    ...
+  }
+}
+```
+''';
+
+class PreferTypingUninitializedVariables extends LintRule {
+  _Visitor _visitor;
+
+  PreferTypingUninitializedVariables()
+      : super(
+            name: 'prefer_typing_uninitialized_variables',
+            description: _desc,
+            details: _details,
+            group: Group.style) {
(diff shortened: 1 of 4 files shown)


lib/src/rules/prefer_typing_uninitialized_variables.dart
```diff
@@ -0,0 +1,86 @@
...
+void aFunction() {
+  var bar; // LINT
+  bar = 5;
+  ...
+}
+```
+
+**GOOD:**
+```
+class BadClass {
```
bwilkerson: "Bad" --> "Good"?

test/rules/prefer_typing_uninitialized_variables.dart
```diff
@@ -0,0 +1,27 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+// test w/ `pub run test -N prefer_typing_uninitialized_variables`
+
+class BadClass {
+  static var bar; // LINT
+  var foo; // LINT
+  final baz; // LINT
+  var a = 5,
+      b; // LINT
```
bwilkerson: Add a field with a type.

test/rules/prefer_typing_uninitialized_variables.dart
```diff
@@ -0,0 +1,27 @@
...
+    bar = 5;
+    print(bar);
+  }
+}
+
+void aFunction() {
+  var bar; // LINT
+  bar = 5;
+  print(bar);
+}
```
bwilkerson: This rule will create lints for top-level variables and variables declared in for loops. There should probably be a couple of tests for those.

lib/src/rules/prefer_typing_uninitialized_variables.dart
```diff
@@ -0,0 +1,86 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:linter/src/analyzer.dart';
+
+const _desc = r'Prefer typing uninitialized variables and fields.';
+
+const _details = r'''
+
```
bwilkerson: Extra blank line?

alexeieleusis: Done.

lib/src/rules/prefer_typing_uninitialized_variables.dart
alexeieleusis: Done.

test/rules/prefer_typing_uninitialized_variables.dart
alexeieleusis: Done.

test/rules/prefer_typing_uninitialized_variables.dart
alexeieleusis: Done.

test/rules/prefer_typing_uninitialized_variables.dart
```diff
@@ -0,0 +1,40 @@
...
+  BadClass(this.baz);
+
+  void method() {
+    var bar; // LINT
+    bar = 5;
+    print(bar);
+  }
+
+  void someMethod() {
+    for (final v in <String>[]) {
```
bwilkerson: Wrong style of for loop :-). I should have been more explicit. (I usually call this a for-in loop to distinguish between the two.) VariableDeclarationList isn't a child of ForInStatement because can declare at most one variable. But that isn't the case in the "normal" for loop:
```
for (var i, j = 0; j < 5; i = j, j++) {
  print(i);
}
```

alexeieleusis: Done.

Thank you!

PR #790, opened 2017-09-30
have public_member_api_docs only apply to sources in the lib/ dir

devoncarew:
- update public_member_api_docs to only apply to sources in the lib/ dir; this will help us transition Flutter to using fewer analysis options files

The heuristic I'm using is a bit fuzzy, but I don't know of a good way to identify where the root of an analysis context is.

I saw that this lint is slow, but don't think this PR regresses it. Probably something to look at (locally, this is the slowest lint).

@pq


devoncarew:
(this will need some updating to fix travis)

lib/src/rules/public_member_api_docs.dart
```diff
@@ -101,9 +104,21 @@ class Visitor extends GeneralizingAstVisitor {
...
+    Uri uri = compilationUnit?.element?.source?.uri;
+
+    // TODO(devoncarew): This will have false positives.
+    return uri != null && uri.pathSegments.contains('lib');
+  }
+
   @override
   visitClassDeclaration(ClassDeclaration node) {
-    if (isPrivate(node.name)) {
+    if (isPrivate(node.name) || !isDefinedInLib(node)) {
```
bwilkerson: It would be faster to do this once in `visitCompilationUnit` and store the result in a field.

lib/src/rules/public_member_api_docs.dart
```diff
@@ -101,9 +104,21 @@ class Visitor extends GeneralizingAstVisitor {
...
+  /// Return true if the given node is declared in a compilation unit that is in
+  /// a `lib/` folder.
+  bool isDefinedInLib(AstNode node) {
+    CompilationUnit compilationUnit = node is CompilationUnit
+        ? node
+        : node.getAncestor((node) => node is CompilationUnit);
+    Uri uri = compilationUnit?.element?.source?.uri;
+
+    // TODO(devoncarew): This will have false positives.
+    return uri != null && uri.pathSegments.contains('lib');
```
bwilkerson: This is assuming that all the compilation units this lint will encounter are in a pub-style package. That isn't necessarily true.

But for unit inside a pub-style package, do we want to exclude compilation units in 'lib/src'?

lib/src/rules/public_member_api_docs.dart
```diff
@@ -59,6 +59,8 @@ documented getters have corresponding undocumented setters. In this case the
 setters inherit the docs from the getters.
 ''';
 
+// TODO(devoncarew): This lint is very slow - we should profile and optimize it.
```
bwilkerson: End-of-line comments between members will sometimes be moved by the sorter we use. If you care about where this comment is you might want to move it to somewhere where it won't be moved.

devoncarew:
> A safer way to identify the root of a pub-style package is to look for a directory containing a pubspec.yaml file.

Gotcha. I'll re-work to include some pubspec.yaml lookup logic.

lib/src/rules/public_member_api_docs.dart
devoncarew: > This is assuming that all the compilation units this lint will encounter are in a pub-style package. That isn't necessarily true.

Perhaps this lint could only be used for those projects though? As in, it doesn't make sense to enable in a project that doesn't use pub conventions.

> But for unit inside a pub-style package, do we want to exclude compilation units in 'lib/src'?

That's very tricky. The issue I'm seeing now is that this lint is reporting violations for directories like `bin/` and `test/` - that can never export their APIs. That's what this PR is trying to solve.

Excluding `lib/src/` is technically correct, but a much trickier issue, since some compilation units in there could be exported by files in `lib/`. Having a correct view of the public api of a package - and ensuring that all public members are documented - is a good goal, but a bit outside the scope of what I want to do here.

lib/src/rules/public_member_api_docs.dart
devoncarew: Makes sense, will do.

kevmoo:
Worth a TODO, but tough to implement: should really only apply to APIs that are exposed (outside of `lib/src`

So the contents of `lib/src/utils.dart` shouldn't be linted – unless something is exported from `lib/foo.dart` or similar...

lib/src/rules/public_member_api_docs.dart
bwilkerson: > Perhaps this lint could only be used for those projects though? As in, it doesn't make sense to enable in a project that doesn't use pub conventions.

I don't see why it wouldn't make sense to enable this rule in a non-pub package, but if we want to restrict it to only work in pub packages we should document that it is only intended to be used in that context (or the relevant restriction, such as "only lints files inside a 'lib' directory").

devoncarew:
Updated to:
- remove the test from test/rules - the lint now needs to lint code in a lib/ folder in order to test it
- move those tests into a new integration test, and create a sample project in test/_data/public_member_api_docs
- change the lib/ detection logic to walk up the directory structure looking for a pubspec.yaml file
- and, cache the look up in the visitor, the first time `visitCompilationUnit` is called

ptal :)

devoncarew:
> Worth a TODO, but tough to implement: should really only apply to APIs that are exposed (outside of lib/src

added -

PR #804, opened 2017-10-19
Create rule avoid_unused_constructor_parameters

aemino:
This rule, as per #799, lints instances within constructors where parameters are defined but not used.

### DIFF
--- a/lib/src/rules/avoid_unused_constructor_parameters.dart
+++ b/lib/src/rules/avoid_unused_constructor_parameters.dart
@@ -11,8 +11,8 @@ const _details = r'''
 
 **BAD:**
 ```
-class Bad {
-  Bad(int unusedParameter, [String unusedPositional]) {};
+class BadOne {
+  Bad(int unusedParameter, [String unusedPositional]);
 }
 ```
 
--- a/lib/src/rules/avoid_unused_constructor_parameters.dart
+++ b/lib/src/rules/avoid_unused_constructor_parameters.dart
@@ -55,8 +55,7 @@ class _Visitor extends SimpleAstVisitor {
     node?.body?.visitChildren(_constructorVisitor);
     node?.initializers?.forEach((i) => i.visitChildren(_constructorVisitor));
 
-    final unusedParameters = _constructorVisitor.unusedParameters;
-    if (unusedParameters.isNotEmpty) rule.reportLint(unusedParameters.first);
+    _constructorVisitor.unusedParameters.forEach(rule.reportLint);
   }
 }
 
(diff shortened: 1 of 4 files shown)


test/rules/avoid_unused_constructor_parameters.dart
```diff
@@ -0,0 +1,77 @@
...
+}
+
+class H {
+  int c;
+
+  H(int a, int b) : c = a + b; // OK because parameters are used in initializer
+}
+
+class I extends H {
+  I(int a, int b) : super(a, b); // OK because paramters are used in initializer
```
a14n: typo: parameters

lib/src/rules/avoid_unused_constructor_parameters.dart
```diff
@@ -0,0 +1,77 @@
...
+  _Visitor(this.rule);
+
+  @override
+  void visitConstructorDeclaration(ConstructorDeclaration node) {
+    final _constructorVisitor = new _ConstructorVisitor(rule, node);
+    node?.body?.visitChildren(_constructorVisitor);
+    node?.initializers?.forEach((i) => i.visitChildren(_constructorVisitor));
+
+    final unusedParameters = _constructorVisitor.unusedParameters;
+    if (unusedParameters.isNotEmpty) rule.reportLint(unusedParameters.first);
```
bwilkerson: It would be better to report every unused parameter, rather than only the first.

test/rules/avoid_unused_constructor_parameters.dart
```diff
@@ -0,0 +1,77 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// test w/ `pub run test -N avoid_unused_constructor_parameters`
```
bwilkerson: Just for completeness, it would be good to include a test for a redirecting constructor (`this(foo)`) and a factory constructor (`= Impl(foo)`).

lib/src/rules/avoid_unused_constructor_parameters.dart
aemino: The test suite expects one error from lines tagged with `// LINT`. Should I just report the entire line? Or is there a way I can get the test suite to expect multiple lint reports?

pq: In cases like that, however artificial it looks, I'd just put each offending statement on its own line.  For example:

```dart

class A {
  A(int x,  // LINT
    int y); // LINT
}
```

aemino:
I implemented the requested changes. Let me know if anything else needs to be changed! 😃

Issue #808, opened 2017-10-20
dartdoc and comment_references lint disagree on [true] and [false]


pq:
_From @kevmoo on October 18, 2017 1:51_

If one enables the [comment_references](http://dart-lang.github.io/linter/lints/comment_references.html) lint, doing`[true]` or `[false]` is fine.

dartdoc reports these as `unresolved-doc-reference`

...should be consistent

_Copied from original issue: dart-lang/dartdoc#1519_

pq:
_From @kevmoo on October 18, 2017 1:56_

FYI @pq

pq:
_From @jcollins-g on October 18, 2017 15:13_

dartdoc does not currently support keywords in comment references, but this could be added.

pq:
_From @devoncarew on October 18, 2017 15:23_

We may want to choose to change the lint here. I'd expect the contents of square brackets to generally be something that resolves to a symbol. For this case, the user should probably write:

    `true`

pq:
_From @jcollins-g on October 18, 2017 15:35_

I can also see the argument for `[]` working for true or false, since functionally those work the same as final global bools and when I'm coding, I don't usually make the distinction in my head that these are keywords.

In general, better synchronization between the linter and dartdoc seems like a good idea if we can find the space for it.

pq:
_From @kevmoo on October 18, 2017 15:54_

It's not like `true` exists as a top-level field in dart:core, though.
There is nothing for dartdoc to link to.

I'd rather we change linter, IMHO

PR #809, opened 2017-10-20
Fix comment_references to flag keywords (#808).

pq:
Update `comment_references` to check for keywords that are not treated as references by the parser but should be flagged by the linter.

Fixes: #808.

@bwilkerson

### DIFF
--- a/lib/src/rules/comment_references.dart
+++ b/lib/src/rules/comment_references.dart
@@ -57,9 +57,8 @@ class Visitor extends SimpleAstVisitor {
     for (Token token in node.tokens) {
       if (!token.isSynthetic) {
         String comment = token.lexeme;
-        int length = comment.length;
         int leftIndex = comment.indexOf('[');
-        while (leftIndex >= 0 && leftIndex + 1 < length) {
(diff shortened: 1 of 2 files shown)


lib/src/rules/comment_references.dart
```diff
@@ -49,10 +50,42 @@ class Visitor extends SimpleAstVisitor {
...
+  visitComment(Comment node) {
+    // Check for keywords that are not treated as references by the parser
+    // but should be flagged by the linter.
+    // Note that no special care is taken to handle embedded code blocks.
+    for (Token token in node.tokens) {
+      if (!token.isSynthetic) {
+        String comment = token.lexeme;
+        int length = comment.length;
+        int leftIndex = comment.indexOf('[');
+        while (leftIndex >= 0 && leftIndex + 1 < length) {
```
bwilkerson: nit: I don't think it's necessary to test `leftIndex + 1 < length`. When that's `false`, `rightIndex` will be negative.

pq: Good catch.  Fixed!

PR #817, opened 2017-10-23
Allow underscores around numbers in `non_constant_identifier_names`.

pq:
Better late than never!

@srawlins.

Fixes: #687.


matanlurey:
@srawlins is on leave, but this is LGTM to me!

pq:
Thanks @matanlurey !

PR #818, opened 2017-10-24
Fix false positive for `avoid_positional_boolean_parameters` on `[]=`

pq:
Fix false positive for `avoid_positional_boolean_parameters` on `[]=` ops.

Fixes: #803.

@a14n @bwilkerson

### DIFF
--- a/lib/src/rules/avoid_positional_boolean_parameters.dart
+++ b/lib/src/rules/avoid_positional_boolean_parameters.dart
@@ -4,7 +4,6 @@
 
 import 'package:analyzer/analyzer.dart';
 import 'package:analyzer/dart/ast/ast.dart';
-import 'package:analyzer/dart/ast/token.dart';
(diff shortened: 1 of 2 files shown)


lib/src/rules/avoid_positional_boolean_parameters.dart
```diff
@@ -84,6 +88,7 @@ class _Visitor extends SimpleAstVisitor {
   visitMethodDeclaration(MethodDeclaration node) {
     if (!node.isSetter &&
         !node.element.isPrivate &&
+        !_indexEqOp(node) &&
```
bwilkerson: Seems like we should we just disable this lint for all operators. Consider:
```
bool operator +(bool value) {}
```

pq: Great idea!

pq: Done. PTAL.

PR #825, opened 2017-10-26
prefer_function_declarations_over_variables: only final declarations or unmutatated vars

### LINKED ISSUE #823: prefer_function_declarations_over_variables should only lint final elements
otherwise you can not change the value of the variable

```dart
var f = () => true; // LINT
f = () => false;
```

 and the change leads to error:

```dart
f() => true; // OK
f = () => false; // ERROR
```

a14n:
Fixes #823

### DIFF
--- a/lib/src/rules/prefer_function_declarations_over_variables.dart
+++ b/lib/src/rules/prefer_function_declarations_over_variables.dart
@@ -53,8 +53,12 @@ class _Visitor extends SimpleAstVisitor {
 
   @override
   visitVariableDeclaration(VariableDeclaration node) {
-    if (node.initializer != null && node.initializer is FunctionExpression) {
-      rule.reportLint(node);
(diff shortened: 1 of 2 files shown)


lib/src/rules/prefer_function_declarations_over_variables.dart
```diff
@@ -53,8 +53,12 @@ class _Visitor extends SimpleAstVisitor {
 
   @override
   visitVariableDeclaration(VariableDeclaration node) {
-    if (node.initializer != null && node.initializer is FunctionExpression) {
-      rule.reportLint(node);
+    if (node?.initializer is FunctionExpression) {
```
bwilkerson: nit: I don't think the `node` can be `null`, so the `?.` shouldn't be needed.

a14n: done

Issue #736, opened 2017-07-12
Add explanation or reasoning for each lint


jacehensley-wf:
It would be nice if each lint had a description around why it was implemented and the benefit it provides. It would make turning on some lints easier to justify if the thought behind it was know.

alexeieleusis:
I am assuming this is in addition to http://dart-lang.github.io/linter/lints/ 
Or expanding those is enough? I think some of them are too brief.

jacehensley-wf:
I think adding to those would be enough. Like why should we use StringBuffer? It's not super obvious that it's for perf gains.

pq:
This is ongoing work.  Let's keep tracking in more targeted follow-ups (like #1025).

PR #830, opened 2017-10-31
Improve rule details and overall rule consistency

aemino:
As suggested by #736...

Though it was rather painful, I went through every single rule and tried to improve consistency and add more description in the details.  I tried to use my best judgement to infer what the preferred standards were, so let me know if something's off.

I forwent adding details to rules if I didn't feel it was necessary to elaborate.

Also, I saw some instances where a "smart apostrophe" was used; I wasn't sure whether or not to use it.

### DIFF
--- a/lib/src/rules/always_put_control_body_on_new_line.dart
+++ b/lib/src/rules/always_put_control_body_on_new_line.dart
@@ -12,11 +12,11 @@ const _details = r'''
 
 From the [flutter style guide](https://flutter.io/style-guide/):
 
-**DO** separate the control structre expression from its statement.
+**DO** separate the control structure expression from its statement.
 
 Don't put the statement part of an `if`, `for`, `while`, `do` on the same line
-as the expression, even if it is short. (Doing so makes it unobvious that there
-is relevant code there. This is especially important for early returns.)
+as the expression, even if it is short.  Doing so makes it unclear that there
+is relevant code there.  This is especially important for early returns.
 
 **GOOD:**
 ```
--- a/lib/src/rules/always_put_control_body_on_new_line.dart
+++ b/lib/src/rules/always_put_control_body_on_new_line.dart
@@ -6,12 +6,13 @@ import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
 import 'package:linter/src/analyzer.dart';
 
-const desc = 'Separate the control structure expression from its statement.';
+const _desc = r'Separate the control structure expression from its statement.';
 
-const details = '''
+const _details = r'''
+
 From the [flutter style guide](https://flutter.io/style-guide/):
 
-**DO** Separate the control structre expression from its statement.
+**DO** separate the control structre expression from its statement.
 
 Don't put the statement part of an `if`, `for`, `while`, `do` on the same line
 as the expression, even if it is short. (Doing so makes it unobvious that there
@@ -41,14 +42,15 @@ else print('ok')
 
 while (condition) i += 1;
 ```
+
 ''';
 
 class AlwaysPutControlBodyOnNewLine extends LintRule {
   AlwaysPutControlBodyOnNewLine()
       : super(
             name: 'always_put_control_body_on_new_line',
-            description: desc,
-            details: details,
+            description: _desc,
+            details: _details,
             group: Group.style);
 
   @override
--- a/lib/src/rules/avoid_classes_with_only_static_members.dart
+++ b/lib/src/rules/avoid_classes_with_only_static_members.dart
@@ -13,6 +13,10 @@ const _details = r'''
 
 **AVOID** defining a class that contains only static members.
 
+Creating classes with the sole purpose of providing utility or otherwise static
+methods is discouraged. Dart allows functions to be exist outside of classes
+for this very reason.
+
 **BAD:**
 ```
 class DateUtils {
--- a/lib/src/rules/avoid_classes_with_only_static_members.dart
+++ b/lib/src/rules/avoid_classes_with_only_static_members.dart
@@ -14,8 +14,8 @@ const _details = r'''
 **AVOID** defining a class that contains only static members.
 
 Creating classes with the sole purpose of providing utility or otherwise static
-methods is discouraged. Dart allows functions to be exist outside of classes
-for this very reason.
+methods is discouraged.  Dart allows functions to exist outside of classes for
+this very reason.
 
 **BAD:**
 ```
--- a/lib/src/rules/avoid_init_to_null.dart
+++ b/lib/src/rules/avoid_init_to_null.dart
@@ -7,9 +7,10 @@ import 'package:analyzer/dart/ast/visitor.dart';
 import 'package:linter/src/analyzer.dart';
 import 'package:linter/src/util/dart_type_utilities.dart';
 
-const desc = r"Don't explicitly initialize variables to null.";
+const _desc = r"Don't explicitly initialize variables to null.";
 
-const details = r'''
+const _details = r'''
+
 From [effective dart]
 (https://www.dartlang.org/effective-dart/usage/#dont-explicitly-initialize-variables-to-null):
 
@@ -19,7 +20,6 @@ In Dart, a variable or field that is not explicitly initialized automatically
 gets initialized to null. This is reliably specified by the language. There's
 no concept of "uninitialized memory" in Dart. Adding `= null` is redundant and
 unneeded.
-
 
 **GOOD:**
 ```
@@ -59,8 +59,8 @@ class AvoidInitToNull extends LintRule {
   AvoidInitToNull()
       : super(
             name: 'avoid_init_to_null',
-            description: desc,
-            details: details,
+            description: _desc,
+            details: _details,
             group: Group.style);
 
   @override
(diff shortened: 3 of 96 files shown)


lib/src/rules/always_put_control_body_on_new_line.dart
```diff
@@ -6,12 +6,13 @@ import 'package:analyzer/dart/ast/ast.dart';
...
-const desc = 'Separate the control structure expression from its statement.';
+const _desc = r'Separate the control structure expression from its statement.';
+
+const _details = r'''
 
-const details = '''
 From the [flutter style guide](https://flutter.io/style-guide/):
 
-**DO** Separate the control structre expression from its statement.
+**DO** separate the control structre expression from its statement.
```
bwilkerson: While you're at it, "structre" --> "structure"

lib/src/rules/avoid_classes_with_only_static_members.dart
```diff
@@ -13,6 +13,10 @@ const _details = r'''
 
 **AVOID** defining a class that contains only static members.
 
+Creating classes with the sole purpose of providing utility or otherwise static
+methods is discouraged. Dart allows functions to be exist outside of classes
```
bwilkerson: "be exist" --> "exist"

lib/src/rules/avoid_return_types_on_setters.dart
```diff
@@ -6,11 +6,14 @@ import 'package:analyzer/dart/ast/ast.dart';
...
 
-const desc = r'Avoid return types on setters.';
+const _desc = r'Avoid return types on setters.';
+
+const _details = r'''
 
-const details = r'''
 **AVOID** return types on setters.
 
+As setters do not return a value, setting the return type of one is redundant.
```
bwilkerson: "setting" --> "declaring"

lib/src/rules/hash_and_equals.dart
```diff
@@ -7,12 +7,15 @@ import 'package:analyzer/dart/ast/visitor.dart';
...
-const desc = r'Always override `hashCode` if overriding `==`.';
+const _desc = r'Always override `hashCode` if overriding `==`.';
 
-const details = r'''
+const _details = r'''
 
 **DO** override `hashCode` if overriding `==`.
 
+Every object in Dart should have a `hashCode`. This is necessary so that
+every Dart object can have its `hashCode` used as a key in a map.
```
bwilkerson: Every object has a hashCode. The reason it's important for `==` and `hashCode` to be consistent is so that a common implementation of hash maps will work correctly, and the reason to override both together is to maintain that consistency.

lib/src/rules/iterable_contains_unrelated_type.dart
```diff
@@ -13,10 +13,12 @@ const _desc = r'Invocation of Iterable<E>.contains with references of unrelated'
...
-**DON'T** Invoke `contains` on `Iterable` with an instance of different type than
-the parameter type since it will invoke `==` on its elements and most likely will
+**DON'T** invoke `contains` on `Iterable` with an instance of different type than
+the parameter type.
+
+Doing this will invoke `==` on its elements and most likely will
 return `false`. Strictly speaking it could evaluate to true since in Dart it
-is possible for an Iterable to contain elements of type unrelated to its
+is possible for an `Iterable` to contain elements of type unrelated to its
 parameter type, but this practice also should be avoided.
```
bwilkerson: I believe that the last statement is not true in strong mode, and hence should be removed.

lib/src/rules/list_remove_unrelated_type.dart
```diff
@@ -8,15 +8,17 @@ import 'package:linter/src/analyzer.dart';
...
-**DON'T** Invoke `remove` on `List` with an instance of different type than
-the parameter type since it will invoke `==` on its elements and most likely will
+**DON'T** Invoke `.remove` on `List` with an instance of different type than
+the parameter type.
+
+Doing this will invoke `==` on its elements and most likely will
 return `false`. Strictly speaking it could evaluate to true since in Dart it
-is possible for an List to contain elements of type unrelated to its
+is possible for a `List` to contain elements of type unrelated to its
 parameter type, but this practice also should be avoided.
```
bwilkerson: I believe that the last statement is not true in strong mode, and hence should be removed.

lib/src/rules/omit_local_variable_types.dart
```diff
@@ -8,12 +8,15 @@ import 'package:analyzer/dart/element/type.dart';
 import 'package:linter/src/analyzer.dart';
 import 'package:linter/src/util/dart_type_utilities.dart';
 
-const _desc = r'Omit the types for local variables.';
+const _desc = r'Consider omitting the types for local variables.';
```
bwilkerson: While the guideline leaves room for choice, the lint rule doesn't. Do we want the description to match the guideline, or what the rule is telling users?

lib/src/rules/prefer_asserts_in_initializer_lists.dart
```diff
@@ -10,9 +10,10 @@ import 'package:analyzer/dart/ast/visitor.dart';
 import 'package:analyzer/dart/element/element.dart';
 import 'package:linter/src/analyzer.dart';
 
-const desc = 'Prefer put asserts in initializer list.';
+const _desc = r'Prefer put asserts in initializer list.';
```
bwilkerson: "put" --> "putting"

lib/src/rules/prefer_foreach.dart
```diff
@@ -8,11 +8,16 @@ import 'package:analyzer/dart/element/element.dart';
 import 'package:linter/src/analyzer.dart';
 import 'package:linter/src/util/dart_type_utilities.dart';
 
-const _desc = r'Use forEach call to only apply a function to all the elements.';
+const _desc = r'Use `.forEach` to only apply a function to all the elements.';
```
bwilkerson: I'd remove the period (both here and below).

aemino: This certainly isn't the only instance where I used \`.methodName\` as a standard for referencing methods. Should I change all instances to be \`methodName\` instead?

bwilkerson: That would be _my_ preference, but I won't block the merge because of it.

lib/src/rules/close_sinks.dart
```diff
@@ -12,8 +12,10 @@ const _desc = r'Close instances of `dart.core.Sink`.';
 
 const _details = r'''
 
-**DO** invoke `close` on instances of `dart.core.Sink` to avoid memory leaks and
-unexpected behaviors.
+**DO** invoke `close` on instances of `dart.core.Sink`.
+
+Closing instances of StreamSubscription prevents memory leaks and
```
alexeieleusis: Sink

lib/src/rules/avoid_types_on_closure_parameters.dart
```diff
@@ -12,6 +12,10 @@ const _details = r'''
 
 **AVOID** annotating types on function expressions.
 
+Annotating types in function expressions is usually unnecessary because
+function expressions are almost always used on multiple values of the same
+type, thus making the practice redundant.
+
```
alexeieleusis: Consider adding something like: Usually the type should be inferred from the context, e.g. the function is provided as a callback argument.

aemino:
I went through all of the rules again and did another sweep, this time focusing on line formatting (making sure lines were cut off at 80 as efficiently as possible) and double-spaces after periods.

I also addressed all of the comments that were made; let me know if you see something else.

I'd also like to clarify -- there are some inconsistencies in the code where a special apostrophe character is used for statements like `DON'T`.  What's the preferred standard in this case?

bwilkerson:
I would prefer to _not_ use special characters. I suspect that they were the result of a copy-paste operation.

I'll try to take a look at the changes tomorrow (unless someone gets to it first).

kevmoo:
Does @aemino have commit access?

aemino:
@kevmoo No, I do not. I'm just a casual contributor. 🙂

kevmoo:
> @kevmoo No, I do not. I'm just a casual contributor. 🙂

Awesome! Just want to make sure this doesn't sit here unmerged. Thanks for your help!

lib/src/rules/avoid_init_to_null.dart
```diff
@@ -7,20 +7,20 @@ import 'package:analyzer/dart/ast/visitor.dart';
...
 From [effective dart]
 (https://www.dartlang.org/effective-dart/usage/#dont-explicitly-initialize-variables-to-null):
 
 **DON'T** explicitly initialize variables to null.
 
 In Dart, a variable or field that is not explicitly initialized automatically
-gets initialized to null. This is reliably specified by the language. There's
-no concept of "uninitialized memory" in Dart. Adding `= null` is redundant and
+gets initialized to null.  This is reliably specified by the language.  There's
+no concept of "uninitialized memory" in Dart.  Adding `= null` is redundant and
```
bwilkerson: "= null" --> "== null"

pq: I agree.  I'll merge this.  @aemino (or anyone 😉 ): feel free to follow up with any fixes.  Tag and I'll be happy to review. 👍

Hixie:
FYI: This PR broke Flutter, because we used to refer directly to the `details` member in `package_names.dart`, which this patch makes private.

Issue #805, opened 2017-10-20
non_constant_identifier_names: Lints new Function syntax


matanlurey:
```dart
// ignore: non_constant_identifier_names
T scope<T>(T Function() run) { /* ... */ }
```

`Function` is not the name of the parameter, so this should be fine.

aemino:
I'm using SDK version `1.25.0-dev.14.0` and I don't have this issue. Also, couldn't you declare that parameter as `(T run())`?

matanlurey:
This is on a later version of the SDK.

I could but that syntax is soft deprecated

aemino:
Ah, thanks for letting me know!

pq:
cc @bwilkerson

pq:
I can no longer reproduce this on bleeding edge so I'm inclined to think that the underlying issue in analyzer has been fixed.  Let's leave this open for a bit and see what shakes out.

Thanks for the report @matanlurey !

pq:
Closing in favor of a test that will fail if a dev build gets pushed that breaks in a similar way: https://github.com/dart-lang/linter/commit/e7fd27d23b50792df113ad59633c74ad58a44e83.

PR #839, opened 2017-11-06
Add test for new function syntax (#805).

pq:
See: #805.

@bwilkerson


test/rules/non_constant_identifier_names.dart
```diff
@@ -60,3 +60,8 @@ foo() {
 Main() => null; //LINT
 
 listen(void onData(Object event)) {}
+
+// New function syntax should be OK (#805).
```
bwilkerson: Perhaps "New" --> "Generic", because it won't be new forever. :-)

PR #845, opened 2017-11-14
update avoid_returning_this

### LINKED ISSUE #842: Document why avoid_returning_this accepts single return this or remove that exception
Why [the lint rule don't lint methods with only one statement](https://github.com/dart-lang/linter/blob/a8dcecd73a747b4aa93f7a81439e3e58d1cc56ac/test/rules/avoid_returning_this.dart#L23-L25) ? (If no valid reason is found the lint should also handle `ExpressionFunctionBody` for `A m() => this;`)

This strange behaviour should be explained in the rule description or removed.

Btw as mentioned in [this comment](https://github.com/flutter/flutter/pull/12719#issuecomment-343013265) it's weird to have the first one below to be ok but not the second:
```dart
Foo resolve(TextDirection textDirection) {
  return this;
}

Foo resolve(TextDirection textDirection) {
  assert(textDirection != null);
  return this;
}
```

a14n:
Improves `avoid_returning_this` and fixes #842

### DIFF
--- a/lib/src/rules/avoid_returning_this.dart
+++ b/lib/src/rules/avoid_returning_this.dart
@@ -17,6 +17,12 @@ const _details = r'''
 Returning `this` from a method is redundant; Dart has a cascade operator which
 allows method chaining universally.
 
+Returning `this` is allowed for:
+
+- operators
+- methods with a return type different of the current class
+- methods defined in parent classes / mixins or interfaces
+
 **BAD:**
 ```
 var buffer = new StringBuffer()
@@ -34,9 +40,6 @@ var buffer = new StringBuffer()
 ```
 
 ''';
-
-bool _hasInheritedMethod(MethodDeclaration node) =>
-    DartTypeUtilities.lookUpInheritedMethod(node) != null;
 
 bool _isFunctionExpression(AstNode node) => node is FunctionExpression;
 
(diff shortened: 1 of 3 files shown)


lib/src/rules/avoid_returning_this.dart
```diff
@@ -17,6 +17,12 @@ const _details = r'''
 Returning `this` from a method is redundant; Dart has a cascade operator which
 allows method chaining universally.
 
+Returning `this` is allowed for:
+
+- operators
+- methods with a return type different of the current class
```
MichaelRFairhurst: Why is this not linted?

I'm imagining someone doing, for instance,

```
class X implements Stream {
  ...
  Stream asStream() => this;
}
```

which IMO should be linted. `X` should simply be used as `useStream(new X())` rather than `useStream(new X().asStream())`.

I suppose maybe a case where I'd want this to be exceptional is where "return this" is a downcast, ie,

```
class X implements Stream<Object> {
  ...
  Stream<int> asIntStream() => this;
}
```

because the method is doing something -- its casting `this` in a way that isn't safe.

That said, even that method is not necessary, as you can do `useIntStream(new X())` because the downcast occurs in either case. At best you could make the argument that the method serves as a reassurance that its safe to perform such an upcast at least sometimes, whereas many upcasts are likely to be accidents.

lib/src/util/dart_type_utilities.dart
```diff
@@ -224,6 +224,16 @@ class DartTypeUtilities {
     }
     return false;
   }
+
+  static bool overridesMethod(MethodDeclaration node) => (node.parent
+          as ClassDeclaration)
+      .element
+      .allSupertypes
```
MichaelRFairhurst: I'm pretty sure you could do something like `new List().from(class.mixins)..addAll(interfaces)..addAll(supertypes)`.any((t) => t.lookup(m.name) != null)`

I'm not a dart performance expert, and don't know how efficiently `lookup` is implemented (or if that's even the right method/it works how you want here), but I suspect it would be faster in at least the negative case of where the method is not found and all inherited method/members must be inspected.

a14n: Done. Thanks.

lib/src/rules/avoid_returning_this.dart
a14n: I haven't change anything regarding this case. I've only add a line to document this behaviour ;)

Feel free to open a new issue to discuss this point. I tend to agree with you for inherited type.

pq:
cc @bwilkerson

pq:
@a14n : Does fixing #842, mean this may get turned on for Flutter?

a14n:
@pq I just tried this updated rule on flutter and it detected no lint. I will activate it once linter is rolled in flutter.

pq:
Super.  Thanks for checking!

PR #847, opened 2017-11-28
avoid_unused_constructor_parameters should not lint redirecting factory constructor

a14n:


pq:
Nice!

Adding @bwilkerson in case he has any thoughts.

PR #867, opened 2018-01-10
add prefer_const_literals_to_create_immutables

a14n:


lib/src/rules/prefer_const_literals_to_create_immutables.dart
```diff
@@ -0,0 +1,160 @@
...
+
+''';
+
+/// The name of `meta` library, used to define analysis annotations.
+String _META_LIB_NAME = 'meta';
+
+/// The name of the top-level variable used to mark a immutable class.
+String _IMMUTABLE_VAR_NAME = 'immutable';
+
+bool _isImmutable(Element element) =>
```
bwilkerson: The class is also public and can take an optional `reason` as an argument. Do we want to cover those cases here as well?

a14n: I created #869 to track handling of Immutable.

bwilkerson: Thanks.

PR #872, opened 2018-01-23
avoid_shadowing

### LINKED ISSUE #597: A lint for shadowing members?
As an extension to https://www.dartlang.org/guides/language/effective-dart/usage#dont-use-this-when-not-needed-to-avoid-shadowing

Would it make sense to add a new lint type for shadowing an inherited member?

```dart
class A {
  @protected
  int get stuff => ... blah ...;
}

class B extends A {
  int stuff = ... blah ...;
  
  [1000 lines later]

  void someMethod() {
    int thingy = stuff; // I thought I was getting super.stuff
  }
}

### PR DESCRIPTION
**Pull request by a14n**
Fixes #597

### DIFF
--- a/lib/src/rules/avoid_shadowing.dart
+++ b/lib/src/rules/avoid_shadowing.dart
@@ -77,8 +77,17 @@ class _Visitor extends SimpleAstVisitor {
 
   @override
   visitVariableDeclarationStatement(VariableDeclarationStatement node) {
-    final variables = node.variables.variables;
+    final variables = node.variables.variables.toList();
     final library = variables.first.element.library;
+
+    // exclude pattern : var name = this.name;
+    variables.removeWhere((variable) {
+      final initializer = variable.initializer;
+      return initializer is PropertyAccess &&
+          initializer.propertyName.name == variable.name.name &&
+          (initializer.target is ThisExpression ||
+              initializer.target is SuperExpression);
+    });
 
     bool skipInstanceMembers = false;
     AstNode current = node;
--- a/lib/src/rules/avoid_shadowing.dart
+++ b/lib/src/rules/avoid_shadowing.dart
@@ -2,8 +2,6 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-library linter.src.rules.avoid_shadowing;
-
 import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
 import 'package:analyzer/dart/element/element.dart';
@@ -89,11 +87,10 @@ class _Visitor extends SimpleAstVisitor {
     });
 
     // exclude pattern : same name as current getter name
-    var currentGetter = _getCurrentGetter(node);
-    if (currentGetter != null) {
-      variables.removeWhere(
-          (variable) => currentGetter.name.name == variable.name.name);
-    }
+    variables.removeWhere(_hasSameNameAsItsGetter);
+
+    // exit on empty variables to check
+    if (variables.isEmpty) return;
 
     bool skipInstanceMembers = false;
     AstNode current = node;
@@ -188,17 +185,34 @@ class _Visitor extends SimpleAstVisitor {
     }
   }
 
-  MethodDeclaration _getCurrentGetter(VariableDeclarationStatement node) {
-    AstNode current = node.parent;
-    while (current != null) {
-      if (current is Block || current is FunctionBody) {
-        current = current.parent;
-      } else if (current is MethodDeclaration && current.isGetter) {
-        return current;
-      } else {
-        break;
-      }
+  /// excludes pattern:
+  ///
+  /// ```dart
+  /// get name {
+  ///   var name;
+  ///   ...
+  /// }
+  /// ```
+  bool _hasSameNameAsItsGetter(VariableDeclaration node) {
+    final name = node.name.name;
+    VariableDeclarationList variableList = node.parent;
+    final variableDeclarationStatement = variableList.parent;
+
+    var current = variableDeclarationStatement.parent;
+    if (current is! Block) return false;
+
+    current = current.parent;
+    if (current is! BlockFunctionBody) return false;
+
+    current = current.parent;
+    if (current is MethodDeclaration && current.isGetter)
+      return current.name.name == name;
+    if (current is FunctionExpression) {
+      final parent = current.parent;
+      if (parent is FunctionDeclaration && parent.isGetter)
+        return parent.name.name == name;
     }
-    return null;
+
+    return false;
   }
 }
--- a/lib/src/rules/avoid_shadowing.dart
+++ b/lib/src/rules/avoid_shadowing.dart
@@ -89,6 +89,13 @@ class _Visitor extends SimpleAstVisitor {
               initializer.target is SuperExpression);
     });
 
+    // exclude pattern : same name as current getter name
+    var currentGetter = _getCurrentGetter(node);
+    if (currentGetter != null) {
+      variables.removeWhere(
+          (variable) => currentGetter.name.name == variable.name.name);
+    }
+
     bool skipInstanceMembers = false;
     AstNode current = node;
     while (current != null) {
@@ -180,5 +187,19 @@ class _Visitor extends SimpleAstVisitor {
         rule.reportLint(variable);
       }
     }
+  }
+
+  MethodDeclaration _getCurrentGetter(VariableDeclarationStatement node) {
+    AstNode current = node.parent;
+    while (current != null) {
+      if (current is Block || current is FunctionBody) {
+        current = current.parent;
+      } else if (current is MethodDeclaration && current.isGetter) {
+        return current;
+      } else {
+        break;
+      }
+    }
+    return null;
   }
 }
--- a/lib/src/rules/avoid_shadowing.dart
+++ b/lib/src/rules/avoid_shadowing.dart
@@ -78,6 +78,7 @@ class _Visitor extends SimpleAstVisitor {
   @override
   visitVariableDeclarationStatement(VariableDeclarationStatement node) {
     final variables = node.variables.variables;
+    final library = variables.first.element.library;
 
     bool skipInstanceMembers = false;
     AstNode current = node;
@@ -109,11 +110,8 @@ class _Visitor extends SimpleAstVisitor {
       }
 
       current = current.parent;
-      if (current == null) {
-        final library = node.variables.variables.first.element.library;
-        _checkLibrary(library, variables);
-      }
     }
+    _checkLibrary(library, variables);
   }
 
   void _checkClass(
--- a/lib/src/rules/avoid_shadowing.dart
+++ b/lib/src/rules/avoid_shadowing.dart
@@ -1,0 +1,177 @@
+// Copyright (c) 2018, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+library linter.src.rules.avoid_shadowing;
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/element.dart';
+import 'package:linter/src/analyzer.dart';
+
+const desc = r'Do not use a name already visible.';
+
+const details = r'''
+
+**DO** Do not use a name already visible as getter or variable.
+
+**BAD:**
+```
+var k = null;
+
+class A {
+  var a;
+}
+
+class B extends A {
+  get b => null;
+
+  m() {
+    var a; // LINT
+    var b; // LINT
+    var k; // LINT
+  }
+}
+```
+
+**GOOD:**
+```
+var k = null;
+
+class A {
+  var a;
+}
+
+class B extends A {
+  get b => null;
+
+  m() {
+    var c; // OK
+    var d; // OK
+    var e; // OK
+  }
+}
+```
+''';
+
+class AvoidShadowing extends LintRule {
+  _Visitor _visitor;
+
+  AvoidShadowing()
+      : super(
+            name: 'avoid_shadowing',
+            description: desc,
+            details: details,
+            group: Group.errors) {
+    _visitor = new _Visitor(this);
+  }
+
+  @override
+  AstVisitor getVisitor() => _visitor;
+}
+
+class _Visitor extends SimpleAstVisitor {
+  final LintRule rule;
+
+  _Visitor(this.rule);
+
+  @override
+  visitVariableDeclarationStatement(VariableDeclarationStatement node) {
+    final variables = node.variables.variables;
+
+    bool skipInstanceMembers = false;
+    AstNode current = node;
+    while (current != null) {
+      if (current is ClassDeclaration) {
+        _checkClass(current, variables, onlyStatics: skipInstanceMembers);
+      } else if (current is ConstructorDeclaration) {
+        _checkParameters(current.parameters, variables);
+        if (current.factoryKeyword != null) {
+          skipInstanceMembers = true;
+        }
+      } else if (current is MethodDeclaration) {
+        _checkParameters(current.parameters, variables);
+        if (current.isStatic) {
+          skipInstanceMembers = true;
+        }
+      } else if (current is FunctionExpression) {
+        _checkParameters(current.parameters, variables);
+      } else if (current is FunctionDeclaration) {
+        _checkParameters(current.functionExpression.parameters, variables);
+      } else if (current is FunctionDeclarationStatement) {
+        _checkParameters(
+            current.functionDeclaration.functionExpression.parameters,
+            variables);
+      }
+
+      if (current.parent is Block) {
+        _checkParentBlock(current, variables);
+      }
+
+      current = current.parent;
+      if (current == null) {
+        final library = node.variables.variables.first.element.library;
+        _checkLibrary(library, variables);
+      }
+    }
+  }
+
+  void _checkClass(
+    ClassDeclaration clazz,
+    List<VariableDeclaration> variables, {
+    bool onlyStatics,
+  }) {
+    for (final variable in variables) {
+      final name = variable.name.name;
+      final getter = clazz.element.lookUpGetter(name, clazz.element.library);
+      if (getter != null && (!onlyStatics || getter.isStatic))
+        rule.reportLint(variable);
+    }
+  }
+
+  void _checkLibrary(
+      LibraryElement library, List<VariableDeclaration> variables) {
+    final topLevelVariableNames = library.units
+        .expand((u) => u.topLevelVariables)
+        .map((e) => e.name)
+        .toList();
+    for (final variable in variables) {
+      final name = variable.name.name;
+      if (topLevelVariableNames.contains(name)) {
+        rule.reportLint(variable);
+      }
+    }
+  }
+
+  void _checkParameters(
+      FormalParameterList parameters, NodeList<VariableDeclaration> variables) {
+    if (parameters == null) return;
+
+    final parameterNames =
+        parameters.parameterElements.map((e) => e.name).toList();
+
+    for (final variable in variables) {
+      final name = variable.name.name;
+      if (parameterNames.contains(name)) {
+        rule.reportLint(variable);
+      }
+    }
+  }
+
+  void _checkParentBlock(
+      AstNode node, NodeList<VariableDeclaration> variables) {
+    final block = node.parent as Block;
+    final names = <String>[];
+    for (final statement in block.statements.takeWhile((n) => n != node)) {
+      if (statement is VariableDeclarationStatement) {
+        names.addAll(statement.variables.variables.map((e) => e.name.name));
+      }
+    }
+    for (final variable in variables) {
+      final name = variable.name.name;
+      if (names.contains(name)) {
+        rule.reportLint(variable);
+      }
+    }
+  }
+}
--- a/lib/src/rules/avoid_shadowing.dart
+++ b/lib/src/rules/avoid_shadowing.dart
@@ -54,20 +54,19 @@ class B extends A {
 ```
 ''';
 
-class AvoidShadowing extends LintRule {
-  _Visitor _visitor;
-
+class AvoidShadowing extends LintRule implements NodeLintRule {
   AvoidShadowing()
       : super(
             name: 'avoid_shadowing',
             description: desc,
             details: details,
-            group: Group.errors) {
-    _visitor = new _Visitor(this);
-  }
+            group: Group.errors);
 
   @override
-  AstVisitor getVisitor() => _visitor;
+  void registerNodeProcessors(NodeLintRegistry registry) {
+    final visitor = new _Visitor(this);
+    registry.addVariableDeclarationStatement(this, visitor);
+  }
 }
 
 class _Visitor extends SimpleAstVisitor {
@@ -78,7 +77,7 @@ class _Visitor extends SimpleAstVisitor {
   @override
   visitVariableDeclarationStatement(VariableDeclarationStatement node) {
     final variables = node.variables.variables.toList();
-    final library = variables.first.element.library;
+    final library = variables.first.declaredElement.library;
 
     // exclude pattern : var name = this.name;
     variables.removeWhere((variable) {
@@ -137,7 +136,8 @@ class _Visitor extends SimpleAstVisitor {
   }) {
     for (final variable in variables) {
       final name = variable.name.name;
-      final getter = clazz.element.lookUpGetter(name, clazz.element.library);
+      final getter = clazz.declaredElement
+          .lookUpGetter(name, clazz.declaredElement.library);
       if (getter != null && (!onlyStatics || getter.isStatic))
         rule.reportLint(variable);
     }
@@ -158,7 +158,7 @@ class _Visitor extends SimpleAstVisitor {
   }
 
   void _checkParameters(
-      FormalParameterList parameters, NodeList<VariableDeclaration> variables) {
+      FormalParameterList parameters, List<VariableDeclaration> variables) {
     if (parameters == null) return;
 
     final parameterNames =
@@ -172,8 +172,7 @@ class _Visitor extends SimpleAstVisitor {
     }
   }
 
-  void _checkParentBlock(
-      AstNode node, NodeList<VariableDeclaration> variables) {
+  void _checkParentBlock(AstNode node, List<VariableDeclaration> variables) {
     final block = node.parent as Block;
     final names = <String>[];
     for (final statement in block.statements.takeWhile((n) => n != node)) {
--- a/test/rules/avoid_shadowing.dart
+++ b/test/rules/avoid_shadowing.dart
@@ -1,0 +1,75 @@
+// Copyright (c) 2018, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: lib/src/rules/avoid_shadowing.dart [hunk 1]
```diff
@@ -0,0 +1,177 @@
...
+            current.functionDeclaration.functionExpression.parameters,
+            variables);
+      }
+
+      if (current.parent is Block) {
+        _checkParentBlock(current, variables);
+      }
+
+      current = current.parent;
+      if (current == null) {
```
bwilkerson: The body of this `if` could be moved outside the `while`.

test/rules/avoid_shadowing.dart
```diff
@@ -0,0 +1,75 @@
+// Copyright (c) 2018, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// test w/ `pub run test -N avoid_shadowing`
+
+var top = null;
```
bwilkerson: I don't see any tests for shadowing parameter names.

Also, I suspect that the rule as written will produce multiple lints for the same location if the name being shadowed is defined in multiple locations. Consider a test such as:
```
class C {
  int x;
  m(int x) {
    int x; // LINT
  }
}
```
(Can we verify that only one lint is produced in this case?)

lib/src/rules/avoid_shadowing.dart
a14n: done

test/rules/avoid_shadowing.dart
a14n: done.

It looks like several report on the same token for the same rule do not result in lint duplications.

a14n:
With the exclusion pattern `var x = this.x;` Flutter has the 82 below lints (121 without the pattern).

- [dev/bots/analyze-sample-code.dart, line 212, col 24](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/dev/bots/analyze-sample-code.dart#L212)
- [dev/tools/dartdoc.dart, line 108, col 22](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/dev/tools/dartdoc.dart#L108)
- [dev/devicelab/lib/framework/adb.dart, line 386, col 16](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/dev/devicelab/lib/framework/adb.dart#L386)
- [examples/catalog/bin/sample_page.dart, line 198, col 28](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/examples/catalog/bin/sample_page.dart#L198)
- [packages/flutter/test/material/user_accounts_drawer_header_test.dart, line 313, col 15](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter/test/material/user_accounts_drawer_header_test.dart#L313)
- [packages/flutter/test/material/user_accounts_drawer_header_test.dart, line 202, col 15](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter/test/material/user_accounts_drawer_header_test.dart#L202)
- [packages/flutter/test/material/time_picker_test.dart, line 547, col 24](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter/test/material/time_picker_test.dart#L547)
- [packages/flutter/test/foundation/diagnostics_test.dart, line 26, col 33](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter/test/foundation/diagnostics_test.dart#L26)
- [packages/flutter/test/widgets/linked_scroll_view_test.dart, line 153, col 20](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter/test/widgets/linked_scroll_view_test.dart#L153)
- [packages/flutter/test/widgets/linked_scroll_view_test.dart, line 163, col 20](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter/test/widgets/linked_scroll_view_test.dart#L163)
- [packages/flutter/lib/src/material/animated_icons/animated_icons.dart, line 106, col 25](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter/lib/src/material/animated_icons/animated_icons.dart#L106)
- [packages/flutter/lib/src/material/tabs.dart, line 795, col 21](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter/lib/src/material/tabs.dart#L795)
- [packages/flutter/lib/src/material/tabs.dart, line 790, col 21](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter/lib/src/material/tabs.dart#L790)
- [packages/flutter/lib/src/material/slider.dart, line 602, col 18](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter/lib/src/material/slider.dart#L602)
- [packages/flutter/lib/src/material/time_picker.dart, line 1497, col 33](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter/lib/src/material/time_picker.dart#L1497)
- [packages/flutter/lib/src/material/checkbox.dart, line 207, col 20](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter/lib/src/material/checkbox.dart#L207)
- [packages/flutter/lib/src/material/input_decorator.dart, line 1181, col 27](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter/lib/src/material/input_decorator.dart#L1181)
- [packages/flutter/lib/src/material/input_decorator.dart, line 1075, col 27](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter/lib/src/material/input_decorator.dart#L1075)
- [packages/flutter/lib/src/painting/decoration_image.dart, line 417, col 18](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter/lib/src/painting/decoration_image.dart#L417)
- [packages/flutter/lib/src/rendering/box.dart, line 527, col 18](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter/lib/src/rendering/box.dart#L527)
- [packages/flutter/lib/src/rendering/box.dart, line 526, col 18](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter/lib/src/rendering/box.dart#L526)
- [packages/flutter/lib/src/rendering/wrap.dart, line 691, col 17](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter/lib/src/rendering/wrap.dart#L691)
- [packages/flutter/lib/src/rendering/wrap.dart, line 690, col 20](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter/lib/src/rendering/wrap.dart#L690)
- [packages/flutter/lib/src/rendering/wrap.dart, line 390, col 9](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter/lib/src/rendering/wrap.dart#L390)
- [packages/flutter/lib/src/rendering/wrap.dart, line 422, col 9](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter/lib/src/rendering/wrap.dart#L422)
- [packages/flutter/lib/src/rendering/wrap.dart, line 689, col 20](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter/lib/src/rendering/wrap.dart#L689)
- [packages/flutter/lib/src/rendering/wrap.dart, line 606, col 9](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter/lib/src/rendering/wrap.dart#L606)
- [packages/flutter/lib/src/rendering/table.dart, line 1156, col 30](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter/lib/src/rendering/table.dart#L1156)
- [packages/flutter/lib/src/rendering/table.dart, line 1157, col 30](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter/lib/src/rendering/table.dart#L1157)
- [packages/flutter/lib/src/rendering/object.dart, line 1123, col 13](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter/lib/src/rendering/object.dart#L1123)
- [packages/flutter/lib/src/rendering/paragraph.dart, line 255, col 16](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter/lib/src/rendering/paragraph.dart#L255)
- [packages/flutter/lib/src/rendering/viewport.dart, line 354, col 14](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter/lib/src/rendering/viewport.dart#L354)
- [packages/flutter/lib/src/rendering/viewport.dart, line 426, col 18](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter/lib/src/rendering/viewport.dart#L426)
- [packages/flutter/lib/src/rendering/viewport.dart, line 447, col 18](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter/lib/src/rendering/viewport.dart#L447)
- [packages/flutter/lib/src/rendering/viewport.dart, line 1402, col 18](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter/lib/src/rendering/viewport.dart#L1402)
- [packages/flutter/lib/src/rendering/viewport.dart, line 473, col 26](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter/lib/src/rendering/viewport.dart#L473)
- [packages/flutter/lib/src/rendering/editable.dart, line 525, col 20](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter/lib/src/rendering/editable.dart#L525)
- [packages/flutter/lib/src/rendering/sliver_multi_box_adaptor.dart, line 197, col 17](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter/lib/src/rendering/sliver_multi_box_adaptor.dart#L197)
- [packages/flutter/lib/src/semantics/semantics.dart, line 964, col 12](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter/lib/src/semantics/semantics.dart#L964)
- [packages/flutter/lib/src/semantics/semantics.dart, line 966, col 19](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter/lib/src/semantics/semantics.dart#L966)
- [packages/flutter/lib/src/semantics/semantics.dart, line 961, col 12](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter/lib/src/semantics/semantics.dart#L961)
- [packages/flutter/lib/src/semantics/semantics.dart, line 963, col 12](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter/lib/src/semantics/semantics.dart#L963)
- [packages/flutter/lib/src/semantics/semantics.dart, line 962, col 12](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter/lib/src/semantics/semantics.dart#L962)
- [packages/flutter/lib/src/semantics/semantics.dart, line 965, col 12](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter/lib/src/semantics/semantics.dart#L965)
- [packages/flutter/lib/src/widgets/pages.dart, line 41, col 31](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter/lib/src/widgets/pages.dart#L41)
- [packages/flutter/lib/src/widgets/overlay.dart, line 569, col 17](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter/lib/src/widgets/overlay.dart#L569)
- [packages/flutter/lib/src/widgets/page_view.dart, line 256, col 18](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter/lib/src/widgets/page_view.dart#L256)
- [packages/flutter/lib/src/widgets/framework.dart, line 3084, col 24](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter/lib/src/widgets/framework.dart#L3084)
- [packages/flutter/lib/src/widgets/icon.dart, line 118, col 25](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter/lib/src/widgets/icon.dart#L118)
- [packages/flutter/lib/src/widgets/scroll_position_with_single_context.dart, line 188, col 32](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter/lib/src/widgets/scroll_position_with_single_context.dart#L188)
- [packages/flutter/lib/src/widgets/gesture_detector.dart, line 539, col 33](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter/lib/src/widgets/gesture_detector.dart#L539)
- [packages/flutter_driver/test/src/extension_test.dart, line 66, col 21](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter_driver/test/src/extension_test.dart#L66)
- [packages/flutter_tools/test/protocol_discovery_test.dart, line 153, col 33](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter_tools/test/protocol_discovery_test.dart#L153)
- [packages/flutter_tools/test/protocol_discovery_test.dart, line 117, col 33](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter_tools/test/protocol_discovery_test.dart#L117)
- [packages/flutter_tools/test/protocol_discovery_test.dart, line 116, col 35](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter_tools/test/protocol_discovery_test.dart#L116)
- [packages/flutter_tools/test/protocol_discovery_test.dart, line 152, col 35](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter_tools/test/protocol_discovery_test.dart#L152)
- [packages/flutter_tools/test/protocol_discovery_test.dart, line 135, col 33](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter_tools/test/protocol_discovery_test.dart#L135)
- [packages/flutter_tools/test/protocol_discovery_test.dart, line 134, col 35](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter_tools/test/protocol_discovery_test.dart#L134)
- [packages/flutter_tools/test/devfs_test.dart, line 479, col 18](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter_tools/test/devfs_test.dart#L479)
- [packages/flutter_tools/test/base/build_test.dart, line 431, col 30](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter_tools/test/base/build_test.dart#L431)
- [packages/flutter_tools/test/base/build_test.dart, line 477, col 34](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter_tools/test/base/build_test.dart#L477)
- [packages/flutter_tools/test/commands/analyze_once_test.dart, line 147, col 23](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter_tools/test/commands/analyze_once_test.dart#L147)
- [packages/flutter_tools/test/ios/mac_test.dart, line 59, col 26](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter_tools/test/ios/mac_test.dart#L59)
- [packages/flutter_tools/lib/src/device.dart, line 147, col 30](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter_tools/lib/src/device.dart#L147)
- [packages/flutter_tools/lib/src/flutter_manifest.dart, line 70, col 22](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter_tools/lib/src/flutter_manifest.dart#L70)
- [packages/flutter_tools/lib/src/flutter_manifest.dart, line 133, col 32](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter_tools/lib/src/flutter_manifest.dart#L133)
- [packages/flutter_tools/lib/src/vmservice.dart, line 1063, col 12](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter_tools/lib/src/vmservice.dart#L1063)
- [packages/flutter_tools/lib/src/vmservice.dart, line 1064, col 26](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter_tools/lib/src/vmservice.dart#L1064)
- [packages/flutter_tools/lib/src/vmservice.dart, line 674, col 17](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter_tools/lib/src/vmservice.dart#L674)
- [packages/flutter_tools/lib/src/vmservice.dart, line 662, col 18](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter_tools/lib/src/vmservice.dart#L662)
- [packages/flutter_tools/lib/src/version.dart, line 237, col 20](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter_tools/lib/src/version.dart#L237)
- [packages/flutter_tools/lib/src/doctor.dart, line 323, col 21](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter_tools/lib/src/doctor.dart#L323)
- [packages/flutter_tools/lib/src/commands/create.dart, line 198, col 20](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter_tools/lib/src/commands/create.dart#L198)
- [packages/flutter_tools/lib/src/commands/create.dart, line 176, col 20](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter_tools/lib/src/commands/create.dart#L176)
- [packages/flutter_tools/lib/src/commands/run.dart, line 305, col 20](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter_tools/lib/src/commands/run.dart#L305)
- [packages/flutter_tools/lib/src/commands/fuchsia_reload.dart, line 359, col 18](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter_tools/lib/src/commands/fuchsia_reload.dart#L359)
- [packages/flutter_tools/lib/src/commands/fuchsia_reload.dart, line 252, col 18](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter_tools/lib/src/commands/fuchsia_reload.dart#L252)
- [packages/flutter_tools/lib/src/commands/fuchsia_reload.dart, line 375, col 38](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter_tools/lib/src/commands/fuchsia_reload.dart#L375)
- [packages/flutter_tools/lib/src/protocol_discovery.dart, line 92, col 11](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter_tools/lib/src/protocol_discovery.dart#L92)
- [packages/flutter_tools/lib/src/protocol_discovery.dart, line 68, col 9](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter_tools/lib/src/protocol_discovery.dart#L68)
- [packages/flutter_tools/lib/src/android/android_studio.dart, line 226, col 18](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter_tools/lib/src/android/android_studio.dart#L226)
- [packages/flutter_tools/lib/src/cache.dart, line 470, col 22](https://github.com/flutter/flutter/blob/159db5ca442f9891a24f2a4273e2fad0884b1d2a/packages/flutter_tools/lib/src/cache.dart#L470)

bwilkerson:
That still seems like a lot, and it leaves me concerned that we haven't really defined correctly what we want to have linted and what we don't. It might be worth adding those results to the original issue to see what the reaction from the Flutter team is.

srawlins:
@a14n What happened to this? Did you check if allowing [@natebosch 's example](https://github.com/dart-lang/linter/issues/597#issuecomment-360321237) helps?

a14n:
>Did you check if allowing @natebosch 's example helps?

Yes there's a test with that sample.

>What happened to this?

I really don't remember why this rule was put aside.

lib/src/rules/avoid_shadowing.dart
```diff
@@ -0,0 +1,204 @@
+// Copyright (c) 2018, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+library linter.src.rules.avoid_shadowing;
```
bwilkerson: Please remove the library directive.

lib/src/rules/avoid_shadowing.dart
```diff
@@ -0,0 +1,204 @@
...
+    // exclude pattern : same name as current getter name
+    var currentGetter = _getCurrentGetter(node);
+    if (currentGetter != null) {
+      variables.removeWhere(
+          (variable) => currentGetter.name.name == variable.name.name);
+    }
+
+    bool skipInstanceMembers = false;
+    AstNode current = node;
+    while (current != null) {
```
bwilkerson: This doesn't seem very efficient, but perhaps in practice it doesn't matter. Have we done any timing measures for this rule?

One easy improvement would be to return before line 98 if `variables` is empty.

bwilkerson:
Ah, I didn't think to re-read the conversation, so I'd forgotten the issue about false positives. It would be interesting to see how many we still have and whether there are other patterns we should allow (before merging this).

bwilkerson:
Thanks! I'm still convinced that we don't have a reasonable definition of what does and does not need to be linted. Until we get that we won't have a rule that's worth adding.

On a meta-level, I have to wonder whether this rule would have enough value (assuming we can define it well) to warrant the amount of effort it will likely take to finish defining the rule. It might be better to put our time toward something with a higher value.

srawlins:
There are two cases which this lint rule does not catch (or at least they're not tested). I've hit both recently in real code :frowning_face: 

#### import prefixes

```dart
import 'dart:math' as m;

void main() {
  var rand = m.Random();
}

void fn(String m) { // lint here
  m.Random x =  null; // dart2js breaks here.
}
```

This even passes the analyzer, and breaks on, .e.g, dart2js compile.

#### A type parameter shadowing a thingy (e.g. class)

```dart
class Foo {}

class Bar<Foo> {} // lint here
```

I could catch this in avoid_shadowing_type_parameters, but I think these two lint rules should be more separated by the thingies they track (the shadowees), rather than the offenders, the shadowers. Seems like the code and performance would be cleaner in each.

mit-mit:

filiph:
This would be a great lint, for a very real problem. Is there a way to resurrect this PR? Or is it already implemented by some other lint? Looking at the available lints today, it looks like we have type parameter shadowing lint, but no regular identifier shadowing lint (at least not by that name).

pq:
I believe this fell due to concerns about an overwhelm of false-positives.  I'm not sure that was ever validated though.  Or maybe the logic could be dialed back to be less aggressive?

/cc @a14n

bwilkerson:
@filiph The reason this PR was closed is because we did not get to a definition of the rule that provided enough value to outweigh the cost of the false positives. While we do allow a small number of false positives from lint rules, we have a very low tolerance for them. This rule, as written in this PR, did not meet that standard.

I believe that the core of the issue is that, while shadowing causes lots of problems, there are far too many places where shadowing is standard practice in the developer community. Perhaps a better path forward would be to follow the pattern of `avoid_shadowing_type_parameters` and identify a small number of uses cases where shadowing is _not_ standard practice and hence could be linted.

filiph:
Thanks for the explanation! 

FWIW, I think a more conservative (== less false positives) lint rule would be very useful, even if it doesn't get into `pedantic` by default. After all, this is a common bug, and a nasty one at that. I had it in my code and caught it just yesterday, Sam says here in the thread he ("recently", at the time) hit two instances in his, Kris has a very real example in https://github.com/dart-lang/language/issues/737, and, looking outside Dart, someone fixed a [pretty major variable shadowing bug](https://github.com/openbsd/src/commit/57915963662663da67aa51b4f944c93678db74ce) in OpenBSD (the UNIX distribution with emphasis on code quality) a week ago.

bwilkerson:
Absolutely! Shadowing causes all lots of bugs and anything we can do to help users find them would be great.

But studies show that developers will stop using any lint that produces too many false positives [1], so there's no value doing this unless we can identify the patterns that won't cause false positives (which is doable, it just takes effort). I'd encourage every reading this thread to open an issue with the specific use cases you've found that you'd like to have a lint for.

[1] Actually, it's worse than that. The study actually shows that a significant portion of users will ignore _all_ lints if any one lint produces too many false positives. So it's not really "no value" it's "negative value".

jpenna:
Arrived here because I have something like this that I'd like to avoid:

```dart

  definitions.forEach((id, definition) {
    // ...other code

    for (var id in definition.downRefIds) { // <-- I'd like to know I'm shadowing something here
      if (definitions[id] != null) {
        definitions[id]!.addUpRefId(id);
      }
    }

    for (var id in definition.upRefIds) { // <-- and here
      definitions[id]!.addDownRefId(id);
    }
  });
```

Coming from JS/TS, there is a rule to complain about it and I like the way they did it there: https://palantir.github.io/tslint/rules/no-shadowed-variable/

PR #874, opened 2018-01-24
add lint prefer_const_declarations

### LINKED ISSUE #795: Prefer const fields (or getters) where posssible
A lint that encourages fields that are effectively const to be made const or getters could be very useful in keeping hot reload scenarios on the rails.  (And may also lead to reduced storage in AOT.)

a14n:
Fixes #795

### DIFF
--- a/lib/src/rules/prefer_const_declarations.dart
+++ b/lib/src/rules/prefer_const_declarations.dart
@@ -6,14 +6,15 @@ import 'package:analyzer/analyzer.dart';
 import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
 import 'package:linter/src/analyzer.dart';
-// ignore: implementation_imports
-
 
 const _desc = r'Prefer const over final for declarations.';
 
 const _details = r'''
 
 **PREFER** using `const` for const declarations.
+
+Const declarations are more hot-reload friendly and allow to use const
+constructors if an instantiation references this declaration.
 
 **GOOD:**
 ```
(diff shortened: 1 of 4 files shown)


lib/src/rules/prefer_const_declarations.dart
```diff
@@ -0,0 +1,91 @@
+// Copyright (c) 2018, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/analyzer.dart';
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:linter/src/analyzer.dart';
+// ignore: implementation_imports
```
bwilkerson: Remove this line? It doesn't appear to apply to anything.

lib/src/rules/prefer_const_declarations.dart
```diff
@@ -0,0 +1,91 @@
...
+
+
+const _desc = r'Prefer const over final for declarations.';
+
+const _details = r'''
+
+**PREFER** using `const` for const declarations.
```
bwilkerson: It would be good to add an explanation of what the rule checks for and why following this rule is a good thing.

devoncarew:
@a14n, fantastic! Thanks for implementing this lint :)

lib/src/rules/prefer_const_declarations.dart
a14n: Done

lib/src/rules/prefer_const_declarations.dart
a14n: Done

a14n:
PTAL

a14n:
>Do you want to add a quick fix for this, or should I?

I can have a look but not quickly.

a14n:
>Do you want to add a quick fix for this, or should I?

https://dart-review.googlesource.com/c/sdk/+/37301
dart-lang/sdk#31993

PR #875, opened 2018-01-27
Type variables, parameters, and return values as Set, not HashSet

srawlins:
There is a question about whether LinkedHashSet will implement HashSet, in Dart 2.0. To reduce the friction, and just for more standard typing, we should type variables, parameters, and return values to Set, rather than HashSet, when there's no reason to hold these objects to act like HashMaps.

CC @lrhn @keertip


pq:
FYI @bwilkerson

srawlins:
All three affected methods are private; no breaking changes; nothing for CHANGELOG (?)

PR #879, opened 2018-02-07
add rule unnecessary_parenthesis

a14n:


a14n:
I already run it on flutter flutter/flutter#14475 and flutter/flutter#14533

bwilkerson:
Sounds good to me.

PR #880, opened 2018-02-07
lint also local variables for prefer_const_declarations

a14n:


test/rules/prefer_const_declarations.dart
```diff
@@ -32,3 +32,17 @@ class A {
...
+
+m() {
+  const o1 = const []; // OK
+  final o2 = []; // OK
+  final o3 = const []; // LINT
+  final o4 = ''; // LINT
+  final o5 = 1; // LINT
+  final o6 = 1.3; // LINT
+  final o7 = null; // LINT
+  final o8 = const {}; // LINT
```
bwilkerson: Can we add a `const` constructor to `A`, a method `int m() => 0`, and the following test cases?
```
final o9 = {};
final o10 = new A();
final o11 = const A();
final o12 = A.o1;
final o13 = A.o2;
final o14 = 011.m();
```

a14n: Done.

a14n:
PTAL

I already run it on flutter codebase (with the old version of the lint on local variable) in flutter/flutter#14358 and flutter/flutter#14416

Rerunning the new version on Flutter I see a strange behaviour because https://github.com/flutter/flutter/blob/7db0564ab695e849602da7391b30f46e90df06eb/packages/flutter/test/painting/alignment_test.dart#L40 has a lint but not the 2 following lines that looks completly the same. Not sure if it comes from the analyzer or flutter.

bwilkerson:
I agree, I would have expected all three lines to be treated the same.

You should be able to rule out flutter as a cause by running the dartanalyzer directly.

The code all looks good to me, it's just a question of being sure we won't break anything. Let me know when you've had a chance to investigate the strange behavior you're seeing.

a14n:
I tried to dive into this strange behaviour but I didn't have enough time to really understand what is exactly happening. It looks quite similar to dart-lang/sdk#31835.

```dart
// test/a.dart
class A {
  const A(this.i);
  final int i;
  static const A i1 = const A(1);
  static const A i2 = const A(2);
}

void main() {
  final a1 = A.i1;
  final a2 = A.i2;
}

// test/b.dart
import 'a.dart';

void main() {
  final a1 = A.i1;
  final a2 = A.i2;
}
```

and the analyze of `b.dart` only lints `a.dart`:

```
/analyzer_cli $ dart --checked bin/analyzer.dart --lints --packages .packages --options ./analysis_options.yaml test/b.dart

  lint • Prefer const over final for declarations at test/a.dart:9:3 • prefer_const_declarations
  lint • Prefer const over final for declarations at test/a.dart:10:3 • prefer_const_declarations
```

I also notice that Travis failed on dev channel. Is it something you will fix on analyzer side ?

a14n:
I think the strange behaviour is related to dart-lang/sdk#31835 and so this PR could be merged.

bwilkerson:
It appears that there are merge conflicts.

a14n:
rebased

a14n:
ping :)

bwilkerson:
Thanks! I missed the "rebased" e-mail.

PR #881, opened 2018-02-07
Removed defunct calls to set `enableAssertInitializer`.

pq:
@bwilkerson


pq:
@bwilkerson : great question regarding the lower-bound.  It looks like that change went in here: https://github.com/dart-lang/sdk/commit/082bbdb6a7295e91001a414ecf45f8b9e44e243a#diff-fe5a2855fb3e7f44ec592292717089c1.  It looks like we should bump to 0.31.0-alpha.2 to get it.

pq:
@bwilkerson lower-bound updated.  PTAL.

PR #889, opened 2018-02-11
Fix grind.dart

pq:
Fixes: https://github.com/dart-lang/linter/issues/886

While this works, the ergonomics aren't great.

I'd prefer `grind rule foo` but I think that's not how grind works.  Also: it would be great to not lose `valueHelp`.

@devoncarew: can you point me to any examples that show help for args?  How is that done in the "new" world?

cc @bwilkerson

### DIFF
--- a/tool/grind.dart
+++ b/tool/grind.dart
@@ -12,14 +12,14 @@ import 'rule.dart';
 main(args) => grind(args);
 
 @Task('Generate lint rule docs.')
-docs(GrinderContext context) {
+docs() {
   TaskArgs args = context.invocation.arguments;
   String dir = args.getOption('dir');
   generateDocs(dir);
 }
 
 @Task('Generate a lint rule stub.')
-rule(GrinderContext context) {
+rule() {
   TaskArgs args = context.invocation.arguments;
   String name = args.getOption('name');
   generateRule(name, outDir: Directory.current.path);
--- a/tool/grind.dart
+++ b/tool/grind.dart
@@ -12,7 +12,7 @@ import 'rule.dart';
 main(args) => grind(args);
 
 @Task('Generate lint rule docs.')
-docs(TaskArgs args) {
+docs(GrinderContext context) {
   TaskArgs args = context.invocation.arguments;
   String dir = args.getOption('dir');
   generateDocs(dir);
--- a/tool/grind.dart
+++ b/tool/grind.dart
@@ -5,33 +5,22 @@
 import 'dart:io';
 
 import 'package:grinder/grinder.dart';
-import 'package:unscripted/unscripted.dart';
 
 import 'doc.dart';
 import 'rule.dart';
 
-main([List<String> args]) {
-  _addTask('rule',
-      parser: (String name) =>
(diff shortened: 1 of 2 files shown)


devoncarew:
Sure! Here's the PR that updated grinder: https://github.com/google/grinder.dart/pull/316

But probably slightly easier to parse, here's an example of passing and using args: https://github.com/google/grinder.dart/#passing-parameters-to-tasks

For the script in this PR, invoking it might look something like: `grind rule --name=foo`.

pq:
Thanks Devon!

What's odd is that if I try to use that API, I get an exception at runtime:

```
$ grind rule --name=foo

NoSuchMethodError: Class 'GrinderContext' has no instance method 'getOption'.
Receiver: Instance of 'GrinderContext'
Tried calling: getOption("name")
```

where the rule task looks like:

```
@Task('Generate a lint rule stub.')
rule(TaskArgs args) {
  String name = args.getOption('name');
  generateRule(name, outDir: Directory.current.path);
}
```

defining the rule to take a context works though:

```
@Task('Generate a lint rule stub.')
rule(GrinderContext context) {
  TaskArgs args = context.invocation.arguments;
  String name = args.getOption('name');
  generateRule(name, outDir: Directory.current.path);
}
```

devoncarew:
Ah, sounds like a bug in grinder, or perhaps in our docs for how to pass in args?

pq:
API questions aside, this does fix the compilation errors (and gets us off `unscripted` to boot).

@bwilkerson @devoncarew PTAL 👍

bwilkerson:
Sorry, but I don't really have the context to review this (one reason I didn't try to fix it).

pq:
No worries!  @devoncarew would you mind taking a quick peek?

tool/grind.dart
```diff
@@ -5,33 +5,22 @@
...
-_addTask(String name, {String description, Parser parser, String valueHelp}) {
-  addTask(new GrinderTask(name, taskFunction: () {
-    String value = context.invocation.positionals.first;
-    parser(value);
-  },
-      description: description,
-      positionals: [new Positional(valueHelp: valueHelp)]));
+@Task('Generate a lint rule stub.')
+rule(GrinderContext context) {
+  TaskArgs args = context.invocation.arguments;
```
devoncarew: I don't think `grinder` intends this to be quite so indirect; can you file a bug to improve this on the grinder side? We should at least update the docs on how to use task args. Thanks!

devoncarew:
Looks like some test failures - not sure if they're related to this PR.

tool/grind.dart
devoncarew: (docs updated here: https://github.com/google/grinder.dart/commit/07507d4dbe7162bd9d685062f24e7a659e783eb5)

devoncarew: And, you should be able to remove the explicit `GrinderContext` param to `rule()` above.

pq: Thanks!

Issue opened: https://github.com/google/grinder.dart/issues/317

PR #895, opened 2018-02-15
add rule use_null_aware_access

a14n:


zoechi:
Doesn't the analyzer produce a warning for this already?
I saw them freqiently for such code. 
Something like "... could be Null ..." (have to check - just on the phone now)
Perhaps it doesn't cover all cases?

zoechi:
image

a14n:
humm, strange, I never saw this warning. Do you have something special about your config ?

a14n:
My bad. I made my tests in a folder that was excluded from analysis :-/

Thanks to notice it.

I will extend the actual rule in sdk to cover all the cases I point out in the tests.

PR #897, opened 2018-02-16
New lint: avoid_relative_lib_imports (#896)

pq:
Fixes: #896

@bwilkerson


lib/src/rules/avoid_relative_lib_imports.dart
```diff
@@ -0,0 +1,76 @@
...
+
+...
+```
+
+''';
+
+class AvoidRelativeLibImports extends LintRule {
+  AvoidRelativeLibImports() : super(
+          name: 'avoid_relative_lib_imports',
+            description: _desc,
```
bwilkerson: Is this formatted?

pq: Whoops!  No.  Fixed now (but there should really be a presubmit!)

PR #900, opened 2018-02-16
Test to validate copyright headers.

pq:
@bwilkerson


pq:
> Should we just include the whole notice?

I thought about that but wanted it to be a quicker test.  (That may have been silly though.)

As for the SDK, it's an interesting point.  I wonder if that shouldn't be part of our presubmit?

PR #905, opened 2018-02-23
factor out const checks

a14n:


a14n:
> Longer term I would like to see these utilities added to the analyzer APIs

:+1: 

BTW I have not completly understood the difference between `ConstantVerifier` and `ConstantVisitor`. I'd tried to only use `ConstantVisitor` (that is public) but the tests were not ok.

bwilkerson:
The `ConstantVisitor` was an early attempt to do some minimal constant evaluation. It still works, but is incomplete (there are lots of expressions it can't evaluate). We kept it around for some uses cases that could safely use it, but I expect it will be removed when we clean up the analyzer API.

a14n:
ditto

Issue #910, opened 2018-03-01
`prefer_const_constructors` doesn't take into account implicit creation


devoncarew:
From https://github.com/flutter/flutter/issues/15023, It's likely that `prefer_const_constructors` doesn't take into account implicit `const` creation.

pq:
Thanks for the report!

I've got this reproduced in the latest Flutter SDK (using linter 0.1.43).  Looking now to see if it's been fixed by @a14n's recent work in the area...

pq:
Confirmed 👍 .

Using linter 0.1.43:

![image](https://user-images.githubusercontent.com/67586/36872977-f34b0b76-1d5b-11e8-9e7f-ab2d1fbc5e66.png)

Using head:

![image](https://user-images.githubusercontent.com/67586/36872996-09aa9972-1d5c-11e8-97ca-5e5cbce37894.png)

I'll look at publishing a new linter in the next few days to get this fix out for integration into the Flutter SDK.

devoncarew:
Awesome! Thanks for digging into it.

pq:
Fixed by @a14n  in 0.1.44 🎉

PR #911, opened 2018-03-01
Tests validating implicit const handling (#910).

pq:
See: #910.

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -1,5 +1,7 @@
 analyzer:
   strong-mode: true
+  language:
+    enablePreviewDart2: true
   exclude:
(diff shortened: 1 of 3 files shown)


pq:
cc @a14n

a14n:
Thanks

a14n:
Should we change the `environment/sdk` constraint in `pubspec.yaml` ? Not sure that we support all the versions.

a14n:
Linter depends on analyzer `^0.31.2-alpha.0` and analyzer-0.31.2-alpha.0 needs a sdk `>=2.0.0-dev` so it looks like we could update the `environment/sdk` of linter to have the same sdk constraint.

PR #918, opened 2018-03-12
Handle undefined classes in rule (sdk issue 32494)

bwilkerson:
The rule prefer_const_literals_to_create_immutables is throwing an exception when presented with invalid code involving an instance creation expression for which the class could not be resolved. This fixes that.


pq:
Nice!

PR #923, opened 2018-03-19
Add avoid_types_as_parameter_names

### LINKED ISSUE #921: Lint for uppercase typedef and function parameter names
What does the following define?

```dart
typedef String camelize(String);

void search(String haystack, callback(String)) {...}
```

When only one identifier is found, it is the _name_ of a parameter, not the _type_. The above is equivalent to:

```dart
typedef String camelize(dynamic String);

void search(String haystack, callback(dynamic String)) {...}
```

How about a lint that detects single-identifier parameters which start with a capital letter?

a14n:
Closes #921

### DIFF
--- a/lib/src/rules/avoid_types_as_parameter_names.dart
+++ b/lib/src/rules/avoid_types_as_parameter_names.dart
@@ -43,6 +43,8 @@ class Visitor extends SimpleAstVisitor {
 
   @override
   visitFormalParameterList(FormalParameterList node) {
+    // TODO(a14n) test that parameter name matches a existing type. No api to do
+    // that for now.
     for (final parameter in node.parameters) {
       if (parameter.element.hasImplicitType &&
           (parameter.identifier.name.startsWith(new RegExp('[A-Z]')) ||
--- a/lib/src/rules/avoid_types_as_parameter_names.dart
+++ b/lib/src/rules/avoid_types_as_parameter_names.dart
@@ -1,0 +1,55 @@
+// Copyright (c) 2018, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:linter/src/analyzer.dart';
+
+const _desc = r'Avoid types as parameter names.';
+
+const _details = r'''
+
+**AVOID** using parameter names that is the same as an existing type.
+
+**BAD:**
+```
+m(f(int));
+```
+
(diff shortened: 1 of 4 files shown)


lib/src/rules/avoid_types_as_parameter_names.dart
```diff
@@ -0,0 +1,55 @@
...
+
+  final LintRule rule;
+
+  @override
+  visitFormalParameterList(FormalParameterList node) {
+    for (final parameter in node.parameters) {
+      if (parameter.element.hasImplicitType &&
+          (parameter.identifier.name.startsWith(new RegExp('[A-Z]')) ||
+              ['num', 'int', 'double', 'bool', 'dynamic']
+                  .contains(parameter.identifier.name))) {
```
a14n: I haven't found how to check if the identifier name matches a type that is visible in scope. @bwilkerson how can I do that ?

test/rules/avoid_types_as_parameter_names.dart
```diff
@@ -0,0 +1,25 @@
...
+typedef void f4(
+  num a, // OK
+  {
+  int, // LINT
+});
+typedef void f5(
+  double a, // OK
+  [
+  bool, // LINT
+]);
```
bwilkerson: We should have at least one test that uses the generic function type syntax:
```
typedef f6 = int Function(int); // OK
```

lib/src/rules/avoid_types_as_parameter_names.dart
bwilkerson: I think this will work: get the `LibraryElement` for the containing library and use `publicNamespace` to get a `Namespace`. You can ask the namespace to look up the name and then test to see whether the returned element is a `TypeDefiningElement`.

a14n: Looks like `cu.element.library.publicNamespace.definedNames` contains only elements defined inside the library. For instance, `publicNamespace.get('int')` returns `null`.

bwilkerson: Oh, right. What you really want we don't expose as public API.

Internally I'd use `new LibraryScope(cu.element.library)` to get a name scope that includes all of the imported names. But that probably won't work in the future as we move analyzer on top of the front_end package, so I hesitate to suggest it. Maybe a comment about what we'd like to do here?

a14n: Comment added.

test/rules/avoid_types_as_parameter_names.dart
a14n: Done

srawlins:
Whoa what!?! Yes! 🎉 🎉 🎉

PR #927, opened 2018-03-19
Add avoid_private_typedef_functions

### LINKED ISSUE #884: Linter that (private) typedefs can be removed in favor of the new function syntax
The Function syntax introduced in Dart 1.24 can be used to replace many typedefs that exist only to spell out types in internal interfaces.

Right now, the only signal that a typedef could be replaced with an inline Function is that the typedef is _private 🙁  but that's a start! If the linter has package-level knowledge, it could provide the same rule for typedefs that are not exported into the "public API."

a14n:
Closes #884

### DIFF
--- a/lib/src/rules/avoid_private_typedef_functions.dart
+++ b/lib/src/rules/avoid_private_typedef_functions.dart
@@ -60,7 +60,14 @@ class Visitor extends SimpleAstVisitor {
 
   _countAndReport(String name, AstNode node) {
     final visitor = new _CountVisitor(name);
-    getCompilationUnit(node).root.visitChildren(visitor);
+    final units = getCompilationUnit(node)
+        .element
+        .library
+        .units
+        .map((cu) => cu.computeNode());
+    for (final unit in units) {
+      unit.visitChildren(visitor);
+    }
     if (visitor.count <= 1) {
       rule.reportLint(node);
     }
@@ -69,7 +76,7 @@ class Visitor extends SimpleAstVisitor {
 
 class _CountVisitor extends RecursiveAstVisitor {
   _CountVisitor(this.type);
-  String type;
+  final String type;
   int count = 0;
 
   @override
--- a/lib/src/rules/avoid_private_typedef_functions.dart
+++ b/lib/src/rules/avoid_private_typedef_functions.dart
@@ -1,0 +1,79 @@
+// Copyright (c) 2018, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:linter/src/analyzer.dart';
+import 'package:linter/src/ast.dart';
+
+const _desc = r'Avoid private typedef functions.';
+
+const _details = r'''
+
+**AVOID** private typedef functions used only once. Prefer inline function
+syntax.
+
+**BAD:**
+```
+typedef void _F();
+m(_F f);
+```
+
+**GOOD:**
+```
+m(void Function() f);
+```
+
+''';
+
+class AvoidPrivateTypedefFunctions extends LintRule {
+  AvoidPrivateTypedefFunctions()
+      : super(
+            name: 'avoid_private_typedef_functions',
+            description: _desc,
+            details: _details,
+            group: Group.style);
+
+  @override
+  AstVisitor getVisitor() => new Visitor(this);
+}
+
+class Visitor extends SimpleAstVisitor {
+  Visitor(this.rule);
+
+  final LintRule rule;
+
+  @override
+  visitFunctionTypeAlias(FunctionTypeAlias node) {
+    if (node.element.isPrivate) {
+      _countAndReport(node.name.name, node);
+    }
+  }
+
+  @override
+  visitGenericTypeAlias(GenericTypeAlias node) {
+    if (node.element.isPrivate) {
+      _countAndReport(node.name.name, node);
+    }
+  }
+
+  _countAndReport(String name, AstNode node) {
+    final visitor = new _CountVisitor(name);
+    getCompilationUnit(node).root.visitChildren(visitor);
+    if (visitor.count <= 1) {
+      rule.reportLint(node);
+    }
+  }
+}
+
+class _CountVisitor extends RecursiveAstVisitor {
+  _CountVisitor(this.type);
+  String type;
+  int count = 0;
+
+  @override
+  visitTypeName(TypeName node) {
+    if (node.name.name == type) count++;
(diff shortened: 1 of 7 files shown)


pq:
Great catch Brian.  I wonder if it's sufficient to add a caveat to the lint docs?

zoechi:
Nowadays parts are heavily used for all kinds of code generation (built_value, built_redux, json_serializable, ...) since build is destined to replace `pub serve` and transformers.

a14n:
PTAL it now handles parts

lib/src/rules/avoid_private_typedef_functions.dart
```diff
@@ -60,7 +60,14 @@ class Visitor extends SimpleAstVisitor {
 
   _countAndReport(String name, AstNode node) {
     final visitor = new _CountVisitor(name);
-    getCompilationUnit(node).root.visitChildren(visitor);
+    final units = getCompilationUnit(node)
+        .element
+        .library
+        .units
+        .map((cu) => cu.computeNode());
```
bwilkerson: I'm concerned about this. The method `computeNode` is generally quite expensive, so invoking it for every `typedef` seems likely to make this lint very slow. We should do some performance testing to determine whether that concern is valid.

a14n: https://travis-ci.org/dart-lang/linter/jobs/355961824#L556

pq: I was just going to link to that!  For posterity, here's what it tells us:

```
---------------------------------------------------
Timings                                          ms
---------------------------------------------------
invariant_booleans                              444
public_member_api_docs                          257
avoid_private_typedef_functions                 142
```

It'd be interesting to disable this check and do another run for comparison?

bwilkerson: What code base(s) does that run the lint rules over, and how many parts do those code bases make use of?

pq: Itself.  We should get more representative for sure.  Moreover we have no parts so not sure what this tells us. :(

a14n: I just ran it on flutter:

```
$ ./bin/cache/dart-sdk/bin/dart ~/dart/dart-lang/linter/bin/linter.dart --benchmark -q -c analysis_options_repo.yaml packages/flutter

---------------------------------------------------
Timings                                          ms
---------------------------------------------------
avoid_private_typedef_functions                1283
invariant_booleans                             1249
public_member_api_docs                          903
prefer_final_fields                             454
unnecessary_this                                236
package_api_docs                                208
avoid_renaming_method_parameters                206
cancel_subscriptions                            190
```

I think I can make an optim if there is only one unit to avoid `computeNode`. Stay tuned.

bwilkerson: A couple of additional optimizations:
- It could cache the list of units the first time they are computed and re-use it.
- It could avoid `computeNode` for the current compilation unit, given that it's the starting point for computing the list.

To be representative, the code base used for measurement also needs to have a representative number of `typedef` declarations.

a14n: humm, after my following optim:

```diff
--- a/lib/src/rules/avoid_private_typedef_functions.dart
+++ b/lib/src/rules/avoid_private_typedef_functions.dart
@@ -60,11 +60,12 @@ class Visitor extends SimpleAstVisitor {

   _countAndReport(String name, AstNode node) {
     final visitor = new _CountVisitor(name);
-    final units = getCompilationUnit(node)
-        .element
-        .library
-        .units
-        .map((cu) => cu.computeNode());
+    final cu = getCompilationUnit(node);
+    final unitElements = cu.element.library.units;
+    print(unitElements.length);
+    final units = unitElements.length == 1
+        ? [cu]
+        : unitElements.map((e) => e.computeNode());
     for (final unit in units) {
       unit.visitChildren(visitor);
     }
```

it looks like `computeNode` is not really a problem (I see the same order of timing).

I guess the time is spent by `_CountVisitor`.

WDYT ?

bwilkerson: I can't tell from the diff where the time is being spent. I'd recommend profiling the code (either in observatory or by using some stopwatches to capture how long each piece takes).

a14n: I added some stopwatches and they confirm the time is mainly spent in `_CountVisitor`.

```dart
  _countAndReport(String name, AstNode node) {
    final visitor = new _CountVisitor(name);
    final cu = getCompilationUnit(node);
    final unitElements = cu.element.library.units;
    print('--');
    print('nb of units: ${unitElements.length}');
    final sw1 = new Stopwatch()..start();
    final units = unitElements.length == 1
        ? [cu]
        : unitElements.map((e) => e.computeNode());
    print('create unit list: ${sw1.elapsedMicroseconds}');
    final sw2 = new Stopwatch()..start();
    for (final unit in units) {
      unit.visitChildren(visitor);
    }
    print('visit units: ${sw2.elapsedMicroseconds}');
    if (visitor.count <= 1) {
      rule.reportLint(node);
    }
  }
```

run on flutter:

```
$ ./bin/cache/dart-sdk/bin/dart ~/dart/dart-lang/linter/bin/linter.dart -q -c analysis_options_repo.yaml packages/flutter
--
nb of units: 1
create unit list: 22
visit units: 753
--
nb of units: 1
create unit list: 21
visit units: 423
--
nb of units: 1
create unit list: 1
visit units: 450
--
nb of units: 17
create unit list: 8
visit units: 593106
--
nb of units: 1
create unit list: 1
visit units: 29
--
nb of units: 1
create unit list: 1
visit units: 95
--
nb of units: 1
create unit list: 22
visit units: 501
--
nb of units: 17
create unit list: 1
visit units: 640530
817 files analyzed, 141183 issues found, in 19555 ms.
```

So I think we can go ahead with the current version in PR.

bwilkerson: Wow! I'm surprised that `computeNode` didn't take more time, but I'm also curious to know why it took an average of ~35000 ms / file to analyze the code in the libraries with 17 units.

But I can sign off on using `computeNode` :-).

a14n: >why it took an average of ~35000 ms / file

microseconds not milliseconds ;)

a14n:
Reminder: I don't have merge access :)

pq:
> Reminder: I don't have merge access :)

That happens at the organization level and I don't have control there but I'm happy to vouch for you!  (And merge manually in the meantime. 👍 )

lib/src/rules/avoid_private_typedef_functions.dart
pq: This is great.  Thanks for the follow-ups!

PR #930, opened 2018-03-19
Tender-loving-care: Travis Edition

matanlurey:
Quality of life updates for travis here:
* Use a pub cache to make "pub get" a bit faster.
* Only run travis on [master], to avoid 2 builds per PR


pq:
Fantastic.  Thanks!

PR #935, opened 2018-03-21
new lint avoid_single_cascade_in_expression_statements

a14n:


lib/src/rules/avoid_single_cascade_in_expression_statements.dart
```diff
@@ -0,0 +1,50 @@
...
+}
+
+class Visitor extends SimpleAstVisitor {
+  Visitor(this.rule);
+
+  final LintRule rule;
+
+  @override
+  visitCascadeExpression(CascadeExpression node) {
+    if (node.cascadeSections.length == 1 && node.parent is ExpressionStatement) {
```
bwilkerson: I'm not sure why this is limited to the statement level. Isn't it just as bad to have `f(o..m())`? What's the use case for limiting this?

a14n: `f(o..m())` will call `f` with `o` but `f(o.m())` will call `f` with the result of `o.m()`.

bwilkerson: Oh, of course. I should have seen that.

a14n:
CI fixed. It can now be merged.

a14n:
Thanks

PR #936, opened 2018-03-22
add avoid_keyword_to_create_instances

a14n:
Fixes dart-lang/sdk#32602

### DIFF
--- a/test/_data/avoid_keyword_to_create_instances/a.dart
+++ b/test/_data/avoid_keyword_to_create_instances/a.dart
@@ -34,4 +34,8 @@ main() {
   const v6 = const A([]); // LINT
   const v7 = A([]); // OK
   final v8 = A(const []); // OK
+  final v9 = const A(const []); // LINT
+  final v10 = const A([]); // OK
+  final v11 = const A(const {}); // LINT
+  final v12 = const A({}); // OK
 }
--- a/test/_data/avoid_keyword_to_create_instances/a.dart
+++ b/test/_data/avoid_keyword_to_create_instances/a.dart
@@ -1,0 +1,37 @@
+// Copyright (c) 2018, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// test w/ `pub run test -N avoid_keyword_to_create_instances`
+
+class A {
+  const A([o]);
+  A.c1();
+}
+
+main() {
+  const A(); // OK
+  new A(); // LINT
+  A(); // OK
+
+  new A.c1(); // LINT
+  A.c1(); // OK
+
+  const A([]); // OK
+  new A([]); // LINT
+  A([]); // OK
+  A(const []); // OK
+
+  const A(A()); // OK
+  A(const A()); // OK
+  A(A()); // OK
+
+  final v1 = A(); // OK
+  final v2 = new A(); // LINT
+  const v3 = const A(); // LINT
+  const v4 = A(); // OK
+  final v5 = const A([]); // OK
+  const v6 = const A([]); // LINT
+  const v7 = A([]); // OK
+  final v8 = A(const []); // OK
+}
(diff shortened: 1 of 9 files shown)


a14n:
Not mergeable until analyzer is upgrade.

bwilkerson:
The tests will need to be updated to reflect the new semantics, but the implementation should not need to change.

a14n:
PTAL

test/_data/avoid_keyword_to_create_instances/a.dart
```diff
@@ -0,0 +1,38 @@
...
+  A(); // OK
+
+  new A.c1(); // LINT
+  A.c1(); // OK
+
+  const A([]); // OK
+  new A([]); // LINT
+  A([]); // OK
+  A(const []); // OK
+  const A([]); // OK
```
bwilkerson: This is a duplicate of line 20.

a14n: removed. Thanks!

bwilkerson:
Based on discussions elsewhere (let me know if you haven't seen them and I'll try to get you a pointer to them) I'm wondering whether we might want to split this into two lints: one to advise removing `const` inside a constant context and one to advise removing `new` where the constructor being invoked is not `const`.

a14n:
>I'm wondering whether we might want to split this into two lints

I can do that but I don't think people will use only one of them.

bwilkerson:
You might be right. We just don't have enough experience using the feature to know, at this point, what people are going to want.

eernstg:
I think @bwilkerson was referring to [this comment](https://github.com/dart-lang/sdk/issues/32798#issuecomment-380043155), about reasons for allowing `new` in some situations.

bwilkerson:
I was, yes. (I just didn't have a convenient way to find it at the time I wrote the comment.)

a14n:
This lint was starting with tests that conformed to _magic const_ . As you can see in https://github.com/dart-lang/linter/pull/936/commits/3962014698ad3cb76634bebaf2683451567dedad#diff-d85039c777133b1ebb3ecb9c9366210fR14 (first version of this lint with magice const) `new A()` was not linted because it is not the same with or without `new`. Actually only the analyzer has changed to remove support of magic const and if/once magic const comes back the lint should magically handle that ( it's all about magic after all :D )

a14n:
PTAL

a14n:
Reminder: I can't merge :)

bwilkerson:
I actually did remember, this time. I was waiting to see whether @eernstg has any additional thoughts.

bwilkerson:
I'm guessing he didn't. Should I merge this now, or are there other changes you wanted to make first?

a14n:
I think it's ok. Trying this lint on flutter I faced flutter/flutter#17429 but it looks like a bug in analysis server. Wdyt?

bwilkerson:
@devoncarew @pq Should we wait to land this until we've found and fixed the referenced issue so that we'll know whether they're the same or different issues?

pq:
I don't feel strongly, @devoncarew ?

devoncarew:
We should figure out whether this will break the next flutter roll; I asked a follow up there but have not investigated the failure.

devoncarew:
We can merge no? We just can't enable this lint on the flutter repo until the other issues are address.

bwilkerson:
Yes, we can merge. But there's no way for users to know that the lint shouldn't be used until other issues have been fixed.

pq:
Sorry to have let this slip.  @a14n: are we still wanting to land this as is?  Any updates?

a14n:
Removing `new` causes issues with `prefer_const_literals_to_create_immutables` and `prefer_const_constructors` so it could be worth to split this lint into 2 lints (one for remonving `new` and an other for removing `const`). Thus we could apply the removing of `const` without waiting the fix on ConstantVerifier. I don't know when @bwilkerson plans to look at a fix for ConstantVerifier.

pq:
OK, cool.  Thanks @a14n.  That's what I thought.  @bwilkerson is on vacation; let's regroup when he's back. 👍

bwilkerson:
I don't know when I will be able to look at that issue. Of course, anyone else is also welcome to look at it. :-)

pq:
@a14n: this looks great.  Thanks for the follow-through!

PR #943, opened 2018-03-26
add avoid_function_type_aliases

a14n:


lib/src/rules/avoid_function_type_aliases.dart
```diff
@@ -0,0 +1,48 @@
...
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:linter/src/analyzer.dart';
+
+const _desc = r'Avoid function type aliases.';
+
+const _details = r'''
+
+**AVOID** function type aliases.
```
bwilkerson: I think we should have some statement of the motivation for this rule. Why should users avoid it?

a14n: It should perhaps say _prefer xxx_.

I don't exactly know how to say that but reading https://github.com/dart-lang/sdk/blob/master/docs/newsletter/20170728.md#function-types and https://github.com/dart-lang/sdk/blob/master/docs/language/informal/generic-function-type-alias.md give the feeling that it's better to use generic function type alias.

> The reason for having two ways to express the same thing is that the new form seamlessly covers non-generic functions as well as generic ones, and developers might prefer to use the new form everywhere, for improved readability.

and 

> While a possibility, we haven't yet decided if we want to deprecate the old syntax.

Any suggestion to summarize that ?

bwilkerson: > It should perhaps say prefer xxx.

I think that's more accurate, yes.

> The reason for having two ways to express the same thing is that the new form seamlessly covers non-generic functions as well as generic ones ...

With all due respect to Eric, that's not my take on it.

I think the reason there are two ways to express the same thing is that we decided not to deprecate the old syntax at this point in time. The reason there's a _new_ syntax is because, with the introduction of generic functions, the old couldn't express all of the possible kinds of parameterization that users might want to express.

> ... developers might prefer to use the new form everywhere, for improved readability.

That's true. Developers _might_ want to do so. What I don't know is whether developers _actually_ want to do so.

But just to be clear, I'm not objecting to adding the lint. I'm just suggesting that we should explain to a user why they might want to enable the lint; I think all of our lints should have an informative statement of what the lint does and what benefit this brings users.

a14n: > I think all of our lints should have an informative statement of what the lint does and what benefit this brings users.

Totally agree.

a14n: @munificent should something be added to the Dart Style guide to prefer _generic function type aliases_ over _function type aliases_ syntax ?

munificent: I'll need to see what the rest of the community things, but my personal opinion is that, yes, we should have a guideline that pushes everyone to exclusively use the new syntax. The new syntax is strictly more expressive than the old and the style guide is heavily in favor of consistency — in places where there are two ways to accomplish the same thing, the style guide picks one and tells you to always use it.

As Dart evolves and adds better notation for things already supported, we are increasingly going to need rules like this that tell people *not* to use certain language features.

PR #946, opened 2018-03-27
add prefer_iterable_whereType

a14n:


lib/src/rules/prefer_iterable_whereType.dart
```diff
@@ -0,0 +1,81 @@
...
+        } else {
+          return;
+        }
+      } else if (body is ExpressionFunctionBody) {
+        expression = body.expression;
+      } else {
+        return;
+      }
+    }
+    expression = expression.unParenthesized;
```
bwilkerson: It the argument is not a `FunctionExpression` (such as a tear-off), then expression will be `null` at this point. Perhaps add a `return` above and a test case.

lib/src/rules/prefer_iterable_whereType.dart
```diff
@@ -0,0 +1,81 @@
...
+  final LintRule rule;
+
+  @override
+  visitMethodInvocation(MethodInvocation node) {
+    if (node.methodName.name != 'where') return;
+    if (!DartTypeUtilities.implementsInterface(
+        node.target.bestType, 'Iterable', 'dart.core')) {
+      return;
+    }
+    final arg = node.argumentList.arguments?.first;
```
bwilkerson: We should test that there is at least one argument. Lints can be run over invalid code.

lib/src/rules/prefer_iterable_whereType.dart
```diff
@@ -0,0 +1,81 @@
...
+    String param;
+    Expression expression;
+    if (arg is FunctionExpression) {
+      param = arg.parameters.parameters.first.identifier.name;
```
bwilkerson: We should check that there is at least one parameter.

lib/src/rules/prefer_iterable_whereType.dart
a14n: Actually I put this statement in the wrong block. Thanks for catching it!

lib/src/rules/prefer_iterable_whereType.dart
a14n: done

lib/src/rules/prefer_iterable_whereType.dart
a14n: done

a14n:
PTAL

lib/src/rules/prefer_iterable_whereType.dart
```diff
@@ -0,0 +1,86 @@
...
+  @override
+  visitMethodInvocation(MethodInvocation node) {
+    if (node.methodName.name != 'where') return;
+    if (!DartTypeUtilities.implementsInterface(
+        node.target.bestType, 'Iterable', 'dart.core')) {
+      return;
+    }
+
+    final args = node.argumentList?.arguments;
+    if (args.length != 1) return;
```
bwilkerson: `args` can be `null` here.

lib/src/rules/prefer_iterable_whereType.dart
```diff
@@ -0,0 +1,86 @@
...
+      if (arg.parameters.parameters.length != 1) return;
+
+      final body = arg.body;
+      Expression expression;
+      if (body is BlockFunctionBody) {
+        final statements = body.block.statements;
+        if (statements.length != 1) return;
+        final statement = body.block.statements.first;
+        if (statement is ReturnStatement) {
+          expression = statement.expression;
```
bwilkerson: Sorry I missed this before, but `expression` can be `null` at this point.

It might be better to drop the `else { return; }` clauses and use `expression?.unParenthesized` on line 76.

lib/src/rules/prefer_iterable_whereType.dart
a14n: oops. Isn't there a lint to show that ? :laughing:

PR #954, opened 2018-03-30
fix issue with avoid_types_as_parameter_names

a14n:
Generic function types were not handled correctly


a14n:
There're already https://github.com/a14n/linter/blob/7e3c8e40f0227e0d9475b6d007acfb468eaf5741/test/rules/avoid_types_as_parameter_names.dart#L9 and https://github.com/a14n/linter/blob/7e3c8e40f0227e0d9475b6d007acfb468eaf5741/test/rules/avoid_types_as_parameter_names.dart#L20

Do you have other cases in mind ?

PR #957, opened 2018-03-30
Add avoid_large_ints

### LINKED ISSUE #955: dart2js lint when integer literals will have rounding errors
Add a lint to indicate that​ certain integer constants are not precisely represented when compiled to JS. For example, 1000000000000000001 is a 59-bit odd number, so is rounded to 1000000000000000000 fit in the effective-53-bits of the double.

a14n:
Fixes #955

### DIFF
--- a/example/all.yaml
+++ b/example/all.yaml
@@ -17,6 +17,7 @@ linter:
     - avoid_empty_else
     - avoid_function_literals_in_foreach_calls
     - avoid_init_to_null
+    - avoid_large_ints
     - avoid_null_checks_in_equality_operators
     - avoid_positional_boolean_parameters
     - avoid_private_typedef_functions
--- a/lib/src/rules.dart
+++ b/lib/src/rules.dart
@@ -18,6 +18,7 @@ import 'package:linter/src/rules/avoid_classes_with_only_static_members.dart';
 import 'package:linter/src/rules/avoid_empty_else.dart';
 import 'package:linter/src/rules/avoid_function_literals_in_foreach_calls.dart';
 import 'package:linter/src/rules/avoid_init_to_null.dart';
+import 'package:linter/src/rules/avoid_large_ints.dart';
 import 'package:linter/src/rules/avoid_null_checks_in_equality_operators.dart';
 import 'package:linter/src/rules/avoid_positional_boolean_parameters.dart';
 import 'package:linter/src/rules/avoid_private_typedef_functions.dart';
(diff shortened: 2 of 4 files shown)


a14n:
/cc @rakudrama for comments on 526c4ccce8bc0489e9aeb70c56b4b475dcada087

From https://github.com/a14n/linter/commit/526c4ccce8bc0489e9aeb70c56b4b475dcada087#r28354339

> Values that are outside this range but are represented exactly as IEEE 754 values should be allowed.
For example, 1,000,000,000,000,000,000 should be allowed, but 1,000,000,000,000,000,001 should not.
> The problem with being too strict is that it will force developers to construct the large values via arithmetic which could introduce rounding errors in the intermediate values.

I can't see the benefits of allowing ints without rounded issue (like `1,000,000,000,000,000,000`). It's likely that such a value will be used in operations and then there will be strange behaviours. For instance:

```dart
int a = 1000000000000000000; // OK
int b = 1000000000000000001; // LINT

a + 1 == a; // WAT ?
```

Do you have a use case where those not-rounded large ints are used without problem?

rakudrama:
@a14n I want to reduce false positives, warnings where there is not a real problem.
False positives are a burden on the developer.
We are 100% sure that `1000000000000000001` is a problem - the developer does not have the program they wrote, and they have a clear place to fix it.
We are _not_ 100% sure that `1000000000000000000` is a problem. It might be, but it might not.

- Large values are used in tests as dummy data that is round-tripped. This use is fine.
- Large values can be used as validation limits (e.g. compared against) without incurring rounding.
- The arithmetic might not be `a + 1`, it could be e.g. nanoseconds and used in calculations like `a / 1000000000` to print seconds.

I think the idea of a 'safe integer' is terribly named concept. Any value might participate in a calculation that has rounding. `a = 1,000,000,001 `is clearly inside the 'safe-integer' range, but `a * a` is rounded.
So should we warn about `1,000,000,001 `too?

Every month I answer a bug report like  *new Int64(1000000000000000001).toString() prints the wrong answer*. I'd like to save them the effort of filing the bug and educate them about the problem without rejecting any correct programs.

a14n:
>We are 100% sure that 1000000000000000001 is a problem.

I was wondering if this is also true for compile time constants (one could arguments that the computation could be done before emitting JS code). The following test gives the answer:

```dart
const a = 1000000000000000001 % 10;
void main() {
  print(a); // 0 on dartpad
}
```

Ok, I will close this PR in favor of a new lint.

PR #959, opened 2018-04-03
add avoid_js_rounded_ints

### LINKED ISSUE #955: dart2js lint when integer literals will have rounding errors

a14n:
Follow up on #957.

Fixes #955

### DIFF
--- a/lib/src/rules/avoid_js_rounded_ints.dart
+++ b/lib/src/rules/avoid_js_rounded_ints.dart
@@ -10,7 +10,7 @@ const _desc = r'Avoid JavaScript rounded ints.';
 
 const _details = r'''
 
-**AVOID** integers that will be rounded.
+**AVOID** integer literals that will be rounded.
 
 When a program is compiled to JavaScript `int` and `double` become JavaScript
 Numbers. Too large integers (`value < Number.MIN_SAFE_INTEGER` or
--- a/lib/src/rules/avoid_js_rounded_ints.dart
+++ b/lib/src/rules/avoid_js_rounded_ints.dart
@@ -10,11 +10,15 @@ const _desc = r'Avoid JavaScript rounded ints.';
 
 const _details = r'''
 
-**AVOID** integer literals that will be rounded.
+**AVOID** integer literals that cannot be represented exactly when compiled to
+JavaScript.
 
 When a program is compiled to JavaScript `int` and `double` become JavaScript
 Numbers. Too large integers (`value < Number.MIN_SAFE_INTEGER` or
 `value > Number.MAX_SAFE_INTEGER`) may be rounded to the closest Number value.
+
+For instance `1000000000000000001` cannot be represented exactly as a JavaScript
+Number, so `1000000000000000000` will be used instead.
 
 **BAD:**
 ```
@@ -52,12 +56,5 @@ class Visitor extends SimpleAstVisitor {
     }
   }
 
-  bool isRounded(int value) {
-    int v = value.abs();
-    while (v > 9007199254740991) {
-      if (v.isOdd) return true;
-      v = v ~/ 2;
-    }
-    return false;
-  }
+  bool isRounded(int value) => value?.toDouble()?.toInt() != value;
 }
--- a/lib/src/rules/avoid_js_rounded_ints.dart
+++ b/lib/src/rules/avoid_js_rounded_ints.dart
@@ -1,0 +1,63 @@
+// Copyright (c) 2018, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:linter/src/analyzer.dart';
+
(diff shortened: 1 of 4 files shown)


a14n:
cc @rakudrama

lib/src/rules/avoid_js_rounded_ints.dart
```diff
@@ -0,0 +1,63 @@
...
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:linter/src/analyzer.dart';
+
+const _desc = r'Avoid JavaScript rounded ints.';
+
+const _details = r'''
+
+**AVOID** integers that will be rounded.
```
bwilkerson: Perhaps "integers" --> "integer literals"

a14n: done

rakudrama: This is more precise: **AVOID** integer literals that cannot be represented exactly when compiled to JavaScript.

lib/src/rules/avoid_js_rounded_ints.dart
```diff
@@ -0,0 +1,63 @@
...
+
+When a program is compiled to JavaScript `int` and `double` become JavaScript
+Numbers. Too large integers (`value < Number.MIN_SAFE_INTEGER` or
+`value > Number.MAX_SAFE_INTEGER`) may be rounded to the closest Number value.
```
rakudrama: Is it possible to put actual values in the message? It would make the message more concrete.
If you could say something like the following, that would make the situation clearer:

1000000000000000001 cannot be represented exactly as a JavaScript Number, so 1000000000000000000 will be used instead.

rakudrama: Note that the nearest representable number is `value.toDouble()`, but this might be a value that cannot be represented as an `int` (when 0x7ff....fff is rounded up to 0x800....000), but you could print it with
`new BigInt.from(value.toDouble()).toString()`

test/rules/avoid_js_rounded_ints.dart
```diff
@@ -0,0 +1,15 @@
...
+
+final i = 1; // OK
+final min = -9007199254740991; // OK
+final max = 9007199254740991; // OK
+
+final minErr = -9007199254740993; // LINT
+final maxErr = 9007199254740993; // LINT
+
+final notRounded = 1000000000000000000; // OK
+final rounded = 1000000000000000001; // LINT
```
rakudrama: Also test with -9223372036854775808
value.abs() for this number is negative on the 64-bit integer VM.
Lucky it is not rounded! (-2^63), so maybe comment that the code works for this edge case.

lib/src/rules/avoid_js_rounded_ints.dart
a14n: done

lib/src/rules/avoid_js_rounded_ints.dart
a14n: >Is it possible to put actual values in the message?

It's not something that linter currently does. It would need some changes if we want to do that. 
@pq @bwilkerson what do you think about customizing messages depending on the code linted ?

test/rules/avoid_js_rounded_ints.dart
a14n: done

@bwilkerson `9223372036854775808` has `node.value == null`. Is it a bug ?

lib/src/rules/avoid_js_rounded_ints.dart
bwilkerson: Actually, I think it's possible to do that today, it's just that nobody has done it so there's no pattern to follow. I'd be fine with merging this change and then refining it, or you can make this change before merging if you'd like.

The trick is that you are not restricted to using `reportLint` to report a lint; I believe you have access to the error reporter that `reportLint` uses so you can create your own LintCode with a message that can be customized by using the substitution mechanism supported by errors. For example, the message might look something like `"The value '{0}' will be rounded to {1} when compiled to JavaScript."`.

bwilkerson:
@rakudrama Does this look good to you?

rakudrama:
Looks good to me.  I would still like to see the user's values used in the message to make the problem concrete (for example, they could replace the original number with the rounded one in mock data, rather than guessing what the rounded result is). I'm ok with that as a followup change.

test/rules/avoid_js_rounded_ints.dart
bwilkerson: It sounds like a bug. Every integer literal that is a valid 64-bit integer ought to have a non-null value.

a14n:
>Did you want to update the error message before we merge this?

I can but the bug I mentioned about `node.value == null` will cause issue.

bwilkerson:
Good point. I'll merge it now and we can update the message after we've fixed that bug. Is it line 15 that illustrates the problem?

a14n:
problem occurs on line 22

>final absNegative = -9223372036854775808;

PR #960, opened 2018-04-03
add avoid_double_and_int_checks

### LINKED ISSUE #956: dart2js lint when using type tests with numeric types
When compiled to JS, integer values are represented as floats. That can lead to some unexpected behavior when using either `is` or `is!` where the type is either `int` or `double`.

There would probably be too many false positives if we flagged every occurrence of these tests, but one case we should definitely catch is where a value is compared first to `double` and then to `int`, such as in:
```
f(num x) {
  if (x is double) {
    ...
  } else if (x is int) {
    ...
  }
}
```

a14n:
Closes #956

### DIFF
--- a/lib/src/rules/avoid_double_and_int_checks.dart
+++ b/lib/src/rules/avoid_double_and_int_checks.dart
@@ -1,0 +1,71 @@
+// Copyright (c) 2018, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:linter/src/analyzer.dart';
+import 'package:linter/src/ast.dart';
+
+const _desc = r'Avoid double and int checks.';
+
+const _details = r'''
+
+**AVOID** to check if type is double or int.
+
+When compiled to JS, integer values are represented as floats. That can lead to
+some unexpected behavior when using either `is` or `is!` where the type is
+either `int` or `double`.
+
+**BAD:**
+```
+f(num x) {
+  if (x is double) {
+    ...
+  } else if (x is int) {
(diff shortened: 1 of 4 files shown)


lib/src/rules/avoid_double_and_int_checks.dart
```diff
@@ -0,0 +1,71 @@
...
+        final typeProvider =
+            getCompilationUnit(node).element.context.typeProvider;
+        final ifExpression = ifCondition.expression;
+        final elseIsExpression = elseCondition.expression;
+        if (ifExpression is SimpleIdentifier &&
+            elseIsExpression is SimpleIdentifier &&
+            ifExpression.name == elseIsExpression.name &&
+            ifCondition.type.type == typeProvider.doubleType &&
+            elseCondition.type.type == typeProvider.intType) {
+          rule.reportLint(node);
```
bwilkerson: The whole `if` statement might cover a lot of code. I think it would be better to report this against the `elseCondition`.

We also might want to provide more details in the error message. Perhaps something like "Avoid checking for the type 'int' after testing for the type 'double' because both are represented as JS Numbers."

lib/src/rules/avoid_double_and_int_checks.dart
```diff
@@ -0,0 +1,71 @@
...
+      final ifCondition = node.condition;
+      final elseCondition = elseStatement.condition;
+      if (ifCondition is IsExpression && elseCondition is IsExpression) {
+        final typeProvider =
+            getCompilationUnit(node).element.context.typeProvider;
+        final ifExpression = ifCondition.expression;
+        final elseIsExpression = elseCondition.expression;
+        if (ifExpression is SimpleIdentifier &&
+            elseIsExpression is SimpleIdentifier &&
+            ifExpression.name == elseIsExpression.name &&
```
bwilkerson: We need to ensure that the identifier is a local variable. If it's the name of a getter, then there's no assurance that the returned value won't change from one invocation to the next.

PR #966, opened 2018-04-09
add avoid_field_initializers_in_const_classes

a14n:
Lint created from https://github.com/dart-lang/sdk/issues/32789#issuecomment-379657135:

>Also, you should never have a final field with an initializer expression in a const class. Instead of `final x = const expr;`, you should write `get x => const expr;` and not allocate a useless field.

### DIFF
--- a/lib/src/rules/avoid_field_initializers_in_const_classes.dart
+++ b/lib/src/rules/avoid_field_initializers_in_const_classes.dart
@@ -1,0 +1,111 @@
+// Copyright (c) 2018, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:linter/src/analyzer.dart';
+
+const _desc = r'Avoid field initializers in const classes.';
+
+const _details = r'''
+
+**AVOID** field initializers in const classes.
+
+Instead of `final x = const expr;`, you should write `get x => const expr;` and
+not allocate a useless field.
+
+**BAD:**
+```
+class A {
+  final a = const [];
+  const A();
+}
+```
+
+**GOOD:**
+```
+class A {
+  get a => const [];
+  const A();
+}
+```
+
+''';
+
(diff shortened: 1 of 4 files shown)


matanlurey:
FWIW, this isn't true. At least when compiled to JS, a final field is much better than a getter.

a14n:
@matanlurey thanks for the information.

So I'm not sure this lint is worth it. @lrhn @bwilkerson WDYT?

lib/src/rules/avoid_field_initializers_in_const_classes.dart
```diff
@@ -0,0 +1,111 @@
...
+        .members
+        .where((e) => e is ConstructorDeclaration)
+        .length;
+    if (constructorCount > 1) return;
+
+    final visitor = new HasParameterReferenceVisitor(constructor
+        .parameters.parameters
+        .map((e) => e.identifier.name)
+        .toList());
+    visitor.visitConstructorFieldInitializer(node);
```
bwilkerson: I think we minimally need to visit `node.expression`. Try
```
class F {
  final a;
  const F(int a) : this.a = 0;
}
```
But that might not be enough. What about
```
class G {
  final g;
  const G(int length) : g = 'xyzzy'.length;
}
```
Name matching is dangerous. :-)

lib/src/rules/avoid_field_initializers_in_const_classes.dart
```diff
@@ -0,0 +1,111 @@
...
+import 'package:linter/src/analyzer.dart';
+
+const _desc = r'Avoid field initializers in const classes.';
+
+const _details = r'''
+
+**AVOID** field initializers in const classes.
+
+Instead of `final x = const expr;`, you should write `get x => const expr;` and
+not allocate a useless field.
```
bwilkerson: Perhaps point out that this is true for the VM, but not for code that will be compiled to JS.

lrhn:
In JavaScript, a getter in the prototype may still be more *space efficient* than a final field on the object itself, ... but a proper optimizing compiler should recognize that there is no distinction and just put a frozen field in the prototype.

In practice we should probably just allow either, and let the compiler figure out that you don't need on-object space to store a constant.

matanlurey:
@lhrn: Sure. Until we do that though it's premature to recommend this pattern. It actively makes JS programs perform worse.

PR #970, opened 2018-04-13
Improve avoid_double_and_int_checks doc example

camsteffen:
I've tweaked the example to add a GOOD case with `x is num` as an alternative. This could be a helpful suggestion in some cases.

### DIFF
--- a/lib/src/rules/avoid_double_and_int_checks.dart
+++ b/lib/src/rules/avoid_double_and_int_checks.dart
@@ -20,10 +20,23 @@ either `int` or `double`.
 
 **BAD:**
 ```
-f(num x) {
+f(dynamic x) {
   if (x is double) {
     ...
   } else if (x is int) {
+    ...
+  } else {
+    ...
+  }
+}
+```
+
+**GOOD:**
+```
+f(dynamic x) {
+  if (x is num) {
+    ...
+  } else {
     ...
   }
 }
--- a/lib/src/rules/avoid_double_and_int_checks.dart
+++ b/lib/src/rules/avoid_double_and_int_checks.dart
@@ -20,12 +20,10 @@ either `int` or `double`.
 
 **BAD:**
 ```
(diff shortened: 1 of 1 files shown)


lib/src/rules/avoid_double_and_int_checks.dart
```diff
@@ -20,11 +20,24 @@ either `int` or `double`.
 
 **BAD:**
 ```
-f(num x) {
+f(dynamic x) {
```
bwilkerson: I think it would be better to leave this as `num`; it's one of the explicit use cases cited in the request for this rule.

camsteffen: I did this to make the GOOD and BAD examples have the same function signature. I don't think it's important to match a specific use case. But alternatively we can leave the BAD example alone and just add the GOOD example.

bwilkerson: > But alternatively we can leave the BAD example alone and just add the GOOD example.

Yeah, I'd prefer that. I think it _is_ important to make the BAD examples as close as possible to the code pattern that actually causes users problems. I don't want someone to not enable the rule because they look at the BAD example and think "I'd never do that, I'd give `x` a type of `num`, so I probably don't need this rule."

> I did this to make the GOOD and BAD examples have the same function signature.

That might well be why the author didn't originally include a GOOD example. It's hard to see how to make them parallel without making the change you made. But I think that it's OK to have examples that are not strictly parallel when there is no straight-forward one-size-fits-all fix.

camsteffen: > I don't want someone to not enable the rule because they look at the BAD example and think "I'd never do that, I'd give x a type of num, so I probably don't need this rule."

That's a very good point. I'll change back the GOOD example.

bwilkerson:
Thanks!

PR #981, opened 2018-04-25
Fail benchmarking on errors.

pq:
Update the build to fail on non-zero exit codes.

NOTE: it's still not properly failing since logged errors in `_computeLints` are not causing a non-zero exit.

@bwilkerson

### DIFF
--- a/tool/travis.sh
+++ b/tool/travis.sh
@@ -17,9 +17,6 @@ echo ""
 if [ "$LINTER_BOT" = "benchmark" ]; then
   echo "Running the linter benchmark..."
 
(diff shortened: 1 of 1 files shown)


pq:
@bwilkerson: I notice that the exception handler in `_computeLints` does not change the exit code.  Should it?

Take a look at the log spam here:

https://travis-ci.org/dart-lang/linter/jobs/371280018

With this change I would expect (and hope) for the benchmark build to fail but for that to happen we'd need the exit code to be non-zero.

Alternatively, I could update the script to scrape the output and assert it doesn't have any logged exceptions?

pq:
Discarding in favor of a fresh take.

PR #984, opened 2018-04-26
Fix benchmarking to fail on errors logged in analysis.

pq:
Fixes benchmarking to properly fail if an error is logged during analysis.

And now things reliably fail when they should.

image

👍 👍 👍

### DIFF
--- a/lib/src/analyzer.dart
+++ b/lib/src/analyzer.dart
@@ -2,10 +2,20 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:async';
+import 'dart:convert';
+import 'dart:io';
+import 'dart:math' as math;
+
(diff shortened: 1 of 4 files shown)


lib/src/analyzer.dart
```diff
@@ -52,3 +64,97 @@ class Analyzer {
...
   }
 }
+
+Future<Iterable<AnalysisErrorInfo>> lintFiles(
+    DartLinter linter, List<File> filesToLint) {
+  // Setup an error watcher to track whether an error was logged to stderr so
+  // we can set the exit code accordingly.
+  ErrorWatchingSink errorWatcher = new ErrorWatchingSink(errorSink);
+  errorSink = errorWatcher;
+  return linter.lintFiles(filesToLint).then((errors) {
```
bwilkerson: Maybe convert to async?

PR #985, opened 2018-04-27
Change most of the lint rules to implement NodeLintRule and node processors.

scheglov:
See [the corresponding analyzer change](https://dart-review.googlesource.com/c/sdk/+/52644) for numbers of performance improvements.

Note, that `_Processors` extend `SimpleAstVisitor`, but we do not pass it anywhere as a visitor. Instead we pass references to separate methods.  But by extending and specifying `@override` we get consistency between visitors and processors.

### DIFF
--- a/lib/src/analyzer.dart
+++ b/lib/src/analyzer.dart
@@ -17,7 +17,15 @@ export 'package:analyzer/src/generated/resolver.dart'
     show InheritanceManager, TypeProvider, TypeSystem;
 export 'package:analyzer/src/generated/source.dart' show LineInfo, Source;
 export 'package:analyzer/src/lint/linter.dart'
-    show DartLinter, LintRule, Group, Maturity, LinterOptions, LintFilter;
+    show
+        DartLinter,
+        LintRule,
+        Group,
+        Maturity,
+        LinterOptions,
+        LintFilter,
+        NodeLintRegistry,
+        NodeLintRule;
 export 'package:analyzer/src/lint/project.dart'
     show DartProject, ProjectVisitor;
 export 'package:analyzer/src/lint/pub.dart' show PubspecVisitor, PSEntry;
--- a/lib/src/rules/avoid_double_and_int_checks.dart
+++ b/lib/src/rules/avoid_double_and_int_checks.dart
@@ -42,7 +42,7 @@ f(dynamic x) {
 
 ''';
 
-class AvoidDoubleAndIntChecks extends LintRule {
+class AvoidDoubleAndIntChecks extends LintRule implements NodeLintRule {
   AvoidDoubleAndIntChecks()
       : super(
             name: 'avoid_double_and_int_checks',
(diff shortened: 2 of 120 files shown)


pubspec.yaml
```diff
@@ -20,3 +20,8 @@ dev_dependencies:
   matcher: ^0.12.0
   path: '>=0.9.0 <2.0.0'
   test: ^0.12.24+1
+dependency_overrides:
```
bwilkerson: I don't think we want to use overrides. We'll need to publish a new version of analyzer and update the dependency.

lib/src/analyzer.dart
```diff
@@ -17,7 +17,16 @@ export 'package:analyzer/src/generated/resolver.dart'
...
+    show
+        DartLinter,
+        LintRule,
+        Group,
+        Maturity,
+        LinterOptions,
+        LintFilter,
+        NodeLintRegistry,
+        NodeLintRule,
+        UnitLintRule;
```
bwilkerson: Should we have `UnitLintRule`? It's only being used in two places, and we could almost as easily just register to process a `CompilationUnit`. It's not clear to me that it's worth the weight of have an alternate interface to implement to save a three line method in two places.

lib/src/rules/invariant_booleans.dart
```diff
@@ -113,8 +114,9 @@ class InvariantBooleans extends LintRule implements UnitLintRule {
             maturity: Maturity.stable);
 
   @override
-  void processUnit(CompilationUnit unit) {
-    new _InvariantBooleansVisitor(this).visitCompilationUnit(unit);
+  void registerNodeProcessors(NodeLintRegistry registry) {
+    var visitor = new _Visitor(this);
```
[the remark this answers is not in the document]
scheglov: Done.

lib/src/rules/unnecessary_this.dart
```diff
@@ -59,8 +60,9 @@ class UnnecessaryThis extends LintRule implements UnitLintRule {
             group: Group.style);
 
   @override
-  void processUnit(CompilationUnit unit) {
-    new _UnnecessaryThisVisitor(this, unit).visitCompilationUnit(unit);
+  void registerNodeProcessors(NodeLintRegistry registry) {
+    var visitor = new _Visitor(this);
```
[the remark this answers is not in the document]
scheglov: `_UnnecessaryThisVisitor` is a `ScopeVisitor`, and it requires source, library, and type provider in its constructor :(

lib/src/rules/avoid_double_and_int_checks.dart
```diff
@@ -51,16 +51,19 @@ class AvoidDoubleAndIntChecks extends LintRule {
             group: Group.style);
 
   @override
-  AstVisitor getVisitor() => new Visitor(this);
+  void registerNodeProcessors(NodeLintRegistry registry) {
+    final visitor = new _Visitor(this);
+    registry.addIfStatement(this, visitor);
```
pq: I've missed a bit of the conversation but this pattern worries me a little for rule authors.  Is there always a one-to-one correlation between registry additions and visit method implementations?

EDIT: I'm more OK with it after some thinking; I'm also guessing the answer may be in general yes but in particular not always.  Documentation would be really useful here (but maybe you have some already ;))

pubspec.yaml
pq: 👍

pq: Alternatively you could consider making a reference to an analyzer git commit

https://www.dartlang.org/tools/pub/dependencies#git-packages

I would still not commit this but it would at least give us an idea of whether the build would be green.

lib/src/rules/avoid_double_and_int_checks.dart
scheglov: Yes, basically we override methods of `SimpleAstVisitor` for nodes we are interested in. And we register our intentions, so that the driving AST visitor does not waste time doing megamorphic invocations for all the other (empty, but to driving visitor cannot know this) methods. So, yes, it is one-to-one.

PR #986, opened 2018-04-30
Error reporting improvements.

pq:
Improved error reporting scope for omit_local_variable_types and prefer_generic_function_type_aliases.

```
[lint] Prefer generic function type aliases.
typedef String _Generator(String libName, String className);
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

=> 

```
typedef String _Generator(String libName, String className);
               ^^^^^^^^^^
```

and

```
[lint] Omit type annotations for local variables.
    for (String rule in sortedRules) {
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

=>
```
    for (String rule in sortedRules) {
         ^^^^^^
```

@bwilkerson @a14n

### DIFF
--- a/lib/src/rules/omit_local_variable_types.dart
+++ b/lib/src/rules/omit_local_variable_types.dart
@@ -79,7 +79,7 @@ class _Visitor extends SimpleAstVisitor<void> {
               DartTypeUtilities.isInterface(type, 'Iterable', 'dart.core'));
       if (iterableInterfaces.length == 1 &&
           iterableInterfaces.first.typeArguments.first == staticType.type) {
-        rule.reportLint(node.loopVariable.type ?? node);
+        rule.reportLint(staticType);
       }
     }
   }


pq:
Ah, Brian is out.

@scheglov PTAL?

lib/src/rules/omit_local_variable_types.dart
```diff
@@ -79,7 +79,7 @@ class _Visitor extends SimpleAstVisitor<void> {
               DartTypeUtilities.isInterface(type, 'Iterable', 'dart.core'));
       if (iterableInterfaces.length == 1 &&
           iterableInterfaces.first.typeArguments.first == staticType.type) {
-        rule.reportLint(node);
+        rule.reportLint(node.loopVariable.type ?? node);
```
scheglov: This looks overly defensive.
We have already tested above at the line `71` that `staticType = node.loopVariable?.type` is not `null`. So, `node.loopVariable` also cannot be `null`.

pq: Good catch.

pq: And better still, we have the value cached already in `staticType`! 🎉

PR #988, opened 2018-04-30
Rewrite collecting mutated fields in 'prefer_final_fields' to use AST visitor.

scheglov:
This makes this lint use about `750 ms` vs. `4500 ms` it used to use during analyzing Flutter repo, i.e. 6 times faster.

But using AST visitor is not really cheap itself, it just better than what was used before. For comparison, if we don't run this visitor at all, the lint uses just `51 ms`. The way to avoid running the whole new visitor is to repeat what we did with lints - run a single AST visitor and let all lints to participate in pre-preprocessing CompilationUnit before running lints generation. More complexity, yes, but faster.

### DIFF
--- a/lib/src/rules/prefer_final_fields.dart
+++ b/lib/src/rules/prefer_final_fields.dart
@@ -134,24 +134,21 @@ class _Visitor extends SimpleAstVisitor<void> {
       return;
     }
 
-    var variables = fields.variables;
-    for (int i = 0; i < variables.length; i++) {
-      final variable = variables[i];
+    fields.variables.forEach((VariableDeclaration variable) {
       final element = variable.element;
-
       if (!element.isPrivate) {
-        continue;
+        return;
       }
 
       if (variable.initializer == null) {
-        continue;
+        return;
       }
 
       if (_mutatedFields.contains(element)) {
-        continue;
+        return;
       }
 
       rule.reportLint(variable);
-    }
+    });
   }
 }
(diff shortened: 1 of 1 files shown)


lib/src/rules/prefer_final_fields.dart
```diff
@@ -5,7 +5,6 @@
 import 'dart:collection';
 
 import 'package:analyzer/dart/ast/ast.dart';
-import 'package:analyzer/dart/ast/standard_resolution_map.dart';
```
pq: Oh, this is awesome.  I didn't like reaching in here. 👍

lib/src/rules/prefer_final_fields.dart
```diff
@@ -119,26 +134,24 @@ class _Visitor extends SimpleAstVisitor<void> {
...
     }
 
-    fields.variables.forEach((VariableDeclaration variable) {
-      if (variable == null ||
-          !resolutionMap
-              .elementDeclaredByVariableDeclaration(variable)
-              .isPrivate) {
-        return;
+    var variables = fields.variables;
+    for (int i = 0; i < variables.length; i++) {
```
pq: Is this measurably faster than `for (var variable in fields.variables)`?

scheglov: It depends on how we measure it :-)

In order to see the difference I changed `visitFieldDeclaration()` to run the cycle over variables `10000` times. The numbers are:

With `forEach(closure)`: 9500 9497 9450 ms.
With index: 5068 5181 5135 ms.

So, there *is* a difference: `0.44 ms` for approximately the half of the Flutter repo.
But it is very small.

I guess it is not worth changing the style.
I will revert this change.

bwilkerson: Except that this is one of the ways in Dart to die by a thousand cuts. If we compared using for-each everywhere against using for everywhere, I suspect we'd notice a much bigger difference.

PR #996, opened 2018-05-17
Add special cases for fixnum to unrelated_type_equality_checks.

davidmorgan:
This suggestion for the unrelated_type_equality_checks lint follows analysis of a large body of Dart code; 80% of the failures for this lint were false positives related to `Int64` or `Int32`, the most common form by far being a check for whether an `Int64` value `== 0`.

The remaining 20% of failures were almost all useful, so with this tweak the lint has a very low false positive rate.

### DIFF
--- a/lib/src/rules/unrelated_type_equality_checks.dart
+++ b/lib/src/rules/unrelated_type_equality_checks.dart
@@ -19,6 +19,9 @@ const _details = r'''
 Comparing references of a type where neither is a subtype of the other most
 likely will return `false` and might not reflect programmer's intent.
 
+`Int64` and `Int32` from `package:fixnum` allow comparing to `int` provided
+the `int` is on the right hand side. The lint allows this as a special case. 
+
 **BAD:**
 ```
 void someFunction() {
@@ -126,11 +129,19 @@ class DerivedClass2 extends ClassBase with Mixin {}
 
 ''';
 
+const String _dartCoreLibraryName = 'dart.core';
+
 bool _hasNonComparableOperands(BinaryExpression node) =>
     !DartTypeUtilities.isNullLiteral(node.leftOperand) &&
     !DartTypeUtilities.isNullLiteral(node.rightOperand) &&
     DartTypeUtilities.unrelatedTypes(
-        node.leftOperand.bestType, node.rightOperand.bestType);
+        node.leftOperand.bestType, node.rightOperand.bestType) &&
+    (node.rightOperand.bestType.name != 'int' ||
+        node.rightOperand.bestType.element?.library?.name !=
+            _dartCoreLibraryName ||
+        !((node.leftOperand.bestType.name == 'Int32' ||
+                node.leftOperand.bestType.name == 'Int64') &&
+            node.leftOperand.bestType.element?.library?.name == 'fixnum'));
 
 class UnrelatedTypeEqualityChecks extends LintRule implements NodeLintRule {
   UnrelatedTypeEqualityChecks()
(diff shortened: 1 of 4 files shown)


lib/src/rules/unrelated_type_equality_checks.dart
```diff
@@ -126,11 +129,19 @@ class DerivedClass2 extends ClassBase with Mixin {}
...
     !DartTypeUtilities.isNullLiteral(node.rightOperand) &&
     DartTypeUtilities.unrelatedTypes(
-        node.leftOperand.bestType, node.rightOperand.bestType);
+        node.leftOperand.bestType, node.rightOperand.bestType) &&
+    (node.rightOperand.bestType.name != 'int' ||
+        node.rightOperand.bestType.element?.library?.name !=
+            _dartCoreLibraryName ||
+        !((node.leftOperand.bestType.name == 'Int32' ||
+                node.leftOperand.bestType.name == 'Int64') &&
+            node.leftOperand.bestType.element?.library?.name == 'fixnum'));
```
scheglov: Please extract `static bool _isCoreInt(DartType)` and `static bool _isFixNumInt32/64` helpers to make code clearer.

lib/src/rules/unrelated_type_equality_checks.dart
```diff
@@ -126,11 +129,19 @@ class DerivedClass2 extends ClassBase with Mixin {}
...
 ''';
 
+const String _dartCoreLibraryName = 'dart.core';
+
 bool _hasNonComparableOperands(BinaryExpression node) =>
     !DartTypeUtilities.isNullLiteral(node.leftOperand) &&
     !DartTypeUtilities.isNullLiteral(node.rightOperand) &&
     DartTypeUtilities.unrelatedTypes(
-        node.leftOperand.bestType, node.rightOperand.bestType);
+        node.leftOperand.bestType, node.rightOperand.bestType) &&
```
scheglov: Please extract `left` and `right` expressions; as well as `leftType` and `rightType`.
We reference them multiple times, and it hurts readability.

test/util/fake_fixnum.dart
```diff
@@ -0,0 +1,5 @@
+library fixnum;
```
scheglov: It might be fragile.
We dropped requirement for library directives long ago.
So, developers are free to remove them - it does not change semantic of the language according to the specification. Ideally we should check for `package:fixnum` URI.

OTOH, we control this package, so we might decide to keep the library directive because it is used.

davidmorgan: How would I check the URI?

I'm happy to make the change; one downside though is that for testing we would then need a dependency on the fixnum package. I guess that would be okay if it's a dev_dependency?

Please let me know. Thanks :)

lib/src/rules/unrelated_type_equality_checks.dart
davidmorgan: Done.

lib/src/rules/unrelated_type_equality_checks.dart
davidmorgan: Done.

test/util/fake_fixnum.dart
scheglov: I think doing this would require some kind of infrastructure to support packages in linter tests. Something like automatically adding all (or requested by the test?) packages in `test/package_mocks/` to `SourceFactory` of the test. You don't need to do this, leave it as is for now.

matanlurey:
I also filed https://github.com/dart-lang/fixnum/issues/38, because it is silly.

test/util/fake_fixnum.dart
davidmorgan: Sounds good to me. Thanks.

davidmorgan:
All done, thanks.

pq:
@scheglov, @bwilkerson : is this good to land?

PR #1001, opened 2018-05-22
add avoid_final_immutable_in_flutter_state

scheglov:
The goal of this lint is to provide static nudge to using code that is Flutter hot-reloadable.

In theory any final fields in `State` could be moved to `Widget`. But if the referenced object is mutable (e.g. `List` is often used in such way in Flutter code), then it cannot be moved to `Widget`. So, we look at `@immutable` presence as a sign that the object is also not mutable.

It occurred to me  static fields are even better - if the field is not constant, it is never reloaded, no matter whether it is defined in `Widget` or `State`. But OTOH static fields don't use space in each instance.

@Hixie @devoncarew


scheglov:
Oh, and let me know if you think that this is too narrow use case, so that it might be not worth to check.

lib/src/util/flutter.dart
```diff
@@ -0,0 +1,31 @@
...
+final _frameworkUri = Uri.parse('package:flutter/src/widgets/framework.dart');
+
+/// Return `true` if the given [element] has the Flutter class `State` as
+/// a superclass.
+bool isState(ClassElement element) =>
+    _hasSupertype(element, _frameworkUri, _STATE_NAME);
+
+/// Return `true` if the given [element] has a supertype with the [requiredName]
+/// defined in the file with the [requiredUri].
+bool _hasSupertype(ClassElement element, Uri requiredUri, String requiredName) {
```
bwilkerson: Seems like this ought to be in `DartTypeUtilities` (in linter/lib/src/util/dart_type_utilities.dart).

Hixie:
This doesn't seem valid. What if it's an @immutable class whose constructor creates (or receives) a mutable object for one of its fields?

scheglov:
Yes, you are right.
`@immutable` does not say that it is transitive.

PR #1004, opened 2018-05-23
add a check to prefer_const_literals_to_create_immutables

a14n:
Fix dart-lang/sdk#33210


lib/src/rules/prefer_const_literals_to_create_immutables.dart
```diff
@@ -79,7 +79,7 @@ class _Visitor extends SimpleAstVisitor<void> {
 
   Iterable<InterfaceType> _getSelfAndInheritedTypes(InterfaceType type) sync* {
     InterfaceType current = type;
-    while (current != null) {
+    while (current != null && current is InterfaceType) {
```
bwilkerson: It would be good to add a TODO comment indicating the issue number being fixed and the fact that this ought not to be necessary (and that we should return to it at some point to try to figure out whether/why it is).

PR #1014, opened 2018-06-07
Add new lint avoid_unused_functions.

davidmorgan:
It is a weaker version of the existing unnecessary_statements lint; it only fails when the expression is a function. This specifically catches cases where the fix is to add parentheses to call the function.

Note: I created this PR by forking the unnecessary_statements lint code and test; it may make sense to refactor to remove the code duplication. But, I thought it would first make sense to see if you agree on the idea of the lint.

When looking at unnecessary_statements we found false positives related to getters with side effects (called to trigger caching, loading, or updating) or getters that might throw (called to check if they throw). The main bug that unnecessary_statements did catch was missing parenthesis. So, this lint is targeted at that precise case.

Thanks!

### DIFF
--- a/example/all.yaml
+++ b/example/all.yaml
@@ -34,6 +34,7 @@ linter:
     - avoid_types_as_parameter_names
     - avoid_types_on_closure_parameters
     - avoid_unused_constructor_parameters
+    - avoid_unused_functions
     - await_only_futures
     - camel_case_types
     - cancel_subscriptions
--- a/lib/src/rules.dart
+++ b/lib/src/rules.dart
@@ -35,6 +35,7 @@ import 'package:linter/src/rules/avoid_slow_async_io.dart';
 import 'package:linter/src/rules/avoid_types_as_parameter_names.dart';
 import 'package:linter/src/rules/avoid_types_on_closure_parameters.dart';
 import 'package:linter/src/rules/avoid_unused_constructor_parameters.dart';
+import 'package:linter/src/rules/avoid_unused_functions.dart';
 import 'package:linter/src/rules/await_only_futures.dart';
 import 'package:linter/src/rules/camel_case_types.dart';
 import 'package:linter/src/rules/cancel_subscriptions.dart';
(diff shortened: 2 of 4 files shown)


a14n:
An other solution could be to extract an `avoid_unused_getter` lint. `unnecessary_statements` assumes that getter have no side effects but it's perhaps a little controversal. Thus `unnecessary_statements` will not handle getters and users will have to opt-in to also check getters. WDYT?

pq:
@davidmorgan : I'm all for this one.

@bwilkerson: thoughts?

pq:
I'm also not terribly worried about false positives in the esoteric cases.

On the other hand, the side-effecting getters case that this is accommodating seems a little fishy.  I recognize that folks are doing this but I have to ask *should they*?  At least on the surface, that seems like a code smell.

For example, this related idiom in Java drives me personally a little crazy:

```java
    // Initialize Foo.  
    FooMananger.getInstance();
```

and would much prefer something like:

```java
    FooMananger.initialize();
```


Sorry for the potential derail!

a14n:
>If we can later figure out how to catch the use of getters that do not have side-effects, then we can add them back in.

Or we could add now a new lint `unnecessary_getter_access` for developers that consider getters should not have side-effects ;)

bwilkerson:
Yes, that would also work. The two lints would be non-overlapping, which is good. They would be strongly related but separate, which is less than ideal, but possibly the best option.

We can partially mitigate the pain caused by having fine grained lints via documentation. For example, each of the rules could suggest that users consider also enabling the other.

pq:
> Honestly, I think the right solution is to update unnecessary_statements to stop reporting the use of getters and to report the narrower case of tear-offs. 

I agree.  Opened https://github.com/dart-lang/linter/issues/1015 to track.  @davidmorgan: would that work for you?

bwilkerson:
Yes, I believe that's what we've decided.

davidmorgan:
Thanks; took a shot at implementing, commented over on the issue. Will close this PR and open a new one.

PR #1026, opened 2018-06-14
new lint: `file_names`

a14n:


pq:
Good question @bwilkerson .  I'm guessing @a14n, you ran this on `flutter`?

Regardless I'll be sure and put a BREAKING CHANGE note in the changelog.

pq:
Ok, hmmm.  The more i think on this, the more I think we really want to break this into a separate lint.   
  I could imagine wanting to enforce one but not the other.  Am I being crazy?

What about adding one called `file_names`?

pq:
Let's split it.  They can both point back to that piece of advice.

Thanks!

lib/src/rules/file_names.dart
```diff
@@ -0,0 +1,60 @@
...
+import 'package:linter/src/utils.dart';
+
+const _desc = r'Name source files using `lowercase_with_underscores`.';
+
+const _details = r'''
+
+**DO** name source files using `lowercase_with_underscores`.
+
+Some file systems are not case-sensitive, so many projects require filenames
+to be all lowercase. Using a separate character allows names to still be
```
pq: separate => separator?

lib/src/rules/library_names.dart
```diff
@@ -7,12 +7,11 @@ import 'package:analyzer/dart/ast/visitor.dart';
...
-    r'Name libraries and source files using `lowercase_with_underscores`.';
+const _desc = r'Name libraries using `lowercase_with_underscores`.';
 
 const _details = r'''
 
-**DO** name libraries and source files using `lowercase_with_underscores`.
+**DO** name libraries using `lowercase_with_underscores`.
 
 Some file systems are not case-sensitive, so many projects require filenames
 to be all lowercase.  Using a separate character allows names to still be
```
pq: I wonder if we should call out `file_names` as a related lint?

lib/src/rules/file_names.dart
a14n: I updated the sentence with the content of https://www.dartlang.org/guides/language/effective-dart/style#do-name-libraries-and-source-files-using-lowercase_with_underscores

lib/src/rules/library_names.dart
a14n: done

pq:
Awesome.  Thank you!

PR #1028, opened 2018-06-14
add lint import_prefix_names

a14n:
https://www.dartlang.org/guides/language/effective-dart/style#do-name-import-prefixes-using-lowercase_with_underscores

### DIFF
--- a/example/all.yaml
+++ b/example/all.yaml
@@ -48,6 +48,7 @@ linter:
     - empty_statements
     - hash_and_equals
     - implementation_imports
+    - import_prefix_names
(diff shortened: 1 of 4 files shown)


pq:
I think we have this one already?

https://github.com/dart-lang/linter/blob/e1a6c5edf96c1cab1e54775ba9e02e45d4413410/lib/src/rules/library_prefixes.dart#L11

But clearly it could be better documented! 😬

a14n:
I just saw it :-( Sorry for the noise.

PR #1031, opened 2018-06-19
bump dependencies

a14n:


pubspec.yaml
```diff
@@ -6,17 +6,17 @@ homepage: https://github.com/dart-lang/linter
...
   meta: ^1.0.2
   source_span: ^1.0.2
   yaml: ^2.1.2
 dev_dependencies:
   cli_util: ^0.1.2
-  dart_style: ^1.0.0
+  dart_style: ^1.1.0
   grinder: ^0.8.0
-  markdown: '>=0.11.0 <2.0.0'
+  markdown: ^2.0.0
```
pq: Is there any downside to just widening the range here?

pubspec.yaml
```diff
@@ -6,17 +6,17 @@ homepage: https://github.com/dart-lang/linter
...
   matcher: ^0.12.0
   path: '>=0.9.0 <2.0.0'
-  test: ^0.12.24+1
+  test: ^1.0.0
```
pq: Similarly here?

a14n:
I don't have problem to narrow dependencies in `dev_dependencies` section as they are not in the dependency graph of a package depending on linter.

Issue #1008, opened 2018-05-25
meta issue tracking lints with issues when using optional new/const


devoncarew:
Follow up from https://github.com/flutter/flutter/issues/17429; we need to ensure that the lints (esp. the ones in use by Flutter) work when using optional new and const. I think this means:

- [ ] removing new and const from the flutter codebase (@munificent has a handy version of dart_style that does this)
- [ ] running `flutter analyze --flutter-repo` and identifying problematic lints (for example, `Prefer const with constant constructors`)
- [ ] addressing the problematic lints (and getting the optional new/const version of them under test)

@pq, can you take this on?

- [ ] prefer_const_literals_to_create_immutables 
- [x] prefer_const_constructors #995

a14n:
Removing optional new/const on flutter examples I only see issues with `prefer_const_literals_to_create_immutables` and `prefer_const_constructors`.

a14n:
Related to #995

pq:
Meta-meta question: do we even want these lints in the context of Dart2?  I'm of two minds.  In particular, I worry that their behavior when correct might nonetheless be surprising. As a general rule, I think the semantics of lints should be straight-forward (and not a place for nuance).  But that's just me...

@hixie: care to chime in?  Starting w/ Dart2, how do you feel about the value of these lints?

Hixie:
For Flutter I'm hoping we can have the following lints:

* avoid using "new"
* avoid using redundant "const"
* use "const" if the entire subexpression could be const

Hixie:
and i would enable the last one for users by default

bwilkerson:
> I worry that their behavior when correct might nonetheless be surprising.

That's a valid concern. The question is whether we can remove the surprise factor by the way the lint is reported (the message). Can we, for example, tell users something like

The keyword `const` is not needed because <the surrounding expression> already requires that this be a constant.

And would that be enough to help users understand why the lint was produced?

pq:
A message like:

> The keyword const is not needed because already requires that this be a constant.

feels like a big step in the right direction.  Great suggestion.

In general my concern is just that with new/const (as w/ types in the face of inference), it's often not obvious to a reader whether an explicit `new`/`const` is gratuitous or actually communicating something.  But this is a meta (if not meta-meta) point.

Back to this point and @Hixie's asks, I wonder whether it makes sense to muscle these existing lints into compliance or to just deprecate them and start fresh.

Hixie:
[...]

PR #1033, opened 2018-06-20
0.1.54

pq:
# 0.1.54

* new `unnecessary_const` lint
* new `unnecessary_new` lint
* fixed errors in `use_to_and_as_if_applicable`
* new `file_names` lint

/cc @bwilkerson @a14n

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+# 0.1.54
+
+* new `unnecessary_const` lint
+* new `unnecessary_new` lint
+* fixed errors in `use_to_and_as_if_applicable`
+* new `file_names` lint
+
 # 0.1.53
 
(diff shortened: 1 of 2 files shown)


pq:
@a14n: you're faster than the bots! ⚡️ 

Thanks for the quick review. 👍

pq:
Oh and thanks a million for these new lints.  Great release!

a14n:
It will be even better once we have #1008 fixed :)

pq:
@a14n : absolutely!  In fact, if you have a chance, could you add any updates you have to that issue?  `unnecessary_new` and `unnecessary_const`  are big steps in the right direction.  Now if only we could tackle the `ConstantVerifier` issues...

a14n:
I splitted the original lint into `unnecessary_const` and `unnecessary_new` because I only faced issues after removing `new`. I didn't see issues by removing `const`.

Issue #995, opened 2018-05-17
prefer_const_constructors false positive


apwilson:
The following code triggers the prefer_const_constructors linter rule on the initial assignment of foo:

```dart
///
class Foo {
  ///
  const Foo();
}

///
class Bar {
  ///
  Foo foo = Foo();    // Triggers prefer_const_constructors

  ///
  set changeFoo(Foo newFoo) {
    foo = newFoo;
  }
}
```

As I thought new/const was optional and inferred, this appears to be a false positive.

matanlurey:
Yes, `new` is optional, but `const` is only optional in a `const` context. Consider:

```dart
main() {
  var foo = /* new */ Foo();
  var foos = const [
    /* const */ Foo(),
  ];
}
```

Sorry about the confusion. I think @leafpetersen is working on an update to this feature for the language site/etc.

apwilson:
So I have a related false positive then:

```
///
class Foo {}

///
class Bar {
  ///
  final Foo foo;

  ///
  const Bar({this.foo});
}

///
class BarFactory {
  ///
  Bar makeBar() {
    return Bar(foo: Foo()); // false positive saying I should make Bar const.  But if I make Bar const, Foo will try to become const and it can't
  }
}

```

a14n:
@apwilson I don't see the lint with [your last example](https://github.com/dart-lang/linter/issues/995#issuecomment-389707091). What's the version of your Dart SDK and the linter package if you have one specified?

apwilson:
Our dart version is: 46ab040e589adc5200370dec7952ce5150850822

a14n:
I can reproduce it on master of linter. It's related to [ConstantVerifier](https://github.com/dart-lang/sdk/blob/4bac17655dfe5a08764e7ac97b0bfb632eb20e30/pkg/analyzer/lib/src/generated/resolver.dart#L1524) that does not trigger error for `const Bar(foo: Foo())`. @bwilkerson could you take a look?

bwilkerson:
I have identifier the cause of the problem. The expression cannot be `const` because `Foo` can't be `const`, but the error indicating the problem is not produced by the `ConstantVerifier`. Hence, it does not appear in the list of "constant related errors" and the lint rule cannot tell that the expression cannot be `const`.

Fixing this problem is a bit bigger than I have time for at the moment.

a14n:
[...]

PR #1035, opened 2018-06-22
[WIP] add tests for optional new/const

a14n:
Related to #995 and #1008

### DIFF
--- a/lib/src/ast.dart
+++ b/lib/src/ast.dart
@@ -335,6 +335,7 @@ class HasConstErrorListener extends AnalysisErrorListener {
     CompileTimeErrorCode.CONST_EVAL_TYPE_NUM,
     CompileTimeErrorCode.CONST_EVAL_THROWS_EXCEPTION,
     CompileTimeErrorCode.CONST_EVAL_THROWS_IDBZE,
+    CompileTimeErrorCode.CONST_WITH_NON_CONST,
     CompileTimeErrorCode.CONST_WITH_NON_CONSTANT_ARGUMENT,
     CompileTimeErrorCode.INVALID_CONSTANT,
     CompileTimeErrorCode.MISSING_CONST_IN_LIST_LITERAL,
--- a/test/rules/prefer_const_constructors.dart
+++ b/test/rules/prefer_const_constructors.dart
@@ -84,3 +84,12 @@ class G {
   static G m2(Map<G, G> m) => new G(m); // OK
   static G m3(G g) => new G({g: g}); // OK
 }
+
+// optional new : https://github.com/dart-lang/linter/issues/995
+class H {}
+class I {
+  final H foo;
+  const I({this.foo});
+
+  I makeI() => I(foo: H()); // OK
+}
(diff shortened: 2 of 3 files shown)


a14n:
Before merging those additionnal tests need to enable [`previewDart2` in DriverOptions](https://github.com/dart-lang/sdk/blob/4ebdb31646c15b896449b039dd3d517547ac9b1a/pkg/analyzer/lib/src/lint/analysis.dart#L88) and upgrade the analyzer dependency to include https://github.com/dart-lang/sdk/commit/b343d0a80d3e8241345945c49692703c82101894.

BTW: it looks like `--preview-dart-2` is now the default in VM. Shouldn't it be also switched in DriverOptions ?

bwilkerson:
Yes, the VM is now Dart 2 by default (on bleeding edge), so we should update the linter package before it is next pulled into the SDK. That would include `DriverOptions` and `pubspec.yaml` changes.

If we need a new version of the analyzer, then as far as I know we can publish one.

pq:
Thanks @bwilkerson! The rub is that `DriverOptions` lives in `analyzer`.  Maybe we can take a look at this together later today?

bwilkerson:
Yep. Let's talk with @devoncarew when you get in (the answer might depend on when we decide to stop supporting `--no-preview-dart-2` in the analyzer).

pq:
@a14n :

> Before merging those additionnal tests need to enable previewDart2 in DriverOptions 

https://github.com/dart-lang/linter/pull/1039 should do this in the tests themselves.

> and upgrade the analyzer dependency to include dart-lang/sdk@b343d0a.

I tried this locally and still saw breaks in `prefer_const_constructors`; is that expected?

a14n:
>I tried this locally and still saw breaks in `prefer_const_constructors`; is that expected?

No. I need to have a look (but perhaps tomorrow).

pq:
Cool.  Thanks @a14n !

a14n:
To make the tests pass you need to override the analyzer package to use the one from the SDK

```diff
diff --git a/pubspec.yaml b/pubspec.yaml
index 469168d..8e1eed3 100644
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -20,3 +20,10 @@ dev_dependencies:
   matcher: ^0.12.0
   path: '>=0.9.0 <2.0.0'
   test: ^1.0.0
+dependency_overrides:
+  analyzer:
+    path: ../sdk/pkg/analyzer
+  front_end:
+    path: ../sdk/pkg/front_end
+  kernel:
+    path: ../sdk/pkg/kernel
```

pq:
@a14n: I thought I'd tried that, sorry!

Should that fix  #1040 as well?

pq:
In any event, since tests are already failing, shouldn't we go ahead and land this change?

a14n:
The fix is to publish a new analyzer package.

>In any event, since tests are already failing, shouldn't we go ahead and land this change?

I think it's fine.

pq:
Great.  Thanks.  We can follow-up on #1040.

/cc @bwilkerson as we'll want to do an analyzer roll.

PR #1045, opened 2018-06-27
fix issue in const error handling

a14n:


a14n:
>

It looks like some errors can not be used to create `AnalysisError`. For instance analyzer shows no error for `new AnalysisError(null, null, null, CompileTimeErrorCode.AMBIGUOUS_EXPORT)` but fails at runtime with `--enable-asserts`.

pq:
@a14n: should I push out a new release w/ this fix?

a14n:
@pq go for a new release (dart-lang/sdk#33647 been related to analyzer it shouldn't impact linter)

pq:
Will do.  Thanks!

bwilkerson:
> ... analyzer shows no error for new AnalysisError(null, null, null, CompileTimeErrorCode.AMBIGUOUS_EXPORT) but fails at runtime with --enable-asserts.

The messaging is poor, but that happens when the error expects arguments and none are provided.

PR #1046, opened 2018-06-27
0.1.55

pq:
# 0.1.55

* fixed an issue in `const` error handling 

/cc @a14n @bwilkerson


a14n:
Should we mention that the package now uses Dart 2?

pq:
@a14n: good catch.  Besides `linter` main changing I don't think this has any user-facing impact, or?

a14n:
>I don't think this has any user-facing impact

I think you're right.

PR #1048, opened 2018-06-27
add lint lines_longer_than_80_chars

a14n:
Lint for [AVOID lines longer than 80 characters.](https://www.dartlang.org/guides/language/effective-dart/style#avoid-lines-longer-than-80-characters)

### DIFF
--- a/lib/src/rules/lines_longer_than_80_chars.dart
+++ b/lib/src/rules/lines_longer_than_80_chars.dart
@@ -1,0 +1,170 @@
+// Copyright (c) 2018, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/token.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:linter/src/analyzer.dart';
+
+const _desc = r'AVOID lines longer than 80 characters.';
+
+const _details = r'''
+
+**AVOID** lines longer than 80 characters
+
+Readability studies show that long lines of text are harder to read because your
+eye has to travel farther when moving to the beginning of the next line. This is
+why newspapers and magazines use multiple columns of text.
+
+If you really find yourself wanting lines longer than 80 characters, our
+experience is that your code is likely too verbose and could be a little more
+compact. The main offender is usually `VeryLongCamelCaseClassNames`. Ask
+yourself, “Does each word in that type name tell me something critical or
+prevent a name collision?” If not, consider omitting it.
+
+Note that dartfmt does 99% of this for you, but the last 1% is you. It does not
+split long string literals to fit in 80 columns, so you have to do that
+manually.
+
+We make an exception for URIs and file paths. When those occur in comments or
+strings (usually in imports and exports), they may remain on a single line even
+if they go over the line limit. This makes it easier to search source files for
+a given path.
+''';
+
+class LinesLongerThan80Chars extends LintRule implements NodeLintRule {
+  LinesLongerThan80Chars()
+      : super(
+            name: 'lines_longer_than_80_chars',
+            description: _desc,
+            details: _details,
+            group: Group.style);
+
+  @override
+  void registerNodeProcessors(NodeLintRegistry registry) {
+    final visitor = new _Visitor(this);
+    registry.addCompilationUnit(this, visitor);
+  }
+}
+
+class _Visitor extends SimpleAstVisitor {
+  final LintRule rule;
+
+  _Visitor(this.rule);
+
+  @override
+  void visitCompilationUnit(CompilationUnit node) {
+    final lineInfo = node.lineInfo;
+
+    final allowedLineVisitor = new _AllowedLongLineVisitor(lineInfo);
+    node.accept(allowedLineVisitor);
+    final allowedCommentVisitor = new _AllowedCommentVisitor(lineInfo);
+    node.accept(allowedCommentVisitor);
+
+    final allowedLines = []
+      ..addAll(allowedLineVisitor.allowedLines)
+      ..addAll(allowedCommentVisitor.allowedLines);
+
+    final lineCount = lineInfo.lineCount;
+    for (int i = 0; i < lineCount; i++) {
+      final start = lineInfo.getOffsetOfLine(i);
+      final end = i == lineCount - 1
+          ? node.end
+          : lineInfo.getOffsetOfLineAfter(start) - 1;
+      final length = end - start;
+      if (length > 80 && !allowedLines.contains(i + 1)) {
+        rule.reporter.reportErrorForOffset(rule.lintCode, start, length);
+      }
+    }
+  }
+}
+
+class _AllowedLongLineVisitor extends RecursiveAstVisitor {
+  _AllowedLongLineVisitor(this.lineInfo);
+
+  final LineInfo lineInfo;
+  final allowedLines = <int>[];
+
+  @override
+  visitStringInterpolation(StringInterpolation node) {
+    if (node.isMultiline) {
+      _handleMultilines(node);
+    } else {
+      final value = node.elements.map((e) {
+        if (e is InterpolationString) return e.value;
+        if (e is InterpolationExpression) return ' ' * e.length;
+      }).join();
+      _handleSingleLine(node, value);
+    }
+  }
+
+  @override
+  visitSimpleStringLiteral(SimpleStringLiteral node) {
+    if (node.isMultiline)
+      _handleMultilines(node);
+    else
+      _handleSingleLine(node, node.value);
+  }
+
+  _handleMultilines(SingleStringLiteral node) {
+    final startLine = lineInfo.getLocation(node.offset).lineNumber;
+    final endLine = lineInfo.getLocation(node.end).lineNumber;
+    for (var i = startLine; i <= endLine; i++) {
+      allowedLines.add(i);
+    }
+  }
+
+  _handleSingleLine(AstNode node, String value) {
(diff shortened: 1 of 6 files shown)


lib/src/rules/lines_longer_than_80_chars.dart
```diff
@@ -0,0 +1,170 @@
...
+    var comment = token.precedingComments;
+    while (comment is CommentToken) {
+      comments.add(comment);
+      comment = comment.next;
+    }
+    return comments;
+  }
+}
+
+bool _looksLikeUriOrPath(String value) => value.contains('/');
```
a14n: Not sure what's the best way to tell a line contains a uri or Path.

a14n:
Probably closes #994

lib/src/rules/lines_longer_than_80_chars.dart
```diff
@@ -0,0 +1,170 @@
...
+  _AllowedCommentVisitor(this.lineInfo);
+
+  final LineInfo lineInfo;
+  final allowedLines = <int>[];
+
+  @override
+  visitNode(AstNode node) {
+    final comments = <CommentToken>[]
+      ..addAll(_getPrecedingComments(node.beginToken))
+      ..addAll(_getPrecedingComments(node.endToken));
```
bwilkerson: This skips comments preceding either the first or last token. It also examines many tokens multiple times. It's probably better to ask the compilation unit for the first token and for each token ask for preceding comments.

lib/src/rules/lines_longer_than_80_chars.dart
```diff
@@ -0,0 +1,170 @@
...
+    for (final comment in comments) {
+      final content = comment.toString();
```
bwilkerson: If the content is less than 80 characters, then I think you don't need to add the line containing it to the list of lines that are allowed to be long, which would save some string manipulations. (See also line 119.)

lib/src/rules/lines_longer_than_80_chars.dart
```diff
@@ -0,0 +1,170 @@
...
+        }
+      }
+    }
+    node.visitChildren(this);
+  }
+
+  List<CommentToken> _getPrecedingComments(Token token) {
+    var comments = <CommentToken>[];
+    var comment = token.precedingComments;
+    while (comment is CommentToken) {
```
bwilkerson: Both `token.preceedingComments` and `CommentToken.next` alway returns either `null` or a `CommentToken`, so this test isn't necessary. You might also be able to avoid creating a new list.

lib/src/rules/lines_longer_than_80_chars.dart
bwilkerson: I'm not either. You could always try parsing it as a URI and/or file path, but that could be fairly expensive.

a14n: I initially test with `Uri.tryParse` but it looks like almost every String can be parsable :-/ I changed the test for a regexp like `/[\S]{5,}` WDYT ?

lib/src/rules/lines_longer_than_80_chars.dart
bwilkerson: I guess I was assuming that the preferred style would be to move such a comment to the line before, which would mean that we _did_ want to flag the line (because it could be "fixed"). For example:
```
if (condition) {
  return some.really.long.expression; // some really long comment describing why we're returning
}
```
would become
```
if (condition) {
  // some really long comment describing why we're returning
  return some.really.long.expression;
}
```
Come to think of it, do we really want to exempt comment lines? Should we flag a line like
```
// some really long comment that someone put on a single line and didn't bother to break up.
```
It's perfectly reasonable for such a comment to be broken before column 80 and wrap to the next line:
```
// some really long comment that someone put on a single line and didn't bother
// to break up.
```

lib/src/rules/lines_longer_than_80_chars.dart
bwilkerson: Ouch! :-)

lib/src/rules/lines_longer_than_80_chars.dart
```diff
@@ -0,0 +1,172 @@
...
+    node.accept(allowedLineVisitor);
+    final allowedCommentVisitor = new _AllowedCommentVisitor(lineInfo);
+    node.accept(allowedCommentVisitor);
+
+    final allowedLines = []
+      ..addAll(allowedLineVisitor.allowedLines)
+      ..addAll(allowedCommentVisitor.allowedLines);
+
+    final lineCount = lineInfo.lineCount;
+    for (int i = 0; i < lineCount; i++) {
```
bwilkerson: It occurs to me that we could significantly improve performance if we moved most of this loop to the beginning of the method. Instead of checking `allowedLines.contains(i + 1)` and reporting a lint, we could add the index of the long line to a collection of long lines and exit early if the collection was empty. We'd need another loop iterating over the collection of long lines to see whether they were allowed.

This would skip two visits of the AST for files that don't contain long lines, which is likely to be most of them for any package that has this rule enabled.

test/rules/lines_longer_than_80_chars.dart
```diff
@@ -0,0 +1,31 @@
...
+
+var short = 'This is a short line'; // OK
+var line80 =
+    '                                                                   '; // OK
+var line81 =
+    '                                                                  '; // LINT
+var long =
+    'This line is a long, very long, very very long, very very very long, very very very very long'; // LINT
+
+var uriIsOk =
```
bwilkerson: We should have some tests for file paths (both Posix and Windows style).

And if we're going to use a pattern, it would be good to have some additional URIs with additional structure to ensure that the pattern covers all of the cases.

lib/src/rules/lines_longer_than_80_chars.dart
```diff
@@ -0,0 +1,172 @@
...
+      final value = lines[i];
+      if (_looksLikeUriOrPath(value)) {
+        final line = lineInfo.getLocation(comment.offset).lineNumber + i;
+        allowedLines.add(line);
+      }
+    }
+  }
+}
+
+final uriRegExp = new RegExp(r'/[\S]{5,}');
```
bwilkerson: I don't speak regex, so perhaps add a comment describing what this pattern is expected to match.

lib/src/rules/lines_longer_than_80_chars.dart
a14n: done

lib/src/rules/lines_longer_than_80_chars.dart
a14n: Done

test/rules/lines_longer_than_80_chars.dart
a14n: Done

lib/src/rules/lines_longer_than_80_chars.dart
```diff
@@ -0,0 +1,191 @@
...
+      if (_looksLikeUriOrPath(value)) {
+        final line = lineInfo.getLocation(comment.offset).lineNumber + i;
+        allowedLines.add(line);
+      }
+    }
+  }
+}
+
+/// String looks like URI if it contains a slash or backslash followed by 5
+/// non-empty chars.
```
bwilkerson: I'm not sure where the 5 came from. More importantly, it seems like this pattern will match some strings that it shouldn't match ('sin = opposite/hypotenuse') and not match some strings that it ought to match ('C:\a directoryPath\with spacesInTheComponents'). Are there any false positives when running this over real code?

lib/src/rules/lines_longer_than_80_chars.dart
```diff
@@ -0,0 +1,191 @@
...
+      lines.add(content.substring(3));
+    } else if (content.startsWith('//')) {
+      lines.add(content.substring(2));
+    } else if (content.startsWith('/*')) {
+      // remove last slash before finding slash
+      lines.addAll(content.substring(2, content.length - 2).split('\n'));
+    }
+    for (var i = 0; i < lines.length; i++) {
+      final value = lines[i];
+      if (_looksLikeUriOrPath(value)) {
```
bwilkerson: Should we also require that the strings have a minimum length before we allow the lines that contain them? For example, the following would probably not be flagged:
```
var errorMessageForUser = 'There is a an error on line ' + lineNumber + ' at column ' + columnNumber + ' of /path/to/input';
```
even though it could easily be split into shorter lines.

test/rules/lines_longer_than_80_chars.dart
```diff
@@ -0,0 +1,28 @@
...
+    'https://www.dartlang.org/guides/language/effective-dart/style#avoid-lines-longer-than-80-characters'; // OK
+
+var posixPathIsOk =
+    '/home/www.dartlang.org/guides/language/effective-dart/style#avoid-lines-longer-than-80-characters'; // OK
+
+var windowsPathIsOk =
+    r'\home\www.dartlang.org\guides\language\effective-dart\style#avoid-lines-longer-than-80-characters'; // OK
```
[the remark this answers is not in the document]
a14n: Done

lib/src/rules/lines_longer_than_80_chars.dart
a14n: I changed the rule to check only the presence of a `/` or `\`. It's quite impossible to tell if a String contains an URI or path because almost everything can be an URI. So I think it's better to have false negative (because a String or comment contains a `/` or `\`) rather than false positive that will force to split a real URI on two lines to remove the lint.

lib/src/rules/lines_longer_than_80_chars.dart
a14n: I hope people that will use this lint will use also `dartfmt`. In this case this example will be reformatted without overflow. So I don't think it's worse to check String length.

PR #1050, opened 2018-06-28
Fix avoid_positional_boolean_parameters for implementing classes

### LINKED ISSUE #786: avoid_positional_boolean_parameters should ignore overridden methods
In the following code, `avoid_positional_boolean_parameters` is reported even in the `Bar` class. I think the rule should only warn the original class that defines the signature.

```dart
abstract class Foo {
  void foo(bool positionalBoolean);
}

class Bar implements Foo {
  @override
  void foo(bool positionalBoolean) {}
}
```

Interestingly, when I change `implements` to `extends`, the warning goes away.

srawlins:
Fix #786

### DIFF
--- a/lib/src/rules/avoid_positional_boolean_parameters.dart
+++ b/lib/src/rules/avoid_positional_boolean_parameters.dart
@@ -66,9 +66,8 @@ class _Visitor extends SimpleAstVisitor<void> {
 
   @override
   void visitCompilationUnit(CompilationUnit node) {
-    LibraryElement library = node == null
-        ? null
-        : resolutionMap.elementDeclaredByCompilationUnit(node)?.library;
+    LibraryElement library =
+        resolutionMap.elementDeclaredByCompilationUnit(node)?.library;
     manager = library == null ? null : new InheritanceManager(library);
   }
 
--- a/lib/src/rules/avoid_positional_boolean_parameters.dart
+++ b/lib/src/rules/avoid_positional_boolean_parameters.dart
@@ -2,8 +2,10 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-import 'package:analyzer/analyzer.dart';
 import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/standard_resolution_map.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/element.dart';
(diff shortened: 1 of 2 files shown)


lib/src/rules/avoid_positional_boolean_parameters.dart
```diff
@@ -102,4 +113,17 @@ class _Visitor extends SimpleAstVisitor<void> {
       DartTypeUtilities.implementsInterface(
           node.identifier.bestType, 'bool', 'dart.core') &&
       !node.isNamed;
+
+  bool _isOverridingMember(Element member) {
```
srawlins: This jazz is largely copied from `lib/src/rules/annotate_overrides.dart`. I'd extract it to be a static utility, but it uses `manager`, instantiated in `visitCompilationUnit`. I wonder if this `manager` might become useful enough to create a parent visitor. `abstract class InheritanceAwareAstVisitor extends AstVisitor` or something...

pq: @bwilkerson: thoughts on that?

lib/src/rules/avoid_positional_boolean_parameters.dart
```diff
@@ -60,8 +60,18 @@ class AvoidPositionalBooleanParameters extends LintRule
 class _Visitor extends SimpleAstVisitor<void> {
   final LintRule rule;
 
+  InheritanceManager manager;
+
   _Visitor(this.rule);
 
+  @override
+  void visitCompilationUnit(CompilationUnit node) {
+    LibraryElement library = node == null
```
bwilkerson: Visit methods should never be invoked with a null argument, so we shouldn't need this test. Did you run into an actual problem, or were you just being defensive?

lib/src/rules/avoid_positional_boolean_parameters.dart
bwilkerson: We haven't decided yet whether InheritanceManager will be part of the public API or whether we'll have a different mechanism for answering questions like this. I'd postpone linter changes until the analyzer API is finalized.

lib/src/rules/avoid_positional_boolean_parameters.dart
srawlins: Literally just copied from `lib/src/rules/annotate_overrides.dart`. Good point. Deleting

lib/src/rules/avoid_positional_boolean_parameters.dart
srawlins: SG

PR #1061, opened 2018-07-03
Update prefer_equal_for_default_values.dart

davidmorgan:
We had feedback "wasn't sure what to do here"--maybe this is clearer?

### DIFF
--- a/lib/src/rules/prefer_equal_for_default_values.dart
+++ b/lib/src/rules/prefer_equal_for_default_values.dart
@@ -7,7 +7,7 @@ import 'package:analyzer/dart/ast/token.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
 import 'package:linter/src/analyzer.dart';
 
-const _desc = r'Prefer equal for default values.';
+const _desc = r'Use equal instead of colon when specifying default values.';
 
 const _details = r'''
 
--- a/lib/src/rules/prefer_equal_for_default_values.dart
+++ b/lib/src/rules/prefer_equal_for_default_values.dart
@@ -7,11 +7,13 @@ import 'package:analyzer/dart/ast/token.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
 import 'package:linter/src/analyzer.dart';
 
-const _desc = r'Use equal instead of colon when specifying default values.';
+const _desc = r'Use `=` to separate a named parameter from its default value.';
 
 const _details = r'''
 
-**DO** Use `=` to define default values.
+From the [style guide](https://www.dartlang.org/guides/language/effective-dart/usage):
+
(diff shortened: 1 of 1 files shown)


a14n:
We should perhaps use [the same sentence as in style guide](https://www.dartlang.org/guides/language/effective-dart/usage#do-use--to-separate-a-named-parameter-from-its-default-value).

>DO use `=` to separate a named parameter from its default value.

davidmorgan:
There's also a suggestion that we somehow include a link to the style guide

https://www.dartlang.org/guides/language/effective-dart/usage#do-use--to-separate-a-named-parameter-from-its-default-value

lib/src/rules/prefer_equal_for_default_values.dart
```diff
@@ -7,7 +7,7 @@ import 'package:analyzer/dart/ast/token.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
 import 'package:linter/src/analyzer.dart';
 
-const _desc = r'Prefer equal for default values.';
+const _desc = r'Use equal instead of colon when specifying default values.';
```
pq: As you suggest, perhaps a link to the style guide in the details?

See, for example:

https://github.com/dart-lang/linter/blob/36bc8fbf15e0156a8f25d87d60cffd41b412f892/lib/src/rules/camel_case_types.dart#L14

davidmorgan:
PTAL; made the language match the style guide, both in the desc and the details, and added a link.

lib/src/rules/prefer_equal_for_default_values.dart
```diff
@@ -7,11 +7,13 @@ import 'package:analyzer/dart/ast/token.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
 import 'package:linter/src/analyzer.dart';
 
-const _desc = r'Prefer equal for default values.';
+const _desc = r'Use `=` to separate a named parameter from its default value.';
```
a14n: I think backticks are not allowed here as the desc appears in output without markdown.

davidmorgan: I think people will still be able to read it :) but if there's an established style for quotes in linter descs, I should follow it...

pq: Hmmm...  We do this elsewhere actually:

https://github.com/dart-lang/linter/blob/36bc8fbf15e0156a8f25d87d60cffd41b412f892/lib/src/rules/prefer_is_not_empty.dart#L18

So I'd just leave the formatting in.

If we want to get serious about unifying the format, we should update the docs.  @a14n : feel free to open an issue if that interests you.

Thanks!

davidmorgan:
Works for me. Thanks!

PR #1064, opened 2018-07-03
enable lint always_declare_return_types

a14n:


pq:
So, I admit I'm not in favor of enabling this lint.  In particular, I think it just adds noise for visitor implementations and is really gratuitous in tests.

a14n:
I tends to always type methods so I don't see those changes as noise (moreover if this can catch bugs). In linter codebase the `Visitor<void>` pattern been almost everywhere I may understand that all those `void` could be seen as noise. However if someone creates a lint using a `Visitor<String>` this would catch runtime errors if a visit method return `int` because its signature was `visitXXX(node)` without return type.

But I'm not an integrist of types (someone that adores `always_specify_types`) and I can close this PR without problems :)

pq:
So, to clarify a little, I'm very much in favor of this rule for any kind of public API where explicit types serve as valuable documentation.  Having said that, the more folks contribute to the linter I realize there's a sense in which it's all a bit public and we should strive for clarity (maybe, at times, at the expense of brevity and shortcuts favored by, possibly, only me) in places I never anticipated it being valuable.  Especially if this change might reduce the chance of bugs.  In that case, best I just get over myself! 😄 

(Funny aside, I *implemented* this rule!)

@bwilkerson : I'm actually curious, do you follow this advice in `analyzer`?  I think maybe except looser wrt to tests?

pq:
FWIW, as an example of where I think this rule is over-the-top, requiring me to define `_countAndReport` as `void` which is indicated to be private and is defined in a **private class**  `_CountVisitor`, feels overly fussy; all of the context for understanding that method is on one screen...  Plus for `main` and for test methods?  Yuck!

a14n:
I'll hesitate to send my next PR where I enable `omit_local_variable_types`, `prefer_const_constructors`, `prefer_const_declarations` and `prefer_final_locals` :rofl:

pq:
@a14n : would you mind opening a separate issue where we can discuss proposed lints?

Great conversations to have! 👍

bwilkerson:
> ... do you follow this advice in analyzer?

I enabled the lint to find out. Outside of tests there are over 150 violations reported, so I guess the answer is 'no'. I didn't look at them to see whether there was a pattern.

Personally, I almost always include a return type outside of `test_` methods, but not all of the code was written or reviewed by me.

pq:
As per the conversation on #1069, @a14n: if you'd like to pursue this one, let's follow-up w/ an issue to further discuss?

Issue #1068, opened 2018-07-06
what lints to enable in the linter codebase


a14n:
There are the following 45 lints that are not enabled in `analysis_options.yaml`.

```yaml
#    - always_declare_return_types
#    - always_put_control_body_on_new_line
#    - always_specify_types
#    - avoid_as
#    - avoid_bool_literals_in_conditional_expressions # not yet tested
#    - avoid_classes_with_only_static_members
#    - avoid_double_and_int_checks # not yet tested
#    - avoid_field_initializers_in_const_classes # not yet tested
#    - avoid_function_literals_in_foreach_calls
#    - avoid_js_rounded_ints # not yet tested
#    - avoid_private_typedef_functions # not yet tested
#    - avoid_relative_lib_imports # not yet tested
#    - avoid_renaming_method_parameters # not yet tested
#    - avoid_single_cascade_in_expression_statements # not yet tested
#    - avoid_types_as_parameter_names # not yet tested
#    - avoid_types_on_closure_parameters
#    - avoid_unused_constructor_parameters # not yet tested
#    - cascade_invocations
#    - close_sinks # https://github.com/dart-lang/linter/issues/268
#    - constant_identifier_names
#    - curly_braces_in_flow_control_structures # not yet tested
#    - file_names # not yet tested
#    - lines_longer_than_80_chars # not yet tested
#    - non_constant_identifier_names
#    - null_closures # not yet tested
#    - omit_local_variable_types
#    - one_member_abstracts
#    - prefer_bool_in_asserts # not yet tested
#    - prefer_const_constructors
#    - prefer_const_declarations # not yet tested
#    - prefer_const_literals_to_create_immutables # not yet tested
#    - prefer_final_locals
#    - prefer_function_declarations_over_variables
#    - prefer_generic_function_type_aliases # not yet tested
#    - prefer_interpolation_to_compose_strings
#    - prefer_iterable_whereType # not yet tested
#    - prefer_typing_uninitialized_variables # not yet tested
#    - public_member_api_docs
#    - sort_constructors_first
#    - sort_unnamed_constructors_first
#    - super_goes_last
#    - type_annotate_public_apis
#    - unnecessary_parenthesis # not yet tested
#    - unnecessary_statements # not yet tested
#    - void_checks # not yet tested
```

Some are not compatible between each other. But a lot of them could be enable to improve the code (and dogfooding our lints).

Those that are commented out should have an explanation to know why.

bwilkerson:
[...]

Issue #1051, opened 2018-06-29
Tools to generate your analysis_options text


srawlins:
I think it would be crazy slick, and not super hard, to write an interactive web tool, or a cmdline tool, that walks a user through each available lint, showing the BAD and the GOOD examples, and asking, "Is this something you want to enforce?", and giving you a `linter:` section for your analysis_options.yaml.

I partly came to this idea because I was wondering how a team/person is supposed to _update_ their list of lint warnings, once they've used the tool for a few months, and 15 more lint rules have been added to the functionality. The tool could start with "what rules do you have enabled today?" to present you with the ones you're missing.

pq:
I *love* this idea.

> your pubspec.yaml.

I think you mean `analysis_options.yaml`?

In practice, I think this would be greatly improved by some default sets a la #288.  Answering 100+ questions would be a drag!

zoechi:
I'm thinking about a config file for a mono-repo
where I define one or more sets of rules I want to use

Some exceptions because of 
- unfixed bugs in linter rules or Dart
- not matching my project style guide
- not matching a specific project type (Angular, Server, Flutter, ...)

And then have an update check that checks if new rules became available since the last check
This check would allow me to interactively accept 
or exclude with a reason (project, project style guide)
and then update an `analysis_options.linter.yaml` in each project
and also the config file that remembers which rule I excluded for what reason.

I would split it out in `analysis_options.linter.yaml` and then import this file in `analysis_options.yaml` for easier handling and avoiding issues with updating manually edited files. 

As far as I remember the analyzer looks upwards for `analysis_options.yaml` files. 
If this is the case then for a mono-repo there could be a general `analysis_options.linter.yaml` with rules used for all sub-projects in the repo root and in each project only the rules that are not shared with the other sub-projects.

bwilkerson:
Not to discourage the original definition, but there are several of enhancements I can see people quickly asking for. (I was typing this when @zoechi's comment came in, just to prove my point :-) A couple that come to mind are:

* Also keep track of the rules that were explicitly rejected so we don't ask about them again.
* Provide a way to review and change previous decisions.

Eventually, I think this turns into the equivalent of a settings dialog, or even more.

[...]

PR #1069, opened 2018-07-06
add missing lints in analysis_options.yaml

a14n:


pq:
Thinking out loud: should we consider adding a validation step to the tests to ensure this stays up to date?  See also #1051, which has I think some relevant conversation.

analysis_options.yaml
```diff
@@ -16,20 +16,30 @@ linter:
     - annotate_overrides
     - avoid_annotating_with_dynamic
 #    - avoid_as
+#    - avoid_bool_literals_in_conditional_expressions # not yet tested
```
pq: Actually, on reading this again, I'm not sure this note is totally accurate.  Maybe a link to #1068 instead?  Maybe something like:

    # under review (see #1068)

a14n:
PTAL

pq:
Super.  Thanks!

Issue #1074, opened 2018-07-12
Publish and document Google rule set.


pq:
As mentioned by @davidmorgan in #1068, a Google-internal ruleset is evolving.  While still a work in progress, I think it'd be valuable to publish it and keep it up to date as a possible jumping off point for external projects.

pq:
I've gone ahead and done the minimal thing but @davidmorgan suggested possibly adding further rationale.

> A secondary issue is that the set that makes sense for external projects may be different. For example, we decided not to enforce empty_statements because we already enforce use of dartfmt, so empty statements already stand out. I guess we can publish the reasoning, too, and people can go from there.

We could consider a small accompanying style guide or maybe better still have brief notes inline?

pq:
First cut done w/ #1075.

PR #1075, opened 2018-07-12
Google internal rules

pq:
Follow-up from: #1068.

See also: #1074.

/cc @bwilkerson @davidmorgan

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,8 +30,7 @@ you could lint your package like this:
 
     $ dartanalyzer --options analysis_options.yaml .
     
-and see any violations of the `annotate_overrides`, `hash_and_equals`, and `prefer_is_not_empty` rules in the console.  In practice you would probably configure quite a few more rules (the full list is [here][lints]).  The evolving list of lints
-enforced internally at Google may serve as inspiration [here][google-lints].
+and see any violations of the `annotate_overrides`, `hash_and_equals`, and `prefer_is_not_empty` rules in the console.  In practice you would probably configure quite a few more rules (the full list is [here][lints]).  The evolving list of lints enforced internally at Google may serve as inspiration [here][google-lints].
 
 If a specific lint warning should be ignored, it can be flagged with a comment.  For example, 
 
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ you could lint your package like this:
 
     $ dartanalyzer --options analysis_options.yaml .
     
-and see any violations of the `annotate_overrides`, `hash_and_equals`, and `prefer_is_not_empty` rules in the console.  In practice you would probably configure quite a few more rules (the full list is [here][lints]).  The evolving list of lints enforced internally at Google may serve as inspiration [here][google-lints].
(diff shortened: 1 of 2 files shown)


bwilkerson:
I don't think that we should publish this unless there is a commitment to maintain it as the internal rule set changes.

If we have such a commitment, then I think that the header comment should be updated to indicate that the rule set is a living artifact that will evolve over time. We might also want a standard copyright notice.

README.md
```diff
@@ -30,7 +30,7 @@ you could lint your package like this:
 
     $ dartanalyzer --options analysis_options.yaml .
     
-and see any violations of the `annotate_overrides`, `hash_and_equals`, and `prefer_is_not_empty` rules in the console.  In practice you would probably configure quite a few more rules (the full list is [here][lints]).
+and see any violations of the `annotate_overrides`, `hash_and_equals`, and `prefer_is_not_empty` rules in the console.  In practice you would probably configure quite a few more rules (the full list is [here][lints]).  The evolving list of lints enforced internally at Google may serve as inspiration [here][google-lints].
```
zoechi: Might be just my personal preference but I dislike links with text `here`.  
Better would be a text that tells what's behind the link.
Why not `... rules ([the full list][lints])`

pq:
> I don't think that we should publish this unless there is a commitment to maintain it as the internal rule set changes.

👍 I don't see why we couldn't commit.

> If we have such a commitment, then I think that the header comment should be updated to indicate that the rule set is a living artifact that will evolve over time. We might also want a standard copyright notice.

Good call.

README.md
pq: I totally agree.  Thanks!

bwilkerson: My writing style is somewhat different, so the tone might not match, but I'd write something like this:
```
... more rules.

To help you choose the rules you want to enable for your package, we have provided both a [complete list of rules][lints] and a list of the [lints that are enforced internally at Google][google-lints].
```

zoechi: perfectly fine for me as well

pq: Updated to @bwilkerson's suggestion. 👍

pq:
OK, some updates.  Word-smithing welcome.

(Off to find ☕️.)

example/google.yaml
```diff
@@ -0,0 +1,23 @@
...
+
+linter:
+  rules:
+    - avoid_empty_else
+    - avoid_return_types_on_setters
+    - avoid_types_as_parameter_names
+    - control_flow_in_finally
+    - no_duplicate_case_values
+    - prefer_contains
+    - prefer_equal_for_default_values
```
davidmorgan: prefer_is_not_empty landed today :)

davidmorgan:
Looks good, thanks!

Since I'm responsible for updating the internal list, I'm happy to commit to keeping the two in sync. I can also add a bit of color, e.g. rejected rules.

example/google.yaml
pq: 🎉

pq:
> Since I'm responsible for updating the internal list, I'm happy to commit to keeping the two in sync. I can also add a bit of color, e.g. rejected rules.

Awesome.  Thanks a million!

PR #1087, opened 2018-07-19
special case with ignore for lines_longer_than_80_chars

### LINKED ISSUE #1081: lines_longer_than_80_chars should not lint `// ignore: ...` lines
image

I don't see a way around this being longer than 80

a14n:
Fixes #1081


lib/src/rules/lines_longer_than_80_chars.dart
```diff
@@ -189,7 +189,12 @@ class _AllowedCommentVisitor extends SimpleAstVisitor {
     if (content.startsWith('///')) {
       lines.add(content.substring(3));
     } else if (content.startsWith('//')) {
-      lines.add(content.substring(2));
+      final commentContent = content.substring(2);
+      if (commentContent.trimLeft().startsWith('ignore:')) {
+        allowedLines.add(lineInfo.getLocation(comment.offset).lineNumber);
+      } else {
+        lines.add(content.substring(2));
```
bwilkerson: Perhaps `lines.add(commentContent);`

PR #1092, opened 2018-07-20
improved description

pq:
Pub analysis flagged this as too short and I listened!

image

Word-smithing welcome (but do keep it above 60 chars! 😉)

/cc @bwilkerson


pq:
Unrelated build breakage tracked in #1093 .

Issue #288, opened 2016-08-17
Meta linter rules


Sfshaza:
The list of [linter rules](http://dart-lang.github.io/linter/lints/) is overwhelmingly long. How about creating two meta rules:
- "Strict Dart" which follows the styles and conventions to the letter
- "Loose Dart" which includes only the most critical rules. I'd like to have this set of rules added to DartPad.

@kevmoo @kwalrath @lukechurch

matanlurey:
+1. It would be nice to have "Loose Dart" more or less match only Dart style guide.

pq:
👍 

At first blush, I see a couple of sub-tasks:

*  [ ] Agree on what rules belong in various groupings (e.g., "strict" vs. "loose" vs. "pedantic" vs. "fly-by-night")
* [ ] Agree on how these groupings should be used.  In particular are the groupings
    1. first-class in the sense that they can be referred to by name in an `.analysis_options` file, or more simply just
    2. captured in exemplar files hosted on our io page (such as the one with all rules [here](http://dart-lang.github.io/linter/lints/options/options.html))

kwalrath:
"Strict" vs. "loose" seems hard to define. Maybe we should be explicit about whether the rules are in [Effective Dart](https://www.dartlang.org/guides/language/effective-dart) and, if so, whether they're DO/DON'T or just PREFER/AVOID/CONSIDER. Here's a first cut (cc @munificent):
* effective-dart-dos-and-donts
* effective-dart-all (an alternative: effective-dart-other, but that's less clear)
* common-errors (this would cover unclosed sinks, e.g.)
* pub-rules
* (any other groups? project-specific ones, like flutter? a way to define project-specific rules? a way to turn off a specific error within a group?)

bwilkerson:
We've talked many times about supporting some kind of user-defined lint group mechanism, and I think it makes a lot of sense. My personal preference would be to support an arbitrary number of groups rather than define a fixed number of hard coded groups, but that would be more work, so it probably won't happen.

One problem you'll run into (by not creating user-definable groups) is the issue of backward compatibility. If you define a "common-errors" group, and then someone writes a new lint that catches a common error so you add it to the group, that can be a breaking change that blocks roles into g3. You basically need to accept responsibility for fixing g3 breakage before adding new lints to any of the groups.

kwalrath:
@bwilkerson if line-item vetoes are possible, then fixing breakages should be easier.

zoechi:
[...]

PR #1106, opened 2018-08-02
Add Effective Dart analysis options list

tenhobi:
As is described in #288, I have tried to set a list of options describing Effective Dart analysis options. I have also included (in comments) unuses options so it will be simplier to see what can be added / what is missing.

All rules are sorted in sections and subsections according to Effective Dart, so it will be easier to sync the rules in time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ you could lint your package like this:
 
     $ dartanalyzer --options analysis_options.yaml .
     
-and see any violations of the `annotate_overrides`, `hash_and_equals`, and `prefer_is_not_empty` rules in the console.  To help you choose the rules you want to enable for your package, we have provided a [complete list of rules][lints], a list of the [lints according to Effective Dart][effective-dart-lints] and a list of the [lints that are enforced internally at Google][google-lints].
+and see any violations of the `annotate_overrides`, `hash_and_equals`, and `prefer_is_not_empty` rules in the console.  To help you choose the rules you want to enable for your package, we have provided a [complete list of rules][lints], an up-to-date list of the [lints according to the Effective Dart guide][effective-dart-lints] and a list of the [lints that are enforced internally at Google][google-lints].
 
 If a specific lint warning should be ignored, it can be flagged with a comment.  For example, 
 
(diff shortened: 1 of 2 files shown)


example/effective-dart.yaml
```diff
@@ -0,0 +1,160 @@
+# List of rules following the Effective Dart guide.
+# https://www.dartlang.org/guides/language/effective-dart
+
```
pq: Can we call out the incompleteness / in-progress-ness of this list?  Similar to what we're doing here:

https://github.com/dart-lang/linter/blob/97a6ec9e061c9bd0c0d85f7587f78d856ecb8ece/example/google.yaml#L7

(Possibly also in the `README`?)

tenhobi: Sure, I will add the notes.

tenhobi:
@pq I have updated the notes.

README.md
```diff
@@ -32,7 +32,7 @@ you could lint your package like this:
 
     $ dartanalyzer --options analysis_options.yaml .
     
-and see any violations of the `annotate_overrides`, `hash_and_equals`, and `prefer_is_not_empty` rules in the console.  To help you choose the rules you want to enable for your package, we have provided both a [complete list of rules][lints] and a list of the [lints that are enforced internally at Google][google-lints].
+and see any violations of the `annotate_overrides`, `hash_and_equals`, and `prefer_is_not_empty` rules in the console.  To help you choose the rules you want to enable for your package, we have provided a [complete list of rules][lints], an up-to-date list of the [lints according to the Effective Dart guide][effective-dart-lints] and a list of the [lints that are enforced internally at Google][google-lints].
```
pq: I wonder if this doesn't suggest that the list is complete...  What about something like:

> a growing list of [lints derived from the Effective Dart guide]

or

> an evolving list of [lints that are enforce advice in the Effective Dart guide]

@bwilkerson : you're good at word-smithing... Any ideas?

bwilkerson: In addition to making it clear that the rule set will be expanded as new rules are added, we probably need to be clear that not everything suggested in the guide has a rule associated with it (and hence the rule set will always be incomplete in some sense).

pq:
Super.  Thanks @tenhobi!

PR #1112, opened 2018-08-06
Simplify public_member_api_docs and adjust some types

srawlins:
* The mapping and keys of `setters` is never used. This change reduces memory and computation.
* Only `add` and `contains` is ever called on `missingDocs`; `Set` will be more performant.


pq:
It looks like we're seeing some integration tests w/ `no-preview-dart-2`:

https://travis-ci.org/dart-lang/linter/jobs/412842018

@srawlins: thoughts?

srawlins:
Rebased and rerunning travis...

srawlins:
Weird, passes locally. How do you debug

```none
Exception while using a _Visitor to visit a ClassDeclarationImpl in CompilationUnitImpl
00:31 +60 -1: integration public_member_api_docs lint lib/ sources and non-lib/ sources [E]
  Expected: <1>
    Actual: <63>
```

?

pq:
Are you running it like this?

     dart --no-preview-dart-2 --enable-asserts test/all.dart

bwilkerson:
We probably need to stop using `--no-preview-dart-2` so that we're using the same semantics that will be used when running inside the analyzer.

srawlins:
Ohhhhhh this uses an omitted `new` :stuck_out_tongue:  I'll wait to land till we're Dart 2.0-only.

pq:
> Ohhhhhh this uses an omitted new 😛 I'll wait to land till we're Dart 2.0-only.

👍

Alternatively, you could add the `new` and we can do a round of cleanup with `dartfmt --fix`.

> We probably need to stop using --no-preview-dart-2 so that we're using the same semantics that will be used when running inside the analyzer.

@devoncarew: do you know when we can get rid of this extra bot?

devoncarew:
We've removed the dart 1 only snapshot of the analysis server, so that's no longer a blocker.

The analyzer and analysis server's tests to also run on a --no-preview-dart-2 bot, but there are plans to remove that. These are the pkg bots (https://logs.chromium.org/logs/dart/buildbucket/cr-buildbucket.appspot.com/8938899712132090608/+/steps/package_unit_tests/0/stdout), and the 'package unit tests' build step.

pq:
Thanks @devoncarew!

srawlins:
@pq this now passes; Dart 1 is dead!

Issue #1102, opened 2018-07-30
Engine and integration test import from `bin`


grouma:
Multiple tests have the following import:

```
import '../bin/linter.dart' as dartlint;
```

This is an invalid import and causes kernel compilation to fail. The required logic should be moved under `lib` and the tests should be updated.

bwilkerson:
How soon does this need to happen?

grouma:
I have a workaround inside Google3 which runs the corresponding binaries from source instead of kernel, so it isn't a huge priority. Ideally the next release it is fixed though.

pq:
I wonder if this is also related to #1104?

I'll take a crack at a refactor.

PR #1114, opened 2018-08-07
move cli main into `lib/`

pq:
Fixes: #1102.

/cc @bwilkerson @grouma @a14n


lib/src/cli.dart
```diff
@@ -0,0 +1,187 @@
...
+  var customPackageRoot = options['package-root'];
+  if (customPackageRoot != null) {
+    lintOptions.packageRootPath = customPackageRoot;
+  }
+
+  var packageConfigFile = options['packages'];
+
+  if (customPackageRoot != null && packageConfigFile != null) {
+    errorSink.write("Cannot specify both '--package-root' and '--packages'.");
+    exitCode = unableToProcessExitCode;
```
kevmoo: pro tip: return the exit code from this method – set it in `bin`. Makes testing MUCH easier.

lib/src/cli.dart
```diff
@@ -0,0 +1,187 @@
...
+        fileCount: linter.numSourcesAnalyzed,
+        fileRoot: commonRoot,
+        showStatistics: stats,
+        machineOutput: options['machine'],
+        quiet: options['quiet'])
+      ..write();
+    // ignore: avoid_catches_without_on_clauses
+  } catch (err, stack) {
+    errorSink.writeln('''An error occurred while linting
+  Please report it at: github.com/dart-lang/linter/issues
```
kevmoo: Move this catch to `bin` – makes testing easier

kevmoo: ...make sure to set the exit code to an error!

lib/src/cli.dart
pq: Great point.  That said, I think these cases are pretty well covered in our integration tests.  Are you making a general point or do you see cases we're missing?

kevmoo:
General point – I find it much easier to reason about code that way.

Issue #1107, opened 2018-08-03
void_checks is too strict


MichaelRFairhurst:
I don't think it should cover this case:

```
void Function(T) f;
T f2(T) { ... }

f = f2;
```

It is very common that functions which have side-effects also return a value, because (for instance) otherwise that value is completely lost:

```
list.removeAt(1); // does not return void, returns the removed item
```

That makes it impossible to assign it to places purely for the side-effect with this lint turned on.

```
Stream<int> f;
f.listen(list.removeAt); // LINT
```

This, to me, is very different than something like:

```
Completer<void>().complete(list.removeAt);
```

In the former case, you are passing a value in a place where it is used, and a part of that value is discarded but there's a lot of indication you knew what you were doing (you passed a unary int function in where a unary int function was expected). And the code in that form is the cleanest it can be.

In the latter case, the entire value you provided is guaranteed (ish*) to be discarded, and there's no indication that there's any relationship between the value asked and the value provided (except in that everything is related to `void`). And the code would be cleaner as:

```
// if "list.removeAt" is an expression with side-effects, it should go here. Otherwise it should be dropped.
Completer<void>().complete(null);
```

I hate to present this contrarian case because its an edge case that was probably fun & challenging to cover (nice work :)) but I unfortunately don't think it's a service to users and should be moved to a different lint if there's disagreement about that, probably.

MichaelRFairhurst:
I can provide numbers on this: I have a large codebase broken into build targets with bazel, and about a fifth of our build targets either are broken from this lint or depend on targets that are broken from this lint.

Removing this case (I think I did so correctly...!) cut it down to about half (so about 10%). Three errors are the cause of 9.5% of that.

I didn't see how much about how the other 10% breaks down (could also be that 9.5% is caused by 3 cases). So the numbers aren't completely fair.

In any case, those are the numbers that I have that I can share :)

eernstg:
*tl;dr Agreed, it is too strict to flag `f = f2`  rd;lt*

**Voidness preservation** didn't make it into Dart 2 as a part of the type system, but we may still introduce it as a lint. Maybe `void_checks` is that lint, already?

[...]

PR #1115, opened 2018-08-07
Relax void checks: allow T Function() to be assigned to void Function().

MichaelRFairhurst:
See #1107. This is likely too strict and should be relaxed.

### DIFF
--- a/test/rules/void_checks.dart
+++ b/test/rules/void_checks.dart
@@ -112,7 +112,7 @@ async_function() {
 
 inference() {
   f(void Function() f) {}
-  f(() // OK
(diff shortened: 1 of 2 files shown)


test/rules/void_checks.dart
```diff
@@ -112,7 +112,7 @@ async_function() {
 
 inference() {
   f(void Function() f) {}
-  f(() //LINT
+  f(() // OK
```
a14n: I'd prefer to keep a LINT here. I think it's fine to relax for method/function references but not for lambda.

PR #1117, opened 2018-08-07
New rule: prefer_relative_imports

### LINKED ISSUE #1101: Add a lint for preferring relative paths within your own package
Proposed lint: `prefer_relative_imports`

This lin is relative to this guide prefer rule: https://www.dartlang.org/guides/language/effective-dart/usage#prefer-relative-paths-when-importing-libraries-within-your-own-packages-lib-directory.

Good:

```
import 'package:other_package/foo.dart';

import 'bar.dart';

...
```

Bad:

```
import 'package:other_package/foo.dart';

import 'package:my_package/bar.dart';

...
```

tenhobi:
Fixes #1101

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -97,6 +97,7 @@ linter:
 #    - prefer_interpolation_to_compose_strings
     - prefer_is_empty
     - prefer_is_not_empty
+    - prefer_relative_imports
 #    - prefer_iterable_whereType # under review (see #1068)
     - prefer_single_quotes
 #    - prefer_typing_uninitialized_variables # under review (see #1068)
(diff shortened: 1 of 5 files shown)


lib/src/rules/prefer_relative_imports.dart
```diff
@@ -0,0 +1,93 @@
...
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:linter/src/analyzer.dart';
+
+const _desc =
+    r"Prefer relative paths when importing libraries within your own package's lib directory.";
+
+const _details = r'''
+
+When referencing a library inside your package’s `lib` directory from another library in that same package, prefer use relative URI.
```
bwilkerson: The word "referencing" implies to me either `import`, `export`, `part` or `part of` directives. Should this be expanded to cover the other cases?

analysis_options.yaml
```diff
@@ -97,6 +97,7 @@ linter:
 #    - prefer_interpolation_to_compose_strings
     - prefer_is_empty
     - prefer_is_not_empty
+    - prefer_relative_imports
```
pq: I don't think we want to enable this here since it will break all over the place.  (We tend to use absolutes in the linter code base.)

PR #1128, opened 2018-08-11
New rule: avoid_void_async. `void f() async`, suggest `Future<void>`.

### LINKED ISSUE #1122: should void_checks lint "void async"?
The language team deliberately decided that it should be OK to make a void async method:

```
void f() async {
  ...
}
```

This makes sense in that a void method _always_ can do async work, it would just take boilerplate otherwise:

```
void f() {
  () async {
    ...
  }();
}
```

This is a valid line of reasoning IMO.

However, the opposite line of reasoning makes less sense:

```
void f() async {
  // why should it be impossible to await this?
  await doWork();
  await doMoreWork();
}
```

Its currently legal to `await void`, but that may be clamped down in future releases. Why, in that case, would it be desirable to make async code that can't be awaited?

While there are a few good reasons (hiding implementation details, allowing non-future extensions, maybe more), these reasons are unlikely, and it seems much more common that people who write `void f() async` meant to write `Future<void> f() async`.

Assuming we lint this, we probably have the best of all worlds:

```
void f() async { // oops! lint catches it
 ...
}

// unless you meant to:

// ignore: void_checks
void f() async {
 ...
}

// which is less boilerplate than would be required if it were _illegal_:

void f() {
  () async {
    ...
  }();
}
```

Linting this will help us make `await void` illegal later. It also would catch the case where you add a `void async` to your public APIs and don't test `await`ing it, which would silently create errors only for people who then attempt to use your package.

MichaelRFairhurst:
Fix #1122.

As per discussion there, this is a *new* rule, not part of void_checks,
because it is much more likely to have low false positives.

In my experince, void_checks is very strict. It serves a useful purpose
in pushing people away from the more complex Null type by making void
useful in a way Null used to be used. However, void_checks is not
required for soundness, and it seems uncommon that the theory here
misses something practical.

For that reason, make a new rule. Users interested in making void behave
as if it were Null in certain ways can enable that, while users who just
want to make sure no async functions are declared wrong can enable this.

@eernstg

### DIFF
--- a/test/mock_sdk.dart
+++ b/test/mock_sdk.dart
@@ -164,7 +164,8 @@ library dart.async;
 
 import 'dart:math';
 
-part 'stream.dart';
+class Stream<T> {}
+abstract class StreamTransformer<S, T> {}
 
 class Future<T> {
   factory Future.delayed(Duration duration, [T computation()]) => null;
@@ -176,8 +177,6 @@ class FutureOr<T> {}
 ''', const <_MockSdkFile>[
     const _MockSdkFile('/lib/async/stream.dart', r'''
 part of dart.async;
-class Stream<T> {}
-abstract class StreamTransformer<S, T> {}
 ''')
   ]);
 
--- a/test/rules/avoid_void_async.dart
+++ b/test/rules/avoid_void_async.dart
@@ -19,13 +19,15 @@ void d() async => null; // LINT
 void e() => null; // OK
 Future<void> f() async => null; // OK
 
-void g() async* {} // LINT
+//TODO(mfairhurst enable this test once it works on appveyor
+//void g() async* {} // LINT
 void h() sync* {} // LINT
 
 Stream<void> i()
 //TODO(mfairhurst enable this test once it works on appveyor
-  //async* {} // OK
-   => null; // OK
+    //async* {} // OK
+    =>
+    null; // OK
 
 Iterable<void> j() sync* {} // OK
 
@@ -51,13 +53,15 @@ class Foo {
   void e() => null; // OK
   Future<void> f() async => null; // OK
 
-  void g() async* {} // LINT
+  //TODO(mfairhurst enable this test once it works on appveyor
+  //void g() async* {} // LINT
   void h() sync* {} // LINT
 
   Stream<void> i()
-  //TODO(mfairhurst enable this test once it works on appveyor
-    //async* {} // OK
-     => null; // OK
+      //TODO(mfairhurst enable this test once it works on appveyor
+      //async* {} // OK
+      =>
+      null; // OK
 
   Iterable<void> j() sync* {} // OK
 
--- a/test/rules/avoid_void_async.dart
+++ b/test/rules/avoid_void_async.dart
@@ -35,6 +35,10 @@ typedef void f1(int x); // OK
 typedef Future<void> f2(int x); // OK
 
 class Foo {
+  static void statica() async {} // LINT
+  static void staticb() {} // OK
+  static Future<void> staticc() async {} // OK
+
   void a() async {} // LINT
   void b() {} // OK
   Future<void> c() async {} // OK
--- a/test/rules/avoid_void_async.dart
+++ b/test/rules/avoid_void_async.dart
@@ -24,6 +24,12 @@ void h() sync* {} // LINT
 Stream<void> i() async* {} // OK
 Iterable<void> j() sync* {} // OK
 
+void get k => null; // OK
+void get l async => null; // LINT
+
+void set m(_) => null; // OK
+set n(_) => null; // OK
+
 typedef void f1(int x); // OK
 
 typedef Future<void> f2(int x); // OK
@@ -40,4 +46,7 @@ class Foo {
   void h() sync* {} // LINT
   Stream<void> i() async* {} // OK
   Iterable<void> j() sync* {} // OK
+
+  void get k => null; // OK
+  void get l async => null; // LINT
 }
--- a/test/rules/avoid_void_async.dart
+++ b/test/rules/avoid_void_async.dart
@@ -19,16 +19,9 @@ void d() async => null; // LINT
 void e() => null; // OK
 Future<void> f() async => null; // OK
 
-//TODO(mfairhurst enable this test once it works on appveyor
-//void g() async* {} // LINT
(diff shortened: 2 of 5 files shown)


test/rules/avoid_void_async.dart
```diff
@@ -0,0 +1,33 @@
...
+
+typedef Future<void> f2(int x); // OK
+
+class Foo {
+  void a() async {} // LINT
+  void b() {} // OK
+  Future<void> c() async {} // OK
+  void d() async => null; // LINT
+  void e() => null; // OK
+  Future<void> f() async => null; // OK
```
bwilkerson: I suppose that for completeness sake it might be good to have `static` versions of all of these.

Also, it would be good to add some examples of getters, setters and operators (the latter only on the instance side, of course).

bwilkerson: This is missing some test cases that I think it would be good to add. I can think of the following:

- declarations marked with `async*`,
- class members marked with `static`,
- getters, setters, and operators.

Also, it might be good to add one declaration marked with `sync` and one with `sync*` (I wouldn't necessarily duplicate those examples everywhere.

MichaelRFairhurst: It looks like setters cannot be asynchronous...the previous setter code was there from always_annotate_return_types.

Added tests for async*, sync*, non async setters, and sync/async void getters & operators, and static.

Great suggestions, thank you Brian! :)

test/rules/avoid_void_async.dart
```diff
@@ -0,0 +1,59 @@
...
+void d() async => null; // LINT
+void e() => null; // OK
+Future<void> f() async => null; // OK
+
+void g() async* {} // LINT
+void h() sync* {} // LINT
+Stream<void> i() async* {} // OK
+Iterable<void> j() sync* {} // OK
+
+void get k => null; // OK
```
bwilkerson: I'm mostly kidding, but should we have a lint for getters whose return type is `void`? Seems kind of pointless if you follow the philosophy that getters should not have side-effects. :-)

MichaelRFairhurst: It would be a good case for an easter egg...generate an "error, expected: bed time, but got: late night code that won't pass review in the morning" :)

pq:
@MichaelRFairhurst : any insights into the test failure on appveyor?:

https://ci.appveyor.com/project/pq/linter#L492

MichaelRFairhurst:
I looked into it briefly and did not see why!

Sorry for merging, I thought appveyor was just broken.

Do you want me to roll back?

pq:
> Do you want me to roll back?

Hmmm...  Maybe?  Alternatively, maybe disable this test (ideally just for windows though I'd have to think how) with a link to a tracking issue?

EDIT: I just kicked the build.  Let's wait and see if it's green.  Fingers crossed this was a transient error...

pq:
OK.  

The **bad news** it's still failing.

The **good news** is it's not merged yet so there's nothing to revert.  That said, lets figure out what's up before landing.

```
00:59 +101 -1: test\rule_test.dart: rule dart avoid_void_async [E]
  Bad state: No definition of type Stream
  #0      SummaryTypeProvider._getType (package:analyzer/src/summary/summary_sdk.dart:360:7)
  #1      SummaryTypeProvider.streamType (package:analyzer/src/summary/summary_sdk.dart:307:21)
  #2      SummaryTypeProvider.streamDynamicType (package:analyzer/src/summary/summary_sdk.dart:300:28)
  #3      ErrorVerifier._checkForIllegalReturnType (package:analyzer/src/generated/error_verifier.dart:3541:27)
  #4      ErrorVerifier.visitFunctionDeclaration (package:analyzer/src/generated/error_verifier.dart:789:7)
  #5      FunctionDeclarationImpl.accept (package:analyzer/src/dart/ast/ast.dart:5135:49)
  #6      NodeListImpl.accept (package:analyzer/src/dart/ast/ast.dart:8043:20)
  #7      CompilationUnitImpl.visitChildren (package:analyzer/src/dart/ast/ast.dart:2405:21)
  #8      RecursiveAstVisitor.visitCompilationUnit (package:analyzer/dart/ast/visitor.dart:707:10)
  #9      ErrorVerifier.visitCompilationUnit (package:analyzer/src/generated/error_verifier.dart:568:18)
  #10     CompilationUnitImpl.accept (package:analyzer/src/dart/ast/ast.dart:2398:49)
  #11     LibraryAnalyzer._computeVerifyErrors (package:analyzer/src/dart/analysis/library_analyzer.dart:443:10)
  #12     LibraryAnalyzer._analyze.<anonymous closure>.<anonymous closure> 
```

One thing to note is that here (for windows) we are running tests like so:

    pub run test -j1

whereas, for linux (e.g., travis), we're running them like this:

    dart --enable-asserts test/all.dart

I wonder if changing the travis test run to use `pub run test` would manifest the same issue?

MichaelRFairhurst:
> The good news is it's not merged yet so there's nothing to revert. That said, lets figure out what's up before landing.

Ah, yep. Great! :)

MichaelRFairhurst:
I'm not the only person using Stream in a test, but I'm the only one testing async*. So I think that's the problem. For now I'm disabling that to confirm.

MichaelRFairhurst:
Yep, just confirmed that the mock SDK (or something) is causing problems with async* specifically.

pq:
Thanks for tracing this!  What's odd is that it passes on travis...

test/mock_sdk.dart
```diff
@@ -164,7 +164,8 @@ library dart.async;
 
 import 'dart:math';
 
-part 'stream.dart';
+class Stream<T> {}
+abstract class StreamTransformer<S, T> {}
```
MichaelRFairhurst: @pq @bwilkerson 

this fixed it, on windows.

Any complaints?

Any insights into the underlying problem? I didn't see any windows paths or anything that would cause it, and, strangely, Stream works _unless the analyzer is creating a Stream to compare against the return type of an `async*` function_...

bwilkerson: > Any insights into the underlying problem?

Not off-hand. I'd have to try debugging it to see what's going on.

pq: I'm comfortable landing this as is.  Maybe open a tracking issue to capture context if we loop back to further investigate?

PR #1131, opened 2018-08-15
update unwaited_futures to check cascades

chaselatta:
During a refactor of some of our code we changed some functions which returned a void to now return a `Future<void>`. The code which called this method was used in a cascade like this:
```dart
object
  ..write('foo') // originally synchronous
  ..save();
```

After the migration, the above example did not trigger the unwaited_futures lint rule and showed up as a bug in our code because we were expecting `write` to finish before `save`. With the change `write` became a future which needed to finish before `save`.

This lint rule visits cascades to verify that there are no unwaited futures which could result in unexpected bugs. The rule still only triggers if you are inside of an async function to match the existing functionality.


bwilkerson:
Let me know if you need me to merge it for you.

chaselatta:
Thanks @bwilkerson, I don't have commit access so will need you to merge for me.

PR #1144, opened 2018-08-29
update always_specify_types to lint generic methods

### LINKED ISSUE #1138: Add a lint to require types on generic methods
Right now, there's no lint that catches (e.g.) the lack of a type argument on the `Iterable.map` method.
It would be nice to be able to know where we're missing them.
Also, ideally such a lint would not trigger if the method in question is labeled with `@optionalTypeArgs`.

cc @a14n

a14n:
Fixes #1138

### DIFF
--- a/lib/src/rules/always_specify_types.dart
+++ b/lib/src/rules/always_specify_types.dart
@@ -93,6 +93,7 @@ class AlwaysSpecifyTypes extends LintRule implements NodeLintRule {
     registry.addDeclaredIdentifier(this, visitor);
     registry.addListLiteral(this, visitor);
     registry.addMapLiteral(this, visitor);
+    registry.addMethodInvocation(this, visitor);
     registry.addSimpleFormalParameter(this, visitor);
     registry.addTypeName(this, visitor);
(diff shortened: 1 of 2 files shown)


bwilkerson:
I'm concerned about the number of false positives. You're correct that the SDK cannot reference anything outside the SDK, making it impossible (currently) to mark generic methods in the SDK as having optional type parameters.

One possibility is to split this rule into three parts:
- requiring type annotations (like for variables, etc.),
- requiring type arguments for classes, and
- requiring type arguments for methods.

That would at least keep a change like this from breaking existing users of the lint who don't want to go quite as far as to add type arguments to all method invocations.

a14n:
>One possibility is to split this rule into three parts

+1

a14n:
Split done in #1153

a14n:
As of https://github.com/dart-lang/linter/pull/1153#issuecomment-420429379 I will re-open this one and maintain the 2 branchs

a14n:
hmm, I can't reopen :-/ I created #1178

PR #1145, opened 2018-08-30
add lint flutter_style_todos

a14n:
Related to https://github.com/flutter/flutter/pull/20837#issuecomment-414489535

### DIFF
--- a/lib/src/rules/flutter_style_todos.dart
+++ b/lib/src/rules/flutter_style_todos.dart
@@ -1,0 +1,68 @@
+// Copyright (c) 2018, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/token.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:linter/src/analyzer.dart';
+
+const _desc = r'Use Flutter TODO format.';
+
+const _details = r'''
+
+**DO** Use Flutter TODO format.
+
+**GOOD:**
+```
+// TODO(username): comment
+```
+
+''';
+
+class FlutterStyleTodos extends LintRule implements NodeLintRule {
+  FlutterStyleTodos()
+      : super(
+            name: 'flutter_style_todos',
+            description: _desc,
+            details: _details,
+            group: Group.style);
+
+  @override
+  void registerNodeProcessors(NodeLintRegistry registry) {
+    final visitor = new _Visitor(this);
+    registry.addCompilationUnit(this, visitor);
+  }
+}
+
+class _Visitor extends SimpleAstVisitor<void> {
+  final LintRule rule;
+
+  _Visitor(this.rule);
+
+  @override
+  void visitCompilationUnit(CompilationUnit node) {
+    var token = node.beginToken;
+    while (token != null) {
+      _getPrecedingComments(token).forEach(_visitComment);
+      if (token == token.next) break;
+      token = token.next;
+    }
+  }
+
+  Iterable<CommentToken> _getPrecedingComments(Token token) sync* {
+    CommentToken comment = token.precedingComments;
+    while (comment != null) {
+      yield comment;
+      comment = comment.next;
+    }
+  }
+
+  void _visitComment(CommentToken node) {
+    final content = node.toString();
+    if (content.startsWith(new RegExp(r'///? ?TODO(\W|$)')) &&
+        !content.startsWith(new RegExp(r'// TODO\([a-zA-Z][^ ]+\): '))) {
+      rule.reportLintForToken(node);
+    }
+  }
+}
--- a/test/_data/flutter_style_todos/a.dart
+++ b/test/_data/flutter_style_todos/a.dart
@@ -1,0 +1,21 @@
+// Copyright (c) 2018, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// test w/ `pub run test -N flutter_style_todos`
+
+// BAD
+//TODO
+// TODO
+// TODO: bla
+// TODO(user) remove
+// TODO:(user): Button
+// TODO(user) : We
+/// nothing here
+/// TODO(user): We
+// TODO(#12357): Bad username
+
+// GOOD
+// TODO(user): We
+// TODO(user-name): We
+// TODO(user1,user2): We
--- a/example/all.yaml
+++ b/example/all.yaml
@@ -49,6 +49,7 @@ linter:
     - empty_constructor_bodies
     - empty_statements
     - file_names
+    - flutter_style_todos
     - hash_and_equals
     - implementation_imports
     - invariant_booleans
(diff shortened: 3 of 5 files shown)


lib/src/rules/flutter_style_todos.dart
```diff
@@ -0,0 +1,68 @@
...
+    while (comment != null) {
+      yield comment;
+      comment = comment.next;
+    }
+  }
+
+  void _visitComment(CommentToken node) {
+    final content = node.toString();
+    if (content.startsWith(new RegExp(r'///? ?TODO(\W|$)')) &&
+        !content.startsWith(new RegExp(r'// TODO\([a-zA-Z][^ ]+\): '))) {
```
Hixie: also allow `0-9` and `-` to be a superset of github usernames.

Hixie: or, i should say, only allow `[a-zA-Z][-a-zA-Z0-9]*`, to be a superset of github usernames but not allow URLs.

Hixie: (Actually github usernames can start with a number... but let's ignore that since we want to make sure people don't put in issue numbers here.)

test/_data/flutter_style_todos/a.dart
```diff
@@ -0,0 +1,21 @@
...
+// TODO:(user): Button
+// TODO(user) : We
+/// nothing here
+/// TODO(user): We
+// TODO(#12357): Bad username
+
+// GOOD
+// TODO(user): We
+// TODO(user-name): We
+// TODO(user1,user2): We
```
Hixie: eh, no need to support commas here i think

lib/src/rules/flutter_style_todos.dart
```diff
@@ -0,0 +1,68 @@
+// Copyright (c) 2018, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/token.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:linter/src/analyzer.dart';
+
+const _desc = r'Use Flutter TODO format.';
```
Hixie: to make this more useful, maybe `r'Use Flutter TODO format: // TODO(username): message, https://URL-to-issue.';`

lib/src/rules/flutter_style_todos.dart
```diff
@@ -0,0 +1,68 @@
...
+    CommentToken comment = token.precedingComments;
+    while (comment != null) {
+      yield comment;
+      comment = comment.next;
+    }
+  }
+
+  void _visitComment(CommentToken node) {
+    final content = node.toString();
+    if (content.startsWith(new RegExp(r'///? ?TODO(\W|$)')) &&
```
Hixie: what are the performance implications of creating a new regexp object each time?

test/_data/flutter_style_todos/a.dart
a14n: I added that because of:
https://github.com/flutter/flutter/blob/7598050fff00f9b15b08c50d8a3e2f613851874e/packages/flutter_tools/bin/fuchsia_tester.dart#L94

lib/src/rules/flutter_style_todos.dart
a14n: done

test/_data/flutter_style_todos/a.dart
Hixie: we should just change that to whoever last edited the line.

a14n:
Here's the list of lint in Flutter with `\bTODO\b`

- [packages/flutter/test/foundation/diagnostics_test.dart#L567](https://github.com/flutter/flutter/tree/a9275d2154b16f2bb4ee6474f579d00f2d04c8d6/packages/flutter/test/foundation/diagnostics_test.dart#L567)
- [packages/flutter_tools/bin/fuchsia_tester.dart#L94](https://github.com/flutter/flutter/tree/a9275d2154b16f2bb4ee6474f579d00f2d04c8d6/packages/flutter_tools/bin/fuchsia_tester.dart#L94)
- [packages/flutter_tools/lib/src/dart/analysis.dart#L165](https://github.com/flutter/flutter/tree/a9275d2154b16f2bb4ee6474f579d00f2d04c8d6/packages/flutter_tools/lib/src/dart/analysis.dart#L165)
- [packages/flutter_tools/lib/src/commands/update_packages.dart#L27]( https://github.com/flutter/flutter/tree/a9275d2154b16f2bb4ee6474f579d00f2d04c8d6/packages/flutter_tools/lib/src/commands/update_packages.dart#L27)

The 2 last are false positives. 

`///*(|.* )TODO(\W|$)` looks fine if we change [packages/flutter_tools/lib/src/commands/update_packages.dart#L27]( https://github.com/flutter/flutter/tree/a9275d2154b16f2bb4ee6474f579d00f2d04c8d6/packages/flutter_tools/lib/src/commands/update_packages.dart#L27).

test/_data/flutter_style_todos/a.dart
```diff
@@ -0,0 +1,22 @@
...
+// test w/ `pub run test -N flutter_style_todos`
+
+// BAD
+//TODO
+// TODO
+// TODO: bla
+// TODO(user) remove
+// TODO:(user): Button
+// TODO(user) : We
+/// nothing here
```
MichaelRFairhurst: DBC: Oh man, I wish this weren't in the "bad" category. This means I can't use it, because I use js style doc comments.

```
/// public docs
class ....
```

a14n: I'm not sure to understand. There's no lint on L14, only on line 15

Hixie:
`\bTODO\b` seems to do pretty good if those are the only problems! The two false positives are tricky, though.

`///*(|.* )TODO(\W|$)` seems equivalent to `//+(.* )?TODO\b` which seems fine I guess.

test/_data/flutter_style_todos/a.dart
MichaelRFairhurst: Ah, should this be moved into "GOOD" then?

bwilkerson:
Can Flutter TODO comments span multiple lines?

Hixie:
Sure:
```
// TODO(foo): bla bla bla
// bla bla bla bla bla bla.
```

bwilkerson:
What I was trying to ask is whether we need to have a test case where not all of the required characters are on the same line:
```
// TODO(super-long-user-name-that-practically-fills-the-whole-line):
// Pick a shorter user name so that I don't have to format TODO comments on multiple lines.
```
Along those lines, what about a non-TODO comment that happens to contain the string "TODO":
```
/// Return all of the TODO comments in the file.
```
And finally, we should test block comments, both doc and non-doc. (I'm guessing that it's OK to put TODOs in block comments, but we should test them either way.)

Hixie:
I definitely don't care about the extreme username case.
For Flutter, we don't use block comments so those are sort of moot as far as I'm concerned.

a14n:
In the current version:
- there's no lint in multiline comment (`/* ... */` or `/** ... */`)
- `/// Return all of the TODO comments in the file.` generates a lint.

bwilkerson:
> ... there's no lint in multiline comment ...

I'm fine with that, if that's how we want the lint to behave, but I think we should explicitly test that (and document it as well).

> ... generates a lint.

I assume that that's _not_ what we want, but again, my point is that we need to decide what we want the lint to do in that case and then add a test to prevent regressions.

pq:
I'm hoping to do a release later today.  @a14n : can I merge this one in?

a14n:
It's ok for me.

PR #1146, opened 2018-09-03
Improve unawaited_futures docs; point to package:pedantic.

davidmorgan:
Not sure quite how you want to word this, but it'd be nice to have a reference to package:pedantic in there somewhere.

Ideally we want it to be in the lint message itself, as a call to action.

That said it's not clear if we want to push for external use of `package:pedantic` yet, so if you prefer we could tweak the error message when we post process analyzer output, and not mention it externally for now.

Please let me know. Thanks!


davidmorgan:
Works for me. Let's see if we get any response at all--positive or negative, internal or external. We can always tweak. Will look at making a bit more noise about it (blog post?) later.

PR #1153, opened 2018-09-04
add 3 lints for always specify types

### LINKED ISSUE #1138: Add a lint to require types on generic methods

a14n:
Follow up on https://github.com/dart-lang/linter/pull/1144#issuecomment-417401664

Fixes #1138

### DIFF
--- a/example/all.yaml
+++ b/example/all.yaml
@@ -6,6 +6,9 @@ linter:
     - always_put_control_body_on_new_line
     - always_put_required_named_parameters_first
     - always_require_non_null_named_parameters
+    - always_specify_type_annotations
+    - always_specify_type_arguments_for_classes
+    - always_specify_type_arguments_for_methods
     - always_specify_types
     - annotate_overrides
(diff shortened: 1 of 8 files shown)


pq:
Sorry to let this one languish!  I admit I'm a little reluctant to see such a proliferation of lints (if it's avoidable).  Besides the argument for completeness, do we have the sense that this level of granularity would benefit a lot of folks?

a14n:
I've just followed https://github.com/dart-lang/linter/pull/1144#issuecomment-417401664 . I don't have a strong opinion on that.
If you prefer a single lint then #1144 is the PR to consider. Let me know what's you decide with @bwilkerson

mit-mit:

PR #1160, opened 2018-09-07
Initial commit of avoid_implementing_value_types.

matanlurey:
Closes https://github.com/dart-lang/sdk/issues/34379.

tl;dr? Enforces the contract of `operator ==`, which states:

> The overriding method must still be an equivalence relation. That is, it must be:
>
> * Symmetric: For all objects o1 and o2, o1 == o2 and o2 == o1 must either both be true, or both be false.

... this is ~mostly not possible if you `implement` class that has an overloaded `==`.

(It is technically possible if _all_ of the members are public state, but then the lint rule would have to try and enforce the method body for `==` has identical semantics, and that seems NP-incomplete).

@lrhn seems to support here: https://github.com/dart-lang/sdk/issues/34379#issuecomment-418706242, and I've gotten only positive feedback via the internal `dart-lints@` group, where I first discussed this idea. I am open to other nouns, i.e. instead of `value_types`, here.


lib/src/rules/avoid_implementing_value_types.dart
```diff
@@ -0,0 +1,122 @@
...
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/element/element.dart';
+import 'package:linter/src/analyzer.dart';
+
+const _desc = r"Don't implement classes that override `==`.";
+
+const _details = r'''**DON'T** implement classes that override `==`.
+
+The `==` operator is contracturally required to be an equivalence relation;
+that is, symmetrically for all objects `o1` and `o2`, `o1 == o2` and `o2 == oq`
```
jonahwilliams: o2 == oq -> o2 == o1

lib/src/rules/avoid_implementing_value_types.dart
```diff
@@ -0,0 +1,122 @@
...
+must either both tbe true, or both be false.
```
jonahwilliams: tbe -> be

lib/src/rules/avoid_implementing_value_types.dart
```diff
@@ -0,0 +1,122 @@
+// Copyright (c) 2018, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/analyzer.dart';
```
bwilkerson: Please replace this import with imports of more specific libraries. (This library will be deprecated in the fairly near future.)

lib/src/rules/avoid_implementing_value_types.dart
```diff
@@ -0,0 +1,122 @@
...
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/element/element.dart';
+import 'package:linter/src/analyzer.dart';
+
+const _desc = r"Don't implement classes that override `==`.";
+
+const _details = r'''**DON'T** implement classes that override `==`.
+
+The `==` operator is contracturally required to be an equivalence relation;
```
bwilkerson: "contracturally" --> "contractually"

test/rules/avoid_implementing_value_types.dart
```diff
@@ -0,0 +1,35 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+// test w/ `pub run test -N avoid_implementing_value_types`
+
+class Size {
+  final int inBytes;
+  const Size(this.inBytes);
+
+  @override
+  bool operator==(Object o) => o is Size && o.inBytes == inBytes;
```
bwilkerson: "inBytes" --> "other.inBytes" (though it doesn't actually affect the tests :-).

test/rules/avoid_implementing_value_types.dart
```diff
@@ -0,0 +1,35 @@
+// Copyright (c) 2018, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// test w/ `pub run test -N avoid_implementing_value_types`
```
bwilkerson: There should be at least one test for each of the old and new mixin styles.

matanlurey: Is the new mixin style supported? Are there other tests that currently do this?

matanlurey:
@bwilkerson:

> I'm not convinced that the premise of this rule is valid, but I'm open to being convinced.

Could you clarify - that you think the contract of `operator ==` is invalid, that you think it should not be enforced, that you think this rule (as currently written) doesn't enforce the contract, or that the rule isn't worth enforcing?

> Have you run this rule over some large code bases to see how many false positives it finds?

I have done a "light" review of google3, i.e. finding cases where `implements T` is used where `T` is a common value type (`Duration`, `DateTime`, a few others), and have mostly found these entirely to be a mistake (i.e. the purpose of this lint) - they are being mocked but mocking isn't necessary for these types of objects.

>  I'd be interested in knowing how many are found in angular, flutter, and the SDK (including the packages defined there).

I'm not intending to turn this on for the SDK or Flutter (this is for Google), so I didn't run it there - I imagine the precedent stands with many of the "Flutter-style" lints that have no applicability outside of the Flutter SDK.

test/rules/avoid_implementing_value_types.dart
matanlurey: Done!

lib/src/rules/avoid_implementing_value_types.dart
matanlurey: Done!

test/rules/avoid_implementing_value_types.dart
bwilkerson: > Is the new mixin style supported?

Analyzer supports the new syntax and generates many of the required diagnostics. It should be far enough along to allow us to start testing in in lints. The front_end package currently only supports the syntax, but that shouldn't limit us.

> Are there other tests that currently do this?

No, we haven't yet gone through the existing lints to determine where we need to add tests for the new mixin syntax.

matanlurey: Then can I push back on this requirement to land the lint? I don't know anything about the new mixin syntax, or how it would (or would not) affect this lint in a meaningful way.

bwilkerson:
> Could you clarify ...

Sure. (It seemed obvious when I wrote it :-/, but clearly wasn't.)

> When using `implements`, you do not inherit the method body of `==`, making it nearly impossible to follow the contract of `==`.

I'm not convinced that the latter part of that statement is true. You _can_ follow the contract of `==` if the information you need is publicly accessible on all instances of implementors of the interface.

I completely agree that subtyping in the presence of `==` is hard to get right, but I think that there are as many pitfalls for classes that `extend` another as there are for classes that `implement` another.

> Classes that override `==` typically are usable directly in tests _without_ creating mocks or fakes as well.

I'm not convinced that's true either. I think it's perfectly possible to write a class that overrides `==` but is better to mock for testing purposes (and I generally avoid mocking whenever possible).

I wonder whether there's a confusion here between value types and objects that implement `==`. I would totally agree that value types should not be implemented / mocked. But implementing `==` doesn't make a class a value types, nor mean that it shouldn't be implemented.

That said, we don't currently have a way to identify a class as a value type, so I'm not sure how to accurately identify value types.

> ... finding cases where `implements T` is used where `T` is a common value type (`Duration`, `DateTime`, a few others) ... so, I don't consider them false-positives ...

I don't either. I think the rule is valid for value types, just not for types that override `==`.

test/rules/avoid_implementing_value_types.dart
bwilkerson: I do think we want a test(s) for the use of mixins, even if it's only the old style. For example:
```
class MixesInSize extends Object with Size { // LINT
}
```
A similar example using the new syntax would be:
```
mixin SizeMixin {
  @override
  bool operator==(Object o) => o is Size && o.inBytes == other.inBytes;

  @override
  int get hashCode => inBytes.hashCode;
}
class MixesInSizeMixin with SizeMixin { // LINT
}
```

matanlurey: The first sample is invalid (`Size` declares a constructor), but I can make a sample that works

matanlurey: Using the new `mixin` syntax causes the file to be flagged in the analyzer as invalid.

matanlurey:
@bwilkerson:

> I'm not convinced that the latter part of that statement is true. You _can_ follow the contract of `==` if the information you need is publicly accessible on all instances of implementors of the interface.

Sure. You _can_ do anything. Lots of the lints in this package were created to avoid an obvious user error, even though you _can_ use it in a valid way. Here are a couple cherry-picked samples, but others also have this property:

* [`hash_and_equals`](https://dart-lang.github.io/linter/lints/hash_and_equals.html)
* [`prefer_void_to_null`](https://dart-lang.github.io/linter/lints/prefer_void_to_null.html)

It's true you _could_ make sure you implement `==` (and `hashCode` in an _identical_ fashion to the classe(s) you are implementing) - but for real user code this is practically impossible. For the lint to identify cases where it _is_ valid it would have to determine if 2 method bodies are semantically identical, I don't want to go down that path.

> I'm not convinced that's true either. I think it's perfectly possible to write a class that overrides `==` but is better to mock for testing purposes (and I generally avoid mocking whenever possible).

Sure, then I guess you wouldn't use this lint? The goal of every lint is not to be used in every code-base, thus being opt-in, right? I'm not sure its worth changing the definition of this lint. Here is a link to the (internal only) thread, where this lint currently has support:
https://groups.google.com/a/google.com/forum/#!topic/dart-lints/goZ_CHg-K8g - Thanks!

matanlurey:
I'm also happy to rename this to `google3_avoid_implementing_value_types` if it side-steps discussions on whether this lint is applicable to every code-bases coding style. Thanks!

lib/src/rules/avoid_implementing_value_types.dart
```diff
@@ -0,0 +1,122 @@
...
+  }
+}
+
+class _Visitor extends SimpleAstVisitor<void> {
+  final LintRule rule;
+
+  _Visitor(this.rule);
+
+  static bool _overridesEquals(ClassElement node) =>
+      node.getMethod('==') != null ||
```
MichaelRFairhurst: edge case: can be abstract

lib/src/rules/avoid_implementing_value_types.dart
```diff
@@ -0,0 +1,122 @@
...
+      _overridesEquals(node.supertype.element) ||
+      node.mixins.any((t) => _overridesEquals(t.element));
+
+  @override
+  void visitClassDeclaration(ClassDeclaration node) {
+    if (node.implementsClause == null) {
+      return;
```
MichaelRFairhurst: should this check if you extend a value class and override ==?

lib/src/rules/avoid_implementing_value_types.dart
```diff
@@ -0,0 +1,122 @@
...
+    }
+    for (TypeName interface in node.implementsClause.interfaces) {
```
MichaelRFairhurst: wouldn't it be fair to say you shouldn't mix in such classes either? Or maybe, that if you mix in a class which provides == you shouldn't override it?

MichaelRFairhurst:
(I should add: I don't think my "approval" means anything in terms of actual approval to merge)

lib/src/rules/avoid_implementing_value_types.dart
matanlurey: Done! And added a test, good catch.

lib/src/rules/avoid_implementing_value_types.dart
matanlurey: I thought about this, and this should probably be a separate lint - if at all. It's possible for example to have a `AnimalValue` and `CatValue`, where you would implement the contract correctly using `extends`.

lib/src/rules/avoid_implementing_value_types.dart
matanlurey: Similar as above, right?

matanlurey:
@MichaelRFairhurst:

> (I should add: I don't think my "approval" means anything in terms of actual approval to merge)

No worries. I don't intend to merge without getting more approval.

lib/src/rules/avoid_implementing_value_types.dart
matanlurey: I did add it. Turns out there isn't a good way to test this, because you can't re-abstract `==`.

MichaelRFairhurst:
> For the lint to identify cases where it is valid it would have to determine if 2 method bodies are semantically identical

I'd add that the linter isn't capable of proving all sorts of correctness properties. It's cool that in this case it can prove it via this lint rule, but if that correctness proof is overly burdensome then it's valid to question whether this rule should be relaxed.

Extreme example: we can't prove that code is free of bugs. Except, we _could_ do so by rejecting all non-empty programs, but that would be a bad rule. And the justification for that rule should not be "how else can we prove your code is free of bugs?"

Obviously examples of where restrictive proofs are good include....type systems as a dang whole.

This lint falls somewhere between those two extremes, like all lints.

As for examples where this rule should not apply, its notable that `Object` declares `==`. You would never write `implements Object`, but, if you did, that would *not* be a bug. Arguably that's because `Object` is an extreme type, but arguably that's what every override of `==` is equivalent to (I left comments about how it seems this should flag overides of `==`) and so counterexamples would then by definition be a larger set than examples! I'll refrain from stating my opinion on which side of those two interpretations I myself take, though.

matanlurey:
I definitely agree on the pedantic `Object` case, but also concur this unlikely to affect real code.

lib/src/rules/avoid_implementing_value_types.dart
bwilkerson: This appears to work for me:
```
class A {
  @override
  bool operator ==(other) {
    return true;
  }
}

abstract class B extends A {
  @override
  bool operator ==(other);
}
```

bwilkerson:
> I'm also happy to rename this to `google3_avoid_implementing_value_types` ...

No, the name indicates what you're trying to lint, and the semantics don't change based on whether the code is internal to Google. But it would be good to add a note to the documentation to the effect that "we don't have any way to identify value types, so this lint uses the fact that a class implements the operator `==` as a proxy for that case".

matanlurey:
Description updated! I added:

```
> _NOTE_: Dart does not have true _value types_, so instead we consider a class
> that implements `==`  as a _proxy_ for identifying value types.
```

lib/src/rules/avoid_implementing_value_types.dart
```diff
@@ -0,0 +1,127 @@
...
+    registry.addClassDeclaration(this, visitor);
+  }
+}
+
+class _Visitor extends SimpleAstVisitor<void> {
+  final LintRule rule;
+
+  _Visitor(this.rule);
+
+  static bool _overridesEquals(ClassElement node) =>
```
bwilkerson: I think the expression `element.lookUpConcreteMethod('==', element.library) != null` would do the same thing as this method, but will handle invalid code containing a cycle in the type hierarchy.

If you do need this method, you'll need to add a `visitedClasses` parameter to break cycles (and consider renaming `node` to `element`).

matanlurey: Nice! I did not know this was an option. Done, PTAL.

lib/src/rules/avoid_implementing_value_types.dart
```diff
@@ -0,0 +1,133 @@
...
+    registry.addClassDeclaration(this, visitor);
+  }
+}
+
+class _Visitor extends SimpleAstVisitor<void> {
+  final LintRule rule;
+
+  _Visitor(this.rule);
+
+  static bool _overridesEquals(
```
bwilkerson: Thanks, but I guess I should have been more clear. The method `lookUpConcreteMethod` will look in the target class and all of it's superclasses (whether introduced via `extends` or `with`). As currently implemented, this method will traverse the hierarchy multiple times. What I was proposing is that you could remove this method completely and replace it's invocation with `element.lookUpConcreteMethod('==', element.library) != null`. On further reflection, though, that's not quite right, because in this case it will always find the implementation in `Object`, so you'll need to check to see whether the result is from `Object`, rather than whether the result is `null`.

matanlurey: Oh ok, thanks. I tried the following it does not work (does not lint what the old code linted):

```dart
  static bool _overridesEquals(ClassElement element) {
    for (final interface in element.interfaces) {
      final interfaceClass = interface.element;
      final equalsMethod = interfaceClass.lookUpConcreteMethod(
        '==',
        element.library,
      );
      if (equalsMethod != null) {
        final implementingType = equalsMethod.enclosingElement.type;
        return !implementingType.isObject;
      }
    }
    return false;
  }
```

Any ideas?

bwilkerson: Wow, I'm still failing to communicate clearly. Huge apologies! Try this:
```
static bool _overridesEquals(ClassElement element) => !element
    .lookUpConcreteMethod('==', element.library)
    ?.enclosingElement
    ?.type
    ?.isObject == false;
```
The invocation of `lookUpConcreteMethod` should always return the implementation of `==` from `Object` (unless there's a concrete implementation lower in the hierarchy), except in the case where there's a cycle in the type hierarchy, hence the null checks.

(I haven't actually run that code, so I might have made a mistake.)

matanlurey: No worries. I'm still learning! 🔖 

Looks like it works! PTAL.

MichaelRFairhurst:
I thought on it more, and while it's probably harder to correctly override
== (and not fully up to the subclass), it's probably not as error prone a
pattern because its easy to implement a "value type" without thinking and
probably most overrides of == will be more deliberate.

So I'm ok with them being separate lints, if at all.

matanlurey:
Oops! Restored comment @bwilkerson!

re: @MichaelRFairhurst - I have no objection to enhancing this lint (adding more cases) or adding separate lints, I'm just currently trying to cover the most common case. I guess you could argue the lint should be `avoid_overriding_non_identity_equals`at the core?

matanlurey:
Thanks all. I'm going to give this to the end of the day before merging if anyone has any more feedback, otherwise I'll squash/merge.

PR #1170, opened 2018-09-18
add lint sort_dependencies

a14n:


lib/src/rules/pub/sort_dependencies.dart
```diff
@@ -0,0 +1,55 @@
...
+const _details = r'''
+**DO** sort dependencies in `pubspec.yaml`.
+
+Sorting list of dependencies makes maintenance easier.
+''';
+
+class SortDependencies extends LintRule {
+  SortDependencies()
+      : super(
+            name: 'sort_dependencies',
```
bwilkerson: "dependencies" is a fairly general term. Perhaps we should name this "sort_pub_dependencies" to be a little more specific.

lib/src/rules/pub/sort_dependencies.dart
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  void _visitDeps(PSDependencyList dependencies) {
+    final depsByLocation = dependencies.toList()
+      ..sort((d1, d2) => d1.name.span.start.compareTo(d2.name.span.start));
+    var previousName = '';
+    for (final dep in depsByLocation) {
+      final name = dep.name.text;
+      if (name.compareTo(previousName) < 0) {
+        rule.reportPubLint(dep.name);
```
bwilkerson: We might want to limit it to reporting a single lint per dependency group. If we've called attention to the fact that a group is out of order, I'd expect that most users would sort the whole list, not just move the one item.

test/integration_test.dart
```diff
@@ -772,6 +772,38 @@ defineTests() {
...
+
+      tearDown(() {
+        collectingOut.buffer.clear();
+        outSink = currentOut;
+        exitCode = 0;
+      });
+
+      test('check order', () async {
+        await cli.run([
+          'test/_data/sort_dependencies',
```
bwilkerson: missing file?

lib/src/rules/pub/sort_dependencies.dart
a14n: Done

test/_data/sort_pub_dependencies/pubspec.yaml
```diff
@@ -0,0 +1,11 @@
...
+
+dependencies:
+  a: any
+  c: any
+  b: any
+
+dev_dependencies:
+  e: any
+  d: any
+  f: any
```
bwilkerson: We should include dependency overrides.

bwilkerson:
Will this require publishing a new version of the analyzer package? (If only the test enhancements I suggested require it, then let's hold off on updating the test.)

pq:
@a14n : is this good to land?

a14n:
>is this good to land?

Yes

PR #1174, opened 2018-09-20
Initial commit of new lint: no_legacy_typedefs

lucyget:


lib/src/rules/no_legacy_typedefs.dart
```diff
@@ -0,0 +1,55 @@
...
+}
+
+class _Visitor extends SimpleAstVisitor {
+  final LintRule rule;
+
+  _Visitor(this.rule);
+
+  @override
+  void visitFunctionTypeAlias(TypeAlias node) {
+    if (node.typedefKeyword.next.isIdentifier) {
```
MichaelRFairhurst: I think you can do `is! GenericTypeAlias` instead of this identifier check?

test/rules/no_legacy_typedefs.dart
```diff
@@ -0,0 +1,21 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+// test w/ `pub run test -N no_legacy_typedefs`
+
+typedef String BadNoParameters(); // LINT
+typedef bool BadOneDynamicParameter(num); // LINT
+typedef bool BadOneParameter(num value); // LINT
+typedef int BadGenericWithDynamicParameters<T>(a, b); // LINT
+typedef int BadGeneric<T>(T a, T b); // LINT
+typedef Foo BadReturnsFoo(); // LINT
```
MichaelRFairhurst: Maybe test one without a return type since the returntype will be `typedefKeyword.next`? Or more complex return types too?

lib/src/rules/no_legacy_typedefs.dart
```diff
@@ -0,0 +1,55 @@
...
+      rule.reportLint(node);
```
bwilkerson: We generally make the highlight region as small as possible, so I'd recommend using `node.name` as the argument.

test/rules/no_legacy_typedefs.dart
bwilkerson: More tests won't hurt, of course, but given how simple I think the visit method can become, there really isn't much to test. :-)

lib/src/rules/no_legacy_typedefs.dart
lucyget: thanks, it's so simple now!

test/rules/no_legacy_typedefs.dart
lucyget: yea, there doesn't seem to be any more to test really

lib/src/rules/no_legacy_typedefs.dart
lucyget: Changed

a14n:
Dup of [prefer_generic_function_type_aliases](https://github.com/dart-lang/linter/blob/master/lib/src/rules/prefer_generic_function_type_aliases.dart) ?

bwilkerson:
I completely missed that, but yes, it is.

lucyget:
Should I close this pull request then?

bwilkerson:
Unfortunately, yes. But I hope that if you have other ideas for lints you'll feel free to contribute them.

lucyget:
No problem, now I know how to write lints!

I would like to suggest that the other lint perhaps be renamed though.

pq:
Thanks for the contribution @lucyget!  Sorry for the false alarm...

PR #1178, opened 2018-09-21
Always specify types generic methods

a14n:


lib/src/rules/always_specify_types.dart
```diff
@@ -164,4 +165,33 @@ class _Visitor extends SimpleAstVisitor<void> {
...
+
+  bool _isAllowed(FunctionTypedElement element) {
+    final returnType = element.returnType;
+
+    // Special case for methods with return type and parameter types identical
+    // and equal to the type parameter on function
+    // eg. T m<T>(T p1, T p2);
+    if (returnType is TypeParameterType) {
+      if (returnType.element.enclosingElement == element &&
+          element.parameters.every((p) => returnType == p.type)) {
```
bwilkerson: I think you're trying to express conditions under which type inference can infer `T`.

If that's true, it ought to be sufficient to have at least one parameter of that type rather than requiring that all parameters are of that type, such as in `T m<T>(T p1, int p2)`. More generally, I think we only need at least one parameter whose type uses that type, such as in `T m<T>(List<T> p1)`.

a14n: Here I've just implemented what @Hixie suggested in https://github.com/dart-lang/linter/issues/1138#issuecomment-417814244

> The following are cases where the type is used for both the arguments and the return value and it's fine whatever the values are, except that in the case of min and max I'd rather we banned num as an option

So @Hixie is the best person to answer.

pq:
Ok.  I hate to back-pedal but after some conversation and more thought, I'm leaning towards going with the original proposal of three separate lints after all.  In the end, I think that's the best approach.  If we do that, we'll want to do at least the following:

1. update the documentation of `always_specify_types` to indicate that it's being deprecated in favor of these more fine-grained rules
2. add an analysis that flags uses of `always_specify_types`
3. settle on a process for deprecating (and ultimately removing) lints

How does that sound to folks?

bwilkerson:
Having been in the same conversation, I would agree. But I'll note that we might want to hold off until we know whether we need / want to add lints for type arguments on methods, which now seems less certain.

a14n:
I agree :+1: 

>But I'll note that we might want to hold off until we know whether we need / want to add lints for type arguments on methods, which now seems less certain.

Why do you say that? Because of https://github.com/dart-lang/linter/issues/1138#issuecomment-423389560 ?

bwilkerson:
Yes. I want to understand what our users want before we try to implement it, but I seem to have less understanding with each passing day :-)

a14n:
Regarding [this comment](https://github.com/dart-lang/linter/issues/1138#issuecomment-423659428) I think the rules implemented in this PR match what is wanted.

mit-mit:

PR #1182, opened 2018-09-25
Fix an NPE in overridden_fields

bwilkerson:
Fixes https://github.com/dart-lang/sdk/issues/34571

### DIFF
--- a/lib/src/rules/overridden_fields.dart
+++ b/lib/src/rules/overridden_fields.dart
@@ -75,7 +75,11 @@ Iterable<InterfaceType> _findAllSupertypesAndMixins(
 
   accumulator.add(interface);
   InterfaceType superclass = interface.superclass;
-  Iterable<InterfaceType> interfaces = [superclass]
+  List<InterfaceType> interfaces = [];
+  if (superclass != null) {
+    interfaces.add(superclass);
+  }
+  interfaces
     ..addAll(interface.element.mixins)
     ..addAll(_findAllSupertypesAndMixins(superclass, accumulator));
   return interfaces.where((i) => i != interface);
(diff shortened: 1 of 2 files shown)


lib/src/rules/overridden_fields.dart
```diff
@@ -75,7 +75,11 @@ Iterable<InterfaceType> _findAllSupertypesAndMixins(
 
   accumulator.add(interface);
   InterfaceType superclass = interface.superclass;
-  Iterable<InterfaceType> interfaces = [superclass]
+  List<InterfaceType> interfaces = [];
+  if (superclass != null) {
+    interfaces.add(superclass);
```
scheglov: Hm... This whole function looks like a copy of `ClassElement.allSupertypes`. This method receives `InterfaceType`, but its client actually uses `classElement.type`, and actually does not care about types - just about presence of an accessor in a supertype.

test/rules/overridden_fields.dart
```diff
@@ -162,3 +162,7 @@ class StaticsNo {
 class VerifyStatic extends StaticsNo {
   static int a;
 }
+
+mixin M on A1 {
```
scheglov: Maybe also positive / negative tests for the lint?

lib/src/rules/overridden_fields.dart
bwilkerson: Nope. This excludes implemented types, looking only along the superclass chain.

test/rules/overridden_fields.dart
bwilkerson: Done

bwilkerson:
Adding support so that the rule correctly reports overrides in mixins required adding enough code that you should probably look at it again.

PR #1189, opened 2018-10-01
unnecessary_this on field initializers

a14n:


a14n:
BTW I wonder when `this` is required in field initializer... Never ?

bwilkerson:
> I wonder when this is required in field initializer... Never ?

Yes, never. The spec says that an initializer of the form `x = e` is equivalent to `this.x = e`.

a14n:
It's not a breaking change: it's a fix of a bug regarding the effective dart rule [DON’T use `this.` when not needed to avoid shadowing](https://www.dartlang.org/guides/language/effective-dart/usage#dont-use-this-when-not-needed-to-avoid-shadowing) :) There's even an example with that case!

a14n:
>Yes, never. The spec says that an initializer of the form x = e is equivalent to this.x = e.

Thanks! Good to know.

bwilkerson:
> There's even an example with that case!

Ah, I didn't realize that. (It's been a long time since I read the guide. :-)

PR #1192, opened 2018-10-03
Fix mixin related bug in avoid_field_initializers_in_const_classes

bwilkerson:
I can't fix the second issue until we publish a new version of analyzer, but I left a TODO so we wouldn't forget. It needs to be fixed before 2.1, but not as high a priority as getting this fix in. I'd like to get this published and pulled into the SDK asap.

### DIFF
--- a/lib/src/rules/avoid_field_initializers_in_const_classes.dart
+++ b/lib/src/rules/avoid_field_initializers_in_const_classes.dart
@@ -99,10 +99,13 @@ class _Visitor extends SimpleAstVisitor<void> {
     if (node.isStatic) return;
     if (!node.fields.isFinal) return;
     // only const class
-    if (node
-        .getAncestor<ClassDeclaration>((e) => e is ClassDeclaration)
-        .declaredElement
-        .constructors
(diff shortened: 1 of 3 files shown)


lib/src/rules/avoid_field_initializers_in_const_classes.dart
```diff
@@ -99,10 +99,13 @@ class _Visitor extends SimpleAstVisitor<void> {
     if (node.isStatic) return;
     if (!node.fields.isFinal) return;
     // only const class
-    if (node
-        .getAncestor<ClassDeclaration>((e) => e is ClassDeclaration)
-        .declaredElement
-        .constructors
+    ClassDeclaration classDeclaration =
+        node.getAncestor<ClassDeclaration>((e) => e is ClassDeclaration);
+    if (classDeclaration == null) {
```
scheglov: Can we use just `node.parent is ClassDeclaration`?
Fields are always direct children of classes.

bwilkerson: Done

PR #1202, opened 2018-10-07
Don't use computeNode() in invariant_booleans.

scheglov:
`computeNode()` is really inefficient, and we need much less info than it provides.


pq:
whoa.

if our benchmark bot is to be trusted, this is a big improvement.  🏁 

**before:**

image

**after:**

image


🌶 🌶 🌶

PR #1206, opened 2018-10-10
add lint voidness_propagation

### LINKED ISSUE #1124: voidness propagation as a lint
Dart 2 was originally going to have "voidness propagation"

Note that while this is illegal:

A:
```
void x;
int y = x; // error! use of void
```

this is not an error until runtime:

B:
```
List<void> x;
List<int> y = x;
```

and worse, this is not an error at all:

C:
```
List<void> x;
List<Object> y = x;
```

A is currently caught by the analyzer,  but B & C were supposed to be handled by the not-landed "voidness_propagation."

Currently B is both a runtime error and can be caught by turning off implicit downcasts, but C is completely silent.

We could add this to the linter in the meantime. It also could be added to void_checks, but maybe it would be better to make it "voidness_propagation" because then users who don't want "void_checks" but _do_ want to be future-proof for voidness_propagation could get what they want and nothing more. (this would be a good way to incrementally roll that out as part of the language).

@eernstg

a14n:
Fixes #1124

### DIFF
--- a/lib/src/rules/voidness_propagation.dart
+++ b/lib/src/rules/voidness_propagation.dart
@@ -85,10 +85,10 @@ class _Visitor extends SimpleAstVisitor<void> {
   void visitMethodDeclaration(MethodDeclaration node) {
     final inheritedMethod = DartTypeUtilities.lookUpInheritedMethod(node);
     if (inheritedMethod != null) {
-      _check(inheritedMethod.returnType, node.returnType.type, node.returnType);
+      _check(inheritedMethod.returnType, node.returnType?.type, node.returnType);
     } else {
       for (final e in DartTypeUtilities.overridenMethods(node)) {
-        _check(e.returnType, node.returnType.type, node.returnType);
+        _check(e.returnType, node.returnType?.type, node.returnType);
       }
     }
   }
--- a/lib/src/rules/voidness_propagation.dart
+++ b/lib/src/rules/voidness_propagation.dart
@@ -85,7 +85,8 @@ class _Visitor extends SimpleAstVisitor<void> {
   void visitMethodDeclaration(MethodDeclaration node) {
     final inheritedMethod = DartTypeUtilities.lookUpInheritedMethod(node);
     if (inheritedMethod != null) {
-      _check(inheritedMethod.returnType, node.returnType?.type, node.returnType);
+      _check(
+          inheritedMethod.returnType, node.returnType?.type, node.returnType);
     } else {
       for (final e in DartTypeUtilities.overridenMethods(node)) {
         _check(e.returnType, node.returnType?.type, node.returnType);
--- a/lib/src/rules/voidness_propagation.dart
+++ b/lib/src/rules/voidness_propagation.dart
@@ -1,0 +1,164 @@
+// Copyright (c) 2018, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/element.dart';
+import 'package:analyzer/dart/element/type.dart';
+import 'package:linter/src/analyzer.dart';
+import 'package:linter/src/util/dart_type_utilities.dart';
+
+const _desc = r'Do not loose voidness.';
+
+const _details = r'''
+
+**DO** not loose voidness. This is likely to lead to a runtime error.
+
+**BAD:**
+
+```dart
+List<void> x;
+List<int> y = x;
+```
+
+```dart
+List<void> x;
+List<Object> y = x;
+```
+
+''';
+
+class VoidnessPropagation extends LintRule implements NodeLintRule {
+  VoidnessPropagation()
+      : super(
+            name: 'voidness_propagation',
+            description: _desc,
+            details: _details,
+            group: Group.style);
+
+  @override
+  void registerNodeProcessors(NodeLintRegistry registry) {
+    final visitor = new _Visitor(this);
+    registry.addAssignmentExpression(this, visitor);
+    registry.addExpressionFunctionBody(this, visitor);
+    registry.addInstanceCreationExpression(this, visitor);
+    registry.addMethodDeclaration(this, visitor);
+    registry.addMethodInvocation(this, visitor);
+    registry.addReturnStatement(this, visitor);
+    registry.addVariableDeclaration(this, visitor);
+  }
+}
+
+class _Visitor extends SimpleAstVisitor<void> {
+  _Visitor(this.rule);
+
+  final LintRule rule;
+
+  @override
+  void visitAssignmentExpression(AssignmentExpression node) {
+    final type = node.leftHandSide?.staticType;
+    _check(type, node.rightHandSide?.staticType, node.rightHandSide);
+  }
+
+  @override
+  void visitExpressionFunctionBody(ExpressionFunctionBody node) {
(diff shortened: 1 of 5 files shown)


pq:
Interesting that we're seeing exceptions in benchmarking:

https://travis-ci.org/dart-lang/linter/jobs/439879861

Thoughts on that?

lib/src/rules/voidness_propagation.dart
```diff
@@ -0,0 +1,164 @@
...
+    if (parameters != null) {
+      _checkArgs(args, parameters);
+    }
+  }
+
+  @override
+  void visitMethodDeclaration(MethodDeclaration node) {
+    final inheritedMethod = DartTypeUtilities.lookUpInheritedMethod(node);
+    if (inheritedMethod != null) {
+      _check(inheritedMethod.returnType, node.returnType.type, node.returnType);
```
bwilkerson: Both here and below, `node.returnType` can return null when the method declaration has no declared return type.

bwilkerson:
@MichaelRFairhurst

pq:
ping: @MichaelRFairhurst 👍

lib/src/rules/voidness_propagation.dart
```diff
@@ -0,0 +1,165 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/element.dart';
+import 'package:analyzer/dart/element/type.dart';
+import 'package:linter/src/analyzer.dart';
+import 'package:linter/src/util/dart_type_utilities.dart';
+
+const _desc = r'Do not loose voidness.';
```
MichaelRFairhurst: nit: "lose"

lib/src/rules/voidness_propagation.dart
```diff
@@ -0,0 +1,165 @@
...
+
+const _details = r'''
+
+**DO** not loose voidness. This is likely to lead to a runtime error.
```
MichaelRFairhurst: same here

lib/src/rules/voidness_propagation.dart
```diff
@@ -0,0 +1,165 @@
...
+  @override
+  void registerNodeProcessors(NodeLintRegistry registry) {
+    final visitor = new _Visitor(this);
+    registry.addAssignmentExpression(this, visitor);
+    registry.addExpressionFunctionBody(this, visitor);
+    registry.addInstanceCreationExpression(this, visitor);
+    registry.addMethodDeclaration(this, visitor);
+    registry.addMethodInvocation(this, visitor);
+    registry.addReturnStatement(this, visitor);
+    registry.addVariableDeclaration(this, visitor);
```
MichaelRFairhurst: I would like to maybe see about giving this lint better support in the analyzer API.

There are many more places that downcasts can occur -- for instance:

```
class DefinesPlus {
  operator +(DefinesPlus other) => null;
}

void main() {
  var x = DefinesPlus();
  void y;
  print(x + y); // downcast from void to DefinesPlus
}
```

The CFE project that lowers Dart to kernel code will include an explicit cast here, and the analyzer can do the same sort of resolution. If we could expose that in the tree, that would make it easier for you to, for instance, find any expressions that are downcasted and then check what the from + to types were. This would get a performance boost (by not having to check every node for downcasts _again_), make this code simpler (hopefully!), and make it easy to have this lint and others like it be consistently conclusive in coverage.

Are you interested in doing that for this lint or do you want me to keep that in my mind as a theoretical? Note: I have not discussed this with the other analyzer teammembers and they may not agree with me on it.

lib/src/rules/voidness_propagation.dart
```diff
@@ -0,0 +1,165 @@
...
+
+  @override
+  void visitAssignmentExpression(AssignmentExpression node) {
+    final type = node.leftHandSide?.staticType;
+    _check(type, node.rightHandSide?.staticType, node.rightHandSide);
+  }
+
+  @override
+  void visitExpressionFunctionBody(ExpressionFunctionBody node) {
+    DartType returnType;
```
MichaelRFairhurst: Do you want to check parameters too to catch this case?

```
void foo(int x) => print(x);

void main {
  void Function(void) f = foo; // fails at runtime
}
```

?

lib/src/rules/voidness_propagation.dart
```diff
@@ -0,0 +1,165 @@
...
+      _checkArgs(args, parameters);
+    }
+  }
+
+  @override
+  void visitMethodDeclaration(MethodDeclaration node) {
+    final inheritedMethod = DartTypeUtilities.lookUpInheritedMethod(node);
+    if (inheritedMethod != null) {
+      _check(
+          inheritedMethod.returnType, node.returnType?.type, node.returnType);
```
MichaelRFairhurst: I thought we already caught this?

```
class B {                                                                        
  int f() => 123;                                                                
}                                                                                
                                                                                 
class C extends B{                                                               
  void f() => null;                                                              
} 
```

gives me "invalid override"

lib/src/rules/voidness_propagation.dart
```diff
@@ -0,0 +1,165 @@
...
+
+  void _check(DartType expectedType, DartType type, AstNode node) {
+    if (expectedType == null || type == null) {
+      return;
+    } else if (expectedType is FunctionType && type is FunctionType) {
+      _check(expectedType.returnType, type.returnType, node);
+    } else if (type is InterfaceType &&
+        expectedType is InterfaceType &&
+        type != expectedType &&
+        type.element == expectedType.element) {
```
MichaelRFairhurst: Hmm, this won't catch `Iterable<void>` being assigned to `List<int>`, or `List<int>` being assigned to `Iterable<void>`.

There are other pathological cases too:

```
abstract class IV implements Iterable<void> {}
abstract class LI implements List<int> {}

void bad(IV iv, LI li) {
  li = iv;
}
```

I guess this is probably the most complex part and the API change I described earlier wouldn't help you solve this...once again, though, this is logic we have implemented in the analyzer. I wonder if we can expose that API to you at all in a way where you can provide custom hooks into the subtype comparison. FWIW the subtype compare code is here: https://github.com/dart-lang/sdk/blob/master/pkg/analyzer/lib/src/generated/type_system.dart#L1480

lib/src/rules/voidness_propagation.dart
```diff
@@ -0,0 +1,165 @@
...
+      for (var i = 0; i < type.typeArguments.length; i++) {
+        if (type.typeArguments[i].isVoid &&
```
MichaelRFairhurst: I would call `_check` on `typeArguments[i]`.

I think the current code misses this:

```
void Function() x;
int Function() y = x;
```

lib/src/rules/voidness_propagation.dart
bwilkerson: > I would like to maybe see about giving this lint better support in the analyzer API.

I'm not sure what kind of support you're thinking about, so it's hard for me to say what I think about it. :-)

I would expect that we already have the "from" and "to" types in the AST. It should just be a matter of testing the relationships between those types. You might be proposing that we encapsulate that test in a `requiresImplicitDowncast` member so that clients don't have to repeat the test in multiple locations and risk getting it wrong, which seems reasonable.

lib/src/rules/voidness_propagation.dart
bwilkerson: We actually _do_ have a public API for subtype comparison: https://github.com/dart-lang/sdk/blob/master/pkg/analyzer/lib/dart/element/type.dart#L154.

mit-mit:

PR #1208, opened 2018-10-16
add lint avoid_returning_null_for_void

a14n:


bwilkerson:
Has this been discussed somewhere? I don't remember the motivation (which should be added to the documentation) and would like to know it for context.

a14n:
>Has this been discussed somewhere?

No

>I don't remember the motivation (which should be added to the documentation) and would like to know it for context.

Basically `return null` was used when `Null` was used instead of `void`. I find it clearer to used `return;` instead of `return null;` in a function with a return type `void`.

lib/src/rules/avoid_returning_null_for_void.dart
```diff
@@ -0,0 +1,102 @@
...
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/element.dart';
+import 'package:analyzer/dart/element/type.dart';
+import 'package:linter/src/analyzer.dart';
+
+const _desc = r"Don't return null for void.";
+
+const _details = r'''
+
+**Don't** return null for void
```
pq: maybe replace w/ 

    **AVOID** returning null for void.

(I think that's more consistent on a quick scan.)

pq: Also, it would be great to have a sentence or two of rationale...

a14n:
PTAL

lib/src/rules/avoid_returning_null_for_void.dart
```diff
@@ -0,0 +1,107 @@
...
+```
+void f1() {
+  return;
+}
+Future<void> f2() async {
+  return;
+}
+```
+
+''';
```
pedromassango, line 42: What is the difference of the "GOOD" and "BAD" statements here?

PR #1215, opened 2018-10-18
Prevent NPE in prefer_iterable_whereType

bwilkerson:


bwilkerson:
> good catch.

I didn't catch it, a user did.

> should we do a quick roll to pull this in?

I think so. As far as I can tell this shouldn't de-stabilize the build.

Issue #628, opened 2017-04-23
prefer_const_constructors issue with generic classes


a14n:
```dart
class A<T> {
  const A();
}
class B<T> {
  const B();
  A<T> get a => new A<T>();
}
```

The change to use const leads to:

>The constant creation cannot use a type parameter

a14n:
This issue has been fixed for a while.

PR #1219, opened 2018-10-21
Add tests for prefer_const_constructors

srawlins:
... to ensure lint does not complain about non-const type-variabled constructor. Related to #628


srawlins:
CC @a14n

PR #1221, opened 2018-10-22
Prefer int literals over double literals

danrubel:
First cut at a prefer int literals lint. There are more situations in which this lint should report a double literal that should be converted, but AFAIK there are no false positives.

### DIFF
--- a/lib/src/rules/prefer_int_literals.dart
+++ b/lib/src/rules/prefer_int_literals.dart
@@ -76,7 +76,7 @@ class _Visitor extends SimpleAstVisitor<void> {
     if (node == null) {
       return false;
     } else if (node is VariableDeclarationList) {
-      return node.type != null;
+      return node.type?.type?.name == 'double';
     } else if (node is ArgumentList) {
       // TODO(danrubel): Determine type associated with this argument
     } else if (node is Expression) {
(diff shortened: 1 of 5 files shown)


lib/src/rules/prefer_int_literals.dart
```diff
@@ -0,0 +1,87 @@
...
+      rule.reportLintForToken(node.literal);
+    }
+  }
+
+  bool isDoubleContext(AstNode child) {
+    final node = child.parent;
+    if (node == null) {
+      return false;
+    } else if (node is VariableDeclarationList) {
+      return node.type != null;
```
bwilkerson: I don't think this is right. What if the declared type is `num` or `Object` (or even `dynamic`)?

lib/src/rules/prefer_int_literals.dart
```diff
@@ -0,0 +1,87 @@
...
+      return;
+    }
+
+    // Ensure that replacing the double would not change the semantics
+    if (isDoubleContext(node)) {
+      rule.reportLintForToken(node.literal);
+    }
+  }
+
+  bool isDoubleContext(AstNode child) {
```
bwilkerson: If the notion of a "double context" is in the spec (either directly or via approved extension), then I'd like to have this method added as a getter (`inDoubleContext`) on DoubleLiteral (and IntegerLiteral?).

lib/src/rules/prefer_int_literals.dart
danrubel: Good point. Fixed.

lib/src/rules/prefer_int_literals.dart
danrubel: I do not know of an "double context" in the spec, but I'm happy to refactor as you see fit.

PR #1222, opened 2018-10-23
Improve prefer_int_literal lint

danrubel:
This improves the prefer_int_literal lint to detect double literals that can be converted in argument lists and binary expressions.

### DIFF
--- a/lib/src/rules/prefer_int_literals.dart
+++ b/lib/src/rules/prefer_int_literals.dart
@@ -66,32 +66,46 @@ class _Visitor extends SimpleAstVisitor<void> {
     }
 
     // Ensure that replacing the double would not change the semantics
-    if (isDoubleContext(node)) {
+    if (canReplaceWithIntLiteral(node)) {
       rule.reportLint(node);
     }
   }
 
-  bool isDoubleContext(AstNode child) {
-    final node = child.parent;
-    if (node == null) {
-      return false;
-    } else if (node is ArgumentList) {
-      if (child is Expression) {
-        return child.staticParameterElement?.type?.name == 'double';
+  /// Determine if the given literal can be replaced by an int literal.
+  bool canReplaceWithIntLiteral(DoubleLiteral literal) {
+    // TODO(danrubel): Consider moving this into analyzer
+    final AstNode parent = literal.parent;
+    if (parent is ArgumentList) {
+      return literal.staticParameterElement?.type?.name == 'double';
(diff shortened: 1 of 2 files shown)


test/rules/prefer_int_literals.dart
```diff
@@ -9,30 +9,59 @@ const double shouldBeInt = 8.0; // LINT
...
   var w = 7.0e2; // OK
   double x = 7.0e2; // LINT
   double y = 7.1e2; // LINT
   double z = 7.576e2; // OK
   A(this.x);
+  namedDouble(String s, {double d}) {}
+  namedDynamic(String s, {dynamic d}) {}
 }
 
 // TODO(danrubel): Report lint in these other situations
```
bwilkerson: Remove this?

lib/src/rules/prefer_int_literals.dart
```diff
@@ -67,20 +67,30 @@ class _Visitor extends SimpleAstVisitor<void> {
...
     if (node == null) {
       return false;
+    } else if (node is ArgumentList) {
+      if (child is Expression) {
+        return child.staticParameterElement?.type?.name == 'double';
+      }
+      return false;
+    } else if (node is AssignmentExpression) {
+      return node.rightHandSide?.staticType?.name == 'double';
+    } else if (node is BinaryExpression) {
```
bwilkerson: I'm having a hard time convincing myself that the logic for binary expressions is correct in the face of user defined operators. Consider
```
class C {
  double operator +(double value) => value;
}
void f(C c) {
  print(c + 2.0);
}
```
I think that should lint because the right operand would need to be auto converted based on the signature of the operator. But `operand` would be `c`, which is not a double, so I think this method would return `false`.

I think for user defined operators we really need to look at the type of the operator's parameter.

I'd have to look at each of the non-user-definable operators to decide whether this is right in those cases. We should probably have tests for the rest of the operators (though one test per operator is probably sufficient).

test/rules/prefer_int_literals.dart
danrubel: Removed.

lib/src/rules/prefer_int_literals.dart
danrubel: Good point. I have revised the logic and added more tests. I've removed support for linting double literals in expressions until I can better understand the various situations.

danrubel:
Revised code and updated tests. PTAL

PR #1228, opened 2018-10-26
Add a rule that reports type parameter shadowing

### LINKED ISSUE #1009: A rule for shadowing type variables
Take this code:

```dart
class A<T> {
  List<T> t;
  
  void fn<T>(List<T> t) {
    print(t);
  }
}
```

Boo. That `fn<T>` is confusing. This will likely be supported by the language going forward though (as per https://github.com/dart-lang/sdk/issues/29057). Should lint. Similar to #597.

srawlins:
Fixes #1009


lib/src/rules/shadow_type_parameters.dart
```diff
@@ -0,0 +1,109 @@
...
+  void fn<U>() {}
+}
+```
+
+''';
+
+class ShadowTypeParameters extends LintRule implements NodeLintRule {
+  ShadowTypeParameters()
+      : super(
+            name: 'shadow_type_parameters',
```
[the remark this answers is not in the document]
srawlins: While we're talking about naming, is it more correct to say "Avoid shadowing type _variables_" or "type _parameters_"?

lib/src/rules/shadow_type_parameters.dart
```diff
@@ -0,0 +1,109 @@
...
+  void visitFunctionExpression(FunctionExpression node) {
+    if (node.typeParameters == null) {
+      return;
+    }
+
+    _checkAncestorParameters(node);
+  }
+
+  // Check the ancestors of [node] for type parameter shadowing.
+  _checkAncestorParameters(AstNode node) {
```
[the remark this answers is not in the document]
srawlins: Done! Thanks for the tip!

lib/src/rules/shadow_type_parameters.dart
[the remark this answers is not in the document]
bwilkerson: > How about `avoid_shadowing_type_parameters`?

Sounds good to me.

> ... is it more correct to say "Avoid shadowing type variables" or "type parameters"?

The spec uses "type variable" 20 times and "type parameter" 114 times. Analyzer consistently uses "type parameter". Hence, I would prefer "parameter".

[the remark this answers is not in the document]
srawlins: Done.

lib/src/rules/avoid_shadowing_type_parameters.dart
```diff
@@ -0,0 +1,101 @@
...
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:linter/src/analyzer.dart';
+
+const _desc = r'Avoid shadowing type parameters.';
+
+const _details = r'''
+
+**DO NOT** shadow type parameters.
```
bwilkerson: "DO NOT" --> "AVOID"?

lib/src/rules/avoid_shadowing_type_parameters.dart
```diff
@@ -0,0 +1,101 @@
...
+        _checkForShadowing(
+            typeParameters, parent.functionExpression.typeParameters);
+      }
+      parent = parent.parent;
+    }
+  }
+
+  // Check whether any of [typeParameters] shadow [ancestorTypeParameters].
+  _checkForShadowing(TypeParameterList typeParameters,
+      TypeParameterList ancestorTypeParameters) {
```
bwilkerson: Consider passing in a list of ancestor type parameter _names_ so that they don't have to be mapped at multiple levels for deeply nested structures. (Not required, just a performance optimization.)

srawlins: Interesting. I'm not 100% sure I know what you're suggesting. Are you saying I could collect the parameter names of all ancestors, in `_checkAncestorParameters`, and call `_checkForShadowing()` outside the while loop, passing in the parameter names? Does that have a performance implication?

bwilkerson: No. I was suggesting something that wouldn't work because I confused `typeParameters` and `ancestorTypeParameters` in my head. Sorry about that!

lib/src/rules/avoid_shadowing_type_parameters.dart
srawlins: Done.

PR #1229, opened 2018-10-27
Improve prefer_int_literal to detect negative double literals

danrubel:


pq:
Thanks Dan!

/cc @bwilkerson (who has most context)

Issue #1231, opened 2018-10-29
prefer_int_literal false positive for max(xDouble, 1.0) ?


a14n:
I'm not really sure but I think `1.0` shouldn't be flagged in `max(xDouble, 1.0)`.

IIUC `max(xDouble, 1)` has a result with a static type `num` which differs with `max(xDouble, 1.0)` that has a result with a static type `double`.

bwilkerson:
Correct. Because `T max<T extends num>(T a, T b);` has a return type that is dependent (in part) on the type of the argument, we can't change the type of the argument without at least potentially changing the static type of the invocation.

danrubel:
As it stands today, the lint does not replace double literal arguments in `Math.max` method calls because the method arguments are not of type `double`. I added a test to this effect in https://github.com/dart-lang/linter/pull/1233.

Please let me know if I missed a situation in which the lint is behaving incorrectly.

PR #1233, opened 2018-10-29
Improve prefer_int_literal to detect double literal in list literal

danrubel:
... and add a test for arguments to Math.max

Fix https://github.com/dart-lang/linter/issues/1230
Fix https://github.com/dart-lang/linter/issues/1231

### DIFF
--- a/lib/src/rules/prefer_int_literals.dart
+++ b/lib/src/rules/prefer_int_literals.dart
@@ -89,6 +89,10 @@ class _Visitor extends SimpleAstVisitor<void> {
     final AstNode parent = expression.parent;
     if (parent is ArgumentList) {
       return expression.staticParameterElement?.type?.name == 'double';
+    } else if (parent is ListLiteral) {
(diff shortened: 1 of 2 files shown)


mateusfccp:
@danrubel

I'm a little puzzled by the test:

```dart
linter_issue_1231() {
  double myDouble = 3;
  max(3.0, 4); // OK
  max(myDouble, 5.0); // LINT
  var value1 = max(myDouble, 6.0); // LINT
  double value2 = max(myDouble, 7.0); // LINT
  return value1 + value2;
}
```

I understand that `max(3.0, 4)` should not trigger the lint; it's the raison d'etre of this issue.

However, why `max(myDouble, 5.0)` should trigger the lint? The expression `max(myDouble, 5.0)` has static type `double`, while `max(myDouble, 5)` has static type `num`. Isn't it exactly what we are trying to avoid?

The same goes for `var value1 = max(myDouble, 6.0);`

PR #1238, opened 2018-10-31
add lint avoid_returning_null_for_future

a14n:


lib/src/rules/avoid_returning_null_for_future.dart
```diff
@@ -0,0 +1,80 @@
...
+import 'package:linter/src/analyzer.dart';
+
+const _desc = r'Avoid returning null for Future.';
+
+const _details = r'''
+
+**AVOID** returning null for Future.
+
+Future being used to represent a potential value, or error, that will be
+available at some time in the future it's almost always wrong to return null for
```
pq: This description is a little hard to follow...  Maybe just simplify it to:

> It is almost always wrong to return null for a `Future`. Most of the time the developer simply forgot to put an `async` keyword on the function.

a14n: Done

pq:
Super.  Thanks!

PR #1243, opened 2018-11-02
Begin making use of the analyzer's LinterContext API.

stereotype441:
This allows us to stop using the Element.context and
Element.computeNode APIs, which will soon be removed.

### DIFF
--- a/lib/src/analyzer.dart
+++ b/lib/src/analyzer.dart
@@ -32,10 +32,11 @@ export 'package:analyzer/src/lint/linter.dart'
         LintRule,
         Group,
         Maturity,
+        LinterContext,
         LinterOptions,
         LintFilter,
         NodeLintRegistry,
-        NodeLintRule;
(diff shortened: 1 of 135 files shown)


stereotype441:
@bwilkerson It's mostly the same.  The only differences are:
- The previous review didn't address uses of .getContents.  See `lib/src/rules/lines_longer_than_80_chars.dart`.
- The previous review didn't address uses of .computeNode.  See `lib/src/rules/avoid_private_typedef_functions.dart`.

If the changes to those two files look good to you, then I think we are good to go.

bwilkerson:
Yes, those lgtm.

bwilkerson:
I agree. I'd like to see us get to one class to extend unless there's a valid reason why we can't or shouldn't.

pq:
As per our conversation, I'm all for landing this as is since we have a good path to a better/simpler world moving forward.

👍

PR #1244, opened 2018-11-02
add lint use_function_type_syntax_for_parameters

### LINKED ISSUE #1204: Ban old syntax for Function-typed function parameters
The old syntax for declaring Function-typed function parameters is confusing.

For example, in `void myTest(void f(int, String))` you would expect `f` to have type `void Function(int, String)` but it actually has type `void Function(dynamic, dynamic)` because the `int` and `String` are treated as parameter names, not types.

All declarations of this form can be replaced with something like `void myTest(void Function(int, String) f)` which is unambiguous.

a14n:
Ensure effective dart rule: [CONSIDER using function type syntax for parameters.](https://www.dartlang.org/guides/language/effective-dart/design#consider-using-function-type-syntax-for-parameters)

Fixes #1204

### DIFF
--- a/lib/src/rules/use_function_type_syntax_for_parameters.dart
+++ b/lib/src/rules/use_function_type_syntax_for_parameters.dart
@@ -1,0 +1,52 @@
+// Copyright (c) 2018, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:linter/src/analyzer.dart';
+
+const _desc = r'Use function type syntax for parameters.';
+
+const _details = r'''
+
+**USE** function type syntax for parameters.
+
+**BAD:**
+```
+Iterable<T> where(bool predicate(T element)) {}
+```
+
+**GOOD:**
+```
+Iterable<T> where(bool Function(T) predicate) {}
+```
+
+''';
+
+class UseFunctionTypeSyntaxForParameters extends LintRule
+    implements NodeLintRule {
+  UseFunctionTypeSyntaxForParameters()
+      : super(
+            name: 'use_function_type_syntax_for_parameters',
+            description: _desc,
+            details: _details,
+            group: Group.style);
+
+  @override
+  void registerNodeProcessors(NodeLintRegistry registry) {
+    final visitor = new _Visitor(this);
+    registry.addFunctionTypedFormalParameter(this, visitor);
+  }
+}
+
+class _Visitor extends SimpleAstVisitor<void> {
+  final LintRule rule;
+
+  _Visitor(this.rule);
+
+  @override
+  void visitFunctionTypedFormalParameter(FunctionTypedFormalParameter node) {
+    rule.reportLint(node);
+  }
+}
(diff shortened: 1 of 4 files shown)


lib/src/rules/use_function_type_syntax_for_parameters.dart
```diff
@@ -0,0 +1,52 @@
...
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:linter/src/analyzer.dart';
+
+const _desc = r'Use function type syntax for parameters.';
+
+const _details = r'''
+
+**USE** function type syntax for parameters.
```
bwilkerson: I'm not sure what the convention is here, but I found one rule that says "**CONSIDER** using ...". It also has a name prefixed with "prefer_". I was just surprised to see "CONSIDER" in the description of the PR and "USE" in the description of the rule.

pq: Yeah.  We're not super consistent here...  At this point I think it's a matter of taste whether to go with "use" or "prefer" (though maybe we should get more crisp about that).  I guess the important thing is to have the name and advice agree as best as possible and this (now) looks OK to me.

pq:
Oh wait, I see what @bwilkerson was saying better now.  Hmmm.  @a14n: how do you feel about making the advice use the same language as the effective dart entry?  I think "consider" is important here since it's less prescriptive than "use".

a14n:
I think lint name shouldn't mention prefer/consider terms. IMHO those kind of wording is relative to what a team want to do (enable a lint or not) and not to what a lint really does.

Tomorrow if the style guide choose to change _CONSIDER using function type syntax for parameters_ to _DO use function type syntax for parameters_ there will be no change to the lint. But what would we do if the lint is named _consider_using_function_type_syntax_for_parameters_?

With that in mind I think lint name should only describe the lint goal and not the opinion about if the lint should be applied or not.

WDYT?

pq:
> I think lint name shouldn't mention prefer/consider terms.

OK, yes.  I completely agree.  Early on I tried to take this approach (w/ names like `camel_case_types`) but we kind of strayed a bit.

How about we keep the name but update the text to agree with Effective Dart?  So:

```
**CONSIDER** using function type syntax for parameters.
```

Thanks for thinking it through!

eernstg:
Just noticed this PR. Did you consider a slightly less breaking approach where old-style function types are accepted by default, but flagged if they omit one or more parameter type annotations?

So `foo(void f(int i))` would be OK, but `foo(void f(i))` would be flagged as dangerous (of course, `foo(void f(int))` may be considered even more obviously wrong, but I think it makes sense to flag all locations where the type annotation is omitted). The diagnostic message could then recommend using the `Function` syntax, or adding a type annotation to the relevant parameter(s).

This would allow us, at some point in the future, to make it an actual error to omit such a type annotation. We could then keep support for the old-style syntax (which is more concise, which is usually given a lot of attention ;-), but never have those crazy parameters named `int`.

a14n:
This lint is to enforce the effective dart rule: [CONSIDER using function type syntax for parameters.](https://www.dartlang.org/guides/language/effective-dart/design#consider-using-function-type-syntax-for-parameters)

What you describe is definitely a good lint but I think it should be another lint. Feel free to file an issue :)

eernstg:
> Feel free to file an issue :)

Did [that](https://github.com/dart-lang/linter/issues/1248). ;-)

However, the point I'm making is that _this_ lint may be a little bit of an overkill, and the lint on this topic that we can push unconditionally is #1248, because it doesn't force developers to change code unless it actually uses that error-prone misfeature of specifying the name and omitting the type annotation for a formal parameter.

I guess this would apply to most rules starting with 'CONSIDER': They are probably worded like that because it is hard to justify a strict enforcement, and hence it's likely to be overkill to mechanically enforce such a rule, e.g., by making it a lint.

pq:
@munificent: I pinged you on the corresponding lint request but context here is useful too.

Thanks in advance for any input!

PR #1247, opened 2018-11-04
fix false positive in `use_setters_to_change_properties` (#1246)

pq:
fixes: #1246 

/cc @bwilkerson

### DIFF
--- a/test/rules/use_setters_to_change_properties.dart
+++ b/test/rules/use_setters_to_change_properties.dart
@@ -8,13 +8,15 @@ abstract class A {
   int _w;
   int _x;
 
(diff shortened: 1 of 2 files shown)


test/rules/use_setters_to_change_properties.dart
```diff
@@ -8,13 +8,15 @@ abstract class A {
   int _w;
   int _x;
 
-  void setW(int w) => this._w = w; // LINT
```
bwilkerson: Should we have both tests (with and without `this`) to ensure that we recognize `_w` in either context?

pq: Not a bad idea.  Done!

PR #1252, opened 2018-11-05
add lint unnecessary_await

a14n:


lib/src/rules/unnecessary_await.dart
```diff
@@ -0,0 +1,97 @@
...
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/type.dart';
+import 'package:linter/src/analyzer.dart';
+
+const _desc = r'Unnecessary await keyword.';
+
+const _details = r'''
+
+Avoid using await on an expression with a type that is assignable to the return
+type.
```
pq: This seems awkward though I admit I'm struggling to improve it.

Maybe something like:

    Avoid returning an awaited expression when the expression type is assignable to a function's return type.

Likely @bwilkerson can do better...

bwilkerson:
Has there been any discussion about adding this rule? I ask because the name sounds much more general than the actual implementation, but I don't know which needs to change unless I understand the purpose for the rule.

a14n:
> Has there been any discussion about adding this rule?

No. It's driven by what I saw here and there. Most of the time a `return await future` where a `return future` was enough.

> the name sounds much more general than the actual implementation

You're right. Thinking about it a little more the lint could be extended to every assignments.

a14n:
> Thinking about it a little more the lint could be extended to every assignments.

Forget it... I need some coffee.

lib/src/rules/unnecessary_await.dart
a14n: Changed with Phil suggestion.

a14n:
> I ask because the name sounds much more general than the actual implementation, but I don't know which needs to change unless I understand the purpose for the rule.

@bwilkerson do you have other cases in mind other than the `return await` targeted by the current version of the lint?

a14n:
testing the lint on flutter I see some test failures because of the following change:

```diff
--- a/packages/flutter/lib/src/services/platform_channel.dart
+++ b/packages/flutter/lib/src/services/platform_channel.dart
@@ -355,7 +355,7 @@ class OptionalMethodChannel extends MethodChannel {
   @override
   Future<dynamic> invokeMethod(String method, [dynamic arguments]) async {
     try {
-      return await super.invokeMethod(method, arguments);
+      return super.invokeMethod(method, arguments);
     } on MissingPluginException {
       return null;
     }
```

The lint should add an exception for function returning `Future<dynamic>` (and perhaps  `Future<Object>`)

bwilkerson:
> do you have other cases in mind other than the return await targeted by the current version of the lint?

No, but if that's really the only case where the lint can be applied then we might want to make that explicit in the name. Something like 'unnecessary_await_in_return'? It's kind of long, but more informative.

> The lint should add an exception for function returning Future<dynamic> (and perhaps Future<Object>)

Wow! That's subtle.

a14n:
> Something like 'unnecessary_await_in_return'

Done

> The lint should add an exception for function returning `Future<dynamic>` (and perhaps `Future<Object>`)

Done

PR #1253, opened 2018-11-05
Enable and fix  prefer_iterable_whereType

listepo:


listepo:
@pq I can take not urgent issue and do it when there is time

PR #1254, opened 2018-11-05
0.1.72

pq:
# 0.1.72

* new lint: `use_function_type_syntax_for_parameters`
* internal changes to migrate towards analyzer's new `LinterContext` API
* fix false positive in `use_setters_to_change_properties`
* implementation improvements (and speed-ups) to `prefer_foreach` and `public_member_api_docs`
* new lint: `avoid_returning_null_for_future`
* new lint: `avoid_shadowing_type_parameters`

/cc @bwilkerson @a14n @stereotype441

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+# 0.1.72
+
+* new lint: `use_function_type_syntax_for_parameters`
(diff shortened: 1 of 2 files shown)


pq:
Spot on.  Not for rolling just yet but I think makes a good functional check-point.  (One we may want to be able to roll back to.)

Thanks!

PR #1263, opened 2018-11-14
bump sdk constraint to 2.1.0

pq:
/cc @bwilkerson


pq:
(of course this can't land until 2.1.0; it will have to wait in the wings until then.)

PR #1265, opened 2018-11-18
remove `prefer_bool_in_asserts`

pq:
/cc @a14n @bwilkerson


pq:
@bwilkerson : absolutely.  Tomorrow I'd like to settle on a deprecation strategy and get started on it. 👍

pq:
I still think we should add a comment to the rule's docs

👍  => #1276

PR #1271, opened 2018-11-20
add count badge

pq:
image

/cc @bwilkerson

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 The Dart Linter package defines lint rules that identify and report on "lints" found in Dart code.  Linting is performed by the Dart
 analysis server and the `dartanalyzer` commandline tool.
 
-
+[![Lint Count](http://dart-lang.github.io/linter/lints/count-badge.svg)](http://dart-lang.github.io/linter/lints/)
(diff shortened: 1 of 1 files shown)


pq:
@a14n : right.  That's the number of documented lints (http://dart-lang.github.io/linter/lints/) but maybe that's not the best number?

pq:
Either way maybe "lints" is ambiguous as a label?

a14n:
Looking at the badge I expected to have the number of lints activated for the project.

pq:
> Looking at the badge I expected to have the number of lints activated for the project.

Aha!  That's another badge I should make. 👍 

This is just a lint count (vanity I know) and a convenient place to hang a link to http://dart-lang.github.io/linter/lints/.  Can you think of another label that would be more clear?

a14n:
>This is just a lint count (vanity I know)

Ahhh ok. I misunderstood what it was.

>Can you think of another label that would be more clear?

Unfortunately no :-(

pq:
Thanks for the feedback @a14n!  I'm going to work on some other badges and will loop you in when there's something to discuss.

PR #1276, opened 2018-11-24
deprecate `prefer_bool_in_asserts`

pq:
Fixes: dart-lang/sdk#57809.

We should sweep and remove all deprecated lints before a major release (#1275).

I'll update docgen to emphasize deprecated-ness in generated docs too (#1274).

### DIFF
--- a/lib/src/rules/prefer_bool_in_asserts.dart
+++ b/lib/src/rules/prefer_bool_in_asserts.dart
@@ -32,8 +32,8 @@ assert(() {
 }());
 ```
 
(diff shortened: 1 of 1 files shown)


lib/src/rules/prefer_bool_in_asserts.dart
```diff
@@ -32,6 +32,11 @@ assert(() {
 }());
 ```
 
+**DEPRECATED:** In Dart 2, `assert`s no longer accept `bool` values so this rule
```
bwilkerson: "accept `bool` values" --> "accept non-`bool` values"

PR #1281, opened 2018-11-26
emphasize non-stable lints in docs

pq:
Emphasize non-stable lints in their generated docs

* adds doc styling to emphasize deprecated and experimental lints
* adds blurb about deprecated lints to lint index description

image

Red and pink may be a subtle distinction but I do want them to pop.

image

image


pq:
I'm not really either to be honest but it's subtle and easily tweaked.  (Not even sure anyone will notice!)

PR #1285, opened 2018-11-27
Fix unnecessary_parenthesis's treatment of function expressions

### LINKED ISSUE #1284: unnecessary_parenthesis requests that an expression-function body unwrap, which breaks the code
Take this code:

```dart
var i = ((b) => b is bool ? b : false)(x);
```

The unnecessary_parenthesis rule warns about the parens here, but removing them breaks the statement.

srawlins:
Fixes #1284 

The new test fails without the change to lib/

### DIFF
--- a/test/rules/unnecessary_parenthesis.dart
+++ b/test/rules/unnecessary_parenthesis.dart
@@ -28,6 +28,9 @@ main() async {
   m(p: (1 + 3)); // LINT
   a..b = (c..d); // OK
   ((x) => x is bool ? x : false)(a); // OK
+  (fn)(a); // LINT
 }
 
 m({p}) => null;
+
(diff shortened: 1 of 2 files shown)


test/rules/unnecessary_parenthesis.dart
```diff
@@ -27,6 +27,7 @@ main() async {
   true ? [] : []..add(''); // OK
   m(p: (1 + 3)); // LINT
   a..b = (c..d); // OK
+  ((x) => x is bool ? x : false)(a); // OK
```
bwilkerson: What about

(getterReturningFunction)(x);

srawlins: Great test! Added.

Issue #1287, opened 2018-11-30
Upgrade `prefer_void_to_null` to stable


pq:
@MichaelRFairhurst: what needs to happen here?

Looking at uses,

https://github.com/search?q=prefer_void_to_null+filename%3Aanalysis_options.yaml&type=Code

a number of files include it while still other have it removed:

```
# TODO: Re-enable prefer_void_to_null when fully supported
#- prefer_void_to_null
```

Thoughts on next steps?

a14n:
FWIW [Flutter already uses this lint](https://github.com/flutter/flutter/blob/5391447fae6209bb21a89e6a5a6583cac1af9b4b/analysis_options.yaml#L142).

pq:
@MichaelRFairhurst: thoughts on this?  I'm inclined to bump to `stable` unless you're actively working on anything that you think should keep this flagged `experimental`?

MichaelRFairhurst:
Nope, upping to stable sgtm!

PR #1292, opened 2018-11-30
promote to stable

pq:
Fixes: #1287.

/cc @MichaelRFairhurst @bwilkerson


pq:
Thank you!

This is some of my favorite kind of house-keeping. It's awesome to see so many folks getting value out of this analysis.  So, cheers!

PR #1295, opened 2018-12-04
[WIP] Add a new lint for unknown escape sequences

joe776:
This adds a new lint to avoid undefined escape sequences like `\z` or unneeded escapes like `\"` in a single-quoted string literal.
The second part is not implemented yet (tests for it fail), and the description needs more work.
If this lint is desirable I will work an that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# x.y.z
+
+* new lint: `avoid_undefined_escape_sequences`
+
 # 0.1.75
 
 * analyzer package dependency bumped to `^0.34.0` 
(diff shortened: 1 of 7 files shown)


pq:
It looks like this is provoking some breakages detected by the benchmark build:
```
Exception while using a AvoidUndefinedAndUnnecessaryEscapeSequences to visit a InterpolationStringImpl in StringInterpolationImpl in ArgumentListImpl in MethodInvocationImpl in ExpressionStatementImpl in BlockImpl in ForEachStatementImpl in BlockImpl in BlockFunctionBodyImpl in MethodDeclarationImpl in ClassDeclarationImpl in CompilationUnitImpl
```
https://travis-ci.org/dart-lang/linter/jobs/465132433

@bwilkerson : any ideas?

joe776:
The problem was, that the string part of an interpolation string may be empty. This seems to get passed as an empty string, and not wrapped in quotes as the other strings are.
So, there might be other subtle errors with handling quotes. `SimpleStringLiteral` has the property `isSingleQuoted`, but I don't see this for `InterpolationString`.
Is it save to assume, that the contents is either empty, or starts with a quote?

pq:
> I still want to look into how to only produce a lint for the escape sequence, not for the whole string; already found LintRule.reporter.reportErrorForOffsetfor that.

`LintRule.reportErrorForToken` may come in handy too?

srawlins:

PR #1297, opened 2018-12-05
Fix unrelated type checks of generic types

### LINKED ISSUE #1085: #unrelated_type_equality_checks false positives with `Future<void>`
This is a placeholder when fixing internal code, but I saw something like:

```dart
import 'dart:async';

Future getFuture() => new Future.value();

void main() {
  Future<void> aVoidFuture;

  var aDynamicFuture = getFuture();
  // LINT: #unrelated_type_equality_checks
  if (aFuture == aVoidFuture) {
    // ...
  }
}
```

I wonder if this is something relatively new (i.e. `void` as a type).

I will try and get a better reproduction case.

/cc @davidmorgan as this came up internally.

srawlins:
Fixes #1085 and https://github.com/dart-lang/sdk/issues/32522

The `unrelated_type_equality_checks` lint, to avoid false positives, treats types that _might_ be related as good enough, such as `fn<A, B>(A a, B b) => a == b`; no lint. This rule was not applied to generics, so the following types on either side of `==` were reported:

* `List<dynamic>` and `List<String>`
* `Future<dynamic>` and `Future<T>`
* `List<A>` and `List<B>` for unbound `A` and `B`.

### DIFF
--- a/lib/src/util/dart_type_utilities.dart
+++ b/lib/src/util/dart_type_utilities.dart
@@ -219,7 +219,32 @@ class DartTypeUtilities {
     return nodes;
   }
 
+  /// Return whether [leftType] and [rightType] are _definitely_ unrelated.
+  ///
+  /// For the purposes of this function, here are some "relation" rules:
+  /// * `dynamic` and `Null` are considered related to any other type.
+  /// * Two equal types are considered related, e.g. classes `int` and `int`,
+  ///   classes `List<String>` and `List<String>`,
+  ///   classes `List<T>` and `List<T>`, and type variables `A` and `A`.
+  /// * Two types such that one is more specific than the other, such as classes
+  ///   `List<dynamic>` and `Iterable<dynamic>`, and type variables `A` and `B`
+  ///   where `A extends B`. The rules of type specificity are documented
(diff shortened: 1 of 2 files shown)


srawlins:
If I should add more tests for this, please let me know!

lib/src/util/dart_type_utilities.dart
```diff
@@ -219,7 +219,32 @@ class DartTypeUtilities {
...
+  ///     and each of their pair-wise type arguments are related, e.g.
+  ///     `List<dynamic>` and `List<int>`, and `Future<T>` and `Future<S>` where
+  ///     `S extends T`.
+  ///   * are unrelated if [leftType]'s supertype is [Object].
+  ///   * are related if their supertypes are equal, e.g. `List<dynamic>` and
+  ///     `Set<dynamic>`.
+  /// * Two types, each representing a type variable, are related if their
+  ///   bounds are related.
+  /// * Otherwise, the types are related.
+  // TODO(srawlins): typedefs :D
```
bwilkerson: I don't understand what this TODO means. Could you elaborate?

srawlins: Oh sorry. I can add more to the comment, but basically I don't think this lint handles typedefs. There certainly aren't any tests :(

PR #1298, opened 2018-12-05
Fix prefer_final_fields with generic super-classes

srawlins:
Fixes https://github.com/dart-lang/sdk/issues/34417

### DIFF
--- a/lib/src/util/dart_type_utilities.dart
+++ b/lib/src/util/dart_type_utilities.dart
@@ -8,6 +8,7 @@ import 'package:analyzer/analyzer.dart';
 import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/element/element.dart';
 import 'package:analyzer/dart/element/type.dart';
+import 'package:analyzer/src/dart/element/member.dart';
 
 typedef bool AstNodePredicate(AstNode node);
 
@@ -17,8 +18,20 @@ class DartTypeUtilities {
       (type is InterfaceType &&
           extendsClass(type.superclass, className, library));
 
-  static Element getCanonicalElement(Element element) =>
-      element is PropertyAccessorElement ? element.variable : element;
+  static Element getCanonicalElement(Element element) {
+    if (element is PropertyAccessorElement) {
+      final variable = element.variable;
+      if (variable is FieldMember) {
+        // A field element defined in a parameterized type where the values of
+        // the type parameters are known.
+        return variable.baseElement;
+      } else {
+        return variable;
+      }
+    } else {
+      return element;
+    }
+  }
 
   static Element getCanonicalElementFromIdentifier(AstNode rawNode) {
     if (rawNode is Expression) {
--- a/lib/src/util/dart_type_utilities.dart
+++ b/lib/src/util/dart_type_utilities.dart
@@ -24,6 +24,11 @@ class DartTypeUtilities {
       if (variable is FieldMember) {
         // A field element defined in a parameterized type where the values of
         // the type parameters are known.
+        //
+        // This concept should be invisible when comparing FieldElements, but a
+        // bug in the analyzer causes FieldElements to not evaluate as
+        // equivalent to equivalent FieldMembers. See
+        // https://github.com/dart-lang/sdk/issues/35343.
         return variable.baseElement;
       } else {
         return variable;
--- a/test/rules/prefer_final_fields.dart
+++ b/test/rules/prefer_final_fields.dart
@@ -73,7 +73,7 @@ class E {
   }
 }
 
-class F{
+class F {
   var _array = new List<int>(5); // LINT
 
   void foo() {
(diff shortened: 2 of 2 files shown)


srawlins:
@pq sorry about the analyzer `src` import, but FieldMember does not appear to be part of the public API? No API docs at https://pub.dartlang.org/documentation/analyzer/latest ...

lib/src/util/dart_type_utilities.dart
```diff
@@ -17,8 +18,20 @@ class DartTypeUtilities {
       (type is InterfaceType &&
           extendsClass(type.superclass, className, library));
 
-  static Element getCanonicalElement(Element element) =>
-      element is PropertyAccessorElement ? element.variable : element;
+  static Element getCanonicalElement(Element element) {
+    if (element is PropertyAccessorElement) {
+      final variable = element.variable;
+      if (variable is FieldMember) {
```
bwilkerson: It's true that FieldMember is not part of the public API. I'd like to understand more about why you needed to get the base element so that we can provide a public API that will meet the need. If it's easy to describe, then perhaps a reply here; otherwise maybe we can VC.

srawlins: Ah yes, sorry. So this rule looks at whether each variable in a `FieldDeclaration` is contained in a list of `_mutatedFields`. Let's take an example source, where `A` and `B` are not generic, and `C` and `D` are generic:

```dart
abstract class A {
  int _current = 0; // can't be final, since test() below changes it
}

class B extends A {
  void test() {
    _current = 1;
  }
}

abstract class C<T> {
  int _current = 0; // can't be final, since test() below changes it
}

class D extends C<int> {
  void test() {
    _current = 1;
  }
}
```

Before this fix, `_mutatedFields` contains a FieldElement for `A._current`, expected, and a FieldMember for `C._current`. But the element in each variable declaration is always a FieldElement, so the `contains` check fails for `C._current`. Here's some debug output:

```
$ dart bin/linter.dart --rules prefer_final_fields 34417.dart 
node is Identifier; staticElement is set _current(int __current) → void
    (a PropertyAccessorElementImpl_ImplicitSetter)
variable is int _current, (a FieldElementImpl)
node is Identifier; staticElement is set C._current(int __current) → void
    (a PropertyAccessorMember)
variable is int _current, (a FieldMember)
Is int _current (a FieldElementImpl) contained in (int _current (a FieldMember), int _current (a FieldElementImpl))
Is int _current (a FieldElementImpl) contained in (int _current (a FieldMember), int _current (a FieldElementImpl))
/Users/srawlins/code/dart-linter/34417.dart 12:7 [lint] Private field could be final.
  int _current = 0; // can't be final, since test() below changes it
      ^^^^^^^^^^^^
```

coming from the debug statements added in this diff:


```diff
diff --git a/lib/src/rules/prefer_final_fields.dart b/lib/src/rules/prefer_final_fields.dart
index 37f0799..23ebe8c 100644
--- a/lib/src/rules/prefer_final_fields.dart
+++ b/lib/src/rules/prefer_final_fields.dart
@@ -145,6 +145,8 @@ class _Visitor extends SimpleAstVisitor<void> {
         return;
       }
 
+      var mutatedFieldsString = _mutatedFields.map((f) => '$f (a ${f.runtimeType})');
+      print('Is ${element} (a ${element.runtimeType}) contained in $mutatedFieldsString');
       if (_mutatedFields.contains(element)) {
         return;
       }
diff --git a/lib/src/util/dart_type_utilities.dart b/lib/src/util/dart_type_utilities.dart
index b4a3e77..e160f06 100644
--- a/lib/src/util/dart_type_utilities.dart
+++ b/lib/src/util/dart_type_utilities.dart
@@ -21,13 +21,14 @@ class DartTypeUtilities {
   static Element getCanonicalElement(Element element) {
     if (element is PropertyAccessorElement) {
       final variable = element.variable;
-      if (variable is FieldMember) {
-        // A field element defined in a parameterized type where the values of
-        // the type parameters are known.
-        return variable.baseElement;
-      } else {
+      print('variable is $variable, (a ${variable.runtimeType})');
+      //if (variable is FieldMember) {
+      //  // A field element defined in a parameterized type where the values of
+      //  // the type parameters are known.
+      //  return variable.baseElement;
+      //} else {
         return variable;
-      }
+      //}
     } else {
       return element;
     }
@@ -37,6 +38,7 @@ class DartTypeUtilities {
     if (rawNode is Expression) {
       final node = rawNode.unParenthesized;
       if (node is Identifier) {
+        print('node is Identifier; staticElement is ${node.staticElement}\n    (a ${node.staticElement.runtimeType})');
         return getCanonicalElement(node.staticElement);
       } else if (node is PropertyAccess) {
         return getCanonicalElement(node.propertyName.staticElement);
```

My fix could totally be a backwards one, or perhaps a bug in analyzer is classifying the two `_current` fields so differently?

bwilkerson: Thanks! I think I understand.

This looks like a bug in analyzer. Members are suppose to act just like the elements on which they are based, and they clearly don't in this case. I'm fine with the work-around, but I'd like it to be documented to be a work-around so that it's clear why the code is there (and that it's safe to remove it once the bug is fixed).

srawlins: OK, that makes sense. Should I file a bug based on the above info? I'm not sure what the bug explicitly is...

I'll put a comment in the PR.

PR #1300, opened 2018-12-06
Clean up deprecation warnings and the copyright notices

bwilkerson:
I also sorted most of the non-test-data files, mostly out of habit, which will probably make this fairly hard to review. I'll back out those changes if you'd like. But I can assure you that I did not intentionally make any semantic altering changes in any of the code.

### DIFF
--- a/bin/linter.dart
+++ b/bin/linter.dart
@@ -1,4 +1,4 @@
-// Copyright (c) 2015, the Dart project authors.  Please see the AUTHORS file
+// Copyright (c) 2015, the Dart project authors. Please see the AUTHORS file
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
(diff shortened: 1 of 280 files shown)


bwilkerson:
> Are you explicit about this in analyzer?

Not really. The people who do most of the work in the packages know, and periodically someone will do a mass re-sort for the things that slipped through. I've always thought it would be good to have a pre-submit, but I don't even know whether that's possible for a subset of the repository.

pq:
@bwilkerson perhaps this is a nudge to finally make a lint for this?  I'd love to see it enforced so we can avoid any future churn.

bwilkerson:
I agree that avoiding churn would be great, and I'd love to have a way to do that. The problem is scoping the check. Test code, for example, probably doesn't want to be under the same restrictions. Lints aren't particularly easy to scope.

PR #1305, opened 2018-12-06
Allow ! arg to be wrapped in parens, even if not required by precedence

### LINKED ISSUE #1283: unnecessary_parenthesis should allow wrapping a `!` argument
E.g. the parens in `var z = !(const [7].contains(5));` are unnecessary, but I think they really enhance readability when the first two tokens to the right of `!` are separated by whitespace.

srawlins:
Fixes #1283

### DIFF
--- a/lib/src/rules/unnecessary_parenthesis.dart
+++ b/lib/src/rules/unnecessary_parenthesis.dart
@@ -72,6 +72,7 @@ class _Visitor extends SimpleAstVisitor<void> {
       if (parent is ConditionalExpression) return;
       if (parent is CascadeExpression) return;
       if (parent is FunctionExpressionInvocation) return;
+      if (parent is PrefixExpression) return;
       if (parent.precedence < node.expression.precedence) {
         rule.reportLint(node);
         return;
--- a/lib/src/rules/unnecessary_parenthesis.dart
+++ b/lib/src/rules/unnecessary_parenthesis.dart
@@ -72,7 +72,12 @@ class _Visitor extends SimpleAstVisitor<void> {
       if (parent is ConditionalExpression) return;
       if (parent is CascadeExpression) return;
       if (parent is FunctionExpressionInvocation) return;
-      if (parent is PrefixExpression) return;
+
+      // A prefix expression (! or -) can have an argument wrapped in
+      // "unnecessary" parens if that argument has potentially confusing
+      // whitespace after its first token.
+      if (parent is PrefixExpression &&
+          _expressionStartsWithWhitespace(node.expression)) return;
       if (parent.precedence < node.expression.precedence) {
         rule.reportLint(node);
         return;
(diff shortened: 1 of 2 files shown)


lib/src/rules/unnecessary_parenthesis.dart
```diff
@@ -72,6 +72,7 @@ class _Visitor extends SimpleAstVisitor<void> {
       if (parent is ConditionalExpression) return;
       if (parent is CascadeExpression) return;
       if (parent is FunctionExpressionInvocation) return;
+      if (parent is PrefixExpression) return;
```
bwilkerson: This seems too permissive. It allows anything to be inside the parens as long as they are preceded by a prefix operator. As an extreme example, consider `!(x)`. It seems like users would want to know about these parens. Both of the test cases you included have a space in them, which is the case listed in the issue. But do we really want to allow every nested expression?

srawlins: That extreme example would be reported by the `if (node.expression is SimpleIdentifier)` check.

But I know what you mean. There's not an easy list of nodes that always/usually have whitespace is there? That's the motivating factor, is to allow people to write "unnecessary" parens in locations where whitespace makes it look like `!` binds tighter than it does. We could make the checks something like:

```dart
if ((parent is PrefixExpression) && (
    node is AsExpression ||
    node is AwaitExpression ||
    node is InstanceCreationExpression ||
    node is IsExpression ||
    node is ListLiteral ||
    node is MapLiteral))
```

bwilkerson:
> There's not an easy list of nodes that always/usually have whitespace is there?

No, there isn't. We'd need to make our own list, something like what you've done below. Only with more complex conditions, such as `node is InstanceCreationExpression && (node.keyword != null || node.parameters.length > 1)`.

A hacky idea that I'm sure to regret: `node.toSource().indexOf(' ') >= 0`. Of course, that won't catch nodes like `f(x).y`, which you might also want to allow having parentheses around.

I suspect that what you want as a condition isn't "has whitespace", but "is complex enough to potentially be confusing", but I don't think we have a very precise definition of "complex enough".

srawlins:
OK I think I can narrow down the list, which should make the rule less surprising and specific:

* `!b as bool`'s precedence is actually `(!b) as bool`, so in `!(b as bool)`, the parens are meaningful.
* `!await Future.value(false)` requires no parens, but an AwaitExpression necessarily has whitespace.
* `-const A().a` requires no parens, but an InstanceCreationExpression with a keyword necessarily has parens. I would say that in the face of whitespace _inside_ the constructor call parens, the unnecessary parens do not clear up wrapping. E.g. `!A(1, 2, c: 3)`. So I'm happy _just_ checking for a keyword here.
* `!b is bool`'s precedence is actually `(!b) is bool` so in `!(b is bool)`, the parens are meaningful.
* For ListLiteral, MapLiteral (and our new baby, SetLiteral), we can guard just on the presence of constKeyword.

This simplifies the check to:

```dart
// A prefix expression (! or -) can have an argument wrapped in "unnecessary" parens if
// that argument definitely has potentially confusing whitespace after its first token.
if ((parent is PrefixExpression) && (
    node is AwaitExpression ||
    (node is InstanceCreationExpression && node.keyword != null) ||
    (node is TypedLiteral && node.constKeyword != null))
```

srawlins:
OK, I committed something like what I wrote above. It got a little more complicated, but I think we safely only allow unnecessary parens around a prefix operator's argument in "sane" cases with definite whitespace.

lib/src/rules/unnecessary_parenthesis.dart
```diff
@@ -81,4 +87,24 @@ class _Visitor extends SimpleAstVisitor<void> {
...
   }
+
+  /// Returns whether [node] "starts" with whitespace.
+  ///
+  /// That is, is there definitely whitespace after the first token in [node]?
+  bool _expressionStartsWithWhitespace(Expression node) =>
+      // As in, `!(await foo)`.
+      node is AwaitExpression ||
+      // As in, `!(new Foo())`.
+      (node is InstanceCreationExpression && node.keyword != null) ||
```
bwilkerson: Do we also want to allow parens if there is more than one argument, such as `Foo(a, b)` or even if there is at least one argument `Foo(a + b)`? (Saying "no" and waiting to see whether we get feedback is a fine answer.)

srawlins: I don't think it's confusing when there is whitespace inside _other_, _inner_ parens, such as in a function call. `-(Foo(a + b))` is not more clear than `-Foo(a + b)`.

srawlins:
Thanks, Brian!

PR #1306, opened 2018-12-10
Improve the messages for always_declare_return_types

bwilkerson:
A minor improvement to the messages, and an example of what it will look like when other lints are improved by creating new lint codes (so that we can discuss it).

### DIFF
--- a/lib/src/rules/always_declare_return_types.dart
+++ b/lib/src/rules/always_declare_return_types.dart
@@ -5,7 +5,6 @@
 import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/token.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
-import 'package:analyzer/src/dart/error/lint_codes.dart';
 import 'package:linter/src/analyzer.dart';
 
 const _desc = r'Declare method return types.';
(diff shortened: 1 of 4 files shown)


lib/src/util/condition_scope_visitor.dart
```diff
@@ -6,8 +6,8 @@ import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/token.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
 import 'package:analyzer/dart/element/element.dart';
-import 'package:analyzer/src/generated/resolver.dart';
 import 'package:linter/src/util/dart_type_utilities.dart';
+import 'package:linter/src/analyzer.dart';
```
pq: Sort this before `package:linter/src/util/..`?

lib/src/rules/always_declare_return_types.dart
```diff
@@ -63,21 +63,35 @@ class AlwaysDeclareReturnTypes extends LintRule implements NodeLintRule {
 }
 
 class _Visitor extends SimpleAstVisitor<void> {
+  static const LintCode functionCode = const LintCode(
+      "always_declare_return_types", // ignore: prefer_single_quotes
```
bwilkerson: The convention in both analyzer and front_end is to always use double quotes for messages so that it's easier to use single quotes inside the message. Having to switch back and forth is a pain, but if we really want to enforce this lint, then I can update the existing strings to get rid of the 'ignore'.

bwilkerson:
I should have separated out the clean up of the other warnings, and am happy to split them out if you'd like.

lib/src/rules/always_declare_return_types.dart
pq: No worries.  I'm Ok with it as is.

pq:
> I should have separated out the clean up of the other warnings, and am happy to split them out if you'd like.

Nah.  This is great.  Thanks!

lib/src/util/condition_scope_visitor.dart
bwilkerson: Done

PR #1327, opened 2018-12-21
test_types_in_equals wordsmithing

pq:
/cc @a14n @bwilkerson


pq:
Oh, that's great.  Thanks for that!

PR #1328, opened 2018-12-27
Fix cascade_invocations false positive when examining complicated expressions

### LINKED ISSUE #1323: Incorrect cascade_invocations lint
*@zanderso commented on Dec 20, 2018, 11:41 PM UTC:*

The analyzer cascade_invocations lint asks for cascading across different references:

test.dart:

```dart
class A {
  int f;
  int g;
}

class B {
  final A a = A();
}

void main() {
  final B b1 = B();
  final B b2 = B();

  b1.a
    ..f = 1
    ..g = 1;
  b2.a.f = 2;
}
```

With analysis_options.yaml:

```yaml
linter:
  rules:
    - cascade_invocations
```

Gives

```
$ dartanalyzer ../cascade_bug_test.dart 
Analyzing ../cascade_bug_test.dart...
  lint • Cascade consecutive method invocations on the same reference at /usr/local/google/home/zra/dart/cascade_bug_test.dart:17:3 • cascade_invocations
1 lint found.

```

Where line 17 is `b2.a.f = 2;` where b2 is a different reference and can't be cascaded with the `b1.a` cascade.

*This issue was moved by [stereotype441](https://github.com/stereotype441) from [dart-lang/sdk#35469](https://github.com/dart-lang/sdk/issues/35469).*

srawlins:
Fixes #1323 

The lint rule uses `getCanonicalElementFromIdentifier` to see if two statements act on the same element, in order to report that they can be combined with a cascade. But unless the expressions are simple, up to a point, the rule cannot know that one statement acts on `b1.a` and the second acts on `b2.a`, even though both act on `A get a`.

### DIFF
--- a/test/rules/cascade_invocations.dart
+++ b/test/rules/cascade_invocations.dart
@@ -109,7 +109,7 @@ void cascadeStatic() {
   StaticFoo.decrement();
 }
 
-// Bug 339
+// https://github.com/dart-lang/linter/issues/339
 class Identifier339 {
   String value;
   String system;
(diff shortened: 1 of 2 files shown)


test/rules/cascade_invocations.dart
```diff
@@ -194,3 +194,40 @@ void function694() {
   Bug694.bar = 2;
   Bug694.foo = 3; // OK
 }
+
+class A {
+  int f;
+  int g;
+
+  void x() {}
```
alexeieleusis: You don't seem to be using these methods. If so, remove them.

alexeieleusis:
Thanks for the PR, quite a contrived use case!

test/rules/cascade_invocations.dart
srawlins: Done. Good catch.

PR #1331, opened 2018-12-28
Fix two false positive bugs with a new element equality utility

### LINKED ISSUE #1329: join_with_return_assignment false positive on complex statements
Take the following code:

```dart
class A {
  int f;
  int g;
}

class B {
  final A a = A();
}

int main() {
  final B b1 = B();
  final B b2 = B();

  b1.a
    ..f = 1
    ..g = 1;
  b2.a.f = 2; // Error here
  return b1.a.f;
}
```

The `join_return_with_assignment` rule reports at the error comment "Join return statement with assignment when possible," because it does not see that `b1.a` is not equal to `b2.a`.

(Same cause as #1323)

srawlins:
Fixes #1329 and #1330

### DIFF
--- a/lib/src/rules/join_return_with_assignment.dart
+++ b/lib/src/rules/join_return_with_assignment.dart
@@ -35,15 +35,15 @@ class A {
 
 ''';
 
-Element _getElementFromAssignmentStatement(Statement node) {
+Expression _getExpressionFromAssignmentStatement(Statement node) {
   final visitor = new _AssignmentStatementVisitor();
   node.accept(visitor);
-  return visitor.element;
+  return visitor.expression;
 }
 
-Element _getElementFromReturnStatement(Statement node) {
+Expression _getExpressionFromReturnStatement(Statement node) {
   if (node is ReturnStatement) {
-    return DartTypeUtilities.getCanonicalElementFromIdentifier(node.expression);
+    return node.expression;
   }
   return null;
 }
@@ -65,11 +65,10 @@ class JoinReturnWithAssignment extends LintRule implements NodeLintRule {
 }
 
 class _AssignmentStatementVisitor extends SimpleAstVisitor {
-  Element element;
+  Expression expression;
   @override
   visitAssignmentExpression(AssignmentExpression node) {
-    element =
-        DartTypeUtilities.getCanonicalElementFromIdentifier(node.leftHandSide);
+    expression = node.leftHandSide;
   }
 
   @override
@@ -84,12 +83,12 @@ class _AssignmentStatementVisitor extends SimpleAstVisitor {
 
   @override
   visitPostfixExpression(PostfixExpression node) {
-    element = DartTypeUtilities.getCanonicalElementFromIdentifier(node.operand);
+    expression = node.operand;
   }
 
   @override
   visitPrefixExpression(PrefixExpression node) {
-    element = DartTypeUtilities.getCanonicalElementFromIdentifier(node.operand);
+    expression = node.operand;
   }
 }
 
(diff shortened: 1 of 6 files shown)


lib/src/rules/join_return_with_assignment.dart
```diff
@@ -107,17 +106,27 @@ class _Visitor extends SimpleAstVisitor<void> {
...
     if (length >= 3) {
       final thirdLastStatement = statements[length - 3];
-      thirdLastElement = _getElementFromAssignmentStatement(thirdLastStatement);
+      thirdLastExpression =
+          _getExpressionFromAssignmentStatement(thirdLastStatement);
     }
-    if (lastElement != null &&
-        secondLastElement != thirdLastElement &&
-        lastElement == secondLastElement) {
+    if (lastExpression != null &&
```
bwilkerson: Unless I'm missing something, `lastExpression` cannot be `null` at this point because of the `if` on line 115. We should remove this condition.

lib/src/rules/prefer_conditional_assignment.dart
```diff
@@ -37,33 +37,30 @@ String get fullName {
...
   }
   if (statement is Block && statement.statements.length == 1) {
-    return _checkStatement(statement.statements.first, element);
+    return _checkStatement(statement.statements.first, condition);
   }
   return false;
 }
 
-Element _getElementInCondition(Expression rawExpression) {
+Expression _getElementInCondition(Expression rawExpression) {
```
bwilkerson: Consider changing the name of the method given that it no longer returns an element.

lib/src/util/dart_type_utilities.dart
```diff
@@ -49,6 +49,69 @@ class DartTypeUtilities {
...
+  ///     a1.b.c; // statement 1
+  ///     a2.b.c; // statement 2
+  ///     a1.b.c; // statement 3
+  ///
+  /// The canonical elements from statements 1 and 2 are different, because a1
+  /// is not the same element as a2.  The canonical elements from statements 1
+  /// and 3 are considered to be equal, even though `A.b` may have side effects
+  /// which alter the returned value.
+  static bool canonicalElementsFromIdentifiersAreEqual(
+      AstNode rawNode1, AstNode rawNode2) {
```
bwilkerson: Consider dropping the prefix "raw" from these. I don't think it adds to the semantic value.

lib/src/util/dart_type_utilities.dart
```diff
@@ -49,6 +49,69 @@ class DartTypeUtilities {
...
+  ///
+  /// The two nodes must both be a [SimpleIdentifier], [PrefixedIdentifier], or
+  /// [PropertyAccess] (otherwise `false` is returned).
+  ///
+  /// If the two nodes are PrefixedIdentifiers, or PropertyAccess nodes, then
+  /// `true` is returned only if their canonical elements are equal, in
+  /// addition to their prefixes' and targets' (respectfully) canonical
+  /// elements.
+  ///
+  /// There is an inherent assumption about pure getters. For example:
```
bwilkerson: Thanks for calling this out! I'm not sure that this is a good assumption to make, but given that it's explicit I'm willing to leave it.

lib/src/util/dart_type_utilities.dart
```diff
@@ -49,6 +49,69 @@ class DartTypeUtilities {
...
+  ///     a2.b.c; // statement 2
+  ///     a1.b.c; // statement 3
+  ///
+  /// The canonical elements from statements 1 and 2 are different, because a1
+  /// is not the same element as a2.  The canonical elements from statements 1
+  /// and 3 are considered to be equal, even though `A.b` may have side effects
+  /// which alter the returned value.
+  static bool canonicalElementsFromIdentifiersAreEqual(
+      AstNode rawNode1, AstNode rawNode2) {
+    if (rawNode1 is! Expression || rawNode2 is! Expression) {
```
bwilkerson: I believe that in all of the current use sites we statically know that the arguments are expressions. Consider changing the type of the parameters to `Expression`, removing this test, and removing the casts below.

lib/src/rules/join_return_with_assignment.dart
```diff
@@ -35,15 +35,15 @@ class A {
...
-Element _getElementFromAssignmentStatement(Statement node) {
+Expression _getExpressionFromAssignmentStatement(Statement node) {
   final visitor = new _AssignmentStatementVisitor();
   node.accept(visitor);
-  return visitor.element;
+  return visitor.expression;
 }
 
-Element _getElementFromReturnStatement(Statement node) {
+Expression _getExpressionFromReturnStatement(Statement node) {
```
pq: Perhaps consider ternary and short-hand:

    => node is ReturnStatement ? node.expression : null;

?

lib/src/util/dart_type_utilities.dart
pq: Aside: I'm intrigued by the idea of `@sideEffects`-related annotations and what they could enable us to do with analysis.  I see there's a conversation started elsewhere so I'll follow with interest!

bwilkerson: I do think it's an incorrect assumption. Getters are not guaranteed to be side-effect free, so we should not assume that they are. Making that assumption leads to false positives, which is a problem in any kind of static analysis.

That said, I'm assuming that the assumption isn't new, and at least now all of the logic based on this assumption is in one place. That's definitely an improvement.

srawlins: I filed an issue for `@sideEffects`, https://github.com/dart-lang/sdk/issues/35513.

lib/src/rules/join_return_with_assignment.dart
srawlins: Done.

PR #1332, opened 2018-12-29
Allow public const field declarations with omitted type

### LINKED ISSUE #1121: type_annotate_public_apis reports missing type for static const
Consider [the following class member declaration](https://github.com/dart-lang/site-www/blob/78ce4759d1102ad4d96cc392e89874fa72df4404/examples/misc/lib/language_tour/classes/misc.dart#L21-L25): 

```dart
class Queue {
  static const initialCapacity = 16;
}
```

The `type_annotate_public_apis` lint causes this issue to be reported:

```nocode
lint • Type annotate public APIs at lib/language_tour/classes/misc.dart:22:16 • type_annotate_public_apis
```

Being a `static const`, its type is fully determined at compile time from the initializer, so a type annotation would be redundant, and certainly should not be _required_ -- as the lint message seems to suggest.

Of course, such a static const w/o an explicit type annotation can end up having type `dynamic`, which might not be desirable, but that is another issue addressed via the `--no-implicit-dynamic` flag IMHO.

cc @munificent @kwalrath @kevmoo

srawlins:
Fixes #1121

### DIFF
--- a/lib/src/rules/type_annotate_public_apis.dart
+++ b/lib/src/rules/type_annotate_public_apis.dart
@@ -75,7 +75,7 @@ class _Visitor extends SimpleAstVisitor<void> {
   _Visitor(this.rule) : v = new _VisitorHelper(rule);
   @override
   void visitFieldDeclaration(FieldDeclaration node) {
-    if (node.fields.type == null) {
+    if (node.fields.type == null && !node.fields.isConst) {
       node.fields.accept(v);
     }
   }
@@ -115,7 +115,7 @@ class _Visitor extends SimpleAstVisitor<void> {
 
   @override
   void visitTopLevelVariableDeclaration(TopLevelVariableDeclaration node) {
-    if (node.variables.type == null) {
+    if (node.variables.type == null && !node.variables.isConst) {
       node.variables.accept(v);
     }
   }
--- a/test/rules/type_annotate_public_apis.dart
+++ b/test/rules/type_annotate_public_apis.dart
@@ -4,53 +4,52 @@
 
 // test w/ `pub run test -N type_annotate_public_apis`
 
-const X = ''; //LINT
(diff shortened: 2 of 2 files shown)


bwilkerson:
This rule was created to support the Flutter style, so I think we should get Ian's input before changing it. Unfortunately, I can't seem to reference him from this project.

a14n:
I think you mix up with `always_specify_types`. `type_annotate_public_apis` was created to enforce a rule from Effective Dart .

bwilkerson:
That seems likely.

So, has the rule from Effective Dart been updated since this rule was written?

pq:
> This rule was created to support the Flutter style, so I think we should get Ian's input before changing it. 

cc @Hixie

Hixie:
I don't think Flutter would use this. I don't understand why we would want to omit the type in a static const declaration or any declaration. The type is part of what makes the code self-documenting.

pq:
Thanks @Hixie!

@srawlins: it sounds like this would render the rule _less_ useful for Flutter so I think we should reconsider the change.  Thoughts?

srawlins:
You missed a few comments, @pq. Flutter never requested this lint; it came from Effective Dart, but the backing recommendation was accidentally removed. Nothing to do with Flutter.

bwilkerson:
So I'll ask the question again. Has Effective Dart been (or do we know that it will be) re-written to have this exception? If the rule is supposed to support the style guide, then I don't think we should make it behave differently.

srawlins:
@bwilkerson No, Bob's issue, https://github.com/dart-lang/site-www/issues/996, is still open.

mit-mit:
Hi @srawlins is this still being worked on? Can we close it?

srawlins:
Lol the issue that this fixes has been fixed.

PR #1333, opened 2018-12-29
Do not report on functions within functions or anything within private classes

### LINKED ISSUE #1108: type_annotate_public_apis reports issues over local non-API functions
Consider the local `good()` function declared in this test:

```dart
test('foo', () {
  good() { ... }
  expect(good, ...);
});
```

The `type_annotate_public_apis` lint causes this issue to be reported:

```nocode
lint • Type annotate public APIs at test/effective_dart_test.dart:28:7 • type_annotate_public_apis
```

Being a local function, `good()` should not be considered in this lint.

cc @munificent @kevmoo

srawlins:
Fixes #1108

### DIFF
--- a/lib/src/rules/type_annotate_public_apis.dart
+++ b/lib/src/rules/type_annotate_public_apis.dart
@@ -75,14 +75,17 @@ class _Visitor extends SimpleAstVisitor<void> {
   _Visitor(this.rule) : v = new _VisitorHelper(rule);
   @override
   void visitFieldDeclaration(FieldDeclaration node) {
-    if (node.fields.type == null) {
+    if (node.fields.type == null && !_isParentClassOrMixinPrivate(node)) {
       node.fields.accept(v);
     }
   }
 
   @override
   void visitFunctionDeclaration(FunctionDeclaration node) {
-    if (!isPrivate(node.name)) {
+    if (!isPrivate(node.name) &&
+        // Only report on top-level functions, not those declared within the
+        // scope of another function.
+        node.parent is CompilationUnit) {
       if (node.returnType == null && !node.isSetter) {
         rule.reportLint(node.name);
       } else {
@@ -104,7 +107,7 @@ class _Visitor extends SimpleAstVisitor<void> {
 
   @override
   void visitMethodDeclaration(MethodDeclaration node) {
-    if (!isPrivate(node.name)) {
+    if (!isPrivate(node.name) && !_isParentClassOrMixinPrivate(node)) {
       if (node.returnType == null && !node.isSetter) {
         rule.reportLint(node.name);
       } else {
@@ -119,6 +122,9 @@ class _Visitor extends SimpleAstVisitor<void> {
       node.variables.accept(v);
     }
   }
+
+  bool _isParentClassOrMixinPrivate(ClassMember node) =>
+      isPrivate((node.parent as ClassOrMixinDeclaration).name);
 }
 
 class _VisitorHelper extends RecursiveAstVisitor {
--- a/test/rules/type_annotate_public_apis.dart
+++ b/test/rules/type_annotate_public_apis.dart
@@ -10,6 +10,11 @@ f() {} //LINT
 
 void g(x) {} //LINT
 
+void h() {
+  void i(x) {} // OK
+  j() {} // OK
+}
+
 typedef Foo(x); //LINT
 
 typedef void Bar(int x);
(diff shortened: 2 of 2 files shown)


kevmoo:
@srawlins – just to be crazy. Could you add the edge case of a public class that extends a private class w/ a public member?

In that case, you DO want the public members to be documented. Since they are all within the same library, it shouldn't be too hard....or maybe add a todo?

srawlins:
Oh, are you saying this should report a lint, @kevmoo ?

```dart
class _A {
  // This may _look_ like it's just a member of a private class, but a ha!
  // There is a _public_ class which extends it (which is a weird thing to do),
  // and so [f] is part of the public API!!
  void f(x, y) {}
}

class B extends _A {}
```

I will be very happy to write a TODO for this 😄 .

lib/src/rules/type_annotate_public_apis.dart
```diff
@@ -119,6 +122,9 @@ class _Visitor extends SimpleAstVisitor<void> {
       node.variables.accept(v);
     }
   }
+
+  bool _isParentClassOrMixinPrivate(ClassMember node) =>
+      isPrivate((node.parent as ClassOrMixinDeclaration).name);
```
pq: This cast looks iffy but maybe it's reasonable not to check first...  @bwilkerson?

srawlins: Yeah I don't think any of the nodes in the tree override `parent` to specify a tighter type. How is this typically handled?

bwilkerson: The question is: will a `ClassMember` ever be a child of something other than a `ClassOrMixinDeclaration`? I don't know. I don't believe that can happen today, but 3 months ago it was true that the parent was always a `ClassDeclaration` and then we added mixins to the language. Who knows what we'll add in the future.

The bottom line is that we can't make this completely future proof. Given that, maybe it's better for the method to fail when our expectations are violated than to have the method silently return the wrong answer.

srawlins: SG. When you say, "maybe it's better for the method to fail..." do you mean maybe its better to leave the method as it is?

srawlins:
OK, I've added that TODO.

lib/src/rules/type_annotate_public_apis.dart
```diff
@@ -75,14 +75,24 @@ class _Visitor extends SimpleAstVisitor<void> {
...
   void visitFieldDeclaration(FieldDeclaration node) {
-    if (node.fields.type == null) {
+    if (node.fields.type == null && !_isParentClassOrMixinPrivate(node)) {
+      // This is an untyped field in a public class or mixin.
+
+      // kevmoo@ points out that a private class with a public field may be
+      // extended by a public class which inherits the private class's
+      // implementation of that field, in which case it becomes public API.
+      // TODO(srawlins): Add logic checking "is this field somehow inherited in
+      // a public class"?
```
bwilkerson: We don't actually have the ability to check for public subclasses. Doing so requires a closed world assumption, and analyzer doesn't support that assumption.

It isn't clear to me that we want to ignore members in private classes, but if we do, then we won't be able to fix the case that Kevin pointed out.

srawlins: Really? How can someone define a public class extending a private one, but that couldn't be visited by the same linter rule visitor?

bwilkerson: You're partially right. I was definitely forgetting that we're talking about private classes that won't be visible outside the library. That said, the visitor won't see code from other parts.

srawlins: Ahh, good point.

Hmm, yeah this is an interesting lint rule. Nowhere is "public API" defined ☹️.

I would be super surprised to receive a lint report that I should document a public member of a private class as "public API", but we actually don't have a bug report on that 😛.

I could punt on the private stuff, and just implement the fix for #1108 here regarding functions-within-functions. WDYT?

bwilkerson: > ... just implement the fix for #1108 here regarding functions-within-functions.

Sounds good to me.

> Nowhere is "public API" defined ...

Actually, it's defined in two ways in two places, which is sort of equivalent. :-)

The spec says:

> A declaration is _private_ iff its name is _private_, otherwise it is _public_.

While `pub` defines what we mean by a public library (https://www.dartlang.org/tools/pub/package-layout#public-directories).

I don't know which definition we want to use here.

srawlins: Ok undone for now.

Issue #1027, opened 2018-06-14
Lint rule for `typedef Foo<T> = Function<T>(T x)`


chalin:
From https://github.com/dart-lang/sdk/issues/33165#issuecomment-397356116:

Consider adding a lint rule for the following situation:

```dart
typedef Foo<T> = Function<T>(T x);
```

cc @eernstg @leafpetersen @kevmoo @kwalrath 

_Maybe_ we also need a lint for the following case, but I'd consider it a distinct issue:

```dart
typedef Foo<T> = Function();
```

(We currently don't complain about `class C<T> {}` so maybe we shouldn't complain about a typedef that doesn't use it type argument.)

chalin:
More generally, I guess that the rule would catch

```dart
typedef Foo<T1,...,Tn> = U0 Function<S1,...,Sk,Ti,Sk+1,...Sm>(U1,...,Uq);
```

where one of the U<sub>i</sub> is T<sub>j</sub> for some _j_ between 1 and _n_.

srawlins:
There is now an `avoid_shadowing_type_parameters` lint rule, which should (but does not) catch this, correct?

I don't understand your general rule, @chalin, from [above](https://github.com/dart-lang/linter/issues/1027#issuecomment-397367443), but that <code>T<sub>i</sub></code> in the type parameters on `Function` should be caught, right?

srawlins:
Dart 2.1.0 ships with linter 0.1.71, but avoid_shadowing_type_parameters is introduced in linter 0.1.72. https://github.com/dart-lang/linter/blob/0.1.76/CHANGELOG.md#0172

Did I just address this issue entirely here then? https://github.com/dart-lang/linter/commit/52cef3a7f755f111ccbe0bf6334b24cd3023fedb

chalin:
When will that fix be released?

srawlins:
It was released in 0.1.77: https://github.com/dart-lang/linter/blob/master/CHANGELOG.md#0177

chalin:
Which SDK would contain 0.1.77?

srawlins:
linter 0.1.78 is in SDK 2.1.1-dev.1.0: https://github.com/dart-lang/sdk/blob/2.1.1-dev.1.0/DEPS

pq:
Sam beat me to it! 👍

PR #1334, opened 2018-12-30
Report shadowed type parameters in generic typedefs

srawlins:
Partial fix for #1027 but I suspect not the whole story...

### DIFF
--- a/lib/src/rules/avoid_shadowing_type_parameters.dart
+++ b/lib/src/rules/avoid_shadowing_type_parameters.dart
@@ -40,8 +40,9 @@ class AvoidShadowingTypeParameters extends LintRule implements NodeLintRule {
   void registerNodeProcessors(NodeLintRegistry registry,
       [LinterContext context]) {
     final visitor = _Visitor(this);
-    registry.addMethodDeclaration(this, visitor);
     registry.addFunctionDeclarationStatement(this, visitor);
+    registry.addGenericTypeAlias(this, visitor);
+    registry.addMethodDeclaration(this, visitor);
   }
 }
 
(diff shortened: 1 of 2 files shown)


lib/src/rules/avoid_shadowing_type_parameters.dart
```diff
@@ -59,6 +60,11 @@ class _Visitor extends SimpleAstVisitor<void> {
     _checkAncestorParameters(functionExpression.typeParameters, node);
   }
 
+  @override
+  void visitGenericTypeAlias(GenericTypeAlias node) {
+    _checkForShadowing(node.functionType.typeParameters, node.typeParameters);
```
bwilkerson: `_checkForShadowing` doesn't check the first parameter to ensure that it isn't `null`. We should either change that (and clean up existing code) or add a null check here.

test/rules/avoid_shadowing_type_parameters.dart
```diff
@@ -51,3 +51,7 @@ class E {
 
   void fn4<T>() {} // OK
 }
+
+typedef Fn1<T> = void Function<T>(T); // LINT
+typedef Fn2<T> = void Function<U>(T); // OK
+typedef Fn3<T> = void Function<U>(U); // OK
```
bwilkerson: Let's add two more test cases:
```
typedef Fn4<T> = void Function(T); // OK
typedef Fn5 = void Function<T>(T); // OK
```

lib/src/rules/avoid_shadowing_type_parameters.dart
srawlins: Ah, good catch. I think I like null-checking before the function here. Done.

test/rules/avoid_shadowing_type_parameters.dart
srawlins: Done, thanks.

Issue #1340, opened 2019-01-03
[docs] badge minimum supporting SDK


pq:
Follow-up from https://github.com/dart-lang/linter/issues/1108#issuecomment-451235677, let's add an SDK badge w/ the number of the minimum supporting SDK.

srawlins:
Oh man, +10, that's a great idea! This would be, like, a String on each lint rule? `introduced: "2.1.0"`?

pq:
@srawlins: exactly.

Thinking a bit about how we'd do it, I suppose we could just grab `DEPS` and grep?

https://github.com/dart-lang/sdk/blob/2.1.0-dev.0.0/DEPS
https://github.com/dart-lang/sdk/blob/2.1.0-dev.1.0/DEPS

etc.

@kevmoo: is there an easy way to enumerate releases (besides querying github for tags...)

pq:
FWIW, I'm working on this now.

Once we have this information, thoughts on where it can best be used?

For starters, I'll add a new field to generated linter docs w/ linter version and minimum SDK.  Anywhere else?

/cc @chalin @srawlins @kevmoo

chalin:
> generated linter docs w/ linter version and minimum SDK

Sounds like a very good place to start.

(Is there still a wish to integrate the linter doc under dartlang.org? #949 cc @kwalrath)

pq:
I'm not sure about dartlang.org integration but happy to discuss it.

In the meantime, I'm close to having things in place to add linter version and minimum SDK to the generated docs.  I'll let you know when there's something to look at.

pq:
Related to Dart SDK, it would interesting to finish the thought by adding the including Flutter SDK as well. It'd be great to have a mapping of Flutter [release tags](https://github.com/flutter/flutter/tags) to version numbers for the Dart SDK they vend in.

Does anyone know if there's an easier way than starting w/ the revision in engine and following the thread from there?

https://github.com/flutter/engine/blob/714856123d5c0602dd798e13437f455074342a5c/DEPS#L34

Or have any other thoughts?

@Hixie @tvolkert @goderbauer : is this mapping maintained anywhere that I'm missing?

PR #1352, opened 2019-01-06
Generate "since" version / SDK info.

pq:
Add generated version info to rule descriptions.

image

Fixes: #1340.

/cc @bwilkerson @srawlins

### DIFF
--- a/tool/doc.dart
+++ b/tool/doc.dart
@@ -12,6 +12,8 @@ import 'package:linter/src/analyzer.dart';
 import 'package:linter/src/rules.dart';
 import 'package:markdown/markdown.dart';
 
+import 'since.dart';
+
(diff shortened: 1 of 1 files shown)


chalin:
Very nice!

> I'm not convinced that the linter version is useful information for users, given that they have no way to control which version of the linter package they're using. The SDK version, though, is very useful, and I think a Flutter version would also be useful when we have that available.

I agree. Maybe the the linter version info could be less prominent? Also, I'm thinking that the Dart SDK label should indicate that it is showing the minimum SDK version? E.g., maybe something like:

    Dart SDK: >=2.0.0. <small>(Since linter v0.1.8)</small>

cc @kwalrath

chalin:
👍 Thanks for the quick fix! :)

PR #1355, opened 2019-01-07
Dart SDK info format tweaks.

pq:
Follow-up from https://github.com/dart-lang/linter/pull/1352#issuecomment-452087141.

Before:

image

After:

image

Thanks for the feedback!

/cc @kwalrath @chalin


chalin:
LGTM, thanks!

PR #1360, opened 2019-01-14
Fix two analysis errors in tests

kevmoo:
Seems weird that these haven't been addressed yet...


kevmoo:
I don't know much about this repo, so please make sure I'm not breaking something I don't understand...

PR #1388, opened 2019-01-21
add flutter_color_instantiation lint

a14n:
Address flutter/flutter#26800

### DIFF
--- a/lib/src/rules/flutter_color_instantiation.dart
+++ b/lib/src/rules/flutter_color_instantiation.dart
@@ -6,13 +6,14 @@ import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
 import 'package:linter/src/analyzer.dart';
 
-const _desc = r'Prefer a 8-hex integer(0xFFFFFFFF) to instantiate Color.';
+const _desc =
+    r'Prefer an 8-digit hexadecimal integer(0xFFFFFFFF) to instantiate Color.';
 
 const _details = r'''
 
-Prefer a 8-hex integer(0xFFFFFFFF) to instantiate Color. Colors have four 8-bit
-channels, which adds up to 32 bits, so Colors are described using a 32 bit
-integer.
+Prefer an 8-digit hexadecimal integer(0xFFFFFFFF) to instantiate Color. Colors
+have four 8-bit channels, which adds up to 32 bits, so Colors are described
+using a 32 bit integer.
 
 **BAD:**
 ```
--- a/lib/src/rules/flutter_color_instantiation.dart
+++ b/lib/src/rules/flutter_color_instantiation.dart
@@ -1,0 +1,65 @@
+// Copyright (c) 2019, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:linter/src/analyzer.dart';
+
+const _desc = r'Prefer a 8-hex integer(0xFFFFFFFF) to instantiate Color.';
+
+const _details = r'''
+
+Prefer a 8-hex integer(0xFFFFFFFF) to instantiate Color. Colors have four 8-bit
+channels, which adds up to 32 bits, so Colors are described using a 32 bit
+integer.
+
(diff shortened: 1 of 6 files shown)


pq:
Are there other places with a similar preference?  If so, I wonder if we couldn't annotate `Color` and elsewhere so we could make this a little more general purpose?

Hixie:
yay!

lib/src/rules/flutter_color_instantiation.dart
```diff
@@ -0,0 +1,65 @@
+// Copyright (c) 2019, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:linter/src/analyzer.dart';
+
+const _desc = r'Prefer a 8-hex integer(0xFFFFFFFF) to instantiate Color.';
```
bwilkerson: "a 8-hex" --> "an 8-digit hexadecimal"

lib/src/rules/flutter_color_instantiation.dart
```diff
@@ -0,0 +1,65 @@
...
+
+const _details = r'''
+
+Prefer a 8-hex integer(0xFFFFFFFF) to instantiate Color. Colors have four 8-bit
```
bwilkerson: ditto

pq:
> I haven't found other places in Flutter where 8-digit hexadecimal is used.

Thanks.

Are there other API contracts like this we could consider grouping under a single `flutter_contracts` header (or something)?  Thinking ahead, do we think there will be lots of micro-rules like these? And if so, do we want to be able to individually opt-in/out or would it be sufficient to treat them as a single group?

a14n:
>Are there other API contracts like this we could consider grouping under a single flutter_contracts header (or something)?

Nothing I'm aware of related to hexa ints.

>Thinking ahead, do we think there will be lots of micro-rules like these?

No idea. @Hixie might know.

>And if so, do we want to be able to individually opt-in/out or would it be sufficient to treat them as a single group?

I don't really like regrouping unrelated things under a single lint. I'd rather create a tag/group _flutter_ if their number grows in future.

Hixie:
We have a whole list of lints in our wishlist:
  https://github.com/flutter/flutter/issues/1220
(Some of those ideas are a bit out of date, and there's probably other things we could add.)

pq:
Thanks @Hixie!  That list is a walk down memory lane. 😄 

No obvious groupings are jumping out at me so maybe we can just think a little on alternative names.

Ideas:

1. `flutter_color_hex_initializers`
2. `flutter_color_initializers`
3. `flutter_color_constructor_args`
4. `flutter_colors` (possibly leaving the door open to this becoming a larger "use flutter colors appropriately" meta-lint)

Thoughts?

bwilkerson:
I'm not fond of the `flutter_` prefix. I suppose it's there to say "if we had a flutter-specific plugin this would be in it". But if we were putting this in a flutter specific plugin would we really prefix the lints in that plugin?

Despite the fact that we used that prefix in one other rule (which I wish we hadn't), I'm thinking something like `use_full_hex_values_for_colors` or `avoid_short_hex_values_for_colors`. Something that better indicates what the rule checks for and doesn't use our character budget on a prefix that doesn't really help users understand the lint.

Hixie:
FWIW, I have no strong opinion either way on the name. The prefix `flutter_` would make it clearer to people writing other frameworks that this lint is hard-coded to `dart:ui`, though, and that they can't make use of it. (I guess that's an argument in favour of making it more generic, too.)

bwilkerson:
To be clear, I have no problem marking our framework-specific lints as being such. I do question whether the name is the best way to mark them.

a14n:
So what should I do? Rename to `flutter_use_full_hex_values_for_colors`?

bwilkerson:
Like I said, I'd like us to consider adding support for categorizing lints that's independent of the name.

a14n:
`use_full_hex_values_for_flutter_colors` ? I think it's important to know this rule apply to `Color` class in Flutter (and not on `color` property in `dart:html`)

pq:
> `use_full_hex_values_for_flutter_colors` ? 

My favorite by far.  Let's run with it.

a14n:
Renamed to `use_full_hex_values_for_flutter_colors`

PR #1389, opened 2019-01-24
More allowances in unnecessary_parenthesis

### LINKED ISSUE #1311: unnecessary_parenthesis should not be so strict around cascades
There are currently some allowances for "unnecessary" parenthesis around cascades:

```dart
// a..b=(c..d) is OK
if (node.expression is CascadeExpression &&
    node.thisOrAncestorMatching(
            (n) => n is Statement || n is CascadeExpression)
        is CascadeExpression) {
  return;
}
```

This says "If the parenthesized thing is a cascade, _and_ the parentheses are in some expression hierarchy with a cascade above them, then allow the parens."

I think this should be changed to an "or":

* If the parenthesized thing is a cascade, leave it alone.
* If the parentheses have a cascade somewhere above them as an ancestor, leave it alone.

Here are some (simplified) examples I've seen, where this lint rule complains today, which highlight common "unnecessary" parentheses patterns, which I think we should allow:

```dart
final response = await _somethingOrOtherClient.sendThingy(
    SendThingyRequest()
      ..foo = foo
      ..verificationBar = (isSomethingSelected
          ? SendThingyRequest_VerificationBar.ONE
          : SendThingyRequest_VerificationBar.TWO));
```

Here the parens defensively guard against unknown cascade vs conditional (`?:`) precedence. They arguably make the code more readable, even when dartfmt helps as well with proper indentation.

```dart
GeneralSomethingOrOtherGraph newGraph = GeneralSomethingOrOtherGraph()
  ..setting = (setting.clone()
    ..categorySetting = (setting.categorySetting.clone())
    ..categorySetting.foo = true
    ..explicitBar = false);
```

These parens again guard against the precedence misunderstandings with nested cascades, method calls and assignments.

```dart
ReportSomethingEncoder get eventSomethingEncoder => _forStringField('foo',
    getValue: (definition) =>
        definition.graphDefinition.showThingyTable ? 'x' : '',
    setValue: (definition, value) => definition.graphDefinition =
        (GraphDefinition()..showThingyTable = value == 'x'));
```

Here the parens guard against precedence between the inner cascade and an outer assignment.

Rather than play more games with how to allow cascades, I propose to just allow "unnecessary" parens around any cascade, and below any cascade.

srawlins:
Fixes #1311

### DIFF
--- a/lib/src/rules/unnecessary_parenthesis.dart
+++ b/lib/src/rules/unnecessary_parenthesis.dart
@@ -60,7 +60,7 @@ class _Visitor extends SimpleAstVisitor<void> {
     }
 
     // a..b=(c..d) is OK
-    if (node.expression is CascadeExpression &&
+    if (node.expression is CascadeExpression ||
         node.thisOrAncestorMatching(
                 (n) => n is Statement || n is CascadeExpression)
             is CascadeExpression) {
--- a/test/rules/unnecessary_parenthesis.dart
+++ b/test/rules/unnecessary_parenthesis.dart
@@ -14,7 +14,8 @@ main() async {
   print(1); // OK
   print((1)); // LINT
   if (a && b || c && d) true; // OK
-  if ((a && b) || c && d) true; // OK because it may be hard to know all precedence rules
+  // OK because it may be hard to know all of the precedence rules.
+  if ((a && b) || c && d) true; // OK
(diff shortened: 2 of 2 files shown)


srawlins:
CC @a14n, lemme know what you think. I'd be very happy to add more test cases if you are worried about this becoming too permissive.

PR #1406, opened 2019-02-01
update description

pq:
Nudged by `pana`:

image

Word-smithing appreciated!

/cc @bwilkerson

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -3,7 +3,10 @@ version: 0.1.79
 
 author: Dart Team <misc@dartlang.org>
 
-description: Style linter for Dart.
+description: >-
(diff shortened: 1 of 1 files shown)


pubspec.yaml
```diff
@@ -3,7 +3,10 @@ version: 0.1.79
 
 author: Dart Team <misc@dartlang.org>
 
-description: Style linter for Dart.
+description: >-
+  A linter for Dart, bundled with the SDK, that helps enforce coding style,
+  flags potentially difficult to maintain, error-prone or non-performant
+  source and generally nudges you towards writing more idiomatic Dart.
```
bwilkerson: Perhaps add explicit text to the effect that we don't expect other packages to use this package?

PR #1407, opened 2019-02-01
add example readme

pq:
Nudged by `pana`:

image

/cc @bwilkerson

### DIFF
--- a/example/readme.md
+++ b/example/readme.md
@@ -1,0 +1,12 @@
+# Examples
+
+Some example analysis options files.
+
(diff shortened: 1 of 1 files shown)


isoos:
> I wonder if there's a natural way to give special treatment to pseudo-packages like this?

It looks like that this example file would fit into the pub site's presentation of the package very nicely, and in the end I'd consider the `pana` suggestion in this specific case as "working as intended". :) Thanks for doing the PR!

pq:
ha.  Point taken.  Thanks @isoos!

PR #1408, opened 2019-02-01
migrate to inheritance_manager2

### LINKED ISSUE #1405: fix deprecated analyzer API references
`lookupInheritance` and `lookupMember` are flagged as deprecated:

https://github.com/dart-lang/linter/blob/db09b38aa1bff0e9200bc0c77d8e785e6cf76122/lib/src/rules/annotate_overrides.dart#L83

https://github.com/dart-lang/linter/blob/db09b38aa1bff0e9200bc0c77d8e785e6cf76122/lib/src/rules/public_member_api_docs.dart#L169

The docs don't say what the preferred approach is.

Analysists: can you advise?

/cc @stereotype441 @bwilkerson @scheglov

pq:
Fixes #1405

/cc @scheglov @bwilkerson

### DIFF
--- a/lib/src/rules/annotate_overrides.dart
+++ b/lib/src/rules/annotate_overrides.dart
@@ -56,7 +56,7 @@ class AnnotateOverrides extends LintRule implements NodeLintRule {
   @override
   void registerNodeProcessors(NodeLintRegistry registry,
       [LinterContext context]) {
-    final visitor = new _Visitor(this);
+    final visitor = new _Visitor(this, context);
     registry.addCompilationUnit(this, visitor);
     registry.addFieldDeclaration(this, visitor);
     registry.addMethodDeclaration(this, visitor);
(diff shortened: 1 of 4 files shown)


lib/src/rules/annotate_overrides.dart
```diff
@@ -80,15 +80,21 @@ class _Visitor extends SimpleAstVisitor<void> {
...
 
   @override
   void visitCompilationUnit(CompilationUnit node) {
     LibraryElement library = node == null
         ? null
         : resolutionMap.elementDeclaredByCompilationUnit(node)?.library;
-    manager = library == null ? null : new InheritanceManager(library);
+    manager = library == null
+        ? null
+        : new InheritanceManager2(library.context.typeSystem);
```
scheglov: `LinterContext` has `typeSystem`.
Ideally it should also have `inheritanceManager` instance as well.

pq: Cool.  Thanks!

PR #1421, opened 2019-02-08
update prefer_collection_literals to support sets

### LINKED ISSUE #1413: update prefer_collection_literals to recognize set literals
Follow-up from #1411.  Issue tracking updates to `prefer_collection_literals`.

pq:
/cc @bwilkerson 

Fixes #1413.

### DIFF
--- a/lib/src/rules/prefer_collection_literals.dart
+++ b/lib/src/rules/prefer_collection_literals.dart
@@ -57,7 +57,7 @@ class _Visitor extends SimpleAstVisitor<void> {
     if (node.methodName.name != 'toSet') {
       return;
     }
-    if (node.target is ListLiteral) {
+    if (node.target is ListLiteral || node.target is ListLiteral2) {
       rule.reportLint(node);
     }
   }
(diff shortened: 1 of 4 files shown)


lib/src/rules/prefer_collection_literals.dart
```diff
@@ -40,23 +43,47 @@ class PreferCollectionLiterals extends LintRule implements NodeLintRule {
...
+        rule.reportLint(node);
+      }
+      return;
+    }
+
+    // Sets.
+    if (DartTypeUtilities.isClass(node.staticType, 'Set', 'dart.core') ||
+        DartTypeUtilities.isClass(
+            node.staticType, 'LinkedHashSet', 'dart.collection')) {
+      if (constructorName != 'identity' && constructorName != 'unmodifiable') {
```
bwilkerson: It would be safer to make this a whitelist rather than a blacklist. As unlikely as it seems, it means that we won't start linting new `Set` constructors accidentally.

lib/src/rules/prefer_collection_literals.dart
```diff
@@ -40,23 +43,47 @@ class PreferCollectionLiterals extends LintRule implements NodeLintRule {
...
-        (DartTypeUtilities.isClass(node.staticType, 'List', 'dart.core') ||
-            DartTypeUtilities.isClass(node.staticType, 'Map', 'dart.core') ||
-            DartTypeUtilities.isClass(
-                node.staticType, 'LinkedHashMap', 'dart.collection'))) {
+  void visitMethodInvocation(MethodInvocation node) {
+    // ['foo', 'bar', 'baz'].toSet();
+    if (node.methodName.name != 'toSet') {
+      return;
+    }
+    if (node.target is ListLiteral) {
```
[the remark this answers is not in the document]
pq: Good catch.  Thank you!

lib/src/rules/prefer_collection_literals.dart
pq: Agreed.  Fixed.

PR #1423, opened 2019-02-09
New rule: Give Deprecation message

### LINKED ISSUE #1073: provide a lint if a `@deprecated` annotation has no description of the deprecation
It's useful when deprecating functionality to indicate to the user what the recommended alternative is. So, instead of:

```dart
@deprecated
void foo() {
  ...
}
```

we should prefer:

```dart
@Deprecated('Prefer to use the bar() method')
void foo() {
  ...
}
```

/cc @kevmoo

srawlins:
Fixes #1073.

New lint rule; description is in the rule; it's not complicated 😛.


lib/src/rules/give_deprecation_message.dart
```diff
@@ -0,0 +1,60 @@
+// Copyright (c) 2018, the Dart project authors. Please see the AUTHORS file
```
bwilkerson: nit: year

test/rules/give_deprecation_message.dart
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) 2018, the Dart project authors. Please see the AUTHORS file
```
bwilkerson: nit: year

lib/src/rules/give_deprecation_message.dart
```diff
@@ -0,0 +1,60 @@
...
+""")
+void oldFunction(arg1, arg2) {}
+```
+
+''';
+
+class GiveDeprecationMessage extends LintRule implements NodeLintRule {
+  GiveDeprecationMessage()
+      : super(
+            name: 'give_deprecation_message',
```
bwilkerson: I'm not fond of 'give'. Perhaps something like 'provide' or 'include'?

lib/src/rules/give_deprecation_message.dart
```diff
@@ -0,0 +1,60 @@
+// Copyright (c) 2018, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:linter/src/analyzer.dart';
+
+const _desc = r'Give a deprecation message, via `Deprecated("message")`.';
```
bwilkerson: Please don't use markdown in the description. It won't be translated before being displayed, except in the generated documentation.

lib/src/rules/give_deprecation_message.dart
srawlins: Curiously, I wrote this code in 2018, but waited to send a PR until the analyzer dep was bumped to 0.35.0, so the git commit's date (as seen above) is from Dec 30, 2018. Should the year still read 2019?

lib/src/rules/give_deprecation_message.dart
srawlins: Should I just remove the backticks? Add an `@`?

lib/src/rules/give_deprecation_message.dart
bwilkerson: I'm not a lawyer, so I don't know whether it's suppose to be the date of file creation of the date that it's first committed into the repo, or whether it actually matters. Your call.

lib/src/rules/give_deprecation_message.dart
bwilkerson: Seems like a reasonable choice to me, yeah.

lib/src/rules/give_deprecation_message.dart
pq: +1. "provide" or "include" sound a little better to me too.

test/rules/give_deprecation_message.dart
pq: FWIW: I'm OK with whatever you decide about the date; I might lean towards 2019 but don't care strongly either way.

lib/src/rules/give_deprecation_message.dart
srawlins: Done. 2019.

lib/src/rules/give_deprecation_message.dart
srawlins: Done.

lib/src/rules/give_deprecation_message.dart
srawlins: Done.

test/rules/give_deprecation_message.dart
srawlins: Done. 2019.

PR #1428, opened 2019-02-11
fixes for const typed literals

pq:
Work-in-progress fixes for #1427.

Note that these tests are not expected to pass on travis until a new `analyzer` package is published w/ set-literals support on by default (https://github.com/dart-lang/sdk/issues/35905).

/cc @bwilkerson

### DIFF
--- a/lib/src/rules/prefer_const_declarations.dart
+++ b/lib/src/rules/prefer_const_declarations.dart
@@ -3,6 +3,7 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/token.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
 import 'package:linter/src/analyzer.dart';
 import 'package:linter/src/ast.dart';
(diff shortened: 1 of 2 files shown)


lib/src/rules/prefer_const_declarations.dart
```diff
@@ -80,6 +81,10 @@ class _Visitor extends SimpleAstVisitor<void> {
     if (!node.isFinal) return;
     if (node.variables.every((declaration) =>
         declaration.initializer != null &&
+        (declaration.initializer is! TypedLiteral ||
+            (declaration.initializer is TypedLiteral &&
+                declaration.initializer.beginToken?.keyword ==
+                    Keyword.CONST)) &&
```
pq: I'm sure there's a better way to do this?

test/rules/prefer_const_declarations.dart
```diff
@@ -61,4 +63,13 @@ m() {
 
   // https://github.com/dart-lang/sdk/issues/32745
   final b, c = 1; // OK
+
+  var s = {}; // OK
+  final Set<int> ids = {}; // OK
+  final Set<int> ids2 = <int>{}; // OK
```
pq: Test cases derived from flutter.

Others welcome!

lib/src/rules/prefer_const_declarations.dart
bwilkerson: I'm not thinking of one at the moment, but the condition on line 85 appears to be redundant. Also, you might want to assign `declaration.initializer` to a local variable in order to shorten the code a bit.

test/rules/prefer_const_declarations.dart
bwilkerson: List literals are also `TypedLiteral`s, so perhaps a test case involving `<int>[]`.

PR #1429, opened 2019-02-11
in progress docs for testing w/ a try-job

pq:
In progress; ideas welcome!

/cc @a14n @bwilkerson @srawlins

### DIFF
--- a/doc/testing.md
+++ b/doc/testing.md
@@ -1,0 +1,9 @@
+# Testing against Flutter
+
+When validating a new lint or experimenting with new language experiments, testing the linter
+against flutter is recommended to help expose false positives and other unexpected behavior.
+
+## Try-Bot testing
+1. update the `"linter_tag"` in the SDK [`DEPS`](https://github.com/dart-lang/sdk/blob/master/DEPS)
+   to point to your linter commit
+2. add “flutter-analyze-try” to your chosen try-jobs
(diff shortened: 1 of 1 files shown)


doc/testing.md
```diff
@@ -0,0 +1,9 @@
+# Testing against Flutter
+
+When validating a new lint or experimenting with new language experiments, testing the linter
+against flutter is recommended to help expose false positives and other unexpected behavior.
+
+## Try-Bot testing
```
bwilkerson: "Try-bot Testing"? Also, I think the steps below ought to be capitalized and punctuated as full sentences.

I think having information like this is great.

I don't know whether this is the easiest/best way to test against flutter, but if it is we probably need to provide a pointer to the docs explaining how to create a CL against the SDK. (They'll have to use Gerrit to choose a try bot.) Also, given that it's often a new rule being tested we probably need to explain how to update the analysis options file used by that bot so that the rule is run.

That said, it might be better to explain how to test rules against other packages and then let users run the rule against the flutter package the way they would against any other package. I think they'd still need to update flutter's analysis options file, which will require a pointer to which file is used, but they might not need to learn Gerrit to do it.

pq: I have no idea if this is easiest to be honest...

@a14n : care to add any info on more local testing?

a14n: To test locally I usually run:

```shell
flutter> bin/cache/dart-sdk/bin/dart $LINTER_HOME/bin/linter.dart --rules my-new-rule dev examples packages | grep '\[lint\]'
```

mit-mit:
Hi @pq can we close this?

pq:
Yeah.  Let's re-open when someone has the steam to do it right.

PR #1434, opened 2019-02-12
Add another corner relaxation for unnecessary_parenthesis

srawlins:
The unnecessary_parenthesis rule has  some exceptions for unary operators next to whitespace-y keywords, but was still complaining about this:

```dart
!(const [7]).contains(42);
```

This fixes that.

### DIFF
--- a/lib/src/rules/unnecessary_parenthesis.dart
+++ b/lib/src/rules/unnecessary_parenthesis.dart
@@ -85,8 +85,8 @@ class _Visitor extends SimpleAstVisitor<void> {
       if (parent is MethodInvocation) {
         Expression target = parent.target;
         if (parent.parent is PrefixExpression &&
-            target is ParenthesizedExpression &&
-            _expressionStartsWithWhitespace(target.expression)) return;
+            target == node &&
+            _expressionStartsWithWhitespace(node.expression)) return;
       }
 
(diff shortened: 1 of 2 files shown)


lib/src/rules/unnecessary_parenthesis.dart
```diff
@@ -78,6 +78,17 @@ class _Visitor extends SimpleAstVisitor<void> {
...
       if (parent is PrefixExpression &&
           _expressionStartsWithWhitespace(node.expression)) return;
+
+      // Another case of the above exception, something like
+      // `!(const [7]).contains(5);`, where the _parent's_ parent is the
+      // PrefixExpression.
+      if (parent is MethodInvocation) {
+        Expression target = parent.target;
+        if (parent.parent is PrefixExpression &&
+            target is ParenthesizedExpression &&
```
bwilkerson: While I think that this is technically correct, I think we want to use `target == node` (rather than `target is ParenthesizedExpression`) here. I think that will make it clear that the interesting condition is that the node we're looking at is the target of the invocation. (I think it's technically correct because I can't think of any other child of a `MethodInvocation` that _could_ be a parenthesized expression, but if that ever changes then this would be broken.)

srawlins: Ah, good catch! Done.

PR #1435, opened 2019-02-12
Add support for reporting on parenthesized !isEmpty / !isNotEmpty

srawlins:
While there is a lint rule for unnecessary parentheses, (a) we can't expect everyone has it enabled, and (b) it allows some exceptions. But that shouldn't allow these `!isEmpty` cases to hide!

### DIFF
--- a/lib/src/rules/prefer_is_not_empty.dart
+++ b/lib/src/rules/prefer_is_not_empty.dart
@@ -5,6 +5,7 @@
 import 'package:analyzer/dart/ast/ast.dart'
     show
         AstNode,
+        ParenthesizedExpression,
         PrefixExpression,
         PrefixedIdentifier,
         PropertyAccess,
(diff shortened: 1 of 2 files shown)


lib/src/rules/prefer_is_not_empty.dart
```diff
@@ -90,17 +91,20 @@ class _Visitor extends SimpleAstVisitor<void> {
...
-      return;
-    }
-    PrefixExpression prefixExpression =
-        isEmptyAccess.parent as PrefixExpression;
-    // Should be !
-    if (prefixExpression.operator.type != TokenType.BANG) {
-      return;
+
+    // Walk up any parentheses above the isEmpty / isNotEmpty.
+    AstNode isEmptyParent = isEmptyAccess.parent;
```
bwilkerson: Consider using `isEmptyAccess.unParenthesized`.

srawlins: I'm unable to figure this out... unParenthesized returns the things inside the parens, not outside them.

So in the expression, `(((foo)))`, I am holding onto `foo` (isEmptyAccess), and I'm trying to back out of the parens, but unParenthesized does the opposite.

bwilkerson: Sorry, yes, I missed that. What you're doing is the right (only) way to do this.

PR #1437, opened 2019-02-13
Fix code that is now triggering lints

bwilkerson:
Just some code cleanup.

### DIFF
--- a/lib/src/rules/directives_ordering.dart
+++ b/lib/src/rules/directives_ordering.dart
@@ -241,7 +241,7 @@ class _Visitor extends SimpleAstVisitor<void> {
 
   @override
   void visitCompilationUnit(CompilationUnit node) {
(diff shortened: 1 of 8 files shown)


pq:
As discussed in person, we should wait to land this until `2.2.0-dev.1` is available to travis.

In the meantime, we could consider disabling `prefer_collection_literals`...  Alternatively we could all just patch in these changes locally.  No big opinion here. 🤷‍♂️

pq:
I think we can land this one now. 👍

pq:
Aside: the coverage report is curious... 🤔

bwilkerson:
Agreed. I don't know why the coverage would have dropped. Perhaps whatever we're using to measure it doesn't understand set literals and hence isn't able to compute coverage correctly?

PR #1439, opened 2019-02-14
support for running tests w/ experiments enabled

pq:
Plumbing for running test configurations.

Currently failing badly but that's unsurprising and so runs are disabled until control-flow-collections are further along.

See: https://github.com/dart-lang/linter/issues/1438

/cc @bwilkerson

### DIFF
--- a/test/configs/control-flow-collections/analysis_options.yaml
+++ b/test/configs/control-flow-collections/analysis_options.yaml
@@ -1,0 +1,3 @@
+analyzer:
+  enable-experiment:
+    - control-flow-collections
--- a/test/rule_test.dart
+++ b/test/rule_test.dart
@@ -31,17 +31,28 @@ main() {
 }
 
 final String ruleDir = p.join('test', 'rules');
+final String testConfigDir = p.join('test', 'configs');
 
 /// Rule tests
 defineRuleTests() {
   // TODO: if ruleDir cannot be found print message to set CWD to project root
   group('rule', () {
     group('dart', () {
-      for (var entry in new Directory(ruleDir).listSync()) {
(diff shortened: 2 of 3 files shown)


test/configs/control-flow-collections/analysis_options.yaml
```diff
@@ -0,0 +1,3 @@
+analyzer:
+  enable-experiment:
+    - control-flow-collections
```
bwilkerson: Given that the two are effectively merged at this point, you might want to also add 'spread-collections' to the list. Not a big deal if you want to wait.

test/rule_test.dart
```diff
@@ -31,17 +31,28 @@ main() {
...
   // TODO: if ruleDir cannot be found print message to set CWD to project root
   group('rule', () {
     group('dart', () {
-      for (var entry in new Directory(ruleDir).listSync()) {
-        if (entry is! File || !isDartFile(entry)) continue;
-        var ruleName = p.basenameWithoutExtension(entry.path);
-        testRule(ruleName, entry, debug: true);
-      }
+      // Rule tests run with default analysis options.
+      testRules(ruleDir, analysisOptions: null);
```
bwilkerson: When we enable the code below we might want to wrap this in a group titled '(default)'.

test/configs/control-flow-collections/analysis_options.yaml
pq: I'm hoping that these will get collapsed into one before long and we can just replace it then.

test/rule_test.dart
pq: Sure.

pq:
Thanks!

Issue #1414, opened 2019-02-04
DartTypeUtilities.extendsClass infinite regress


pq:
```
Exception
Dart analysis server, SDK version 2.1.1-dev.2.0, server version 1.22.1, error: Exception while using a NullClosures to visit a InstanceCreationExpressionImpl in ExpressionStatementImpl in BlockImpl in BlockFunctionBodyImpl in FunctionExpressionImpl in FunctionDeclarationImpl in CompilationUnitImpl
Stack Overflow
#0      ClassElementImpl.supertype (package:analyzer/src/dart/element/element.dart:896:3)
#1      InterfaceTypeImpl.superclass (package:analyzer/src/dart/element/type.dart:1550:44)
#2      DartTypeUtilities.extendsClass (package:linter/src/util/dart_type_utilities.dart:18:29)
#3      DartTypeUtilities.extendsClass (package:linter/src/util/dart_type_utilities.dart:18:11)
#4      DartTypeUtilities.extendsClass (package:linter/src/util/dart_type_utilities.dart:18:11)
#5      DartTypeUtilities.extendsClass (package:linter/src/util/dart_type_utilities.dart:18:11)
#6      DartTypeUtilities.extendsClass (package:linter/src/util/dart_type_utilities.dart:18:11)
#7      DartTypeUtilities.extendsClass (package:linter/src/util/dart_type_utilities.dart:18:11)
#8      DartTypeUtilities.extendsClass (package:linter/src/util/dart_type_utilities.dart:18:11)
#9      DartTypeUtilities.extendsClass (package:linter/src/util/dart_type_utilities.dart:18:11)
#10     DartTypeUtilities.extendsClass (package:linter/src/util/dart_type_utilities.dart:18:11)
#11     DartTypeUtilities.extendsClass (package:linter/src/util/dart_type_utilities.dart:18:11)
#12     DartTypeUtilities.extendsClass (package:linter/src/util/dart_type_utilities.dart:18:11)
#13     DartTypeUtilities.extendsClass (package:linter/src/util/dart_type_utilities.dart:18:11)
#14     DartTypeUtilities.extendsClass (package:linter/src/util/dart_type_utilities.dart:18:11)
#15     DartTypeUtilities.extendsClass (package:linter/src/util/dart_type_utilities.dart:18:11)
#16     DartTypeUtilities.extendsClass (package:linter/src/util/dart_type_utilities.dart:18:11)
[...]

PR #1444, opened 2019-02-22
test for cycles in extendsClass

pq:
Fixes: #1414

/cc @bwilkerson

### DIFF
--- a/lib/src/util/dart_type_utilities.dart
+++ b/lib/src/util/dart_type_utilities.dart
@@ -20,7 +20,7 @@ class DartTypeUtilities {
       type is InterfaceType &&
       seenTypes.add(type.element) &&
       (isClass(type, className, library) ||
-          (_extendsClass(type.superclass, seenTypes, className, library)));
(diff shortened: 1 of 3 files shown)


lib/src/util/dart_type_utilities.dart
```diff
@@ -13,9 +13,14 @@ typedef bool AstNodePredicate(AstNode node);
...
-      (type is InterfaceType &&
-          extendsClass(type.superclass, className, library));
+      _extendsClass(type, new Set<ClassElement>(), className, library);
+
+  static bool _extendsClass(DartType type, Set<ClassElement> seenTypes,
+          String className, String library) =>
+      type is InterfaceType &&
+      seenTypes.add(type.element) &&
+      (isClass(type, className, library) ||
+          (_extendsClass(type.superclass, seenTypes, className, library)));
```
bwilkerson: nit: You could remove the parens around the invocation of `_extendsClass`.

pq: 👍 Thanks!

Issue #414, opened 2017-02-09
Warn on call to Map.remove with unrelated type


a14n:
```dart
final map = <String,String>{};
map.remove(1); // LINT
map.remove(""); // OK
```

srawlins:
Seems like this can just be folded into [list_remove_unrelated_type](https://github.com/dart-lang/linter/blob/master/lib/src/rules/list_remove_unrelated_type.dart)? Can't imagine why someone would want to use one and not the other.

a14n:
But the name of the lint is completely unrelated to map.

srawlins:
I don't think that the name should prevent us from re-using it. We can create an alias for `list_remove_unrelated_type` called `remove_unrelated_type` (or better yet, rename the rule and create an alias for the old name).

Is there a reason to not make a breaking change? Just bump the appropriate version number.

a14n:
>Is there a reason to not make a breaking change? Just bump the appropriate version number.

Unfortunately linter package is not like any other package. It's bundled in the sdk with the analyzer. So a simple new release of the sdk may come with a new version of linter. Actually the version number of linter is only used to [link it into the sdk](https://github.com/dart-lang/sdk/blob/f5cd6ccfb73799818ef3c88ea4b4d9e582f63504/DEPS#L98).

Note that it would be a normal package if it'd [become an anlyzer plugin](https://github.com/dart-lang/linter/issues/1097).

bwilkerson:
> Note that it would be a normal package ...

Yes, but that comes with other costs, and it isn't clear to me that that's the best path forward for our users.

srawlins:
Since @pq has been working on deprecation recently, what say we deprecate `list_remove_unrelated_type` and rename to `collection_remove_unrelated_type`?

bwilkerson:
Sounds reasonable. And if we do that, then I'm going to want us to have a way to implement a quick-fix to change the name for users, which our current deprecation support doesn't include.

munificent:
Other operations we might want to either roll into the existing lints or define new lints for:

```dart
Map.containsKey(Object? key)
Map.containsValue(Object? value)
Map.operator[Object? value]
Set.contains(Object? value)
Set.lookup(Object? object)
Set.remove(Object? value)
```

lrhn:
The complete list (as I see it) would be:
```dart
Iterable.contains(Object?)  // inherited by List/Set/Queue too
List.remove(Object?)
Set.remove(Object?)
[...]

PR #1458, opened 2019-02-28
Document and tidy UnrelatedTypesProcessors

srawlins:
I'm looking into #1307 and #414, but it took me forever to understand how UnrelatedTypesProcessors works.

I thought I'd commit my doc comments and a bit of tidying first before mailing a large refactoring.

### DIFF
--- a/test/rules/iterable_contains_unrelated_type.dart
+++ b/test/rules/iterable_contains_unrelated_type.dart
@@ -26,7 +26,7 @@ void someFunction4() {
 
 void someFunction4_1() {
   List list;
-  if(list.contains(null)) print('someFucntion4_1');
+  if (list.contains(null)) print('someFucntion4_1');
 }
 
 void someFunction5_1() {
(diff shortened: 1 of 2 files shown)


test/rules/iterable_contains_unrelated_type.dart
```diff
@@ -133,22 +134,32 @@ bool takesIterable2(Iterable<String> iterable) => iterable.contains('a'); // OK
...
 abstract class A2 implements Iterable<String> {}
+
 abstract class B2 extends A2 {}
+
 bool takesB2(B2 b) => b.contains('a'); // OK
 
-abstract class SomeIterable<E> implements Iterable<E> {}
+abstract class AddlInterface {}
+
+abstract class SomeIterable<E> implements Iterable<E>, AddlInterface {}
```
srawlins: While investigating `_findImplementedInterfaces`, trying to figure out how it works, I couldn't find any test cases where the accumulator ever contains more than a single type. There were no commas in this test before this change, so I thought it was a good little addition to land.

pq:
Late to the party but chiming in to thank you for fixing this up!

🍻

kevmoo:
> @srawlins: I'm looking into #1307

PR #1459, opened 2019-03-01
migrate to new SetOrMapLiteral API

pq:
Migrates to new `SetOrMapLiteral` API.

A few notes on the unification:

* `always_specify_types` was missing the opportunity to flag set literals; this fixes that 👍 
* I added TODOs in places where existing logic was preserved but might better be widened to include Set literals too 🤔 
* I'm not sure what to do about the test for `node is MapLiteralEntry`:  will that interface go away too?

(A similar migration for `ForStatement2 ` to follow.)

/cc @stereotype441 @bwilkerson

### DIFF
--- a/lib/src/rules/prefer_const_literals_to_create_immutables.dart
+++ b/lib/src/rules/prefer_const_literals_to_create_immutables.dart
@@ -63,7 +63,7 @@ class PreferConstLiteralsToCreateImmutables extends LintRule
       [LinterContext context]) {
     final visitor = new _Visitor(this, context);
     registry.addListLiteral(this, visitor);
-    registry.addMapLiteral(this, visitor);
+    registry.addSetOrMapLiteral(this, visitor);
   }
 }
 
@@ -78,7 +78,12 @@ class _Visitor extends SimpleAstVisitor<void> {
   void visitListLiteral(ListLiteral node) => _visitTypedLiteral(node);
 
   @override
-  void visitMapLiteral(MapLiteral node) => _visitTypedLiteral(node);
+  void visitSetOrMapLiteral(SetOrMapLiteral node) {
+    // todo (pq): should this apply to set literals as well?
+    if (node.isMap) {
+      _visitTypedLiteral(node);
+    }
+  }
 
   Iterable<InterfaceType> _getSelfAndInheritedTypes(InterfaceType type) sync* {
     InterfaceType current = type;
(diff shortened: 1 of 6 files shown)


lib/src/rules/prefer_const_literals_to_create_immutables.dart
```diff
@@ -78,7 +78,12 @@ class _Visitor extends SimpleAstVisitor<void> {
   void visitListLiteral(ListLiteral node) => _visitTypedLiteral(node);
 
   @override
-  void visitMapLiteral(MapLiteral node) => _visitTypedLiteral(node);
+  void visitSetOrMapLiteral(SetOrMapLiteral node) {
+    // todo (pq): should this apply to set literals as well?
```
bwilkerson: I believe the answer is yes.

lib/src/rules/prefer_const_literals_to_create_immutables.dart
```diff
@@ -113,7 +118,8 @@ class _Visitor extends SimpleAstVisitor<void> {
         (node is ParenthesizedExpression ||
             node is ArgumentList ||
             node is ListLiteral ||
-            node is MapLiteral ||
+            node is SetOrMapLiteral ||
+            // todo (pq): check if this needs updating with the SetOrMapLiteral unification
```
bwilkerson: No. We still use MapLiteralEntry to represent a `key : value` structure.

lib/src/rules/prefer_void_to_null.dart
```diff
@@ -92,7 +92,10 @@ class _Visitor extends SimpleAstVisitor<void> {
         final literal = parent.parent;
         if (literal is ListLiteral && literal.elements.isEmpty) {
           return;
-        } else if (literal is MapLiteral && literal.entries.isEmpty) {
+        } else if (literal is SetOrMapLiteral &&
+            // todo (pq): should this apply to set literals as well?
```
bwilkerson: Again, I think the answer is yes. Tests will confirm.

lib/src/rules/unnecessary_const.dart
```diff
@@ -74,7 +74,12 @@ class _Visitor extends SimpleAstVisitor {
   visitListLiteral(ListLiteral node) => _visitTypedLiteral(node);
 
   @override
-  visitMapLiteral(MapLiteral node) => _visitTypedLiteral(node);
+  void visitSetOrMapLiteral(SetOrMapLiteral node) {
+    // todo (pq): should this apply to SetLiterals as well?
```
bwilkerson: Yes.

lib/src/rules/prefer_const_literals_to_create_immutables.dart
pq: Thanks!

pq:
Thanks for the quick feedback @bwilkerson!

PR #1464, opened 2019-03-05
add a pana baseline bot

### LINKED ISSUE #1463: consider a bot that tracks a pana score baseline
As of https://github.com/dart-lang/pana/pull/514 and pana  `0.12.14` we can get score info with the `--scores` flag.   It could be useful to have a "warning bot" that tracks a baseline so we can see when a proposed change would have a negative score impact.

pq:
The output is a little spammy but serviceable as a first cut.

Essentially, this:

* adds a new `pana_baseline` bot, that
* checks current pana scores against a
* baseline stored in `'tool/baseline/pana.json'`

image

Sample run [here](https://travis-ci.org/dart-lang/linter/jobs/502145908).

Fixes #1463.

/cc @devoncarew @bwilkerson 

/fyi @isoos @jonasfj

### DIFF
--- a/tool/pana_baseline.dart
+++ b/tool/pana_baseline.dart
@@ -8,17 +8,17 @@ import 'dart:io';
 const baseLinePath = 'tool/baseline/pana.json';
 
 main() async {
-  print('reading baseline...');
+  print('Reading baseline...');
   var contents = new File(baseLinePath).readAsStringSync();
   var baseline = jsonDecode(contents)['scores'];
   print(baseline);
 
-  print('installing pana...');
+  print('Installing pana...');
   var activate = await Process.run('pub', ['global', 'activate', 'pana']);
   expectOk(activate);
   print(activate.stdout);
 
-  print('running pana...');
+  print('Running pana...');
   var output = await Process.run('pub', [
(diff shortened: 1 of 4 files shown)


tool/pana_baseline.dart
```diff
@@ -0,0 +1,74 @@
...
+  }
+  if (failureReport.isNotEmpty) {
+    print('Baseline check failed: $failureReport');
+    exit(13);
+  }
+  print('Baseline check passed ✅');
+
+  if (currentHealth != baselineHealth ||
+      currentMaintenance != baselineMaintenance) {
+    print('... you have a new baseline! 🎉 Consider updating $baseLinePath');
```
devoncarew: If the score is lower, consider exiting with a non-zero code?

pq:
Updated to fail on new baselines.  👍

tool/pana_baseline.dart
devoncarew: > was to limit failures to potentially bad events

That totally makes sense to me - to only fail if things regress below the baseline.

pq: 👍

PR #1467, opened 2019-03-06
Additional migrations to prepare for the "UI as code" feature.

stereotype441:
See https://dart-review.googlesource.com/c/sdk/+/95665/ for the
complete set of analyzer APIs that are being deprecated.

### DIFF
--- a/lib/src/rules/prefer_foreach.dart
+++ b/lib/src/rules/prefer_foreach.dart
@@ -131,8 +131,8 @@ class _Visitor extends SimpleAstVisitor {
   visitForStatement2(ForStatement2 node) {
     final loopParts = node.forLoopParts;
     if (loopParts is ForEachParts) {
-    final visitor = new _PreferForEachVisitor(rule);
-    node.accept(visitor);
-  }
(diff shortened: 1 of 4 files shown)


lib/src/rules/prefer_foreach.dart
```diff
@@ -128,8 +128,11 @@ class _Visitor extends SimpleAstVisitor {
   _Visitor(this.rule);
 
   @override
-  visitForEachStatement(ForEachStatement node) {
+  visitForStatement2(ForStatement2 node) {
+    final loopParts = node.forLoopParts;
+    if (loopParts is ForEachParts) {
     final visitor = new _PreferForEachVisitor(rule);
     node.accept(visitor);
```
bwilkerson: Maybe it's the display on GitHub, but it looks like the body isn't indented.

pq: Nope.  Good eye.  Caught also by travis:

https://travis-ci.org/dart-lang/linter/jobs/502837949#L1030

stereotype441: Fixed, thanks!

PR #1474, opened 2019-03-07
Fix #1473 by looking at constructor field initializers

### LINKED ISSUE #1473: unnecessary_parenthesis false positive on arrow function on RHS of null-aware in constructor initializers
This is super similar to https://github.com/dart-lang/linter/issues/1395, but apparently different because there is a test case checked in for https://github.com/dart-lang/linter/issues/1395.

Here's a test case:

```dart
class UnnecessaryParenthesis {
  final value;
  UnnecessaryParenthesis.named(Function value)                                                                                           
      : this.value = (value ?? (_) => 42); // OK
}
```

The linter currently reports a lint on that line. But removing the parens formats to:

```dart
class UnnecessaryParenthesis {
  final value;
  UnnecessaryParenthesis.named(Function value) : this.value = value ?? (_) =>                                                            
    42; // OK                                                                                                                          
}
```

Which highlights that the analyzer interprets the arrow to belong to the constructor, not the function literal.

srawlins:
Fixes #1473. This is a rather broad fix 🙁 For example the unnecessary_parenthesis rule would no longer complain about:

```dart
class C {
  C(int a) : a = (1 + 2);
}
```

in which the parens are definitely unnecessary and kind of silly. But taking a look at our two real world examples:

```dart
 class UnnecessaryParenthesis {
  final value;
  // https://github.com/dart-lang/linter/issues/1395
  UnnecessaryParenthesis() : value = (ClassWithFunction()..f = () => 42); // OK
  // https://github.com/dart-lang/linter/issues/1473
  UnnecessaryParenthesis.named(Function value)
      : this.value = (value ?? (_) => 42); // OK
}
```

The offending  arrow which leaps downwards (I think) in precedence is not easy to find... The cascade there is a hint that users could totally put that thing deep.

### DIFF
--- a/lib/src/rules/unnecessary_parenthesis.dart
+++ b/lib/src/rules/unnecessary_parenthesis.dart
@@ -108,7 +108,8 @@ class _Visitor extends SimpleAstVisitor<void> {
   }
 
   bool _containsFunctionExpression(ParenthesizedExpression node) {
-    final containsFunctionExpressionVisitor = _ContainsFunctionExpressionVisitor();
+    final containsFunctionExpressionVisitor =
+        _ContainsFunctionExpressionVisitor();
     node.accept(containsFunctionExpressionVisitor);
     return containsFunctionExpressionVisitor.hasFunctionExpression;
   }
--- a/lib/src/rules/unnecessary_parenthesis.dart
+++ b/lib/src/rules/unnecessary_parenthesis.dart
@@ -68,9 +68,10 @@ class _Visitor extends SimpleAstVisitor<void> {
     }
 
     // Constructor field initializers are rather unguarded by delimiting
-    // tokens, which can get confused with a function literal. See test cases
-    // for issues #1395 and #1473.
-    if (parent is ConstructorFieldInitializer) {
+    // tokens, which can get confused with a function expression. See test
+    // cases for issues #1395 and #1473.
+    if (parent is ConstructorFieldInitializer &&
+        _containsFunctionExpression(node)) {
       return;
     }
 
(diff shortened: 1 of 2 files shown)


lib/src/rules/unnecessary_parenthesis.dart
```diff
@@ -67,6 +67,13 @@ class _Visitor extends SimpleAstVisitor<void> {
       return;
     }
 
+    // Constructor field initializers are rather unguarded by delimiting
+    // tokens, which can get confused with a function literal. See test cases
+    // for issues #1395 and #1473.
+    if (parent is ConstructorFieldInitializer) {
```
bwilkerson: Could we write something like `parent is ConstructorFieldInitializer && containsFunctionExpression(node)`? Seems like we could write a simple visitor that would visit the parenthesized expression's children looking for a closure. That should allow us to fix the bug without hiding valid issues.

srawlins: Wouldn't the visitor have to be very very complete and complex, to find specific nodes arbitrarily deep? I may be looking at it wrong... is there an example of another rule doing this?

bwilkerson: I'm not sure whether there are any other rules doing this, but I think the visitor should be fairly simple. I haven't tested it, but something like the following ought to work:
```
class ClosureFinder extends UnifyingAstVisitor<void> {
  bool hasClosure = false;

  @override
  void visitFunctionExpression(FunctionExpression node) {
    hasClosure = true;
  }

  @override
  void visitNode(AstNode node) {
    if (!hasClosure) {
      node.visitChildren(this);
    }
  }
}
```

srawlins: Ah, very nice! Done.

pq:
Thanks @srawlins!  🍻

PR #1478, opened 2019-03-08
fix description typo

pq:
/cc @bwilkerson

### DIFF
--- a/lib/src/rules/prefer_const_literals_to_create_immutables.dart
+++ b/lib/src/rules/prefer_const_literals_to_create_immutables.dart
@@ -15,7 +15,7 @@ const desc =
 
 const details = '''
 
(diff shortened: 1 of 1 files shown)


lib/src/rules/prefer_const_literals_to_create_immutables.dart
```diff
@@ -15,7 +15,7 @@ const desc =
 
 const details = '''
 
-**PREFER** using `const` for instantiating list and map literal used as
+**PREFER** using `const` for instantiating list and map literals used as
```
bwilkerson: Should this be "list, map and set"?

pq: Yes!  Thanks.

PR #1487, opened 2019-03-15
update `file_names` to skip extension-prefixed .dart files

pq:
Allow (`.css.dart`, `.g.dart`, etc.).

Fixes: #1486

/cc @kevmoo @bwilkerson @matanlurey @davidmorgan

### DIFF
--- a/lib/src/rules/file_names.dart
+++ b/lib/src/rules/file_names.dart
@@ -30,6 +30,13 @@ symbolic imports.
 * `filesystem.dart`
 * `file-system.dart`
 
+Files without a strict `.dart` extension are ignored.  For example:
+
+**OK:**
+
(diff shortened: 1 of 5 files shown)


kevmoo:
Shouldn't be be a bit more strict here?

In Google, we don't have case issues. Just `-`, right? Also, the `-` is only in the extension bit, right?

Wouldn't it be good to be conservative?

matanlurey:
It does not make any sense to spend more time trying to lint the names of generated files.

kevmoo:
Dashes aren't that big of deal. But if a `.dart` file (with any extension)
with wonky casing gets into pub, it can really break things.

Generated files are sometimes committed to pub. I'd push to fix this!

PR #1489, opened 2019-03-15
shorten package description

pq:
On the advice of `pana`:

image

Details are in the README anyway...

Word-smithing welcome.

/cc @bwilkerson @kwalrath

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -4,7 +4,8 @@ version: 0.1.82
 author: Dart Team <misc@dartlang.org>
 
 description: >-
-  A linter for idiomatic Dart, catching common errors at compile time.
+  The implementation of the lint rules supported by the analyzer framework.
+  This package is not intended to be used directly.
 homepage: https://github.com/dart-lang/linter
 documentation: https://dart-lang.github.io/linter/lints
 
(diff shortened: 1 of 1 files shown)


pubspec.yaml
```diff
@@ -4,11 +4,7 @@ version: 0.1.82
 author: Dart Team <misc@dartlang.org>
 
 description: >-
-  A linter for Dart, bundled with the SDK, that helps enforce coding style,
-  flags potentially difficult to maintain, error-prone or poorly performing
-  source and generally nudges you towards writing more idiomatic Dart.
-  This package is vended into the SDK, where it is plugged into the analyzer
-  framework; it is not intended to be used directly.
+  A linter for idiomatic Dart, catching common errors at compile time.
```
bwilkerson: Even though it already said the same, my biggest concern with this is that it sounds like the linter is a stand-alone tool. I'd prefer to make it clear that it isn't. How about something more like:

The implementation of the lint rules supported by the analyzer framework. This package is not intended to be used directly.

pq:
@bwilkerson: I went with your advice.  Take another look?

Thanks!

pq:
Thank you both!

PR #1497, opened 2019-04-03
migrated unit tests; todo cleanups

pq:
/cc @a14n @bwilkerson


pq:
>  I have reservations about deleting tests if they haven't been moved 

Agreed! 👍 

In this case, the tests were moved and not deleted.

image

image

The TODO that suggested we needed more positive and negative cases was stale so I just removed it.

pq:
The deleted code in the integration tests is made unnecessary when run in the context of the rule unit test runner...

PR #1505, opened 2019-04-05
ui as code if element lints

pq:
Introduces:

* `prefer_if_elements_to_conditional_expressions`
* `prefer_if_elements_to_map_literals`

Q: is `if_elements` the right descriptor?

/cc @bwilkerson

### DIFF
--- a/lib/src/rules/prefer_if_elements_to_conditional_expressions.dart
+++ b/lib/src/rules/prefer_if_elements_to_conditional_expressions.dart
@@ -38,12 +38,6 @@ Widget build(BuildContext context) {
 }
 ''';
 
-
-// spread_collections => prefer_spread_collections (?)
-
-// prefer_if_elements_to_map_literals
-// prefer_if_elements_to_conditional_expressions
-
 class PreferIfElementsToConditionalExpressions extends LintRule implements NodeLintRule {
   PreferIfElementsToConditionalExpressions()
       : super(
--- a/lib/src/rules/prefer_if_elements_to_conditional_expressions.dart
+++ b/lib/src/rules/prefer_if_elements_to_conditional_expressions.dart
@@ -10,7 +10,7 @@ const _desc = r'Prefer if elements to conditional expressions where possible.';
 
 const _details = r'''
 When building collections, it is preferable to use `if` elements rather than
-conditionals and repetition.
+conditionals.
 
 **BAD:**
 ```
--- a/lib/src/rules/prefer_if_elements_to_conditional_expressions.dart
+++ b/lib/src/rules/prefer_if_elements_to_conditional_expressions.dart
@@ -1,0 +1,80 @@
+// Copyright (c) 2019, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:linter/src/analyzer.dart';
+
+const _desc = r'Prefer if elements to conditional expressions where possible.';
+
+const _details = r'''
+When building collections, it is preferable to use `if` elements rather than
+conditionals and repetition.
+
+**BAD:**
+```
+Widget build(BuildContext context) {
+  return Row(
+    children: [
+      IconButton(icon: Icon(Icons.menu)),
+      Expanded(child: title),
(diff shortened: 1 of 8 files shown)


lib/src/rules/prefer_if_elements_to_map_literals.dart
```diff
@@ -0,0 +1,119 @@
+// Copyright (c) 2019, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/element.dart';
+import 'package:linter/src/analyzer.dart';
+
+const _desc = r'Prefer if elements to map literals where possible.';
```
bwilkerson: "map literals"? The lint flags uses of the constructor `Map.fromIterable`.

lib/src/rules/prefer_if_elements_to_map_literals.dart
```diff
@@ -0,0 +1,119 @@
...
+
+const _details = r'''
+When building maps from iterables, it is preferable to use `if` and `for` 
+elements rather than conditionals and repetition.
```
bwilkerson: The assist doesn't use if elements, so I think we can drop that part. (That confusion was probably my fault. Looking at the assist I see I wrote "if" where I should have written "for".) Also, I don't know what "rather than conditionals and repetition" means in this context. How about something like:

When building maps from iterables, it is preferable to use `for` elements.

lib/src/rules/prefer_if_elements_to_map_literals.dart
```diff
@@ -0,0 +1,119 @@
...
+  for (var demo in kAllGalleryDemos)
+    '${demo.routeName}': demo.buildRoute,
+};
+```
+''';
+
+class PreferIfElementsToMapLiterals extends LintRule implements NodeLintRule {
+  PreferIfElementsToMapLiterals()
+      : super(
+            name: 'prefer_if_elements_to_map_literals',
```
bwilkerson: Maybe `prefer_if_elements_to_map_fromIterable`?

test/rules/prefer_if_elements_to_map_literals.dart
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) 2019, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// ignore_for_file: prefer_expression_function_bodies
+// ignore_for_file: avoid_positional_boolean_parameters
```
bwilkerson: Do we need this ignore? I don't see any boolean parameters.

lib/src/rules/prefer_if_elements_to_map_literals.dart
pq: Perfect.  Thanks!

lib/src/rules/prefer_if_elements_to_map_literals.dart
pq: Absolutely; sloppy copy-paste. 😬

lib/src/rules/prefer_if_elements_to_map_literals.dart
pq: Yes!  Except maybe `prefer_for_elements_to_map_fromIterable` is better still?

test/rules/prefer_if_elements_to_map_literals.dart
pq: Nope.  You're right.  Another copy-paste goof-up.

lib/src/rules/prefer_if_elements_to_conditional_expressions.dart
```diff
@@ -0,0 +1,75 @@
...
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:linter/src/analyzer.dart';
+
+const _desc = r'Prefer if elements to conditional expressions where possible.';
+
+const _details = r'''
+When building collections, it is preferable to use `if` elements rather than
+conditionals and repetition.
```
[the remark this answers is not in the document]
pq: This was cribbed essentially from the feature spec (https://github.com/dart-lang/language/blob/master/accepted/future-releases/control-flow-collections/feature-specification.md) but I think describes a case we don't lint for in any case so I'll remove.

PR #1521, opened 2019-04-12
fix false positive in prefer_iterable_whereType for is!

### LINKED ISSUE #1520: prefer_iterable_whereType false positive
The code below is flagged by the lint rule prefer_iterable_whereType:
```
var foo = [42].where((num) => num is! int);
```

pq:
Fixes #1520

/cc @bwilkerson


gabrc:
Thanks for the quick fix!

PR #1526, opened 2019-04-16
don’t lint for spreads in const contexts

pq:
/cc @bwilkerson

### DIFF
--- a/lib/src/rules/prefer_spread_collections.dart
+++ b/lib/src/rules/prefer_spread_collections.dart
@@ -4,6 +4,7 @@
 
 import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/src/dart/ast/ast.dart'; // ignore: implementation_imports
 import 'package:linter/src/analyzer.dart';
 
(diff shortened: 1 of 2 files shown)


lib/src/rules/prefer_spread_collections.dart
```diff
@@ -99,7 +99,9 @@ class _Visitor extends SimpleAstVisitor<void> {
     CascadeExpression cascade = invocation.thisOrAncestorOfType();
     NodeList<Expression> sections = cascade.cascadeSections;
     Expression target = cascade.target;
-    if (target is! ListLiteral || sections[0] != invocation) {
+    if (target is! ListLiteral ||
+        (target as ListLiteral).isConst ||
```
bwilkerson: `isConst` return `true` if the literal has a `const` modifier. To catch both that and the case where it's nested inside some larger constant, you need to use `inConstantContext` (which is currently only available on the `Impl` classes).

pq: Nice!  I hadn't seen that.  Thank you!

pq:
OK.  This should work more generally now. 👍

PR #1528, opened 2019-04-17
New Linter rule: unsafe_html

### LINKED ISSUE #1495: Lint rule for unsafe API usage, such as ScriptElement.src=
There is internal desire to report on usage of unsafe APIs. One example is setting the src on a ScriptElement ([ScriptElement.src=](https://api.dartlang.org/dev/2.2.1-dev.2.0/dart-html/ScriptElement/src.html)).

The lint rule would therefore report on the following setter calls: `ScriptElement.src=` and `dynamic.src=`.

There is of course a lot of room for false positives there, so this rule is only recommended for _web_ packages.

srawlins:
Fixes #1495.

Questions for @sigmundch:

* [ ] What is the recommended alternative (currently there is no "**GOOD**" example in the description)?
* [x] Do I need to cover reading from `src`, or just writing to it?


lib/src/rules/unsafe_script_src.dart
```diff
@@ -0,0 +1,74 @@
...
+            name: 'unsafe_script_src',
+            description: _desc,
+            details: _details,
+            group: Group.errors);
+
+  @override
+  void registerNodeProcessors(NodeLintRegistry registry,
+      [LinterContext context]) {
+    final visitor = new _Visitor(this);
+    registry.addPropertyAccess(this, visitor);
```
bwilkerson: I don't believe you need this line.

lib/src/rules/unsafe_script_src.dart
```diff
@@ -0,0 +1,74 @@
...
+    } else if (leftPart is PropertyAccess) {
+      _checkAssignment(leftPart.target, leftPart.propertyName, node);
+    } else if (leftPart is PrefixedIdentifier) {
+      _checkAssignment(leftPart.prefix, leftPart.identifier, node);
+    }
+  }
+
+  void _checkAssignment(Expression target, SimpleIdentifier property,
+      AssignmentExpression assignment) {
+    DartType type = target?.staticType;
```
bwilkerson: Not strictly necessary, but you could delay accessing the type until after you've checked the property name.

lib/src/rules/unsafe_script_src.dart
srawlins: Done!

lib/src/rules/unsafe_script_src.dart
srawlins: Done, thanks!

lib/src/rules/unsafe_script_src.dart
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  void _checkAssignment(Expression target, SimpleIdentifier property,
+      AssignmentExpression assignment) {
+    // It is more efficient to first check if `src` is being assigned, _then_
+    // check if the target is ScriptElement or dynamic.
+    if (property?.name == 'src') {
+      DartType type = target?.staticType;
+      if (DartTypeUtilities.extendsClass(
+              type, 'ScriptElement', 'dart.dom.html') ||
```
sigmundch: Up to you whether to do it now or later, but we will also want to check for a few other types.

My understanding is that `script.src`, `embed.src`, `iframe.src`, `frame.src`, `image.src`, and `a.href` are all exposing the same vulnerability.

lib/src/rules/unsafe_script_src.dart
```diff
@@ -0,0 +1,73 @@
...
+          type.isDynamic) {
```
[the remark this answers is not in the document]
srawlins: Done.

lib/src/rules/unsafe_script_src.dart
srawlins: I cannot find a "frame" element. Do you know what that would be? I've added the other classes though.

srawlins:
After discussing offline, there simply aren't any available alternatives. The lint is provided first to help teams spot unsafe HTML API usage, and review individual violations.

This is ready for review.

test/rules/unsafe_html.dart
```diff
@@ -0,0 +1,41 @@
...
+  script?.src = 'foo.js'; // LINT
+
+  C().src = 'foo.js'; // OK
+  C()..src = 'foo.js'; // OK
+  C()?.src = 'foo.js'; // OK
+
+  dynamic d;
+  d.src = 'foo.js'; // LINT
+  d.href = 'foo.js'; // LINT
+  (script as dynamic).src = 'foo.js'; // LINT
```
sigmundch: consider also testing one more line:

```dart
  (C() as dynamic).src = 'foo.js'; // LINT
```

(Since `C` is the only one above that doesn't give a lint already)

lib/src/rules/unsafe_html.dart
```diff
@@ -0,0 +1,86 @@
...
+      _checkAssignment(leftPart.prefix, leftPart.identifier, node);
+    }
+  }
+
+  void _checkAssignment(Expression target, SimpleIdentifier property,
+      AssignmentExpression assignment) {
+    // It is more efficient to first check if `src` (or `href`) is being
+    // assigned, _then_ check if the target of an interesting  type.
+    if (property?.name == 'src') {
+      DartType type = target?.staticType;
```
sigmundch: seems like if `target` is null `type` will be null and you'll get a NSM error on the next line below.

Maybe we can just check whether `property` or `target` are null at the beginning of this method and return quickly? (then we can remove all the `?` here and below)

lib/src/rules/unsafe_html.dart
```diff
@@ -0,0 +1,86 @@
...
+  final LintRule rule;
+
+  _Visitor(this.rule);
+
+  @override
+  visitAssignmentExpression(AssignmentExpression node) {
+    final parent = node.parent;
+    final leftPart = node.leftHandSide.unParenthesized;
+    if (parent is CascadeExpression && leftPart is PropertyAccess) {
+      _checkAssignment(parent.target, leftPart.propertyName, node);
```
bwilkerson: Mot required, but if I understand what you're trying to do here, I think you could combine this and the next block by using `leftPart.realTarget`.

lib/src/rules/unsafe_html.dart
```diff
@@ -0,0 +1,86 @@
...
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/type.dart';
+import 'package:linter/src/analyzer.dart';
+import 'package:linter/src/util/dart_type_utilities.dart';
+
+const _desc = r'Avoid unsafe HTML APIs.';
+
+const _details = r'''
```
kevmoo: This could use some more details, IMHO. What should folks do instead?

srawlins: As I mention above, there simply aren't any available alternatives. The lint is provided first to help teams spot unsafe HTML API usage, and review individual violations.

Should I codify that here?

kevmoo: There **are** safe ways to do this, right? It's really not "don't set src=", it's more "don't set src= without being careful".

So it's kinda a weird lint

PR #1531, opened 2019-04-17
fix issue 1518

### LINKED ISSUE #1518: Issue with lint unnecessary_await_in_return
In the code below, await is necessary.
```
import 'dart:async';

Future<num> foo() async => 1;

// The lint unnecessary_await_in_return reports an issue here, but the await is necessary
// as an implicit downcast is necessary.
// We cannot change the code to `Future<int> bar() => foo()`
Future<int> bar() async => await foo();

void main() async => print(await bar());
```

a14n:
Fix #1518

### DIFF
--- a/lib/src/rules/unnecessary_await_in_return.dart
+++ b/lib/src/rules/unnecessary_await_in_return.dart
@@ -96,7 +96,7 @@ class _Visitor extends SimpleAstVisitor<void> {
     }
     if (returnType != null &&
         returnType.isDartAsyncFuture &&
-        type.isAssignableTo(returnType)) {
+        type.isSubtypeOf(returnType)) {
       rule.reportLintForToken((expression as AwaitExpression).awaitKeyword);
     }
   }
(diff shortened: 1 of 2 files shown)


lib/src/rules/unnecessary_await_in_return.dart
```diff
@@ -96,7 +96,7 @@ class _Visitor extends SimpleAstVisitor<void> {
     }
     if (returnType != null &&
         returnType.isDartAsyncFuture &&
-        type.isAssignableTo(returnType)) {
+        type.isSubtypeOf(returnType)) {
```
pq: Looks good to me but I'll defer to @bwilkerson for the final call since he may be aware of other implications.

bwilkerson: It seems reasonable to me, and the tests appear to show that it's the right test, but I haven't spent any time to try to _prove_ to myself that it's correct. I say, go for it.

pq:
Thanks for taking a look @bwilkerson !

PR #1535, opened 2019-04-22
0.1.87

pq:
# 0.1.87

* fixed `unnecessary_await_in_return` false positive
* fixed `unrelated_type_equality_checks` false negative with functions
* fixed `prefer_spread_collections` to not lint in const contexts
* fixed false positive in `prefer_iterable_whereType` for `is!`
* fixed false positive in `prefer_collection_literals` for constructors with params

/cc @bwilkerson @srawlins


pq:
Waiting to land, pending #1537.

PR #1536, opened 2019-04-22
remove the control flow collections test config

pq:
No longer adding value w/ UI-as-code enabled by default.

/cc @bwilkerson


pq:
Test breakage to be addressed in: #1537.

Issue #1537, opened 2019-04-22
prefer_const_constructors_in_immutables needs updates for 0.36.1


pq:
With the `package:analyzer` v 0.36.1, we're seeing new failures in `prefer_const_constructors_in_immutables` tests.  Specifically, this is no longer linting:

```dart
@immutable
class H {
  final f;
  H(f) : f = f ?? f == null; // LINT
}
```

See: https://travis-ci.org/dart-lang/linter/jobs/523234741

/cc @bwilkerson @a14n

pq:
Marking as a P1, since we'll want to fix this before publishing a new linter (#1535).

pq:
After looking this over w/ @scheglov, it appears that the hack used in `_hasOnlyConstExpressionsInIntializerList` took advantage of accidental behavior in the analyzer and will no longer work --- the rub is that these programmatic changes to the AST are no longer being reflected in the element model.  We discussed a few options.  One involves changing `ConstantVerifier` to allow clients to state that they want an element to be treated as constant.  Maybe there are others?

@a14n: any thoughts?

To move ahead with the release, I'll update the tests to note that these cases are no longer being caught and link back here for updating once this issue is definitively addressed.

a14n:
Note that the same kind of hack is used in `unnecessary_const` lint.

a14n:
>Note that the same kind of hack is used in `unnecessary_const` lint.

And `prefer_const_literals_to_create_immutables` and `prefer_const_constructors` and `prefer_const_constructors_in_immutables`.

pq:
Ah.  Thanks for that added context @a14n.  Without looking, it's interesting that we're not seeing failures there too.  Maybe we're not testing those paths?  Or maybe this functionality is still (accidentally) working in those cases?

Either way, it seems like this is significant enough that we should figure out how to get this information in a more reliable way.  @bwilkerson mentioned that `AstNode` used to have this info.  Maybe we could add that back?  Or provide some other means?

@scheglov: any thoughts?

scheglov:
I will rewrite it without looking into the enclosing `ConstructorElement`.

scheglov:
https://dart-review.googlesource.com/c/sdk/+/100277

I have not tried if this fixed linter.

pq:
Thank you @scheglov!

With that fix, the existing tests pass and this now gets flagged:

```dart
@immutable
class D {
  final _a;
[...]

PR #1539, opened 2019-04-23
analyzer 0.36.1 updates for prefer_const_constructors_in_immutables

pq:
See: #1537.

/cc @bwilkerson @a14n


pq:
Pending fixes in a future analyzer I think we're stuck w/ something like this.

@bwilkerson: PTAL?

PR #1544, opened 2019-04-29
add sort_child_properties_last

### LINKED ISSUE #1542: new lint: sort_child_properties_last
Name TBD but the idea is to nudge people to sorting child/children properties last in their build methods to improve UI-as-Code feature readability.

@jacob314: could you add some more context?  Maybe some concrete exemplars?

pq:
Adds new lint `sort_child_properties_last`.

Currently limited to `child` and `children` properties but after some analysis we could consider broadening.

Fixes #1542

/cc @jacob314 @a14n @bwilkerson

### DIFF
--- a/lib/src/rules/sort_child_properties_last.dart
+++ b/lib/src/rules/sort_child_properties_last.dart
@@ -101,19 +101,23 @@ class _Visitor extends SimpleAstVisitor {
     }
 
     var arguments = node.argumentList.arguments;
-    if (arguments.length < 2) {
+    if (arguments.length < 2 || isChildArg(arguments.last)) {
       return;
     }
 
     for (int i = 0; i < arguments.length - 1; ++i) {
-      var arg = arguments[i];
-      if (arg is NamedExpression) {
-        var name = arg.name.label.name;
-        if ((name == 'child' || name == 'children') &&
-            isWidgetProperty(arg.staticType)) {
-          rule.reportLint(arg);
-        }
+      if (isChildArg(arguments[i])) {
+        rule.reportLint(arguments[i]);
       }
     }
+  }
+
+  static bool isChildArg(Expression e) {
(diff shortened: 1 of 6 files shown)


lib/src/rules/sort_child_properties_last.dart
```diff
@@ -0,0 +1,119 @@
...
+    if (!isWidgetType(node.staticType)) {
+      return;
+    }
+
+    var arguments = node.argumentList.arguments;
+    if (arguments.length < 2) {
+      return;
+    }
+
+    for (int i = 0; i < arguments.length - 1; ++i) {
```
bwilkerson: Not sure whether it matters, but there's a possible optimizations that could be made here: generally the desired argument _is_ last, so if we check for it before iterating we can generally avoid looking at every other argument.

pq: oh, smart!  i'll play with that.  thanks!

pq: updated. 👍

lib/src/rules/sort_child_properties_last.dart
```diff
@@ -0,0 +1,123 @@
...
+    ),
+    floatingActionButton: FloatingActionButton(
+      onPressed: _incrementCounter,
+      tooltip: 'Increment',
+      child: Icon(Icons.add),
+    ),
+  ),
+);
+```
+''';
```
rokob, line 74: These are not the same layouts, you moved mainAxisAlignment from inside the Column to being a property on Center. I couldn't for the life of me figure out how this lint could imply that but I'm assuming it was just a mistake.

rokob: https://github.com/dart-lang/linter/pull/2051

PR #1547, opened 2019-05-03
Add lint prefer_double_quotes

### LINKED ISSUE #1096: Add prefer_double_quotes linter
For those of us using languages with double-quotes (e.g. C#) on the back-end, and using Dart on the front-end, it's quite annoying to have to switch between two quote demarcations. There already exists a `prefer_single_quotes` linter, but there is no counterpart for double quotes. This is a request to add one, which should be fairly straightforward given the existence of the `prefer_single_quotes` linter.

a14n:
Fixes #1096

### DIFF
--- a/lib/src/rules/prefer_single_quotes.dart
+++ b/lib/src/rules/prefer_single_quotes.dart
@@ -5,6 +5,7 @@
 import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
 import 'package:linter/src/analyzer.dart';
+import 'package:meta/meta.dart';
 
 const _desc =
     r"Prefer single quotes where they won't require escape sequences.";
@@ -54,7 +55,7 @@ class PreferSingleQuotes extends LintRule implements NodeLintRule {
   @override
   void registerNodeProcessors(NodeLintRegistry registry,
       [LinterContext context]) {
-    final visitor = new _Visitor(this);
+    final visitor = new QuoteVisitor(this, useSingle: true);
     registry.addSimpleStringLiteral(this, visitor);
     registry.addStringInterpolation(this, visitor);
   }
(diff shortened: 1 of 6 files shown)


pq:
It looks like this is failing the integration tests, since `p5.dart` contains single quotes:

```dart
import 'package:p6/p6_lib.dart';
```

Maybe this is as simple as adding an analysis_options file in `_data/p5`?

lib/src/rules/prefer_single_quotes.dart
```diff
@@ -117,7 +122,8 @@ class _Visitor extends SimpleAstVisitor<void> {
 
   @override
   void visitSimpleStringLiteral(SimpleStringLiteral string) {
-    if (string.isSingleQuoted || string.value.contains("'")) {
+    if (useSingle && (string.isSingleQuoted || string.value.contains("'")) ||
```
pq: Perhaps extract a function that can be used here and below?

kevmoo:
No! Don't let folks go the other way ❗️

kevmoo:
...I'm kinda joking...but also kinda serious. It's great if push the community in one direction. Who is asking for this?

bwilkerson:
See #1096.

a14n:
>It looks like this is failing the integration tests, since p5.dart contains single quotes

I really don't understand why it fails because there's no place where this new `prefer_double_quotes` is used :-/

pq:
Yes. Right.  That's a good point... I'm unsure now but I'm wondering if, in the absence of an options file, the linter CLI isn't defaulting to enabling all the lints? 😬

lib/src/rules/prefer_single_quotes.dart
a14n: The conditions are not exactly the same.

test/integration_test.dart
```diff
@@ -105,8 +105,9 @@ defineTests() {
       group('.packages', () {
         test('basic', () async {
           // Requires .packages to analyze cleanly.
-          await cli
-              .run(['test/_data/p5', '--packages', 'test/_data/p5/_packages']);
+          await cli.runLinter(
+              ['test/_data/p5', '--packages', 'test/_data/p5/_packages'],
+              LinterOptions([]));
```
pq: 👍

PR #1549, opened 2019-05-06
fix curly_braces_in_flow_control_structures

### LINKED ISSUE #1548: curly_braces_in_flow_control_structures isn't catching all cases
In the following code, only the last block complains with `curly_braces_in_flow_control_structures`:

Screenshot 2019-05-06 at 5 20 57 pm

Code is here:

https://github.com/flutter/devtools/blob/9bcab46e323c867664e90ee4954ff57ae507a44b/packages/devtools/lib/src/ui/analytics.dart#L280

Even if I fix the last one, it doesn't warn on the other cases:

Screenshot 2019-05-06 at 5 21 44 pm

a14n:
Fixes #1548

### DIFF
--- a/lib/src/rules/curly_braces_in_flow_control_structures.dart
+++ b/lib/src/rules/curly_braces_in_flow_control_structures.dart
@@ -87,7 +87,9 @@ class _Visitor extends SimpleAstVisitor {
   @override
   void visitIfStatement(IfStatement node) {
     var elseStatement = node.elseStatement;
-    if (elseStatement == null) {
+    var parent = node.parent;
+    if (elseStatement == null &&
+        !(parent is IfStatement && parent.elseStatement == node)) {
       if (node.thenStatement is Block) return;
 
       final unit = node.root as CompilationUnit;
--- a/lib/src/rules/curly_braces_in_flow_control_structures.dart
+++ b/lib/src/rules/curly_braces_in_flow_control_structures.dart
@@ -99,9 +99,7 @@ class _Visitor extends SimpleAstVisitor {
       }
     } else {
       _check(node.thenStatement);
-      if (elseStatement is IfStatement) {
-        _check(elseStatement.thenStatement);
-      } else {
+      if (elseStatement is! IfStatement) {
         _check(elseStatement);
       }
     }
--- a/lib/src/rules/curly_braces_in_flow_control_structures.dart
+++ b/lib/src/rules/curly_braces_in_flow_control_structures.dart
@@ -87,9 +87,7 @@ class _Visitor extends SimpleAstVisitor {
   @override
   void visitIfStatement(IfStatement node) {
     var elseStatement = node.elseStatement;
-    var parent = node.parent;
-    if (elseStatement == null &&
-        !(parent is IfStatement && parent.elseStatement == node)) {
+    if (elseStatement == null) {
       if (node.thenStatement is Block) return;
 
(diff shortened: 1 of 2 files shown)


lib/src/rules/curly_braces_in_flow_control_structures.dart
```diff
@@ -86,17 +86,22 @@ class _Visitor extends SimpleAstVisitor {
...
       if (unit.lineInfo.getLocation(node.rightParenthesis.end).lineNumber !=
           unit.lineInfo.getLocation(node.thenStatement.end).lineNumber) {
         rule.reportLint(node.thenStatement);
       }
-    } else if (node.elseStatement is! IfStatement) {
+    } else {
       _check(node.thenStatement);
-      _check(node.elseStatement);
+      if (elseStatement is IfStatement) {
+        _check(elseStatement.thenStatement);
```
bwilkerson: If `elseStatement` is an if statement, then I believe we will end up checking the then statement twice: once here and once when `visitIfStatement` is invoked with `elseStatement` as the argument. I think this just wants to be
```
if (elseStatement is! IfStatement) {
  _check(elseStatement);
}
```

a14n: yes the elseStatement is checked twice but it's not the same check. Actually the `if (elseStatement == null)` should also check if `node.parent` is a IfStatement to avoid this case.

bwilkerson: It wasn't the else statement I was concerned about. Consider the following code:
```dart
if (condA)
  stmtA;
else if (condB)
  stmtB;
else
  stmtC;
```
When this visit method is visiting the if whose condition is `condA` the else statement will not be null, so it will invoke `_check` on `stmtA` (line 101). Because the else statement is another if statement (the one whose condition is `condB`), it will then check that if statement's then statement, which is `stmtB`.

Later, when this visit method is visiting the if statement whose condition is `condB`, it will invoke `_check` on `stmtB` (line 101).

a14n: You're right. Thanks. Updated.

lib/src/rules/curly_braces_in_flow_control_structures.dart
```diff
@@ -86,17 +86,22 @@ class _Visitor extends SimpleAstVisitor {
 
   @override
   void visitIfStatement(IfStatement node) {
-    if (node.elseStatement == null) {
+    var elseStatement = node.elseStatement;
+    var parent = node.parent;
+    if (elseStatement == null &&
+        !(parent is IfStatement && parent.elseStatement == node)) {
```
bwilkerson: I think this condition can (and maybe should) be removed. I believe it was attempting to guard against a case that can no longer occur.

bwilkerson:
@a14n @pq Does this impact Flutter? If so, we might want to run this over Flutter and fix any new lints before we commit it so that it doesn't block us from rolling a new version of the linter into the sdk. It might also have implications for rolling internally.

pq:
@bwilkerson: I was going to ask the same thing.  I won't have time to do this today (or maybe tomorrow) but happy to look later this week, unless you can in the meantime @a14n?

As always, thanks for your contributions!  💎

a14n:
`curly_braces_in_flow_control_structures` is not currently activated. It could change with flutter/flutter#31448

PR #1551, opened 2019-05-07
update to new linter context API

pq:
see: #1537

/cc @scheglov @a14n

### DIFF
--- a/lib/src/rules/prefer_const_constructors_in_immutables.dart
+++ b/lib/src/rules/prefer_const_constructors_in_immutables.dart
@@ -3,12 +3,10 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'package:analyzer/dart/ast/ast.dart';
-import 'package:analyzer/dart/ast/token.dart';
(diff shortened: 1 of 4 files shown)


lib/src/rules/prefer_const_constructors_in_immutables.dart
```diff
@@ -130,18 +130,6 @@ class _Visitor extends SimpleAstVisitor<void> {
...
-    try {
-      hasConstError = hasErrorWithConstantVerifier(context, node);
-    } finally {
-      // restore const keyword
-      node.constKeyword = null;
-    }
-
-    return !hasConstError;
-  }
+  bool _hasOnlyConstExpressionsInIntializerList(ConstructorDeclaration node) =>
```
scheglov: I would inline this method - it does not have much functionality.

pq: Aha, yeah.  No reason to have it anymore...  Thanks!

PR #1552, opened 2019-05-07
pedantic 1.6.0 fixes [TBR]

pq:
/cc @bwilkerson 

fyi @devoncarew (who predicted as much...)


pq:
TBR'ing to unblock #1551.

PR #1553, opened 2019-05-07
de-dup lints defined in pedantic

pq:
/cc @bwilkerson


a14n:
BTW It would be great if analyzer warns on those duplications.

pq:
> BTW It would be great if analyzer warns on those duplications.

I've been thinking about this too.  Part of me thinks it should be a lint though so that it can be opted out of. 

@bwilkerson: WDYT?

bwilkerson:
I think it depends on whether users will want to have duplications between their options file and an included file. I haven't thought of a compelling use case for doing so, but that doesn't mean there isn't one.

PR #1563, opened 2019-05-14
Fix #1533 with new exceptions

### LINKED ISSUE #1533: False positive for unnecessary_parenthesis
**Describe the issue**
A false positive for unnecessary_parenthesis

**To Reproduce**

`(String).toString();`
`(AnyType).toString();`

With parentheses the toString() converts the Type to a String and does not call toString() on the type which is the case if the parentheses were to be removed.

srawlins:


bwilkerson:
I have some concerns about how specific these tests are, and wonder whether we couldn't have made them more general, but other than that LGTM.

PR #1564, opened 2019-05-14
Fix #1456 by adding another exception

### LINKED ISSUE #1456: prefer_contains false positive
_From @sma on February 26, 2019 20:31_

This code snippet triggers a false _"Use contains instead of indexOf"_ analyzer hint as I'm using `next` inside the while loop and therefore `contains` is no option.

```dart
while ((next = text.indexOf('{', last)) != -1) {
  ...
}
```

Dart version: `Dart VM version: 2.2.0 (Tue Feb 26 15:04:32 2019 +0100) on "macos_x64"`

_Copied from original issue: dart-lang/sdk#36042_

srawlins:


pq:
@bwilkerson: any reservations?

PR #1565, opened 2019-05-14
widen application of prefer_null_aware_operators

a14n:
Remove restriction on `prefer_null_aware_operators` (was only applied for local variables).

### DIFF
--- a/lib/src/rules/prefer_null_aware_operators.dart
+++ b/lib/src/rules/prefer_null_aware_operators.dart
@@ -5,7 +5,6 @@
 import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/token.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
-import 'package:analyzer/dart/element/element.dart';
 import 'package:linter/src/analyzer.dart';
 
 const _desc = r'Prefer using null aware operators.';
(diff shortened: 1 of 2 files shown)


a14n:
Tested on Flutter I haven't seen false positives.

bwilkerson:
Sorry for the delay. Most of the Dart team is at an off-site this week.

Interested to know how many cases it found, and the nature of those cases (for example, what percentage of the cases were references to parameters, for example).

I have no concerns with widening the lint to flag parameters, but, while the probability of finding code like this is small, any code that includes a getter or a method invocation could potentially have side effects such that the suggested transformation would change the semantics. I'll wait for @pq to weigh in on whether it's worth the risk of false positives.

Also, does our quick fix / assist for this case need to be updated?

a14n:
>Interested to know how many cases it found, and the nature of those cases (for example, what percentage of the cases were references to parameters, for example).

There are only 3 lints for this update. You can find them in flutter/flutter#32711

pq:
@a14n: reading https://github.com/flutter/flutter/pull/32711#issuecomment-492930932, are you dropping this?

a14n:
Yes and no :smile:

This linter PR is still valid and should be merged (actually https://github.com/dart-lang/linter/pull/1567#issuecomment-492650767 and https://github.com/dart-lang/linter/pull/1567#issuecomment-492825814 should have been done on this PR).

The rollback I made on flutter is to avoid changing the current behaviour. I will revisit flutter with this lint once NNBD has landed. That said with this lint PR the next linter roll in flutter will need to comment out `prefer_null_aware_operators` in analysis_options.yaml.

pq:
Ah.  Thanks for the clarification!

>> That said with this lint PR the next linter roll in flutter will need to comment out prefer_null_aware_operators in analysis_options.yaml.

Would you mind updating Flutter before this lands?  I'm hoping to publish and integrate as early as tomorrow and this would break the `flutter-analyze-try` bot blocking a land in the SDK.

a14n:
I updated flutter/flutter#32711 to disable `prefer_null_aware_operators`

a14n:
@pq flutter/flutter#32711 is now merged

pq:
Super.  Thanks for the prompt follow-through!

PR #1567, opened 2019-05-14
add lint prefer_if_null_operators

a14n:


lib/src/rules/prefer_if_null_operators.dart
```diff
@@ -0,0 +1,75 @@
+// Copyright (c) 2019, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/token.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/element.dart';
```
pq: I think this is unused?

Removing it should fix the travis break:

https://travis-ci.org/dart-lang/linter/jobs/532359228

a14n: Oops, thanks

bwilkerson:
I had to search our existing lints because I had trouble believing that we hadn't already implemented this rule.

The only concern I have here is the possibility that the expression being de-duped might have side-effects, causing false positives.

a14n:
>I had to search our existing lints because I had trouble believing that we hadn't already implemented this rule.

Same here :)

I use this new lint in flutter/flutter#32711.

bwilkerson:
Dan's comments make me wonder whether this actually brings value to users. Personally, I'd like to see how that discussion plays out before landing this change.

a14n:
Actually Dan's comments are applicable to the current lint behaviour. `a != null ? a.b.c : null` will be linted today if `a` is a local variable. The fix will be to write `a?.b?.c`. So Dan's comment already applied.

PR #1569, opened 2019-05-17
New package descriptions lint rule

### LINKED ISSUE #1450: lint that flags package descriptions not between 60 and 180 characters
From the linter pana report:

image

srawlins:
Fixes #1450 .

I'm not sure if my description of the rule is great. Mostly I'm not sure why < 60 characters gets punished.

### DIFF
--- a/lib/src/rules/pub/package_descriptions.dart
+++ b/lib/src/rules/pub/package_descriptions.dart
@@ -1,0 +1,43 @@
+// Copyright (c) 2015, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:linter/src/analyzer.dart';
+import 'package:linter/src/utils.dart';
+
+const _desc = r'Use brief package descriptions.';
+
+const _details = r'''
+
+**DO** keep package descriptions between 60 and 180 characters long.
+
+Search engines display only the first part of the description. Try to keep the
+value of the `description` field in your package's `pubspec.yaml` file between
+60 and 180 characters.
+''';
+
+class PubPackageDescriptions extends LintRule {
+  PubPackageDescriptions()
+      : super(
+            name: 'package_descriptions',
+            description: _desc,
+            details: _details,
+            group: Group.pub);
+
+  @override
+  PubspecVisitor getPubspecVisitor() => new Visitor(this);
+}
+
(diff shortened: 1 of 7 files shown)


lib/src/rules/pub/package_descriptions.dart
```diff
@@ -0,0 +1,43 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:linter/src/analyzer.dart';
+import 'package:linter/src/utils.dart';
+
+const _desc = r'Use brief package descriptions.';
+
+const _details = r'''
+
+**DO** keep package descriptions between 60 and 180 characters long.
```
bwilkerson: I think it would be good to include the fact that this rule exists because of a policy defined by `pana` (though it should be described in terms of `pub.dev`, as most users won't know what `pana` is).

Have we coordinated with the maintainers of pana so that they know that this lint needs to be updated if they change their policy?

I agree that the 60 character limit is odd; perhaps you could create an issue in `pana` asking the question there.

lib/src/rules/pub/package_descriptions.dart
```diff
@@ -0,0 +1,43 @@
+// Copyright (c) 2019, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:linter/src/analyzer.dart';
+import 'package:linter/src/utils.dart';
```
pq: Unused import?

FWIW: our pana baseline bot flags this (though it's noisy!):

https://travis-ci.org/dart-lang/linter/jobs/533694054

pq:
> I think it would be good to include the fact that this rule exists because of a policy defined by pana (though it should be described in terms of pub.dev, as most users won't know what pana is).

+1.

The best I could find is this doc here:

https://pub.dev/help#scoring

but I don't think description length is rationalized there -- but maybe should?

And actually, looking at https://dart.dev/tools/pub/pubspec#description there's sort of contradicting advice:

> This is optional for your own personal packages, but if you intend to publish your package you must provide a description. This should be relatively short—a few sentences, maybe a whole paragraph—and tells a casual reader what they might want to know about your package.

A whole paragraph is almost certain to be more than 60 chars...


fyi @jonasfj

srawlins:
Nvm, this requires more than a simple contribution.

PR #1582, opened 2019-05-22
Stop using deprecated analyzer API left over from "UI as code" migration.

stereotype441:
Since analyzer version 0.36.0, we no longer need to override
visitForStatement2.  visitForStatement is equivalent.

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -1,5 +1,5 @@
 name: linter
-version: 0.1.91-dev
+version: 0.1.90
 
 author: Dart Team <misc@dartlang.org>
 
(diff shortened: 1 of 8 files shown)


pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: linter
-version: 0.1.90
+version: 0.1.91-dev
```
pq: Could you drop this back to `0.1.90`?  I'll bump before release and this will fix the broken bots.

stereotype441: > Could you drop this back to `0.1.90`? I'll bump before release and this will fix the broken bots.

Done.  PTAL

pq:
Thanks @stereotype441 !

Issue #1587, opened 2019-05-28
Avoid prefer_initializing_formals when parameter is named and the name is different from the field


dramos07:
This example here is outside of the "when possible". If the parameter is named and the name is different from the field, then an initializing formal would force you to change the API of the class. In that case, you shouldn't use it.

```dart
class Foo {
  final bool isDisabled;
  Foo({bool disabled}) : this.isDisabled = disabled;
}
var foo = Foo(disabled: true);
```

The only 2 current possible fixes for prefer_initializing_formals:

### Solutions 1: Change the field to matched the named argument
```dart
class Foo {
  final bool disabled; // Goes against https://dart.dev/guides/language/effective-dart/design#prefer-a-non-imperative-verb-phrase-for-a-boolean-property-or-variable
  Foo({this.disabled});
}
var foo = Foo(disabled: true);
```

### Solutions 2: Change the named argument to match the field
```dart
class Foo {
  final bool isDisabled;
  Foo({this.isDisabled});
}
var foo = Foo(isDisabled: true); // Goes against https://dart.dev/guides/language/effective-dart/design#consider-omitting-the-verb-for-a-named-boolean-parameter
```

Either option contradicts the guidelines, but more importantly either would force you to change the API of the class.

pq:
Thanks for the write-up!  I'm on the fence about whether the lint should not fire in this case.  What if a team opts in to this lint but does not subscribing entirely to the naming guidelines?  Another nuance is:

> either would force you to change the API of the class.

which is a great point.  Unfortunately we have a pretty under-developed notion of what constitutes an API proper and so it would be tricky to know when this would be an issue or not.

In practice, I guess I would think this might be a case where you'd want to suppress this warning specifically in this instance but not sure it would make sense to change the logic in general since knowing whether a change would be breaking and whether the naming conventions are being strictly followed requires situational awareness.

Having said all this, I'm still mulling it all over...  Thanks for the compelling example!

/cc @bwilkerson for thoughts.

bwilkerson:
> ... what constitutes an API ...

I don't really want to debate what an API is right now, but I think it's fairly clear that both changes are breaking changes.

Given that, I'm strongly leaning toward agreeing that the rule should not create a lint in this case.

pq:
> I don't really want to debate what an API is right now,

Ok, but **_I_** can't resist. 😄 

[...]

PR #1590, opened 2019-05-30
Avoid prefer_initializing_formals when parameter is named and the name is different from the field

dramos07:
If the argument is a named argument only lint if the name is the same as to avoid changes to the API to either fix the field or the name arguments for more details see the issue provided below.

Fixes # http://go/gh/dart-lang/linter/issues/1587

### DIFF
--- a/lib/src/rules/prefer_initializing_formals.dart
+++ b/lib/src/rules/prefer_initializing_formals.dart
@@ -35,6 +35,41 @@ class Point {
 }
 ```
 
+**BAD:**
+```
+class Point {
+  num x, y;
+  Point({num x, num y}) {
+    this.x = x;
+    this.y = y;
+  }
+}
+```
+
+**GOOD:**
+```
+class Point {
+  num x, y;
+  Point({this.x, this.y});
+}
+```
+
+**NOTE**
+Named parameters must match with the field name in order to be consider by the
+lint to avoid having to update either the field or the named argument and to
+allow for the API to accomodate for different guidelines for named arguments
+and fields currently proposed in https://dart.dev/guides/language/effective-dart
+thus the following example won't trigger the lint:
+
+```
+class Point {
+  bool isEnabled;
+  Point({bool enabled}) {
+    this.isEnabled = enable; // OK
+  }
+}
+```
+
 ''';
 
 Iterable<AssignmentExpression> _getAssignmentExpressionsInConstructorBody(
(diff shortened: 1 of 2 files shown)


pq:
I voiced some reservations in #1587 but don't feel super strongly.  One thing I would ask though is perhaps update the docs to explain how this special case is supported?  (Maybe with an example?)

Thanks @dramos07 for the the thoughtful conversation! 👍

dramos07:
Thanks @pq I have update the docs let me know what you think!

lib/src/rules/prefer_initializing_formals.dart
```diff
@@ -35,6 +35,41 @@ class Point {
...
+**GOOD:**
+```
+class Point {
+  num x, y;
+  Point({this.x, this.y});
+}
+```
+
+**NOTE**
+Named parameters must match with the field name in order to be consider by the
```
pq: => considered

lib/src/rules/prefer_initializing_formals.dart
```diff
@@ -35,6 +35,41 @@ class Point {
...
+lint to avoid having to update either the field or the named argument and to
+allow for the API to accomodate for different guidelines for named arguments
```
pq: => accommodate

lib/src/rules/prefer_initializing_formals.dart
```diff
@@ -35,6 +35,41 @@ class Point {
...
+and fields currently proposed in https://dart.dev/guides/language/effective-dart
+thus the following example won't trigger the lint:
+
```
pq: /cc @bwilkerson for word-smithing ideas.

(After 3 tries, it's time for a life-line!)

bwilkerson: The biggest improvement would be to split this up into multiple sentences. Consider something like the following:

This rule will not generate a lint for named parameters unless the parameter name and the field name are the same. The reason for this is that resolving such a lint would require either renaming the field or renaming the parameter, and both of those actions would potentially be a breaking change. For example, the following will not generate a lint:

pq: Excellent.  Thanks @bwilkerson!

@dramos07: does that sound OK to you?

dramos07: That was definitely an improvement thanks @bwilkerson and @pq

PR #1593, opened 2019-05-31
Loosen `avoid_function_literals_in_foreach_calls` to allow method chained calls

### LINKED ISSUE #365: loosen `prefer_each` to allow chained calls
From [Effective Dart](https://www.dartlang.org/guides/language/effective-dart/usage#avoid-using-iterableforeach-with-a-function-literal).

`forEach()` functions are widely used in JavaScript because the built in for-in loop doesn’t do what you usually want. In Dart, if you want to iterate over a sequence, the idiomatic way to do that is using a loop.

dramos07:
Loosens the lint to allow chained calls such as:
```dart
people
      .where((person) => person != null)
      .map((person) => '$person!')
      .forEach((person) => print(person)); // OK

Person()
    .children
    .firstWhere((person) => person != null)
    .children
    .forEach((person) => print(person)); // OK

Person().children.first.children.forEach((person) => print(person)); // OK
```
Fixes #365 
@pq @munificent

### DIFF
--- a/lib/src/rules/avoid_function_literals_in_foreach_calls.dart
+++ b/lib/src/rules/avoid_function_literals_in_foreach_calls.dart
@@ -54,32 +54,30 @@ class _Visitor extends SimpleAstVisitor<void> {
 
   @override
   void visitMethodInvocation(MethodInvocation node) {
-    var exp = node.target;
-    var chainCount = 0;
-    while (exp is PrefixedIdentifier ||
-        exp is MethodInvocation ||
-        exp is PropertyAccess) {
-      if (exp is PrefixedIdentifier) {
-        chainCount++;
-        exp = (exp as PrefixedIdentifier).prefix;
-      } else if (exp is MethodInvocation) {
-        chainCount++;
-        exp = (exp as MethodInvocation).target;
-      } else if (exp is PropertyAccess) {
-        chainCount++;
-        exp = (exp as PropertyAccess).target;
-      }
-      if (chainCount > 1) {
-        return;
-      }
-    }
     if (node.target != null &&
         node.methodName.token.value() == 'forEach' &&
         node.argumentList.arguments.isNotEmpty &&
         node.argumentList.arguments[0] is FunctionExpression &&
         DartTypeUtilities.implementsInterface(
-            node.target.staticType, 'Iterable', 'dart.core')) {
+            node.target.staticType, 'Iterable', 'dart.core') &&
+        !_hasMethodChaining(node)) {
       rule.reportLint(node.function);
     }
   }
+}
+
+bool _hasMethodChaining(MethodInvocation node) {
+  var exp = node.target;
+  while (exp is PrefixedIdentifier ||
+      exp is MethodInvocation ||
+      exp is PropertyAccess) {
+    if (exp is PrefixedIdentifier) {
+      exp = (exp as PrefixedIdentifier).prefix;
+    } else if (exp is MethodInvocation) {
+      return true;
+    } else if (exp is PropertyAccess) {
+      exp = (exp as PropertyAccess).target;
+    }
+  }
+  return false;
 }
--- a/lib/src/rules/avoid_function_literals_in_foreach_calls.dart
+++ b/lib/src/rules/avoid_function_literals_in_foreach_calls.dart
@@ -54,6 +54,25 @@ class _Visitor extends SimpleAstVisitor<void> {
 
   @override
   void visitMethodInvocation(MethodInvocation node) {
+    var exp = node.target;
+    var chainCount = 0;
+    while (exp is PrefixedIdentifier ||
+        exp is MethodInvocation ||
+        exp is PropertyAccess) {
+      if (exp is PrefixedIdentifier) {
+        chainCount++;
+        exp = (exp as PrefixedIdentifier).prefix;
+      } else if (exp is MethodInvocation) {
+        chainCount++;
+        exp = (exp as MethodInvocation).target;
+      } else if (exp is PropertyAccess) {
+        chainCount++;
+        exp = (exp as PropertyAccess).target;
+      }
+      if (chainCount > 1) {
+        return;
+      }
+    }
     if (node.target != null &&
         node.methodName.token.value() == 'forEach' &&
         node.argumentList.arguments.isNotEmpty &&
--- a/test/rules/avoid_function_literals_in_foreach_calls.dart
+++ b/test/rules/avoid_function_literals_in_foreach_calls.dart
@@ -27,11 +27,18 @@ void main() {
       .map((person) => person.toUpperCase())
       .forEach((person) => print('$person!')); // OK
 
+  people
+      .where((person) => person != null)
+      .map((person) => person.toUpperCase())
+      .forEach(print); // OK
+
   Person()
       .children
       .firstWhere((person) => person != null)
       .children
       .forEach((person) => print('$person!')); // OK
+
+  Person().children.forEach(print); // OK
 
   Person()
       .children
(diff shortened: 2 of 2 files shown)


lib/src/rules/avoid_function_literals_in_foreach_calls.dart
```diff
@@ -54,6 +54,25 @@ class _Visitor extends SimpleAstVisitor<void> {
 
   @override
   void visitMethodInvocation(MethodInvocation node) {
+    var exp = node.target;
+    var chainCount = 0;
+    while (exp is PrefixedIdentifier ||
+        exp is MethodInvocation ||
+        exp is PropertyAccess) {
+      if (exp is PrefixedIdentifier) {
```
pq: In trying to convince myself that this is right I found myself wanting a bit of text in the lint description to describe the exception.

That said, @scheglov, @bwilkerson: is there a better way to do this?

test/rules/avoid_function_literals_in_foreach_calls.dart
```diff
@@ -17,4 +21,17 @@ void main() {
...
+      .map((person) => '$person!')
+      .forEach((person) => print(person)); // OK
+
+  Person()
+      .children
+      .firstWhere((person) => person != null)
+      .children
+      .forEach((person) => print(person)); // OK
+
+  Person().children.first.children.forEach((person) => print(person)); // OK
```
pq: Interesting.  I thought the proposal was limited to method-chains.

IOW, I would probably still prefer this one:

```dart
for (var person in Person().children.first.children) {
  print(person);
}
```

(which feels different than the `.where(..).map(..)` case...)

/cc @sjmcdowall @munificent

lib/src/rules/avoid_function_literals_in_foreach_calls.dart
bwilkerson: A better way to do what, specifically? There isn't any utility method that will count the number of method invocations in a chain, if that's what you're asking.

pq: Yep.  That's what I was trying to ask (but without ☕️).

My biggest remaining question is whether the hole should be poked open for field/getter accesses (as implemented) vs. only for chained function calls (as I thought).

Any opinion?

test/rules/avoid_function_literals_in_foreach_calls.dart
bwilkerson: I agree. It seems unnecessary and possibly undesirable to me to special case getters, but I'll defer to Bob on that question.

dramos07: So I initially went for only method chaining which mitigated the where/map issues, but then I felt a bit in the middle for long field/getter chaining... I'm OK either way so I'll defer to you guys.

munificent: I don't think I have a useful opinion either way here. I think it's good for the linter to *not* yell at uses of `forEach()` at the end of method chains, but I don't have a strong sense of when I think it *should* complain.

dramos07: OK since the initial issue/request was mainly focus on loosen the lint on method calls such as where/map maybe we could start limiting to only method-chains? I can update to that if we all agree

sjmcdowall: FWIW -- I agree @dramos07  :)

bwilkerson: That sounds reasonable to me. If the getter chains are causing a real issue, then we can always revisit (though changing it later will be a breaking change).

pq: Limiting it to method chains SGTM.  Thanks!

test/rules/avoid_function_literals_in_foreach_calls.dart
```diff
@@ -17,4 +21,17 @@ void main() {
   people.forEach((person) => print(person)); // LINT
 
   people.forEach(print); // OK
+
+  people
+      .where((person) => person != null)
+      .map((person) => '$person!')
+      .forEach((person) => print(person)); // OK
```
lrhn: The example may be confusing because it hits another lint (https://dart-lang.github.io/linter/lints/unnecessary_lambdas.html), and should be written as just `.forEach(print)`.
(It is literally the example given as bad for `unnecessary_lambdas`).

It's always hard to make a small meaningless example without actually being meaningless :)

lib/src/rules/avoid_function_literals_in_foreach_calls.dart
dramos07: I thinks is addressed in the other conversations

test/rules/avoid_function_literals_in_foreach_calls.dart
dramos07: Updated the code to only account for method chained calls, thanks everyone!!

test/rules/avoid_function_literals_in_foreach_calls.dart
dramos07: Good call, I have updated the example slightly to avoid the `unnecessary_lambdas` lint thanks!

test/rules/avoid_function_literals_in_foreach_calls.dart
```diff
@@ -12,26 +12,30 @@ void main() {
...
-      .forEach((person) => print(person)); // OK
+      .map((person) => person.toUpperCase())
+      .forEach((person) => print('$person!')); // OK
 
   Person()
       .children
       .firstWhere((person) => person != null)
       .children
-      .forEach((person) => print(person)); // OK
+      .forEach((person) => print('$person!')); // OK
```
pq: Maybe just one more test case?

```dart
Person().children.forEach(print); // LINT
```

dramos07: Sure but did you mean "// OK" instead? that would be the case since it is using lambda just like:

`people.forEach(print); // OK`

dramos07: Added 2 more test cases for tear-offs + chaining!!

pq: Sorry to keep muddying the waters!

I guess what's confusing me is that I would expect this to be limited to `MethodInvocation`s based on the original proposal?

```dart
bool _hasMethodChaining(MethodInvocation node) {
  var exp = node.target;
  while (exp is PrefixedIdentifier ||
      exp is MethodInvocation ||
      exp is PropertyAccess) {
    if (exp is PrefixedIdentifier) {
      exp = (exp as PrefixedIdentifier).prefix;
    } else if (exp is MethodInvocation) {
      return true;
    } else if (exp is PropertyAccess) {
      exp = (exp as PropertyAccess).target;
    }
  }
  return false;
}
```

bwilkerson: Yes, but the lint is only produced if there is a closure being passed to `forEach`, and the case you suggested (`Person().children.forEach(print);`) doesn't contain one.

pq: Got it.  It was a bad example! 😬 

I had another confusion too I realize and was mis-reading `_hasMethodChaining` but get it now.

Thanks for bearing with me!

pq:
@dramos07: is this one ready to land?

lexaknyazev:
Should the case with null-aware `forEach` invocation be also covered? Like:
```dart
arguments?.forEach((arg) => arg.promote());
```

With `curly_braces_in_flow_control_structures`, an alternative feels too verbatim:
```dart
if (arguments != null) {
  for (final arg in arguments) {
    arg.promote();
  }
}
```

davidmorgan:
Could we merge this? Thanks :)

pq:
Sorry for the delay on this one; I wasn't 100% sure it was ready to land.

PR #1595, opened 2019-06-03
fix const constructor checking w/ type params

### LINKED ISSUE #1592: [prefer_const_constructors] does not work for all Flutter Icon's
## Description

Coming from [StackOverflow](https://stackoverflow.com/q/56397809/6509751), it seems like [prefer_const_constructors](https://dart-lang.github.io/linter/lints/prefer_const_constructors.html) does not detect most instances of [`Icon`](https://github.com/flutter/flutter/blob/bfa1d25bf9f51f19dec62e1f8e9c25e6ec6cf894/packages/flutter/lib/src/widgets/icon.dart#L51).

## Reproduction

For me personally, it does seem to work for **some** instances. I successfully got hints for `Icon(Icons.delete_sweep)`, `Icon(Icons.cancel)` and `Icon(Icons.send)`, however, there is no apparent reason for this behavior since [all constants in `Icons` are assigned the same way](https://github.com/flutter/flutter/blob/bfa1d25bf9f51f19dec62e1f8e9c25e6ec6cf894/packages/flutter/lib/src/material/icons.dart#L37).  
For the OP on StackOverflow, there are no warning at all for instances of `Icon`.

```dart
main() {
  runApp(WidgetsApp(home: Icon(Icons.add), color: Color(0xffffffff)));
}
```
This warns me for the instance of `Color` but not for the instance of `Icon`, even though it should warn me for both.

No warnings like this:

```dart
runApp(WidgetsApp(home: Icon(Icons.add), color: const Color(0xffffffff)));
```

pq:
Fixes #1592

/cc @scheglov 

/fyi @a14n

### DIFF
--- a/lib/src/rules/prefer_const_constructors.dart
+++ b/lib/src/rules/prefer_const_constructors.dart
@@ -3,10 +3,9 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'package:analyzer/dart/ast/ast.dart';
-import 'package:analyzer/dart/ast/token.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/element.dart';
 import 'package:linter/src/analyzer.dart';
-import 'package:linter/src/ast.dart';
 
 const _desc = r'Prefer const with constant constructors.';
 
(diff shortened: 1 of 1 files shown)


lib/src/rules/prefer_const_constructors.dart
```diff
@@ -86,21 +85,25 @@ class _Visitor extends SimpleAstVisitor<void> {
...
-      try {
-        hasConstError = hasErrorWithConstantVerifier(context, node);
-      } finally {
-        // restore old keyword
-        node.keyword = oldKeyword;
-      }
-
-      if (!hasConstError) {
+      if (context.canBeConst(node)) {
+        // A work-around to address: #1592; remove once fixed in analyzer.
```
a14n: Add a TODO or an issue to track?

PR #1602, opened 2019-06-05
Linter: add 'avoid_relative_imports'

### LINKED ISSUE #1601: Add 'avoid_relative_imports'
This lint is necessary on codebases where there are often files being
moved around. The lack of consistent
'package:my.project.folder/mypackage.dart' makes it very painful to
detect when files are imported since not even intellij tells you that.

This lint is not supposed to be used for everyone since there are a lot
of valid cases for relative imports but I would prefer to have it as
warning or lints, not errors.

Opposite of https://github.com/dart-lang/linter/issues/1101

Regarding Effective Dart:
https://www.dartlang.org/guides/language/effective-dart/usage#prefer-relative-paths-when-importing-libraries-within-your-own-packages-lib-directory.

This rule makes it much harder move files around in codebases of thousands of files. Mixing fully qualified and relative imports is inconsistent. 

This lint does conflict with codegenerated files (web/main.template.dart in the case of Angular main entry point) and also PageLoader (my_po.g.dart).

manuelF:
This lint is necessary on codebases where there are often files being
moved around. The lack of consistent
'package:my.project.folder/mypackage.dart' makes it very painful to
detect when files are imported since not even intellij tells you that.

This lint is not supposed to be used for everyone since there are a lot
of valid cases for relative imports but I would prefer to have it as
warning or lints, not errors.

This is the opposite of #1101 since we prefer to have fully qualified types.

# Description

Fixes #1601

### DIFF
--- a/lib/src/rules.dart
+++ b/lib/src/rules.dart
@@ -26,6 +26,7 @@ import 'package:linter/src/rules/avoid_null_checks_in_equality_operators.dart';
 import 'package:linter/src/rules/avoid_positional_boolean_parameters.dart';
 import 'package:linter/src/rules/avoid_private_typedef_functions.dart';
 import 'package:linter/src/rules/avoid_relative_lib_imports.dart';
+import 'package:linter/src/rules/avoid_relative_imports.dart';
 import 'package:linter/src/rules/avoid_renaming_method_parameters.dart';
 import 'package:linter/src/rules/avoid_return_types_on_setters.dart';
 import 'package:linter/src/rules/avoid_returning_null.dart';
@@ -179,6 +180,7 @@ void registerLintRules() {
     ..register(new AvoidPositionalBooleanParameters())
     ..register(new AvoidPrivateTypedefFunctions())
     ..register(new AvoidRelativeLibImports())
+    ..register(new AvoidRelativeImports())
     ..register(new AvoidRenamingMethodParameters())
     ..register(new AvoidReturningNull())
     ..register(new AvoidReturningNullForFuture())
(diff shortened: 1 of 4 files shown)


lib/src/rules/avoid_relative_imports.dart
```diff
@@ -0,0 +1,81 @@
...
+import 'package:linter/src/analyzer.dart';
+
+const _desc = r'Avoid relative imports for all files`.';
+
+const _details = r'''*DO* avoid relative imports for files in `lib/`.
+
+Mixing relative and absolute imports makes it complicated to refactor across
+multiple packages.
+
+This is a more strict version of 'avoid_relative_lib_imports'.
```
bwilkerson: What this rule doesn't do (which avoid_relative_lib_imports does implicitly) is to guard against a false positive when a library outside the `lib` directory imports another library outside the `lib` directory. This happens, for example, in some test code. In those cases a relative path is all that can be used to reference the library.

lib/src/rules.dart
```diff
@@ -179,6 +180,7 @@ void registerLintRules() {
     ..register(new AvoidPositionalBooleanParameters())
     ..register(new AvoidPrivateTypedefFunctions())
     ..register(new AvoidRelativeLibImports())
+    ..register(new AvoidRelativeImports())
```
pq: Sort before `AvoidRelativeLibImports()`?

2ZeroSix:
@pq These changes are merged here #2126, and this pr can be safely closed.

pq:
Thanks @2ZeroSix !

PR #1605, opened 2019-06-11
Loosen prefer_initializing_formals to not include different params vs field types

dramos07:
The code below triggers the lint but they are using different field/param type and the actual fix will change the constructor argument to accept the same type as the field and thus will cause the "add" call to fail of course we could cast but it seems a bit odd
```dart
class CaseWithDifferentType {
  Iterable<int> ids;
  CaseWithDifferentType(List<int> ids){
    this.ids = ids; // LINT
    ids.add(1);
  }
}
```
The only current fix with casting:
```dart
class CaseWithDifferentType {
  Iterable<int> ids;
  CaseWithDifferentType(this.ids){
    (ids as List).add(1);
  }
}
```
This pull request will loosen this edge case to also match the type before linting so that this won't be an issue.

@pq @bwilkerson

### DIFF
--- a/lib/src/rules/prefer_final_fields.dart
+++ b/lib/src/rules/prefer_final_fields.dart
@@ -65,6 +65,23 @@ class GoodMutable {
 }
 ```
 
+**BAD:**
+```
+class AssignedInAllConstructors {
+  var _label; // LINT
+  AssignedInAllConstructors(this._label);
+  AssignedInAllConstructors.withDefault() : _label = 'Hello';
+}
+```
+
+**GOOD:**
+```
+class NotAssignedInAllConstructors {
+  var _label; // OK
+  NotAssignedInAllConstructors();
+  NotAssignedInAllConstructors.withDefault() : _label = 'Hello';
+}
+```
 ''';
 
 class PreferFinalFields extends LintRule implements NodeLintRule {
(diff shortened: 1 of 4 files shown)


lib/src/rules/prefer_initializing_formals.dart
```diff
@@ -88,14 +88,14 @@ Iterable<ConstructorFieldInitializer>
             ConstructorDeclaration node) =>
         node.initializers.whereType<ConstructorFieldInitializer>();
 
-Element _getLeftElement(AssignmentExpression assignment) =>
+FieldElement _getLeftElement(AssignmentExpression assignment) =>
```
bwilkerson: I don't think this is a safe change to make. This method can be called with any assignment in a constructor and the left hand side won't necessarily resolve to a field. For example, it might be an assignment to a local variable.

lib/src/rules/prefer_initializing_formals.dart
```diff
@@ -88,14 +88,14 @@ Iterable<ConstructorFieldInitializer>
...
     DartTypeUtilities.getCanonicalElementFromIdentifier(
         assignment.leftHandSide);
 
 Iterable<Element> _getParameters(ConstructorDeclaration node) =>
     node.parameters.parameters.map((e) => e.identifier.staticElement);
 
-Element _getRightElement(AssignmentExpression assignment) =>
+VariableElement _getRightElement(AssignmentExpression assignment) =>
```
bwilkerson: I don't think this is a safe change to make either. The right hand side of an assignment isn't always a variable. For example, consider a tear-off, which I believe will resolve to an ExecutableElement.

test/rules/prefer_initializing_formals.dart
```diff
@@ -228,3 +228,39 @@ class GoodCaseWithDifferentNamedArgsInitializer {
...
+  GoodCaseWithDifferentType(List<int> ids) {
+    this.ids = ids; // OK
+    ids.add(12);
+  }
+}
+
+class GoodCaseWithDifferentTypeNamed {
+  Iterable<int> ids;
+
+  GoodCaseWithDifferentTypeNamed({List<int> ids}) {
```
bwilkerson: Was this intended to be a named constructor? (I'm not sure what "Named" means in this context.)

test/rules/prefer_initializing_formals.dart
```diff
@@ -228,3 +228,39 @@ class GoodCaseWithDifferentNamedArgsInitializer {
...
+class GoodCaseWithDifferentTypeInitializer {
+  Iterable<int> ids;
+
+  GoodCaseWithDifferentTypeInitializer(List<int> ids)
+      : this.ids = ids { // OK
+    ids.add(12);
+  }
+}
+
+class GoodCaseWithDifferentTypeNamedInitializer {
```
bwilkerson: Ditto

lib/src/rules/prefer_initializing_formals.dart
dramos07: Indeed in fact I did not needed to change this one!

lib/src/rules/prefer_initializing_formals.dart
dramos07: Yep added as a check before using it instead.

test/rules/prefer_initializing_formals.dart
dramos07: Sorry my bad it was actually NamedParams updated

test/rules/prefer_initializing_formals.dart
dramos07: Same thanks.

lib/src/rules/prefer_final_fields.dart
```diff
@@ -135,21 +152,44 @@ class _Visitor extends SimpleAstVisitor<void> {
...
+            constructor.initializers.any((initializer) =>
+                _containedInInitializer(element, initializer)) ||
+            constructor.parameters.parameters
+                .any((formal) => _containedInFormal(element, formal));
+
+        final classDeclaration = node.parent;
+        final constructors = (classDeclaration is ClassDeclaration
+            ? classDeclaration.members
+                .whereType<ConstructorDeclaration>()
+                .toSet()
```
bwilkerson: I don't think there's any advantage to using a `Set` rather than a `List` (or even an `Iterable`), and there might be a small performance penalty for doing so.

lib/src/rules/prefer_final_fields.dart
```diff
@@ -135,21 +152,44 @@ class _Visitor extends SimpleAstVisitor<void> {
...
+bool _containedInInitializer(
+        Element element, ConstructorInitializer initializer) =>
+    initializer is ConstructorFieldInitializer &&
+    DartTypeUtilities.getCanonicalElementFromIdentifier(
+            initializer.fieldName) ==
+        element;
+
+bool _containedInFormal(Element element, FormalParameter formal) =>
+    formal is FieldFormalParameter &&
+    formal.declaredElement.toString() == element.toString();
```
bwilkerson: It ought to work to compare the elements without converting them to strings first and it ought to be safer in that we can't end up with two different elements that happen to produce the same string incorrectly comparing to be equal.

srawlins:
Changes were requested in 2019 which haven't been addressed. I'm closing this just to tidy the issue tracker. If you would like to update the branch with main, and address comments, then please re-open with changes.

PR #1611, opened 2019-06-13
 Adjust the 'prefer_final_fields' lint to cover more cases

### LINKED ISSUE #1606: prefer_final_fields doesn't cover fields only initialize in constructors
The current documentation claims that a field that is only initialize either using initializing formals or initializer list should be linted (see `_someOther`):

```dart
class MultipleMutable {
  var _label = 'hola mundo! GoodMutable', _offender = 'mumble mumble!'; // LINT
  var _someOther; // LINT
  MultipleMutable() : _someOther = 5;
  MultipleMutable(this._someOther);

  void changeLabel() {
    _label= 'hello world! GoodMutable';
  }
}
```
But that is currently not the case, in fact the code actually skips any field that is not initialize when the field is declared. So this needs to be fixed or at the very least the documentation should be updated to match the code.

@pq @bwilkerson

dramos07:
The current lint only covers fields which have a default value set to them which excludes cases like:

```dart
class AssignedInAllConstructors {
  var _label; // Currently OK
  AssignedInAllConstructors(this._label);
  AssignedInAllConstructors.withDefault() : _label = 'Hello';
}
```
These change tries to aim for linting fields assigned only once (either by default or in all the constructors)"

Fixes #1606

@pq @bwilkerson

### DIFF
--- a/lib/src/rules/prefer_final_fields.dart
+++ b/lib/src/rules/prefer_final_fields.dart
@@ -163,11 +163,9 @@ class _Visitor extends SimpleAstVisitor<void> {
                 .any((formal) => _containedInFormal(element, formal));
 
         final classDeclaration = node.parent;
-        final constructors = (classDeclaration is ClassDeclaration
-            ? classDeclaration.members
-                .whereType<ConstructorDeclaration>()
-                .toSet()
-            : <ConstructorDeclaration>{});
+        final Iterable constructors = (classDeclaration is ClassDeclaration
+            ? classDeclaration.members.whereType<ConstructorDeclaration>()
+            : []);
         final isFieldInConstructors = constructors.any(fieldInConstructor);
         final isFieldInAllConstructors = constructors.every(fieldInConstructor);
 
@@ -190,6 +188,8 @@ bool _containedInInitializer(
             initializer.fieldName) ==
         element;
 
-bool _containedInFormal(Element element, FormalParameter formal) =>
-    formal is FieldFormalParameter &&
-    formal.declaredElement.toString() == element.toString();
+bool _containedInFormal(Element element, FormalParameter formal) {
+  final formalField = formal.identifier.staticElement;
+  return formalField is FieldFormalParameterElement &&
+      formalField.field == element;
+}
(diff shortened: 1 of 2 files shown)


lib/src/rules/prefer_final_fields.dart
```diff
@@ -135,21 +152,44 @@ class _Visitor extends SimpleAstVisitor<void> {
...
+            constructor.initializers.any((initializer) =>
+                _containedInInitializer(element, initializer)) ||
+            constructor.parameters.parameters
+                .any((formal) => _containedInFormal(element, formal));
+
+        final classDeclaration = node.parent;
+        final constructors = (classDeclaration is ClassDeclaration
+            ? classDeclaration.members
+                .whereType<ConstructorDeclaration>()
+                .toSet()
```
bwilkerson: I don't believe that that there's any advantage to converting this to a `Set` as opposed to either a `List` or `Iterable`, and there might be a small performance penalty for doing so.

lib/src/rules/prefer_final_fields.dart
```diff
@@ -135,21 +152,44 @@ class _Visitor extends SimpleAstVisitor<void> {
...
+bool _containedInInitializer(
+        Element element, ConstructorInitializer initializer) =>
+    initializer is ConstructorFieldInitializer &&
+    DartTypeUtilities.getCanonicalElementFromIdentifier(
+            initializer.fieldName) ==
+        element;
+
+bool _containedInFormal(Element element, FormalParameter formal) =>
+    formal is FieldFormalParameter &&
+    formal.declaredElement.toString() == element.toString();
```
bwilkerson: I don't think it's necessary to convert the elements to strings before comparing them, and might open the door for a false positive if two elements that are not equal happen to generate the same string.

lib/src/rules/prefer_final_fields.dart
dramos07: Left it as Iterable thanks!!

lib/src/rules/prefer_final_fields.dart
dramos07: I was actually out of ideas on how to get to the FieldElement from the FormalParameter but I digging a bit deeper I found out how.

PR #1617, opened 2019-06-18
Apply the same restrictions for LinkedHashMap as the ones for LinkedHashSet

dramos07:
Currently `LinkedHashSet` excludes the following cases for the lint:

```dart
void main() {
  LinkedHashSet<int> linkedHashSet =  LinkedHashSet.from([1, 2, 3]); // OK
  LinkedHashSet<String> sss1 = <int, LinkedHashSet<String>>{}.putIfAbsent(3, () => LinkedHashSet<String>()); // OK
 
  printSet(LinkedHashSet<int>()); // LINT
  printHashSet(LinkedHashSet<int>()); // OK
}
void printSet(Set<int> ids) => print('$ids!');
void printHashSet(LinkedHashSet<int> ids) => printSet(ids);
```
For consistency we want the same cases to be expected for `LinkedHashMap`:

```dart
void main() {
  LinkedHashMap linkedHashMap = LinkedHashMap(); // OK
  LinkedHashMap<String, String> lhm = <int, LinkedHashMap<String,String>>{}.putIfAbsent(3, () => LinkedHashMap<String, String>()); // OK

  printMap(LinkedHashMap<int, int>()); // LINT
  printHashMap(LinkedHashMap<int, int>()); // OK
}
void printMap(Map map) => print('$map!');
void printHashMap(LinkedHashMap map) => printMap(map);
```
@pq @bwilkerson

### DIFF
--- a/lib/src/rules/prefer_collection_literals.dart
+++ b/lib/src/rules/prefer_collection_literals.dart
@@ -34,7 +34,7 @@ var coordinates = <int,int>{};
 
 **EXCEPTIONS:**
 
-There are cases with LinkedHasSet or LinkedHashMap where a literal constructor
+There are cases with `LinkedHashSet` or `LinkedHashMap` where a literal constructor
 will trigger a type error so those will be excluded from the lint.
 
 ```
--- a/lib/src/rules/prefer_collection_literals.dart
+++ b/lib/src/rules/prefer_collection_literals.dart
@@ -32,6 +32,28 @@ var ids = <int>{};
 var coordinates = <int,int>{};
 ```
 
+**EXCEPTIONS:**
+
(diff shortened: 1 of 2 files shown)


lib/src/rules/prefer_collection_literals.dart
```diff
@@ -32,6 +32,28 @@ var ids = <int>{};
 var coordinates = <int,int>{};
 ```
 
+**EXCEPTIONS:**
+
+There are cases with LinkedHasSet or LinkedHashMap where a literal constructor
```
pq: LinkedHasSet => LinkedHashSet

Also, consider formatting these as code? As in \`LinkedHashSet\`?

dramos07: Sure thanks

PR #1626, opened 2019-06-27
avoid_print

### LINKED ISSUE #88: Linter for print statements
Published code should use logging rather than `print()` statements, so consumers can decide how debugging should be handled.

pq:
(Nearly) the simplest possible lint for `print`s.

Refinements welcome.

(In particular, do we want to allow calls in `Logger`s?)

Fixes #88

/cc @bwilkerson

### DIFF
--- a/lib/src/rules/avoid_print.dart
+++ b/lib/src/rules/avoid_print.dart
@@ -44,7 +44,7 @@ class _Visitor extends SimpleAstVisitor {
   @override
   visitMethodInvocation(MethodInvocation node) {
     bool isDartCore(MethodInvocation node) =>
-        node.staticInvokeType?.element?.library?.name == 'dart.core';
+        node.methodName.staticElement?.library?.name == 'dart.core';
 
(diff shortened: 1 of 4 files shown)


lib/src/rules/avoid_print.dart
```diff
@@ -0,0 +1,53 @@
...
+
+class _Visitor extends SimpleAstVisitor {
+  final LintRule rule;
+
+  _Visitor(this.rule);
+
+  @override
+  visitMethodInvocation(MethodInvocation node) {
+    bool isDartCore(MethodInvocation node) =>
+        node.staticInvokeType?.element?.library?.name == 'dart.core';
```
bwilkerson: I think this should be `node.methodName.staticElement?.library`. The `staticInvokeType` is a `FunctionType` and those don't always have elements, whereas the suggested replacement should always work for valid code.

pq: Ah, great.  Thanks!

PR #1630, opened 2019-06-28
fix pedantic scoring to work w/ includes [tbr]

pq:
Scoring broke with the new "include" scheme for versioned pedantic options.  This fixes that.

/cc @bwilkerson


pq:
Brian is out but can safely land TBR.

PR #1632, opened 2019-07-01
initial fix excludes (‘a’ edition)

pq:
Starting w/ the `a`s.

/cc @bwilkerson


tool/score/fix_excludes.yaml
```diff
@@ -0,0 +1,17 @@
...
+# semantics.
+#
+# Rules in this list will be excluded from scoring.
+
+- avoid_as
+- avoid_catching_errors
+- avoid_classes_with_only_static_members
+- avoid_implementing_value_types
+- avoid_positional_boolean_parameters
+- avoid_returning_null_for_future
```
bwilkerson: We have an existing fix to add `async` to the function body that we could consider using for this one, but I'm not sure it's worth the effort.

tool/score/fix_excludes.yaml
```diff
@@ -0,0 +1,17 @@
...
+- avoid_returning_this
+- avoid_slow_async_io
```
bwilkerson: We could have a fix to convert the async member to the sync member (and remove the `await` in the process). I don't know that it's worth doing, however.

tool/score/fix_excludes.yaml
```diff
@@ -0,0 +1,17 @@
...
+- avoid_types_as_parameter_names
+# todo: consider a wrap w/ future assist?
+- await_only_futures
```
bwilkerson: I was going to suggest a possible fix of removing the `await` keyword, but I checked and we already have that fix. Seems like we ought to not exclude lints for which there are already fixes. :-)

pq: Agreed. 😄 And thanks for checking!

tool/score/fix_excludes.yaml
pq: Maybe not.  For this pass though I'd like to be a bit more conservative and just rule out things that would be obviously crazy.

tool/score/fix_excludes.yaml
pq: Doesn't sound like a bad proposal actually.  I'll add it back in.

PR #1635, opened 2019-07-03
remove (usually) unneeded import

pq:
This is gets added to all rule stubs and is usually unnecessary.

/cc @srawlins @bwilkerson


pq:
> it does have the advantage that it point folks to the utility library so that they're less likely to spend time re-implementing these

That was why I had it in the first place.  The problem is there are now a few utilities and odds are good we won't anticipate the right ones...

Another alternative (and probably either way) would be better documentation in the generated scaffold.  We could comment out but describe utilities in imports for example?

More to think about. 🤔

pq:
Oh and FWIW the pana breakage should be addressed by: #1636

PR #1637, opened 2019-07-03
some implicit cast tidying

pq:
Baby-steps towards NNBD preparedness.

/cc @bwilkerson


pq:
FWIW: the pana baseline will be broken until we can update to the new analyzer which can't happen until there's a new compatible `dart_style`...

PR #1640, opened 2019-07-05
more casts (nnbd prep)

pq:
/cc @bwilkerson

### DIFF
--- a/lib/src/rules/avoid_setters_without_getters.dart
+++ b/lib/src/rules/avoid_setters_without_getters.dart
@@ -73,7 +73,7 @@ class _Visitor extends SimpleAstVisitor<void> {
 
   @override
   void visitClassDeclaration(ClassDeclaration node) {
-    final methods = node.members.where(isMethod) as Iterable<MethodDeclaration>;
+    final methods = node.members.where(isMethod);
     for (MethodDeclaration method in methods) {
       if (method.isSetter &&
           !_hasInheritedSetter(method) &&
--- a/lib/src/rules/avoid_setters_without_getters.dart
+++ b/lib/src/rules/avoid_setters_without_getters.dart
@@ -73,7 +73,7 @@ class _Visitor extends SimpleAstVisitor<void> {
 
   @override
   void visitClassDeclaration(ClassDeclaration node) {
-    final methods = node.members.where(isMethod);
+    final methods = node.members.where(isMethod) as Iterable<MethodDeclaration>;
     for (MethodDeclaration method in methods) {
       if (method.isSetter &&
           !_hasInheritedSetter(method) &&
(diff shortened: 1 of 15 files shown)


a14n:
just curious: are you using some tools to prepare NNBD and to get potential issues?

test/mock_sdk.dart
```diff
@@ -377,7 +377,7 @@ class  ScriptElement {
       String libraryPath = provider.convertPath(library.path);
       if (filePath == libraryPath) {
         try {
-          resource.File file = provider.getResource(filePath);
+          resource.File file = provider.getResource(filePath) as resource.File;
```
bwilkerson: Consider using `getFile` rather than `getResource`, both here and below.

test/rule_test.dart
```diff
@@ -72,11 +72,11 @@ defineRuleTests() {
     group('pub', () {
       for (var entry in Directory(p.join(ruleDir, 'pub')).listSync()) {
         if (entry is! Directory) continue;
-        Directory pubTestDir = entry;
+        Directory pubTestDir = entry as Directory;
```
bwilkerson: You could remove the casts by reversing the sense of the tests:
```
if (entry is Directory) {
  for (var child in entry.listSync()) {
    if (child is File && isPuspecFile(child)) {
      ...
    }
  }
}
```

test/rule_test.dart
```diff
@@ -319,19 +319,16 @@ AnnotationMatcher matchesAnnotation(
         String message, ErrorType type, int lineNumber) =>
     AnnotationMatcher(Annotation(message, type, lineNumber));
 
-testEach(Iterable<Object> values, bool f(String s), Matcher m) {
-  values.forEach((s) => test('"$s"', () => expect(f(s), m)));
-}
-
-testEachInt(Iterable<Object> values, bool f(int s), Matcher m) {
+testEach<T>(Iterable<T> values, bool f(T s), Matcher m) {
```
bwilkerson: Could we just import the definition from `test_utils.dart` rather than duplicating it here?

test/mock_sdk.dart
pq: Thanks!

pq:
@a14n: pretty manual.  Just tracking down implicit casts for now.

lib/src/rules/avoid_setters_without_getters.dart
```diff
@@ -73,8 +73,8 @@ class _Visitor extends SimpleAstVisitor<void> {
 
   @override
   void visitClassDeclaration(ClassDeclaration node) {
-    final methods = node.members.where(isMethod);
-    for (MethodDeclaration method in methods) {
+    for (var member in node.members.where(isMethod)) {
```
leafpetersen: @munificent It might be nice if pattern matching/destructuring provided a nicer way to handle this idiom.

PR #1641, opened 2019-07-08
implicit cast cleanup

pq:
Final implicit case cleanup and options update to disallow moving forward.

/cc @bwilkerson

### DIFF
--- a/lib/src/cli.dart
+++ b/lib/src/cli.dart
@@ -99,12 +99,12 @@ Future runLinter(List<String> args, LinterOptions initialLintOptions) async {
 
   var configFile = options['config'];
   if (configFile != null) {
-    var config = LintConfig.parse(readFile(configFile));
+    var config = LintConfig.parse(readFile(configFile as String));
     lintOptions.configure(config);
   }
 
   var lints = options['rules'];
-  if (lints != null && !lints.isEmpty) {
+  if (lints is Iterable<String> && lints.isNotEmpty) {
     var rules = <LintRule>[];
     for (var lint in lints) {
       var rule = Registry.ruleRegistry[lint];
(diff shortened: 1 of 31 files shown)


lib/src/cli.dart
```diff
@@ -99,12 +99,12 @@ Future runLinter(List<String> args, LinterOptions initialLintOptions) async {
 
   var configFile = options['config'];
   if (configFile != null) {
-    var config = LintConfig.parse(readFile(configFile));
+    var config = LintConfig.parse(readFile(configFile as String));
```
bwilkerson: This could be done in the preceding if condition (`configFile is String`).

lib/src/rules/invariant_booleans.dart
```diff
@@ -172,6 +174,7 @@ class _InvariantBooleansVisitor extends ConditionScopeVisitor {
     final elements = _getElementsInExpression(node);
     Iterable<Expression> conjunctions = getTrueExpressions(elements).toSet();
     Iterable<Expression> negations = getFalseExpressions(elements).toSet();
-    return TestedExpressions(node, conjunctions, negations);
+    return TestedExpressions(
+        node, LinkedHashSet.from(conjunctions), LinkedHashSet.from(negations));
```
bwilkerson: Given that we're using `toSet` above (lines 175 and 176), do we really need to call these constructors? Could we just change the types of these variables to `Set<Expression>`?

lib/src/rules/unnecessary_lambdas.dart
```diff
@@ -90,14 +90,17 @@ class _Visitor extends SimpleAstVisitor<void> {
       final statement = body.block.statements.single;
       if (statement is ExpressionStatement &&
           statement.expression is InvocationExpression) {
-        _visitInvocationExpression(statement.expression, node);
+        _visitInvocationExpression(
+            statement.expression as InvocationExpression, node);
```
bwilkerson: We could reduce the number of casts by assigning `statement.expression` to a local before testing the type (thus allowing promotion to work).

tool/crawl.dart
```diff
@@ -137,9 +137,10 @@ Future<String> findSinceLinter(String lint) async {
 
 Future<int> _readLatestMinorVersion() async {
   var contents = await File('pubspec.yaml').readAsString();
-  YamlMap pubspec = loadYamlNode(contents);
+  YamlMap pubspec = loadYamlNode(contents) as YamlMap;
   // 0.1.79 or 0.1.79-dev
-  return int.parse(pubspec['version'].split('.').last.split('-').first);
+  return int.parse(
+      pubspec['version'].split('.').last.split('-').first as String);
```
bwilkerson: Consider moving the cast earlier: `(pubspec['version'] as String).split('.')`. It would allow stronger checking of the code following the cast.

lib/src/rules/invariant_booleans.dart
pq: Great catch.  Thanks!

tool/crawl.dart
pq: Good idea.  In fact, I wonder if we couldn't add a hint in cases like this...

bwilkerson: The problem is that `pubspec['version']` is `dynamic`, which means we can't know which `split` is being invoked. We'd basically need to work backwards by finding all declarations of `first` and look for those that return a `String`, then look for all implementations of `split` that return one of the types that implement one of the declarations of `first`, then ..., and then hope that there's only one choice of types to which to `pubspec['version']` could be cast in order to make it all type check correctly.

PR #1648, opened 2019-07-12
test delint

pq:
Cleanup unrelated lints.

/cc @srawlins @bwilkerson


pq:
> It might be better to consider excluding the test data from analysis altogether, and certainly from linting. WDYT?

We're actually already doing that.  (But analysis is still triggered for open files.)

I don't feel strongly about this one so I'll close.  If that changes I'll revisit.

Thanks!

Issue #1655, opened 2019-07-19
false positive in `prefer_asserts_in_initializer_lists`


kevmoo:
This should be fine, but is not. Seems a lot like https://github.com/dart-lang/linter/issues/1557 (only comes up if there is a type parameter)

```dart
abstract class H<T> {
  H() {
    assert(_assert()); // OK
  }

  bool _assert() => true;
}
```

Dart VM version: 2.5.0-dev.1.0 (Tue Jul 9 15:27:01 2019 +0200) on "macos_x64"

kevmoo:
```
info: Prefer putting asserts in initializer list. (prefer_asserts_in_initializer_lists at [linter] test/rules/prefer_asserts_in_initializer_lists.dart:130)
```

kevmoo:
CC @pq

PR #1656, opened 2019-07-19
Fix false positive in prefer_asserts_in_initializer_lists

kevmoo:
Fixes https://github.com/dart-lang/linter/issues/1655

### DIFF
--- a/lib/src/rules/prefer_asserts_in_initializer_lists.dart
+++ b/lib/src/rules/prefer_asserts_in_initializer_lists.dart
@@ -101,8 +101,8 @@ class _AssertVisitor extends RecursiveAstVisitor {
   bool _hasMethod(MethodElement element) {
     final type = classElement.type;
     final name = element.name;
-    return type.lookUpMethod(name, element.library) == element ||
-        type.lookUpInheritedMethod(name) == element;
+    return element == type.lookUpMethod(name, element.library) ||
+        element == type.lookUpInheritedMethod(name);
   }
 }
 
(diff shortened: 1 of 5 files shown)


lib/src/rules/prefer_asserts_in_initializer_lists.dart
```diff
@@ -101,8 +101,8 @@ class _AssertVisitor extends RecursiveAstVisitor {
   bool _hasMethod(MethodElement element) {
     final type = classElement.type;
     final name = element.name;
-    return type.lookUpMethod(name, element.library) == element ||
-        type.lookUpInheritedMethod(name) == element;
+    return element == type.lookUpMethod(name, element.library) ||
```
kevmoo: The `==` impl here is not symmetric, which is pretty annoying.

CC @scheglov

kevmoo: CC @bwilkerson

scheglov: Yeah, this stinks.

The reason why it is not symmetrical is that `type.lookUpMethod` might return `MethodMember`, which is `MethodElement` with type information (i.e. type parameters replaced with specific type arguments from the `type`). And `MethodMember` or `Member` at all does not implement `operator==`. It is not obvious what `==` means for members. Is it just comparison of base elements? Or also types of the member and the given element?

kevmoo:
> I don't understand why "-dev" was removed, but you understand pub better than I do, so I'll trust that it's the right thing to do.

Some of us use `-dev` to signal the repo is not ready for publish.

But that breaks the `pana` test. It doesn't seem to be the flow on this repo.

pq:
This is awesome. Thanks!

PR #1662, opened 2019-07-25
Move some testing support and tests of said support into test/util

srawlins:
Summary:

* Move Annotation and helper into `test/util/annotation.dart`,
* Move AnnotationMatcher and helper into `test/util/annotation_matcher.dart`,
* Move sanity tests of both Annotation and AnnotationMatcher to `test/util/annotation_test.dart`,
* Move all file-system/ResourceProvider code into `test/util/test_resource_provider.dart`. A lot of ResourceProvider overhead was localized to ~20 lines of code and is better suited to a single-purpose utility file.


test/rule_test.dart
```diff
@@ -204,57 +197,10 @@ defineRuleUnitTests() {
   });
 }
 
-/// Test framework sanity
+// Test framework sanity.
```
bwilkerson: We use a sorter on our code, but the sorter doesn't always preserve the locations of non-doc comments. Please either restore the slash or move the comments inside the method so that they won't be moved away from the method unintentionally (here and elsewhere).

srawlins: Oh interesting. Done!

pq:
Thanks @srawlins !

PR #1666, opened 2019-07-28
Replace DartType.isSubtypeOf() with TypeSystem.isSubtypeOf().

scheglov:
All type operations in general should be done using `TypeSystem`.
We want to remove duplicate, older implementations in `DartTypeImpl`.

### DIFF
--- a/lib/src/rules/prefer_is_empty.dart
+++ b/lib/src/rules/prefer_is_empty.dart
@@ -118,7 +118,7 @@ class _Visitor extends SimpleAstVisitor<void> {
             null &&
         typeSystem.mostSpecificTypeArgument(type, typeProvider.mapType) ==
             null &&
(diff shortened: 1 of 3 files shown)


test/rule_test.dart
```diff
@@ -241,6 +241,9 @@ testRules(String ruleDir, {String analysisOptions}) {
 testRule(String ruleName, File file,
     {bool debug = false, String analysisOptions}) {
   registerLintRules();
+  if (ruleName != 'unnecessary_await_in_return') {
+    return;
```
srawlins: Should there be a comment here about why this test is being skipped?

scheglov: No, it should not be here.
I forgot to revert after debugging the failure.
Will do before landing.

scheglov: Done.

PR #1667, opened 2019-08-01
Do not report parens around set-or-map literal as method target

srawlins:
Currently the unnecessary_parenthesis rule reports the following statement:

```dart
({1, 2, 3}).forEach(print);
```

But Dart cannot parse the statement if the parens are removed; it thinks an anonymous block is being declared, and fails to parse.

This PR stops reporting that case.

### DIFF
--- a/lib/src/rules/unnecessary_parenthesis.dart
+++ b/lib/src/rules/unnecessary_parenthesis.dart
@@ -106,15 +106,19 @@ class _Visitor extends SimpleAstVisitor<void> {
       // PrefixExpression.
       if (parent is MethodInvocation) {
         Expression target = parent.target;
-
-        // Something like `({1, 2, 3}).forEach(print);`.
-        // The parens cannot be removed because then the curly brackets are not
-        // interpreted as a set-or-map literal.
-        if (target == node && node.expression is SetOrMapLiteral) return;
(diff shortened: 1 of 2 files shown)


lib/src/rules/unnecessary_parenthesis.dart
```diff
@@ -106,6 +106,12 @@ class _Visitor extends SimpleAstVisitor<void> {
       // PrefixExpression.
       if (parent is MethodInvocation) {
         Expression target = parent.target;
+
+        // Something like `({1, 2, 3}).forEach(print);`.
+        // The parens cannot be removed because then the curly brackets are not
+        // interpreted as a set-or-map literal.
+        if (target == node && node.expression is SetOrMapLiteral) return;
```
bwilkerson: I think this disables the lint in too many places. Consider:
```
print({}.length);
```
No parentheses are needed here because it's not at the beginning of a statement. I think this condition should include ` && parent.parent is ExpressionStatement`.

srawlins: Thanks for this check, Brian! Your `.length` example also uncovered a PropertyAccess shortcoming in my impl. All fixed now.

PR #1668, opened 2019-08-02
cast fix; travis build fix

pq:
Fixes a cast issue and updates travis to make sure it breaks for things like this in the future.

image

(Not clear why `dartanalyzer` wasn't failing on this and I'm researching that separately.

/cc @srawlins @bwilkerson

### DIFF
--- a/lib/src/rules/unnecessary_parenthesis.dart
+++ b/lib/src/rules/unnecessary_parenthesis.dart
@@ -115,7 +115,7 @@ class _Visitor extends SimpleAstVisitor<void> {
       // The parens cannot be removed because then the curly brackets are not
       // interpreted as a set-or-map literal.
       if (parent is PropertyAccess || parent is MethodInvocation) {
-        Expression target = (parent as dynamic).target;
+        var target = (parent as dynamic).target;
         if (target == node &&
             node.expression is SetOrMapLiteral &&
             parent.parent is ExpressionStatement) return;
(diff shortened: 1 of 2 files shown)


srawlins:
Do you have no-implicit-casts turned on or something, in your IntelliJ?

lib/src/rules/unnecessary_parenthesis.dart
```diff
@@ -115,7 +115,7 @@ class _Visitor extends SimpleAstVisitor<void> {
       // The parens cannot be removed because then the curly brackets are not
       // interpreted as a set-or-map literal.
       if (parent is PropertyAccess || parent is MethodInvocation) {
-        Expression target = (parent as dynamic).target;
+        var target = (parent as dynamic).target;
```
srawlins: Normally I would not like this fix, as it might invoke dynamic later (just a personal distaste), but I see its just doing an `==`, so its clean enough. My target access is kind of  hacky anyway. PropertyAccess and MethodInvocation both just _happen_ to have a field named target, and I took advantage of that.

bwilkerson: An alternative would be to introduce a `_getTarget` helper that returned the target or `null`.

pq: Yeah.  It's not an ideal fix.

> Do you have no-implicit-casts turned on or something, in your IntelliJ?

Actually we do in analysis options.

https://github.com/dart-lang/linter/blob/f3413e972bbe8015009ec1450245cc63d51963bf/analysis_options.yaml#L5

It's a puzzle as to why travis isn't flagging it without the explicit add.

PR #1677, opened 2019-08-19
update since cache

pq:
/cc @bwilkerson


pq:
FWIW: the pana baseline breakage is because we can't bump to the latest analyzer release until there's a compatible `dart_style`

https://travis-ci.org/dart-lang/linter/jobs/573858121#L379-L385

(See https://github.com/dart-lang/dart_style/issues/839.)

PR #1678, opened 2019-08-19
Add a lint to avoid operator == and hashCode on mutable classes

### LINKED ISSUE #83: Lint for hashCode implementation referencing mutable fields
http://blog.mgm-tp.com/2012/03/hashset-java-puzzler/

krisgiesing:
Fixes #83


krisgiesing:
@pq Can you review this, or is there someone else who can?

I'm curious whether this is the right way to lint for this issue. I started down a different path of detecting mutation of non-final fields in the operator == and hashCode implementations, but that gets complicated if the mutation happens in a different method.

pq:
Happy to take a look.  Adding @bwilkerson too. 👍 

And thanks!

lib/src/rules.dart
```diff
@@ -22,6 +22,8 @@ import 'package:linter/src/rules/avoid_function_literals_in_foreach_calls.dart';
 import 'package:linter/src/rules/avoid_implementing_value_types.dart';
 import 'package:linter/src/rules/avoid_init_to_null.dart';
 import 'package:linter/src/rules/avoid_js_rounded_ints.dart';
+import 'package:linter/src/rules/avoid_mutable_fields_in_equality_operators.dart';
```
pq: This one is missing from this PR I think?

lib/src/rules.dart
```diff
@@ -176,6 +178,8 @@ void registerLintRules() {
     ..register(AvoidImplementingValueTypes())
     ..register(AvoidInitToNull())
     ..register(AvoidJsRoundedInts())
+    ..register(AvoidMutableFieldsInEqualityOperators())
```
pq: Here too.

lib/src/rules/avoid_operator_equals_on_mutable_classes.dart
```diff
@@ -0,0 +1,120 @@
+// Copyright (c) 2017, the Dart project authors. Please see the AUTHORS file
```
pq: => 2019

test/rules/avoid_operator_equals_on_mutable_classes.dart
```diff
@@ -0,0 +1,30 @@
+// Copyright (c) 2017, the Dart project authors. Please see the AUTHORS file
```
pq: => 2019

lib/src/rules.dart
krisgiesing: Whoops, that is left over from a different approach. I need to remove it.

krisgiesing: Done

lib/src/rules.dart
krisgiesing: Done

lib/src/rules/avoid_operator_equals_on_mutable_classes.dart
krisgiesing: Done

test/rules/avoid_operator_equals_on_mutable_classes.dart
krisgiesing: Done

lib/src/rules/avoid_operator_equals_on_mutable_classes.dart
```diff
@@ -0,0 +1,120 @@
...
+
+const _desc = r'AVOID overloading operator == on classes not marked `@immutable`.';
+
+const _details = r'''
+
+**AVOID** overloading operator == on classes not marked `@immutable`.
+
+If a class is not immutable, overloading operator == and hashCode can lead to
+unpredictable and undesirable behavior when used in collections. See
+for more information.
```
Hixie: missing reference

krisgiesing: Done

lib/src/rules/avoid_operator_equals_on_mutable_classes.dart
```diff
@@ -0,0 +1,121 @@
...
+    if (node.name.token?.type == TokenType.EQ_EQ || isHashCode(node)) {
+      final ClassElement classElement = _getClassForMethod(node);
+      if (classElement != null && !_hasImmutableAnnotation(classElement)) {
+        rule.reportLintForToken(node.firstTokenAfterCommentAndMetadata);
+      }
+    }
+  }
+
+  ClassElement _getClassForMethod(MethodDeclaration node) {
+    AstNode maybeClass = node.parent;
```
bwilkerson: Consider using `AstNode.thisOrAncestorOfType` rather than re-implementing the lookup loop.

lib/src/rules/avoid_operator_equals_on_mutable_classes.dart
```diff
@@ -0,0 +1,121 @@
...
+    InterfaceType current = type;
+    Set<ClassElement> seenTypes = <ClassElement>{};
+    while (current != null && seenTypes.add(current.element)) {
+      yield current;
+      current = current.superclass;
+    }
+  }
+
+  bool _hasImmutableAnnotation(ClassElement clazz) {
+    final inheritedAndSelfTypes = _getSelfAndInheritedTypes(clazz.type);
```
bwilkerson: Consider using `ClassElement.allSupertypes` to get the supertypes of `clazz`.

krisgiesing:
> This rule requires use of the `@immutable` annotation in order to suppress the lint, but it's possible that not all users will want to use it. Have you considered defining a slightly different rule that disallows references to non-final fields within `==` and `hashCode`? It seems like that rule would cover the cases you have in mind while making the rule more generally useful.

Yes, that was my first approach, as mentioned above. The complication there is that the reference to the non-final field might occur in a method called by `==` or `hashCode` rather than those methods themselves. I can even concoct cases where the reference occurs in a different class, although that's probably an extremely uncommon pattern.

Getting this exactly right seems like it would involve replicating a lot of the logic of @immutable itself (though constrained only to the implementation used by `==` and `hashCode`). We could cover only the more common cases, instead, and then carefully document the limitations of the lint. I'm interested in hearing your thoughts about this tradeoff.

krisgiesing:
By the way, the difference in approach here isn't just theoretical; in a recent code review, we tried to analyze the impact of defining `==` on a particular class, and we ended up marking the class `@immutable` simply because reasoning about whether the `==` check would behave properly was distracting from the main purpose of the class.

bwilkerson:
> Yes, that was my first approach, as mentioned above.

If I understood your first approach, I think there's a subtle difference. I understood your first approach to be to look inside these methods for references to fields, and then to look everywhere to see whether the field is mutated somewhere. I was proposing a slight variation on that: look inside the methods for references to fields, and flag the method if any of those fields are non-final. My suggestion has the advantage that you don't need to look outside the methods in order to determine whether or not to flag them.

krisgiesing:
> > Yes, that was my first approach, as mentioned above.
> 
> If I understood your first approach, I think there's a subtle difference. I understood your first approach to be to look inside these methods for references to fields, and then to look everywhere to see whether the field is mutated somewhere. I was proposing a slight variation on that: look inside the methods for references to fields, and flag the method if any of those fields are non-final. My suggestion has the advantage that you don't need to look outside the methods in order to determine whether or not to flag them.

I meant look inside the methods for references to fields, and flag the method if any of those fields are non-final.

The complication is introduced in the first clause: how deep should I look inside the method? Just the method body itself, or into any and all other methods called from within that method? The first is what I meant by covering only the more common cases and documenting the limitations of the lint. The second starts to seem like it requires the same machinery as what `@immutable` itself does.

Again, this distinction isn't just theoretical to me. The code review I mentioned above had `==` and `hashCode` defined on a base class where those methods called a `get value` method declared as abstract on the base. Subclasses could override that `get value` to return a reference to a non-final field.

bwilkerson:
Ok, yes. If we need to worry about subclasses overriding getters, then you're right; it isn't practical to prove immutability without the annotation.

pq:
Looks like you need to reformat?

https://travis-ci.org/dart-lang/linter/jobs/574031326

krisgiesing:
Yup, working on the other refactoring that @bwilkerson suggested also.

lib/src/rules/avoid_operator_equals_on_mutable_classes.dart
krisgiesing: Done, thanks! I didn't know about that method.

lib/src/rules/avoid_operator_equals_on_mutable_classes.dart
krisgiesing: Done. Note that I copied this logic from `prefer_const_constructors_in_immutables.dart` and `prefer_const_literals_to_create_immutables.dart`, so perhaps those classes should be updated as well?

krisgiesing:
All review comments should be addressed now (and the Travis tests passed). Please take another look, thanks!

lib/src/rules/avoid_operator_equals_on_mutable_classes.dart
```diff
@@ -0,0 +1,103 @@
...
+import 'package:analyzer/dart/element/element.dart';
+import 'package:linter/src/analyzer.dart';
+import 'package:linter/src/ast.dart';
+
+const _desc =
+    r'AVOID overloading operator == on classes not marked `@immutable`.';
+
+const _details = r'''
+
+**AVOID** overloading operator == on classes not marked `@immutable`.
```
bwilkerson: This and the `_desc` above should probably be expanded to include hashCode in the list of members being checked for.

lib/src/rules/avoid_operator_equals_on_mutable_classes.dart
```diff
@@ -0,0 +1,103 @@
...
+bool _isImmutable(Element element) =>
+    element is PropertyAccessorElement &&
+    element.name == _IMMUTABLE_VAR_NAME &&
+    element.library?.name == _META_LIB_NAME;
+
+class AvoidOperatorEqualsOnMutableClasses extends LintRule
+    implements NodeLintRule {
+  AvoidOperatorEqualsOnMutableClasses()
+      : super(
+            name: 'avoid_operator_equals_on_mutable_classes',
```
bwilkerson: Should this be 'avoid_equals_and_hash_code_on_mutable_classes'?

lib/src/rules/avoid_operator_equals_on_mutable_classes.dart
```diff
@@ -0,0 +1,103 @@
...
+```
+class B {
+  final String key;
+  const B(this.key);
+  @override operator ==(other) => other is B && other.key == key;
+  @override int hashCode() => key.hashCode;
+}
+```
+
+''';
```
kjeremy: As someone relatively new to dart this suggests to me that the error is really that `B` could be marked `@immutable` but isn't.

lib/src/rules/avoid_operator_equals_on_mutable_classes.dart
krisgiesing: Done

lib/src/rules/avoid_operator_equals_on_mutable_classes.dart
krisgiesing: Done

lib/src/rules/avoid_operator_equals_on_mutable_classes.dart
krisgiesing: I was trying to emphasize that the lint is literally checking for the `@immutable` annotation rather than doing an analysis of the mutability of the class: here, the lint will trigger even though the class is not mutable. If you think it's clearer I can remove the `final` keyword from the field, or include both examples.

krisgiesing: I added another example and some explanatory text.

krisgiesing:
@pq I don't have write access; can you merge this for me? That assumes @kjeremy is good with the updated wording on the examples; please let me know if that's not the case.

kjeremy:
Thanks I think this is a lot clearer.

PR #1682, opened 2019-08-20
extension support for avoid_setters_without_getters

pq:
See: https://github.com/dart-lang/linter/issues/1683.

/cc @bwilkerson


pq:
Dropping as per discussion in #1683

PR #1687, opened 2019-08-21
support for class decls, enums and mixins

pq:
/cc @bwilkerson


pq:
Ha!  I just came to the same conclusion. 😄 

I got this rule confused w/ `public_member_api_docs`.

Thanks!

pq:
> There's still a change needed for members in extensions in that the definition of isPrivate needs to change for them.

On closer look, I don't even think we need to change this since it's just checking the `SimpleIdentifier` and not (as maybe it should) take the parent's visibility into consideration. 🤔

bwilkerson:
Yeah, even with classes, a non-private method in a private class probably shouldn't be getting flagged. That said, if no one is complaining, then it might not be important enough to spend time fixing it. Still, it wouldn't hurt to open an issue; we might get user feedback to give us more information.

pq:
https://github.com/dart-lang/linter/issues/1689. 👍

Issue #1683, opened 2019-08-20
Linter Extension-Method Experiment Support


pq:
Tracking issue for `extension-method` support.

**Lints to vet/update.**

- [x] `always_declare_return_types`
- [ ] `always_specify_types` (#1706)
- [x] ~`annotate_overrides`~
- [x] `avoid_positional_boolean_parameters`
- [x] `avoid_renaming_method_parameters`
- [x] `avoid_returning_this` (assumes method parents are classes or mixins)
- [x] ~`avoid_setters_without_getters`~
- [x] `public_member_api_docs`
- [x] `prefer_constructors_over_static_methods` (assumes method parents are classes or mixins)
- [x] `slash_for_doc_comments`
- [x] `type_annotate_public_apis`
- [x] `use_setters_to_change_properties`
- [x] ~`void_checks`~
...

**New lints.**

- [ ] `camel_case_extensions`
  - [ ] style guide update: https://github.com/dart-lang/site-www/pull/1846
  - [x] lint experiment: #1685

pq:
After some more thinking, I'm less sure about whether `avoid_setters_without_getters` should lint extensions.

@bwilkerson: thoughts?

bwilkerson:
I would recommend that it not lint extensions unless we get a request for it from users.

pq:
Closing this in favor of targeted issues moving forward

PR #1691, opened 2019-08-21
public_member_api_docs extension support

pq:
See: #1683.

/cc @bwilkerson


lib/src/rules/public_member_api_docs.dart
```diff
@@ -283,6 +285,57 @@ class _Visitor extends SimpleAstVisitor {
     }
   }
 
+  @override
+  visitExtensionDeclaration(ExtensionDeclaration node) {
```
bwilkerson: This should also check (static) fields.

pq: They get checked in `visitFieldDeclaration`.  (I've added a testcase too.)

PR #1695, opened 2019-08-22
make use_setters_to_change_properties extension-aware

pq:
See: #1683 

/cc @bwilkerson


test/rules/experiments/use_setters_to_change_properties/use_setters_to_change_properties.dart
```diff
@@ -0,0 +1,53 @@
+// Copyright (c) 2017, the Dart project authors. Please see the AUTHORS file
```
bwilkerson: nit: year This appears to be a new file, but if it isn't then the year might be correct.

pq: It's a copy of an existing file that will get (re)merged when this is no longer an experiment.  I've opted to leave copyrights as-is but we can certainly change that if desired!

bwilkerson: That sounds like the right choice to me.

PR #1697, opened 2019-08-23
make type utils extension-safe

pq:
Protects utilities from calls against methods defined on extensions.

See: #1683 

/cc @bwilkerson @scheglov

### DIFF
--- a/lib/src/util/dart_type_utilities.dart
+++ b/lib/src/util/dart_type_utilities.dart
@@ -189,10 +189,9 @@ class DartTypeUtilities {
       expression?.unParenthesized is NullLiteral;
 
   static PropertyAccessorElement lookUpGetter(MethodDeclaration node) {
-    final parent = node.parent;
-    if (parent is ClassOrMixinDeclaration) {
-      return parent.declaredElement
-          .lookUpGetter(node.name.name, node.declaredElement.library);
+    final parent = node.declaredElement.enclosingElement;
+    if (parent is ClassElement) {
+      return parent.lookUpGetter(node.name.name, node.declaredElement.library);
     }
     // todo (pq): consider extensions
     return null;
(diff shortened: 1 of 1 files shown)


lib/src/util/dart_type_utilities.dart
```diff
@@ -188,30 +188,47 @@ class DartTypeUtilities {
   static bool isNullLiteral(Expression expression) =>
       expression?.unParenthesized is NullLiteral;
 
-  static PropertyAccessorElement lookUpGetter(MethodDeclaration node) =>
-      (node.parent as ClassOrMixinDeclaration)
-          .declaredElement
+  static PropertyAccessorElement lookUpGetter(MethodDeclaration node) {
+    final parent = node.parent;
+    if (parent is ClassOrMixinDeclaration) {
```
scheglov: Given that we ask `declaredElement`, it might be better to use `node.declaredElement.enclosingElement` and then check if it is `ClassElement` or `ExtensionElement`.

pq: Good idea. 👍

lib/src/util/dart_type_utilities.dart
```diff
@@ -188,43 +188,63 @@ class DartTypeUtilities {
...
-  static PropertyAccessorElement lookUpGetter(MethodDeclaration node) =>
-      (node.parent as ClassOrMixinDeclaration)
-          .declaredElement
-          .lookUpGetter(node.name.name, node.declaredElement.library);
+  static PropertyAccessorElement lookUpGetter(MethodDeclaration node) {
+    final parent = node.declaredElement.enclosingElement;
+    if (parent is ClassElement) {
+      return parent.lookUpGetter(node.name.name, node.declaredElement.library);
+    }
+    // todo (pq): consider extensions
```
bwilkerson: It should be easy to add an else-if to handle `ExtensionElement`.

lib/src/util/dart_type_utilities.dart
```diff
@@ -188,43 +188,63 @@ class DartTypeUtilities {
...
-          .declaredElement
-          .lookUpInheritedConcreteMethod(
-              node.name.name, node.declaredElement.library);
+      MethodDeclaration node) {
+    final parent = node.declaredElement.enclosingElement;
+    if (parent is ClassElement) {
+      return parent.lookUpInheritedConcreteGetter(
+          node.name.name, node.declaredElement.library);
+    }
+    // todo (pq): consider extensions
```
bwilkerson: Extensions don't inherit anything, so returning `null` seems like the right behavior here. I think you can remove the todo from the "inherited" versions of these methods.

PR #1705, opened 2019-08-30
Use InheritanceManager3, and take the existing instance from LinterContext.

scheglov:


scheglov:
Well, potentially reusing the instance makes it faster.
But I have not done any benchmarking for linter.

pq:
Cool.  Thanks!

PR #1723, opened 2019-09-20
Make await_only_futures mention the incorrectly-awaited type.

davidmorgan:
On behalf of a googler without a github account :)

The more detailed message is useful for us in cleaning up google3 because we need to fix awaiting a method that returns `void` differently from everything else. That can actually have an effect, and we then need to check if it was intended that the method return `Future<void>`.

What do you think, please?

### DIFF
--- a/lib/src/rules/await_only_futures.dart
+++ b/lib/src/rules/await_only_futures.dart
@@ -47,6 +47,9 @@ class AwaitOnlyFutures extends LintRule implements NodeLintRule {
 }
 
 class _Visitor extends SimpleAstVisitor<void> {
+  static const LintCode _errorCode = LintCode('await_only_futures',
+      "'await' applied to '{0}', which is not a 'Future'.");
+
   final LintRule rule;
 
   _Visitor(this.rule);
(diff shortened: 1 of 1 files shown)


lib/src/rules/await_only_futures.dart
```diff
@@ -62,7 +62,9 @@ class _Visitor extends SimpleAstVisitor<void> {
         DartTypeUtilities.extendsClass(type, 'Future', 'dart.async') ||
         DartTypeUtilities.implementsInterface(type, 'Future', 'dart.async') ||
         DartTypeUtilities.isClass(type, 'FutureOr', 'dart.async'))) {
-      rule.reportLintForToken(node.awaitKeyword);
+      rule.reportLintForToken(node.awaitKeyword,
+          errorCode: LintCode('await_only_futures',
```
bwilkerson: I'd make this a `const` invocation (and probably assign it to a static field).

lib/src/rules/await_only_futures.dart
```diff
@@ -62,7 +62,9 @@ class _Visitor extends SimpleAstVisitor<void> {
...
+              'Await applied to $type, which is not a Future'));
```
bwilkerson: Minor changes to match the conventions we have for diagnostic messages (specifically, quoting snippets of Dart code):

"'await' applied to '$type', which is not a 'Future'."

or, if you assign the lint code to a static field (and use the named argument `arguments` to pass in the type when invoking `reportLintForToken`), then it would be:

"'await' applied to '{0}', which is not a 'Future'."

lib/src/rules/await_only_futures.dart
davidmorgan: Done.

lib/src/rules/await_only_futures.dart
davidmorgan: Done.

davidmorgan:
Done, thanks!

PR #1727, opened 2019-09-20
Upgrade to analyzer 0.38.3 and fix deprecation warnings.

scheglov:
https://github.com/dart-lang/linter/pull/1725 is in a strange state where it is blocked, because Travis runs it with analyzer `0.38.3`, but the PR decided to not upgrade to `0.38.3` because it itself does not have to. So, here is a separate PR to upgrade and fix deprecations.

### DIFF
--- a/lib/src/rules/prefer_is_empty.dart
+++ b/lib/src/rules/prefer_is_empty.dart
@@ -8,6 +8,7 @@ import 'package:analyzer/dart/ast/visitor.dart';
 import 'package:analyzer/dart/element/element.dart';
 import 'package:analyzer/dart/element/type.dart';
 import 'package:linter/src/analyzer.dart';
+import 'package:linter/src/util/dart_type_utilities.dart';
 
 const alwaysFalse = 'Always false because length is always greater or equal 0.';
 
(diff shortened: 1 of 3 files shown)


lib/src/rules/prefer_is_empty.dart
```diff
@@ -111,13 +112,8 @@ class _Visitor extends SimpleAstVisitor<void> {
...
     // Should be subtype of Iterable, Map or String.
-    TypeProvider typeProvider = context.typeProvider;
-    TypeSystem typeSystem = context.typeSystem;
-
-    if (typeSystem.mostSpecificTypeArgument(type, typeProvider.iterableType) ==
-            null &&
-        typeSystem.mostSpecificTypeArgument(type, typeProvider.mapType) ==
-            null &&
+    if (!DartTypeUtilities.implementsInterface(type, 'Iterable', 'dart.core') &&
+        !DartTypeUtilities.implementsInterface(type, 'Map', 'dart.core') &&
```
pq: I thought using `typeSystem` was the preferred way?

scheglov: In general - yes, this specific API - no.

It should be something like (currently non-API `InterfaceTypeImpl`) `InterfaceType asInstanceOf(ClassElement targetElement)`. Passing raw `Map` type is weird, we should use `ClassElement`.

pq: Thanks!

PR #1729, opened 2019-09-21
get resource provider from session

pq:
I assume this is better than `PhysicalResourceProvider.INSTANCE`?

/cc @scheglov @bwilkerson


pq:
Thanks!

PR #1733, opened 2019-09-23
New lint avoid_async_from_sync

### LINKED ISSUE #836: unawaited_futures doesn't work when method/function is not marked `async`
```dart
Future<Null> sendInvitations() async { ... }
```
```dart
String doSomething() {
    inviteController
      ..sendInvitations() // <<<==== would expect a hint
      ..updateSearchTerm('');
}
```
but I don't get one.

Changing it to

```dart
Future<String> doSomething() async {
```
makes the lint show up.

JankoLancer:
# Description

This lint is necessary when we want to know places where we call async functions from non-async functions. This can be a lurking bug. In my case, I never actually want to do this. I want to make the enclosing function async, await the call, and repeat as necessary.

Fixes #836

### DIFF
--- a/lib/src/rules/avoid_async_from_sync.dart
+++ b/lib/src/rules/avoid_async_from_sync.dart
@@ -66,7 +66,7 @@ class _Visitor extends SimpleAstVisitor<void> {
     for (final expr in node.cascadeSections) {
       if (expr.staticType?.isDartAsyncFuture == true &&
           _isEnclosedInSyncFunctionBody(expr) &&
-          !(expr is AssignmentExpression)) {
+          expr is! AssignmentExpression) {
         rule.reportLint(expr);
       }
     }
--- a/lib/src/rules/avoid_async_from_sync.dart
+++ b/lib/src/rules/avoid_async_from_sync.dart
@@ -41,10 +41,15 @@ void main() {
 
 class AvoidAsyncFromSync extends LintRule implements NodeLintRule {
   AvoidAsyncFromSync()
-      : super(name: 'avoid_async_from_sync', description: _desc, details: _details, group: Group.style);
+      : super(
+            name: 'avoid_async_from_sync',
+            description: _desc,
+            details: _details,
+            group: Group.style);
 
   @override
-  void registerNodeProcessors(NodeLintRegistry registry, LinterContext context) {
+  void registerNodeProcessors(
+      NodeLintRegistry registry, LinterContext context) {
     final visitor = _Visitor(this);
     registry.addExpressionStatement(this, visitor);
     registry.addCascadeExpression(this, visitor);
@@ -75,7 +80,8 @@ class _Visitor extends SimpleAstVisitor<void> {
     var type = expr?.staticType;
     if (type?.isDartAsyncFuture == true) {
       // Ignore a couple of special known cases.
-      if (_isFutureDelayedInstanceCreationWithComputation(expr) || _isMapPutIfAbsentInvocation(expr)) {
+      if (_isFutureDelayedInstanceCreationWithComputation(expr) ||
+          _isMapPutIfAbsentInvocation(expr)) {
         return;
       }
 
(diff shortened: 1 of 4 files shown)


JankoLancer:
@googlebot I signed it!

lib/src/rules/avoid_async_from_sync.dart
```diff
@@ -0,0 +1,117 @@
...
+  final LintRule rule;
+
+  _Visitor(this.rule);
+
+  @override
+  void visitCascadeExpression(CascadeExpression node) {
+    for (final expr in node.cascadeSections) {
+      if (expr.staticType?.isDartAsyncFuture == true &&
+          _isEnclosedInSyncFunctionBody(expr) &&
+          !(expr is AssignmentExpression)) {
```
bwilkerson: Stylistically, we prefer `expr is! AssignmentExpression`.

test/rules/avoid_async_from_sync.dart
```diff
@@ -0,0 +1,81 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+// test w/ `pub run test -N avoid_async_from_sync`
+
+import 'dart:async';
+
+Future fut() => null;
+
+foo1() {
+  fut(); //LINT
```
bwilkerson: Consider adding a test for a getter that returns a `Future`.

davidmorgan:
Thanks for looking at this!

There's one other case I can think of that we might want to allow, which is when there is a chain of `then`. The last thing in the chain returns a `Future`, but the fact that there is a chain probably means it's being handled correctly already.

e.g. we probably want to allow

```
File('/tmp/a').readAsString().then(print);
```

A stricter variant would be to only allow this as the _last_ thing in the method.

What do you think?

davidmorgan:
The difference with `unawaited_futures` is that in an `async` method you never need to use `then`; you can always use `await`. In a sync method, it's valid to use `then`, and should work well.

Looking through a random sample of ten uses of `then` in Google-internal Dart code, I find:

5 are the last thing in a sync method
3 are somewhere else in a sync method
2 are in an async method

We see approximately 10% as much usage of `then` as of `async`. So while it's relatively rare, there is a lot of usage in terms of absolute numbers, and it would be expensive to tag it all `unawaited` to satisfy this lint.

My feeling is that it's probably best to ignore any Future resulting from a `then` call, regardless of where it is in the method.

JankoLancer:
Hi, sorry for delay.
Can you help me to resolve this and finish PR, it is hard for me to test on real project this lint, cause I can't run it in IDE.

bwilkerson:
As for testing, I believe you should be able to use the command-line interface in the linter package (linter/bin/linter.dart) to run your rule over some existing package. It's not quite as convenient as viewing the results in an IDE, but it should work.

Testing the rule in an IDE would involve building an SDK that has been updated to point to your local copy of the linter package.

davidmorgan:
Brian, any thoughts on use of `then`? I'm worried that this lint as is will push people towards something we don't want: writing `unawaited` everywhere in synchronous methods.

If we do want to go in that direction, it almost seems like we should be telling people to just not use synchronous methods with `then`, rather than using `then`+`unawaited`. WDYT, please?

bwilkerson:
Both `await` and `then` serve similar purposes: to handle the result of an asynchronous operation. The biggest difference is that `then` returns a `Future` so that they can be chained.

I believe that the general principle we decided on when designing 'unawaited_futures' was to require either one or the other unless `unawaited` was invoked, and I think we should do the same thing here.

Or, stated slightly differently, I think that any synchronous method that uses `then` should have the same lints generated for it by this rule as it would have from 'unawaited_futures' if we added `async` to the body. I think that any differences between these two lints is likely to be at least confusing, if not wrong.

bwilkerson:
I was confused by your response, so I tried an example. The confusion, of course, was on my end. I didn't realize, or had forgotten, that we don't allow `then` chains in asynchronous functions.

But I agree that we have to allow `then` chains in synchronous functions. There's no other way to handle the value returned from the initial future, and there's no way to terminate the chain, so I think we have to assume that if there's at least one `then` then the user knows what they're doing.

I don't have a strong opinion about whether to allow `then` anywhere or only at the end, but in the absence of a strong argument for restricting it I'd lean toward allowing it anywhere.

davidmorgan:
Sounds good to me, thanks.

edwardaux:
Not sure whether to comment in here or in #836, but anyway... just wanted to see if there's any chance of movement on this one? I just spent two hours chasing down a defect that would have been prevented by this lint.

I totally understand that including it in the core or recommended files may not appropriate due to the false positives it would introduce into existing codebases, however, having this an available lint for people to opt into if they like seems like a net win. 

I mean, if I go to the trouble of enabling this lint in my codebase, then I'm 100% opting in to annotating any false positives in my project. For me personally, the benefit is definitely worth it.

I'm more than happy to deal with the conflicts in this branch if I can get some consensus that moving forward is possible.

srawlins:
Changes were requested in 2019 which haven't been addressed. I'm closing this just to tidy the issue tracker. If you would like to update the branch with main, and address comments, then please re-open with changes.

PR #1734, opened 2019-09-23
Lints that warn about compile-time errors of invalid FFI usage

sjindel-google:
We have a number of custom compile-time errors implemented in the CFE's VM target which halt compilation if the FFI is used incorrectly.

We would like to surface these errors earlier through an Analyzer lint.

This PR replicates some of the FFI checks. I ran into some issues:

1. How can we make this lint on-by-default? It should be on-by-default because the custom compile-time errors that it warns about cannot be disabled by the user.
2. How can I check whether an Expression in the source is constant?
3. To reproduce some of the type tests in the CFE, I need to construct a DartType which is parallel to a DartType containing FFI phantom types. How can I access the ClassElements which are known to be declared in the core (or FFI) libraries?

/cc @mkustermann @dcharkes @mraleph

### DIFF
--- a/example/all.yaml
+++ b/example/all.yaml
@@ -56,6 +56,7 @@ linter:
     - empty_catches
     - empty_constructor_bodies
     - empty_statements
+    - ffi
     - file_names
     - flutter_style_todos
     - hash_and_equals
--- a/lib/src/rules.dart
+++ b/lib/src/rules.dart
@@ -57,6 +57,7 @@ import 'package:linter/src/rules/directives_ordering.dart';
 import 'package:linter/src/rules/empty_catches.dart';
 import 'package:linter/src/rules/empty_constructor_bodies.dart';
 import 'package:linter/src/rules/empty_statements.dart';
+import 'package:linter/src/rules/ffi.dart';
 import 'package:linter/src/rules/file_names.dart';
 import 'package:linter/src/rules/flutter_style_todos.dart';
 import 'package:linter/src/rules/hash_and_equals.dart';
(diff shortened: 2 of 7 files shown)


bwilkerson:
> How can we make this lint on-by-default? It should be on-by-default because the custom compile-time errors that it warns about cannot be disabled by the user.

I agree that these should be enabled by default.

In the analyzer, we made a distinction between lints, which are disabled by default, and hints, which are enabled by default. I believe that we should implement these diagnostics as hints in order for them to be enabled by default.

> How can I check whether an Expression in the source is constant?

By using the getter `ExpressionImpl.inConstantContext`. I think we should make this part of the public API for `Expression`.

> How can I access the ClassElements which are known to be declared in the core (or FFI) libraries?

Good question. I don't remember any public API for getting accesses to the core libraries directly.

As an ugly work around, assuming you have access to a `ClassElement` for something else, you can walk the superclass chain until you get to the class element for `Object`, then ask it for its `library`, which will be the element representing `dart:core`.

Alternatively, you could use the `AnalysisSession` to access the `analysisContext` (a `DriverBasedAnalysisContext`), use that to access the `driver`, the the `sourceFactory` and then `dartSdk`, from which you can access all of the libraries in the SDK.

Of course, none of these are quite as ugly if the diagnostics are implemented as hints because hints are implemented in the 'analyzer' package so they are "allowed" to access internal implementation details.

sjindel-google:
@bwilkerson Thanks! Is there an example of a hint in `pkg/analyzer` I can imitate?

bwilkerson:
We don't implement hints in as fine-grained a way as we do in the linter. Instead we have classes (usually visitors) that find groups of related hints all at the same time. A couple of examples are `BestPracticesVerifier` and `DeadCodeVerifier`. I'd propose a class named `FfiVerifier` (though you might want to break the capitalization convention; that's kind of ugly).

The visitors are all called from `LibraryAnalyzer._computeHints`.

srawlins:

Issue #214, opened 2016-04-01
Use relative paths within lib


donny-dont:
Given the following directory structure

```
lib
  src
    b_1.dart
  a.dart
  b.dart
  c.dart
```

For c.dart

BAD

```
import 'package:foo/b.dart;
```

GOOD

```
import 'b.dart';
```

For b_1.dart

BAD

```
import 'package:foo/a.dart';
```

GOOD

```
import '../a.dart';
```

The DDC seems to mess up the output if you follow the BAD options so a lint to enforce this should be a good thing.

donny-dont:
For test, example, tool, whatever else in the dir using the package signature is the way to go. Not sure if that as well needs to be enforced.

bwilkerson:
@vsmenon Sounds like there might be a bug in DDC. It's actually better (read necessary) to use the same style of import everywhere, so if DDC is having trouble using this style of import we should look at fixing it.

donny-dont:
From talking to @kwalrath it sounded like this is also a style issue @bwilkerson. Though @munificent could also probably chime in.

vsmenon:
Yes, that looks like a bug in DDC.

PR #1738, opened 2019-09-27
implement a prefer_relative_imports lint

devoncarew:
# Description

- implement a `prefer_relative_imports` lint. This lint will trigger if you have a package: self-reference from within code in the `lib/` directory.

Fixes # (issue)
- fixes https://github.com/dart-lang/linter/issues/1101
- fixes https://github.com/dart-lang/linter/issues/214

@pq


lib/src/rules/prefer_relative_imports.dart
```diff
@@ -0,0 +1,129 @@
...
+When mixing relative and absolute imports it's possible to create confusion
+where the same member gets imported in two different ways. One way to avoid
+that is to ensure you consistently use relative imports for files withing the
+`lib/` directory.
+
+**GOOD:**
+
+```
+import 'bar.dart';
+...
```
bwilkerson: Personally I'd remove the "..." (here and below) because I don't think it adds to the example. Your call.

lib/src/rules/prefer_relative_imports.dart
```diff
@@ -0,0 +1,129 @@
...
+      if (!isPackage(uri)) return false;
+    } on FormatException catch (_) {
+      return false;
+    }
+
+    // Is the package: import referencing the current package?
+    var segments = uri.pathSegments;
+    if (segments.isEmpty) return false;
+
+    // ignore: prefer_conditional_assignment
```
bwilkerson: This seems like a place where this lint ought not to be generating a diagnostic. We should look at fixing the lint.

devoncarew: This might be from an earlier version of the impl of the lint - I'll remove if this ignore is no longer relevant.

lib/src/rules.dart
```diff
@@ -165,11 +166,10 @@ void registerLintRules() {
     ..register(AlwaysSpecifyTypes())
     ..register(AnnotateOverrides())
     ..register(AvoidAnnotatingWithDynamic())
-    ..register(AvoidBoolLiteralsInConditionalExpressions())
-    ..register(AvoidTypesOnClosureParameters())
     ..register(AvoidAs())
-    ..register(AvoidCatchingErrors())
+    ..register(AvoidBoolLiteralsInConditionalExpressions())
```
pq: 💎 

thanks for (re)sorting!

lib/src/rules/prefer_relative_imports.dart
```diff
@@ -0,0 +1,129 @@
...
+        // ignore: avoid_catches_without_on_clauses
+      } catch (_) {}
+      parsedPubspec = _parseYaml(content);
+    }
+
+    return parsedPubspec['name'] == segments[0];
+  }
+
+  @override
+  visitCompilationUnit(CompilationUnit node) {
```
pq: consider adding a `void` return type here (or removing the one declared below)?

lib/src/rules/prefer_relative_imports.dart
```diff
@@ -0,0 +1,129 @@
...
+    registry.addImportDirective(this, visitor);
+  }
+}
+
+class _Visitor extends SimpleAstVisitor<void> {
+  final PreferRelativeImports rule;
+
+  bool isInLibFolder;
+  File pubspecFile;
+  Map parsedPubspec;
```
pq: In the future I guess we might consider storing this in `ast.dart` or somewhere so that other lints can take advantage of it?  Or: do you envision this something we will ultimately get from the analysis context? 

@bwilkerson: thoughts on that?

bwilkerson: I agree that this kind of support should be shared. I think it should be available from the analyzer (though I haven't yet thought about a concrete API for it) because we're likely to want it for hints as well.

pq: @bwilkerson: if you're keeping a running "technical debt" list, could you consider adding it there?   (Or suggesting how I can best capture the idea?)

Thanks!

bwilkerson: I don't really have such a list. I suppose an issue is probably best.

lib/src/rules/prefer_relative_imports.dart
devoncarew: done

lib/src/rules/prefer_relative_imports.dart
pq: 👍 https://github.com/dart-lang/sdk/issues/38616

PR #1750, opened 2019-10-06
Added new lint rule - avoid leading underscore for non private identifiers

### LINKED ISSUE #1676: New rule for using a leading underscore for identifiers that aren't private
As per [Effective Dart](https://dart.dev/guides/language/effective-dart/style#dont-use-a-leading-underscore-for-identifiers-that-arent-private):

> There is no concept of “private” for local variables, parameters, or library prefixes. When one of those has a name that starts with an underscore, it sends a confusing signal to the reader. To avoid that, don’t use leading underscores in those names.

```dart
// BAD

void fn(String _s) {
  int _n = _s.length;
}
```

jawahars16:
As per Effective Dart:

> There is no concept of “private” for local variables, parameters, or library prefixes. When one of those has a name that starts with an underscore, it sends a confusing signal to the reader. To avoid that, don’t use leading underscores in those names.

A new lint rule has been added to enforce no leading underscores for non private fields.

Fixes #1676

### DIFF
--- a/lib/src/rules/avoid_leading_underscore_for_non_private_identifiers.dart
+++ b/lib/src/rules/avoid_leading_underscore_for_non_private_identifiers.dart
@@ -1,4 +1,4 @@
-// Copyright (c) 2015, the Dart project authors. Please see the AUTHORS file
+// Copyright (c) 2019, the Dart project authors. Please see the AUTHORS file
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
@@ -13,13 +13,23 @@ const _details = r'''
 
 **DON’T** use a leading underscore for identifiers that aren’t private.
 
-Dart uses a leading underscore in an identifier to mark members and top-level declarations as private. This trains users to associate a leading underscore with one of those kinds of declarations. They see “_” and think “private”.
+Dart uses a leading underscore in an identifier to mark members and 
+top-level declarations as private. This trains users to associate a 
+leading underscore with one of those kinds of declarations. 
+They see “_” and think “private”.
 
-There is no concept of “private” for local variables, parameters, or library prefixes. When one of those has a name that starts with an underscore, it sends a confusing signal to the reader. To avoid that, don’t use leading underscores in those names.
+There is no concept of “private” for local variables, parameters, 
+or library prefixes. When one of those has a name that starts with 
+an underscore, it sends a confusing signal to the reader. To avoid that,
+don’t use leading underscores in those names.
 
-**Exception**: An unused parameter can be named _, __, ___, etc. This happens in things like callbacks where you are passed a value but you don’t need to use it. Giving it a name that consists solely of underscores is the idiomatic way to indicate the value isn’t used.
+**Exception**: An unused parameter can be named _, __, ___, etc. 
+This happens in things like callbacks where you are passed a value 
+but you don’t need to use it. Giving it a name that consists solely 
+of underscores is the idiomatic way to indicate the value isn’t used.
 
 **BAD**
+
 ```
 print(string _name) {
   int _size = _name.length;
@@ -27,10 +37,17 @@ print(string _name) {
 ```
 
 **GOOD:**
+
 ```
 print(string name) {
   int size = name.length;
 }
+```
+
+**OK:**
+
+```
+[1,2,3].map((_) => print('Hello'));
 ```
 
 ''';
@@ -51,6 +68,8 @@ class AvoidLeadingUnderscoreForNonPrivateIdentifiers extends LintRule
     registry.addFormalParameterList(this, visitor);
     registry.addVariableDeclarationStatement(this, visitor);
     registry.addImportDirective(this, visitor);
+    registry.addDeclaredIdentifier(this, visitor);
+    registry.addCatchClause(this, visitor);
   }
 }
 
@@ -64,20 +83,17 @@ class _Visitor extends SimpleAstVisitor<void> {
       return;
     }
 
-    if (isJustUnderscoresOK && isJustUnderscores(id.name)) {
-      return;
-    }
-
-    if (hasLeadingUnderscores(id.name)) {
+    if (hasLeadingUnderscore(id.name)) {
+      if (isJustUnderscoresOK && isJustUnderscores(id.name)) {
+        return;
+      }
       rule.reportLint(id);
     }
   }
 
   @override
   void visitImportDirective(ImportDirective node) {
-    if (node.prefix != null) {
-      checkIdentifier(node.prefix);
-    }
+    checkIdentifier(node.prefix);
   }
 
   @override
@@ -88,9 +104,20 @@ class _Visitor extends SimpleAstVisitor<void> {
   @override
   void visitFormalParameterList(FormalParameterList node) {
     node.parameters.forEach((FormalParameter p) {
-      if (p is! FieldFormalParameter) {
+      if (p is! FieldFormalParameter && !p.isOptionalNamed) {
         checkIdentifier(p.identifier, isJustUnderscoresOK: true);
       }
     });
+  }
+
+  @override
+  void visitDeclaredIdentifier(DeclaredIdentifier node) {
+    checkIdentifier(node.identifier);
+  }
+
+  @override
+  void visitCatchClause(CatchClause node) {
+    checkIdentifier(node.exceptionParameter);
+    checkIdentifier(node.stackTraceParameter);
   }
 }
--- a/lib/src/rules/avoid_leading_underscore_for_non_private_identifiers.dart
+++ b/lib/src/rules/avoid_leading_underscore_for_non_private_identifiers.dart
@@ -1,0 +1,96 @@
+// Copyright (c) 2015, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:linter/src/analyzer.dart';
+import 'package:linter/src/utils.dart';
+
+const _desc = r'Avoid leading underscore for identifiers that aren’t private.';
+
+const _details = r'''
+
+**DON’T** use a leading underscore for identifiers that aren’t private.
+
+Dart uses a leading underscore in an identifier to mark members and top-level declarations as private. This trains users to associate a leading underscore with one of those kinds of declarations. They see “_” and think “private”.
+
+There is no concept of “private” for local variables, parameters, or library prefixes. When one of those has a name that starts with an underscore, it sends a confusing signal to the reader. To avoid that, don’t use leading underscores in those names.
+
+**Exception**: An unused parameter can be named _, __, ___, etc. This happens in things like callbacks where you are passed a value but you don’t need to use it. Giving it a name that consists solely of underscores is the idiomatic way to indicate the value isn’t used.
+
+**BAD**
+```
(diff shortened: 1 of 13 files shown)


lib/src/rules/avoid_leading_underscore_for_non_private_identifiers.dart
```diff
@@ -0,0 +1,96 @@
+// Copyright (c) 2015, the Dart project authors. Please see the AUTHORS file
```
pq: nit: year

lib/src/rules/avoid_leading_underscore_for_non_private_identifiers.dart
```diff
@@ -0,0 +1,96 @@
...
+}
+```
+
+**GOOD:**
+```
+print(string name) {
+  int size = name.length;
+}
+```
+
```
pq: maybe consider adding an "OK" example for the unused parameter case?

test/rules/avoid_leading_underscore_for_non_private_identifiers.dart
```diff
@@ -0,0 +1,54 @@
+// Copyright (c) 2016, the Dart project authors. Please see the AUTHORS file
```
pq: nit: year

pq:
Also,  fyi @kwalrath for updating doc references (when this lands).

lib/src/rules/avoid_leading_underscore_for_non_private_identifiers.dart
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    if (hasLeadingUnderscores(id.name)) {
+      rule.reportLint(id);
+    }
+  }
+
+  @override
+  void visitImportDirective(ImportDirective node) {
+    if (node.prefix != null) {
```
bwilkerson: Given that `checkIdentifier` already checks for `null`, this check isn't needed. I'd remove it.

lib/src/rules/avoid_leading_underscore_for_non_private_identifiers.dart
```diff
@@ -0,0 +1,96 @@
...
+
+  @override
+  void visitVariableDeclarationStatement(VariableDeclarationStatement node) {
+    node.variables.variables.forEach((v) => {checkIdentifier(v.name)});
+  }
+
+  @override
+  void visitFormalParameterList(FormalParameterList node) {
+    node.parameters.forEach((FormalParameter p) {
+      if (p is! FieldFormalParameter) {
```
bwilkerson: This should not check optional named parameters because there's already a diagnostic for those and we don't want the lint to duplicate that reporting.

lib/src/utils.dart
```diff
@@ -34,6 +36,9 @@ bool isIdentifier(String name) => _identifier.hasMatch(name);
 /// Returns `true` of the given [name] is composed only of `_`s.
 bool isJustUnderscores(String name) => _underscores.hasMatch(name);
 
+/// Returns `true` if the given [name] has leading `_`s.
+bool hasLeadingUnderscores(String name) => _leadingUnderscores.hasMatch(name);
```
bwilkerson: It's only the first underscore that matters, so it would be more efficient to test `name.startsWith('_')` than to use a regular expression.

lib/src/rules/avoid_leading_underscore_for_non_private_identifiers.dart
```diff
@@ -0,0 +1,96 @@
...
+            details: _details,
+            group: Group.style);
+
+  @override
+  void registerNodeProcessors(
+      NodeLintRegistry registry, LinterContext context) {
+    final visitor = _Visitor(this);
+    registry.addFormalParameterList(this, visitor);
+    registry.addVariableDeclarationStatement(this, visitor);
+    registry.addImportDirective(this, visitor);
```
bwilkerson: I believe this also needs to visit `DeclaredIdentifier` in order to detect cases like `for (var _x in y)` and `CatchClause` to detect cases like `} catch (_e) {`. And we should add tests for all such cases.

lib/src/rules/avoid_leading_underscore_for_non_private_identifiers.dart
```diff
@@ -0,0 +1,96 @@
...
+  final LintRule rule;
+
+  _Visitor(this.rule);
+
+  checkIdentifier(SimpleIdentifier id, {bool isJustUnderscoresOK = false}) {
+    if (id == null) {
+      return;
+    }
+
+    if (isJustUnderscoresOK && isJustUnderscores(id.name)) {
```
bwilkerson: Given that most identifiers don't start with an underscore, it will be more performant to check for a single initial underscore first and only check for the only underscores case after that.

lib/src/rules/avoid_leading_underscore_for_non_private_identifiers.dart
```diff
@@ -0,0 +1,96 @@
+// Copyright (c) 2015, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:linter/src/analyzer.dart';
+import 'package:linter/src/utils.dart';
+
+const _desc = r'Avoid leading underscore for identifiers that aren’t private.';
```
bwilkerson: Both of these are more food for thought than a request for a change.

It isn't so much that they "aren't" private, but that they "can't be" private (or conversely that they can't be public). Maybe a better way to say it is "local" identifiers.

Along those lines, the name of the lint is kind of long. Some of the other lint rules we have that are related to naming conventions have names like "file_names", "library_names", "package_names", "constant_identifier_names", and "non_constant_identifier_names". Perhaps for consistency something like "local_variable_names"? The disadvantage of this is that "constant_identifier_names" and "non_constant_identifier_names" already check local variables (but not for leading underscores), so it might be somewhat confusing. No concrete suggestion, just a notion that a shorter name would be better.

lib/src/rules/avoid_leading_underscore_for_non_private_identifiers.dart
```diff
@@ -0,0 +1,96 @@
...
+
+const _details = r'''
+
+**DON’T** use a leading underscore for identifiers that aren’t private.
+
+Dart uses a leading underscore in an identifier to mark members and top-level declarations as private. This trains users to associate a leading underscore with one of those kinds of declarations. They see “_” and think “private”.
```
bwilkerson: nit: these lines should be wrapped to fit within 80 columns.

lib/src/rules/avoid_leading_underscore_for_non_private_identifiers.dart
jawahars16: Updated copyright year.

lib/src/rules/avoid_leading_underscore_for_non_private_identifiers.dart
jawahars16: Added OK case for unused callback parameter.

test/rules/avoid_leading_underscore_for_non_private_identifiers.dart
jawahars16: Updated copyright year.

lib/src/rules/avoid_leading_underscore_for_non_private_identifiers.dart
jawahars16: Removed unwanted NULL check.

lib/src/rules/avoid_leading_underscore_for_non_private_identifiers.dart
jawahars16: Skipped optional named parameters in this check.

lib/src/utils.dart
jawahars16: Removed regex and updated code.

lib/src/rules/avoid_leading_underscore_for_non_private_identifiers.dart
jawahars16: Added check for `for` loop variables and `catch` parameters

lib/src/rules/avoid_leading_underscore_for_non_private_identifiers.dart
jawahars16: Updated code.

lib/src/rules/avoid_leading_underscore_for_non_private_identifiers.dart
jawahars16: I also had the same confusion. Initially I wanted to name like "avoid_leading_underscore_for_local_identifiers". But this rule also check for library prefixes, which I am not sure whether categorised as local identifier.

How about we use this rule only for local identifiers and move library prefix check to existing "library_prefixes" rule? Or shall we create new one like "avoid_leading_underscore_for_library_prefixes".

lib/src/rules/avoid_leading_underscore_for_non_private_identifiers.dart
jawahars16: Wrapped lines to fit within 80 columns.

lib/src/rules/avoid_leading_underscore_for_non_private_identifiers.dart
bwilkerson: > But this rule also check for library prefixes, which I am not sure whether categorised as local identifier.

No, we generally don't consider a library prefix to be a local variable. We generally use "local" to mean local to a function / method.

Perhaps we should flip the sense of the name and call it 'private_names' and describe it as a rule to ensure that underscore is only used for identifiers that would be public if the underscore were absent.

> How about we ... move library prefix check to existing "library_prefixes" rule?

I considered that possibility. We could also move the local variable checking into "constant_identifier_names" and "non_constant_identifier_names".

The problem is that changing any of these three rules is potentially a breaking change. If we have users that have enabled those rules but currently have local variables with underscores, then when they upgrade their SDK they'll start seeing lints being reported that might cause their builds to stop working. We generally avoid making breaking changes like that (we can't always, such as for some bug fixes, but we try).

Similarly, if some users intentionally use underscores for some reason, then that makes these rules unusable for them.

I think overall it's better to create a new rule the way you did than to risk breaking existing users.

jawahars16: I realised that `non_private_identifiers` is not the right term to use here. And updating existing rules is also not a good idea. So I come up with two lint rules,

- avoid leading underscore for local identifiers
- avoid leading underscore for library prefixes

I updated the PR accordingly. Let me know if you have any suggestions.

pq: Thanks for the thoughtful follow-up!

I feel inclined to focus on avoiding leading underscores for local identifiers here and then investigate updating `library_prefixes` to check for underscores.  Before we do that though I'd want to do a bit of investigating to see if it's really what we want to do.  (For example, is this something that we do internally?  cc @srawlins who may know....)

@jawahars16: have you seen this naming in practice?

Rockvole:
I came to this repo to add a request for this lint rule and found this issue already existed.
Adding my 2c that this is a useful rule.

I spent ages in my flutter project trying to figure out why my "private" variable inside my build method was having no effect on the button text in DropdownButton when changed inside the setState() function.

A warning that this _ variable is local and not private would have given me a clue and saved me quite a bit of time.

pq:
Closing in favor of #3062.

@jawahars16: do reply if you'd like attribution in `AUTHORS`.

Thanks again for the contribution!

PR #1751, opened 2019-10-08
Fixed - avoid_print rule fails to catch tear-off usage

### LINKED ISSUE #1696: avoid_print fails to catch tear-off usage
Example:

```dart
Future.value('hello').then(print);
```

If the `avoid_print` rule is on, no lint warning is raised.

jawahars16:
The lint rule `avoid_print` is not catching the scenario when `print` function passed as an argument. This PR includes code that added additional check to validate function arguments for `print`.

Fixes #1696

### DIFF
--- a/lib/src/rules/avoid_print.dart
+++ b/lib/src/rules/avoid_print.dart
@@ -46,8 +46,8 @@ class _Visitor extends SimpleAstVisitor {
     if (expression is SimpleIdentifier) {
       final Element element = expression.staticElement;
       if (element is FunctionElement &&
-          element?.name == 'print' &&
-          element?.library?.name == 'dart.core') {
+          element.name == 'print' &&
+          element.library.isDartCore) {
         rule.reportLint(expression);
       }
     }
--- a/lib/src/rules/avoid_print.dart
+++ b/lib/src/rules/avoid_print.dart
@@ -61,7 +61,7 @@ class _Visitor extends SimpleAstVisitor {
     if (node.methodName.name == 'print' && isDartCore(node)) {
       rule.reportLint(node.methodName);
     }
-    
+
     node.argumentList.arguments.forEach(_validateArgument);
   }
 }
--- a/lib/src/rules/avoid_print.dart
+++ b/lib/src/rules/avoid_print.dart
@@ -4,6 +4,7 @@
 
 import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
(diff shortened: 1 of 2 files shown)


lib/src/rules/avoid_print.dart
```diff
@@ -41,6 +42,17 @@ class _Visitor extends SimpleAstVisitor {
 
   _Visitor(this.rule);
 
+  _validateArgument(Expression expression) {
+    if (expression is SimpleIdentifier) {
+      final Element element = expression.staticElement;
+      if (element is FunctionElement &&
+          element?.name == 'print' &&
+          element?.library?.name == 'dart.core') {
```
pq: perhaps: `element.libary.isDartCore`?

lib/src/rules/avoid_print.dart
```diff
@@ -41,6 +42,17 @@ class _Visitor extends SimpleAstVisitor {
 
   _Visitor(this.rule);
 
+  _validateArgument(Expression expression) {
+    if (expression is SimpleIdentifier) {
+      final Element element = expression.staticElement;
+      if (element is FunctionElement &&
+          element?.name == 'print' &&
```
pq: since you just verified `element is FunctionElement` I think you can drop the `?`, simplifying this to:

    element.name == 'print'

pq:
Thanks a million for your contributions!

lib/src/rules/avoid_print.dart
jawahars16: Changed to `isDartCore`

lib/src/rules/avoid_print.dart
jawahars16: Removed unwanted null checks.

bwilkerson:
I have no idea why it's failing for your PR, but it's clearly a bug. I'll have a fix out in a minute.

bwilkerson:
PR out for review: https://github.com/dart-lang/linter/pull/1752

bwilkerson:
PR landed.

pq:
Restarting build...

pq:
@jawahars16: thank you!  I'd like to an attribution to `AUTHORS`,

https://github.com/dart-lang/linter/blob/master/AUTHORS

How would you like to appear?

jawahars16:
> @jawahars16: thank you! I'd like to an attribution to `AUTHORS`,
> 
> https://github.com/dart-lang/linter/blob/master/AUTHORS
> 
> How would you like to appear?

Thank you. 😄 

Jawahar <jawahars_16@live.in>

PR #1752, opened 2019-10-08
Fix an NPE in CamelCaseExtensions

bwilkerson:


pq:
Ah, this for unnamed extensions?

Maybe add a test case to `test/rules/camel_case_extensions.dart`?

```dart
extension on Object { }  //OK
```

bwilkerson:
Test added. Thanks!

PR #1753, opened 2019-10-08
flutter_html initial impl

pq:
See: https://github.com/dart-lang/sdk/issues/38691

An early jumping off point for conversation.

Things (at least) to resolve:

- [ ] is the web context heuristic right? and
- [ ] are the tests validating it?
- [ ] can/should we cache pubspecs?
- [ ] is flagging import directives sufficient?

🤔 

...

Docs could use word-smithing too.  Mainly just placeholding...

/cc @devoncarew @bwilkerson 

Any input very welcome!


example/all.yaml
```diff
@@ -57,6 +57,7 @@ linter:
     - empty_constructor_bodies
     - empty_statements
     - file_names
+    - flutter_html
```
devoncarew: `flutter_avoid_dart_html`? `flutter_warn_dart_html`? `flutter_use_of_html_libraries`?

This lint would be mostly about warning for use of `dart:html` (and related libraries) in a context where they won't be available (Flutter mobile).

lib/src/rules/flutter_html.dart
```diff
@@ -0,0 +1,108 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:linter/src/analyzer.dart';
+import 'package:linter/src/ast.dart';
+import 'package:yaml/yaml.dart';
+
+const _desc = r'Avoid using web-only libraries outside Flutter web packages';
```
devoncarew: s/packages/projects?

lib/src/rules/flutter_html.dart
```diff
@@ -0,0 +1,108 @@
...
+
+const _details = r'''Avoid using web packages, `dart:html`, `dart:js` and 
+`dart:js_util` in non-web Flutter packages.  These packages are not supported
```
devoncarew: `These libraries...`

lib/src/rules/flutter_html.dart
```diff
@@ -0,0 +1,108 @@
...
+    final pubspecFile = locatePubspecFile(node);
+    if (pubspecFile == null) {
+      isInFlutterApp = false;
+      return;
+    }
+
+    var parsedPubspec;
+    try {
+      final content = pubspecFile.readAsStringSync();
+      parsedPubspec = _parseYaml(content);
```
devoncarew: This is eagerly parsing the pubspec - should we wait until we need to validate the first import reference?

lib/src/rules/flutter_html.dart
```diff
@@ -0,0 +1,108 @@
...
+  bool isInFlutterApp = true;
+  bool isInFlutterWebContext = false;
+
+  final rule;
+
+  _Visitor(this.rule);
+
+  @override
+  void visitCompilationUnit(CompilationUnit node) {
+    // todo (pq): consider caching for library?
```
devoncarew: I'm not sure of the lifecycle of these visitors; should we reset `isInFlutterApp` to false at the start of this method, and only set it to true if we recognize the pubspec as using flutter?

example/all.yaml
pq: Awesome.  Thanks for the suggestions.  I went with the last suggestion but easily swayed...

lib/src/rules/flutter_html.dart
pq: Ah, yeah.  I'll make things a little lazier.  Thanks!

lib/src/rules/flutter_html.dart
pq: I *believe* we get a fresh visitor for each compilation unit...  Will check though.

lib/src/rules/flutter_html.dart
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    isInFlutterWebContext = pubspecFile.parent.getChild('web').exists ||
+        ((parsedPubspec['flutter'] ?? const {})['plugin'] ?? const {})['web'] !=
+            null;
+  }
+
+  @override
+  void visitImportDirective(ImportDirective node) {
+    if (!isInFlutterApp || isInFlutterWebContext) return;
```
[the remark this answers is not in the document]
pq: Right.  String checks are definitely cheaper than parsing pubspecs.  Good call!

devoncarew:
Looks like a CI issue:

```
Rule description for avoid_web_libraries_in_flutter should end with a '.'
```

lib/src/rules/avoid_web_libraries_in_flutter.dart
```diff
@@ -0,0 +1,111 @@
...
+
+Web library access is allowed in:
+
+* projects meant to run on the web (e.g., have a `web/` directory)
+* plugin packages that declare `web` as a supported context
+
+otherwise, imports of `dart:html`, `dart:js` and  `dart:js_util` are flagged.
+''';
+
+const _webLibs = [
```
devoncarew: Should this be a set?

bwilkerson: Not clear whether the time to perform a hashed lookup would be faster than a linear search with only three items. It might even be faster to write something like
```
int uriLength = uri.length;
if ((uriLength == 9 && uri == 'dart:html') ||
    (uriLength == 7 && uri == 'dart:js') ||
    (uriLength == 12 && uri == 'dart:js_util')) {
  // ...
}

bwilkerson:
Do we want to answer the question "is flagging import directives sufficient?" before committing, or commit it and let people start playing with it?

pq:
> Do we want to answer the question "is flagging import directives sufficient?" before committing, or commit it and let people start playing with it?

I'm inclined to land and iterate but happy to go with the flow...

lib/src/rules/avoid_web_libraries_in_flutter.dart
pq: Good deal.  Will update!

PR #1768, opened 2019-10-14
integrate extension method tests

pq:
no need to sandbox this test code anymore...

/cc @bwilkerson @srawlins


test/rules/slash_for_doc_comments.dart
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2015, the Dart project authors. Please see the AUTHORS file
+// Copyright (c) 2019, the Dart project authors. Please see the AUTHORS file
```
bwilkerson: In the sdk code we generally don't update the year. The reasoning is that some portions of the file have been, or might have been, copyrighted since then. I don't know what you want to do here, so any decision is fine.

pq: Good catch.  That was just a copy paste.  Will fix. 👍

PR #1769, opened 2019-10-15
nnbd test harness

pq:
Adds a test harness that runs rule tests with the `non-nullable` experiment enabled.

To start, I ported a few tests from the `pedantic` ruleset.  These tests are currently identical to those being tested w/o nnbd.

My intention is to evolve them to include nnb-migrated code so we can get early feedback on possible issues.

/cc @scheglov @srawlins

### DIFF
--- a/test/rules/experiments/nnbd/rules/avoid_init_to_null.dart
+++ b/test/rules/experiments/nnbd/rules/avoid_init_to_null.dart
@@ -15,7 +15,10 @@ foo({p: null}) {} //LINT
 class X {
   static const nil = null; //OK
   final nil2 = null; //OK
-  int x = null; //LINT
+
+  // TODO(pq): ints are not nullable so we'll want to update the lint here
+  // since it will produce a compilation error.
+  // int x = null; //LINT
   int y; //OK
   int z;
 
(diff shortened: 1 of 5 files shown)


pq:
fyi @MichaelRFairhurst who might have ideas for things we might start testing...

test/rules/experiments/nnbd/rules/avoid_init_to_null.dart
```diff
@@ -0,0 +1,40 @@
...
+var z = 1; //OK
+const nil = null; //OK
+final nil2 = null; //OK
+foo({p: null}) {} //LINT
+
+
+class X {
+  static const nil = null; //OK
+  final nil2 = null; //OK
+  int x = null; //LINT
```
scheglov: This is a compile-time error.

test/rules/experiments/nnbd/rules/null_closures.dart
```diff
@@ -0,0 +1,51 @@
...
+
+void future_wait() {
+  // Future.wait is a _static_ function with a _named_ argument.
+  Future.wait([], cleanUp: null); // LINT
+  Future.wait([], cleanUp: (_) => print('clean')); // OK
+}
+
+void list_generate() {
+  // List.generate is a _constructor_ with a _positional_ argument.
+  new List.generate(3, null); // LINT
```
scheglov: Don't know for sure how the migrated SDK will do this, but I would think that generator cannot be `null`.

pq:
`TODO`s added.

Thanks!

PR #1771, opened 2019-10-16
update to relative imports

pq:
🤘 

/cc @devoncarew


lib/src/rules/avoid_catching_errors.dart
```diff
@@ -4,8 +4,8 @@
 
 import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
-import 'package:linter/src/analyzer.dart';
-import 'package:linter/src/util/dart_type_utilities.dart';
+import '../analyzer.dart';
```
devoncarew: organize imports?

lib/src/rules/avoid_catches_without_on_clauses.dart
```diff
@@ -4,7 +4,7 @@
 
 import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
-import 'package:linter/src/analyzer.dart';
+import '../analyzer.dart';
```
devoncarew: organize imports?

lib/src/rules/avoid_bool_literals_in_conditional_expressions.dart
```diff
@@ -4,7 +4,7 @@
 
 import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
-import 'package:linter/src/analyzer.dart';
+import '../analyzer.dart';
```
devoncarew: organize imports?

pq:
> It looks like the CI found two files that need formatting.

Ironically, one of them was the format validation code itself!

image

😬

PR #1773, opened 2019-10-16
basic nnbd awareness for `avoid_init_to_null`

pq:
First cut.

Lint description: http://dart-lang.github.io/linter/lints/avoid_init_to_null.html

(Used in `pedantic` and `flutter`.)

Test ideas and suggestions welcome!

See: #1770 

/cc @stereotype441 @scheglov 

/fyi @munificent

### DIFF
--- a/lib/src/rules/avoid_init_to_null.dart
+++ b/lib/src/rules/avoid_init_to_null.dart
@@ -69,6 +69,7 @@ class AvoidInitToNull extends LintRule implements NodeLintRule {
   void registerNodeProcessors(
       NodeLintRegistry registry, LinterContext context) {
     final visitor = _Visitor(this, context);
+    registry.addCompilationUnit(this, visitor);
     registry.addVariableDeclaration(this, visitor);
     registry.addDefaultFormalParameter(this, visitor);
   }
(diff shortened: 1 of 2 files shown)


lib/src/rules/avoid_init_to_null.dart
```diff
@@ -66,21 +68,30 @@ class AvoidInitToNull extends LintRule implements NodeLintRule {
...
 }
 
 class _Visitor extends SimpleAstVisitor<void> {
   final LintRule rule;
+  final LinterContext context;
 
-  _Visitor(this.rule);
+  final bool nnbdEnabled;
+  _Visitor(this.rule, this.context)
+      : nnbdEnabled = context.analysisOptions.contextFeatures
```
scheglov: This is not fully correct.
Individual libraries might specify their language versions, so `FeatureSet`.
The feature set from the context is the default value when not overridden on the library level.

So, ask the unit for its `CompilationUnit.featureSet` instead.

stereotype441: It's possible for NNBD to be enabled in some files and not others.  The right way to check for this is using CompilationUnit.featureSet or LibraryElement.isNonNullableByDefault

lib/src/rules/avoid_init_to_null.dart
```diff
@@ -89,7 +100,9 @@ class _Visitor extends SimpleAstVisitor<void> {
     if (!node.isConst &&
         !node.isFinal &&
         DartTypeUtilities.isNullLiteral(node.initializer)) {
-      rule.reportLint(node);
+      if (isNullable(node.declaredElement.type)) {
```
scheglov: The enclosing `if` combines multiple checks with `&&`.
So, probably this additional check could be added there.

Same for the similar check for parameters above.

test/rules/experiments/nnbd/rules/avoid_init_to_null.dart
```diff
@@ -18,12 +24,20 @@ class X {
...
   // TODO(pq): ints are not nullable so we'll want to update the lint here
   // since it will produce a compilation error.
-  // int x = null; //LINT
+  int x = null; //OK (compilation error)
+  int? xx = null; //LINT
   int y; //OK
-  int z;
+  int z; //OK
+
+  X({int a: null}) //OK (compilation error)
```
scheglov: Test for nullable parameter?

lib/src/rules/avoid_init_to_null.dart
pq: super thanks!

pq: excellent.  Paul pointed this out too.  Thank you!

test/rules/experiments/nnbd/rules/avoid_init_to_null.dart
pq: See `{this.xx: null}` below.  Is that what you mean?

scheglov: Ah, you reused fields to also test parameters.
Clever! :-)

PR #1775, opened 2019-10-18
flutter mocking

pq:
/cc @scheglov


bwilkerson:
It would be cool if we could integrate these into `PackageMixin`.

Issue #1560, opened 2019-05-09
Add lints to help users avoid common Widget anti-patterns.


jacob314:
Motivation from a Flutter user group post: "I've been seeing this a lot with new devs. Don't needlessly use Containers. Wrapping a widget in Container and setting no other parameters does absolutely nothing but make your code more complicated."
These lints could help Flutter examples on the web be more idiomatic with less unneeded boilerplate.

Initial list of lints with obvious quick fixes we could add:
1. Container with no parameters other than child --> Remove the container.
2. Center() widget as a leaf node.
3. Wrapping a widget in a Padding even though the widget itself has padding settings. 
4. Using multiple widgets to do something Container can do.
5. Your idea here??

pq:
These ideas are great.  Do you imagine folks would want to opt in to some but not all of these advices?   To what extent can/should we bundle them in a single rule?  1-3 seem like a natural bundle...  4 might deserve it's own.  Thoughts on names?

Also, motivating examples (ideally from the wild) would be welcome for documentation too.

bwilkerson:
I could imagine cases when teams might define conventions that would negate some of 1-3 but not others, so I'd leave them as separate lints.

Perhaps it's just my unfamiliarity with flutter, but I need more concrete information to understand 4.

jacob314:
4 is the most complex one. Container is basically a de-sugaring for a number of simple combinations of nested padding, alignment, and decoration widgets. Think of it like "div" in CSS. 
Its build method shows what that desugaring is.
For example 
```
Align(
  alignment: alignment
  child : Padding(
    padding: padding,
    child: widget
  ),
)
```
is identical to 
```dart
Container(alignment: alignment, padding: padding, child: widget)
```
Here is the Build method that converts the parameters of a container into the desugaring. To be safe, you could only suggest the lints if the underlying widgets are exactly in the same order they are listed in the build method.

```dart
  Widget build(BuildContext context) {
    Widget current = child;

    if (child == null && (constraints == null || !constraints.isTight)) {
      current = LimitedBox(
        maxWidth: 0.0,
        maxHeight: 0.0,
        child: ConstrainedBox(constraints: const BoxConstraints.expand()),
      );
    }

    if (alignment != null)
      current = Align(alignment: alignment, child: current);

    final EdgeInsetsGeometry effectivePadding = _paddingIncludingDecoration;
    if (effectivePadding != null)
      current = Padding(padding: effectivePadding, child: current);

    if (decoration != null)
[...]

PR #1776, opened 2019-10-18
avoid_unnecessary_containers

pq:
See: #1560.

/cc @jacob314 @scheglov

### DIFF
--- a/lib/src/util/flutter_utils.dart
+++ b/lib/src/util/flutter_utils.dart
@@ -17,6 +17,10 @@ var _collectionInterfaces = <InterfaceTypeDefinition>[
 bool isWidgetType(DartType type) =>
     DartTypeUtilities.implementsInterface(type, 'Widget', '');
 
+bool isContainerType(DartType type) =>
+    DartTypeUtilities.implementsInterface(type, 'Container', '');
+
+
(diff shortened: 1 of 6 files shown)


lib/src/util/flutter_utils.dart
```diff
@@ -13,9 +14,8 @@ var _collectionInterfaces = <InterfaceTypeDefinition>[
   InterfaceTypeDefinition('LinkedHashSet', 'dart.collection'),
 ];
 
-// todo (pq): consider caching lookups
-bool isWidgetType(DartType type) =>
-    DartTypeUtilities.implementsInterface(type, 'Widget', '');
+bool isContainerType(DartType type) =>
+    DartTypeUtilities.implementsInterface(type, 'Container', '');
```
scheglov: Is this the semantic you want - not just exactly `Container`, but also its subclasses?
Why not `isInterface`?

scheglov: See for example `isExactWidgetTypeContainer` is DAS.

pq: hmm, yes.  i think you're right.  thanks.

pq: Ok, fixed.  PTAL?

PR #1778, opened 2019-10-18
fix mocking of `package:meta`

pq:
Long time in the coming...

/cc @scheglov


pq:
Yeah.  Feels good to finally 😬 be getting to this...  Thanks so much for your help!

PR #1779, opened 2019-10-20
New lint: prefer_is_not_operator

### LINKED ISSUE #1522: prefer_is_not_type
Code like:
```
if (!(foo is Foo)) ...
```
should be replaced with:
```
if (foo is! Foo) ...
```

jawahars16:
Fixed #1522 

When checking if the object has not specified type, it is preferable to use 'is!' operator.

**BAD:**
```
if (!(foo is Foo)) {
  ...
}
```

**GOOD:**
```
if (foo is! Foo) {
  ...
}
```

### DIFF
--- a/lib/src/rules/prefer_is_not_operator.dart
+++ b/lib/src/rules/prefer_is_not_operator.dart
@@ -51,13 +51,19 @@ class _Visitor extends SimpleAstVisitor<void> {
 
   @override
   void visitIsExpression(IsExpression node) {
+    // Return if it is `is!` expression
+    if (node.notOperator != null) {
+      return;
+    }
+
     // Check whether is expression is inside parenthesis
     if (node.parent is ParenthesizedExpression) {
-      final isExpressionInsideParenthesis = node.parent;
+      final parenthesizedExpression = node.parent;
+      final prefixExpression = parenthesizedExpression.parent;
       // Check for NOT (!) operator
-      if (isExpressionInsideParenthesis.parent.beginToken.type ==
-          TokenType.BANG) {
-        rule.reportLint(isExpressionInsideParenthesis.parent);
+      if (prefixExpression is PrefixExpression &&
+          prefixExpression.operator.type == TokenType.BANG) {
+        rule.reportLint(parenthesizedExpression.parent);
       }
     }
   }
--- a/lib/src/rules/prefer_is_not_operator.dart
+++ b/lib/src/rules/prefer_is_not_operator.dart
@@ -1,0 +1,62 @@
+// Copyright (c) 2019, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/ast/ast.dart';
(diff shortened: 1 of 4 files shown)


lib/src/rules/prefer_is_not_operator.dart
```diff
@@ -0,0 +1,64 @@
...
+  void visitIsExpression(IsExpression node) {
+    // Check whether is expression is inside parenthesis
+    if (node.parent is ParenthesizedExpression) {
+      final isExpressionInsideParenthesis = node.parent;
+      // Check for NOT (!) operator
+      if (isExpressionInsideParenthesis.parent.beginToken.type ==
+          TokenType.BANG) {
+        rule.reportLint(isExpressionInsideParenthesis.parent);
+      }
+    }
```
pq: I think this may be more robust?

```dart
    final prefixExpression =
        node.parent.thisOrAncestorOfType<PrefixExpression>();
    if (prefixExpression?.operator?.type == TokenType.BANG) {
      rule.reportLint(prefixExpression);
    }
```

(Deferring to @scheglov and @bwilkerson who know the AST model better...)

scheglov: I'd go with pure `.parent` checks.
Something like (not tested):
```dart
if (node.notOperator == null) {
  final parent = node.parent;
  if (parent is ParenthesizedExpression) {
    final parent2 = parent.parent;
    if (parent2 is PrefixExpression && parent2.operator.type = TokenType.BANG) {
      ...report
    }
  }
}
```

jawahars16: Yes checking for `is!` in the beginning is better. 

And I also prefer checking for `ParanthesizedExpression` before doing NOT operator check for the parenthesized expression. 

```dart
  void visitIsExpression(IsExpression node) {
    // Return if it is `is!` expression
    if(node.notOperator != null) {
      return;
    }

    // Check whether is expression is inside parenthesis
    if (node.parent is ParenthesizedExpression) {
      final isExpressionInsideParenthesis = node.parent;
      final prefixExpression = isExpressionInsideParenthesis.parent;
      // Check for NOT (!) operator
      if (prefixExpression is PrefixExpression &&
          prefixExpression?.operator?.type == TokenType.BANG) {
        rule.reportLint(isExpressionInsideParenthesis.parent);
      }
    }
  }
```

Let me know if you have any comments, so that I can update the PR.

pq: SGTM!

scheglov: The name `isExpressionInsideParenthesis` does not sound right. It is not "inside" parenthesis, it is parenthesis. Maybe just `parenthesizedExpression`.

We have already checked that `prefixExpression is PrefixExpression`, so it cannot be `null`, so `prefixExpression?.operator` is not necessary, just `prefixExpression.operator` will work. Correspondingly the second `?.type` also does not have to be null aware.

pq:
The `pana` baseline breakage is unrelated (see: https://github.com/dart-lang/dart_style/issues/866 and https://github.com/dart-lang/test/issues/1102) and should get resolved shortly.

@jawahars16: thanks again for the contribution!

PR #1794, opened 2019-10-28
avoid_returning_null_for_void only checks null literal

### LINKED ISSUE #1787: avoid_returning_null_for_void false positives
The lint fires for this:

```dart
  void clearFilter() => _filteredItems = null;
```

Effective Dart explicitly calls this out as something that's reasonable to do

https://dart.dev/guides/language/effective-dart/usage#consider-using--for-simple-members
"You can also use => on members that don’t return a value."
```
set x(num value) => center = Point(value, center.y);
```

so I think an update is in order to not fire on `=>` methods.

What do you think, please?

a14n:
# Description

Relax `avoid_returning_null_for_void` to check only `null` literal (and not expression having Null type)

Fixes #1787

### DIFF
--- a/lib/src/rules/avoid_returning_null_for_void.dart
+++ b/lib/src/rules/avoid_returning_null_for_void.dart
@@ -77,7 +77,7 @@ class _Visitor extends SimpleAstVisitor<void> {
   }
 
   void _visit(AstNode node, Expression expression) {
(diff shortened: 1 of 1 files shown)


pq:
Thanks!

@a14n : is this good to land?

a14n:
> @a14n : is this good to land?

Yes

Issue #1786, opened 2019-10-24
avoid_renaming_method_parameters doesn't fire


davidmorgan:
The lint doesn't fire anywhere in google3, and I couldn't get it to fire by hand, even using the example in the lint docs. Any ideas please?

```dart
abstract class A {
  m(a);
}

abstract class B extends A {
  m(b);
}
```

a14n:
I can see the lint on your example. What's the version of Dart SDK?

davidmorgan:
Thanks! I dug in a bit more. The lint (using `dartanalyzer --lint` on the command line) only works if there is a `pubspec.yaml` file.

Since we don't use `pubspec.yaml` in google3 it's not there and the lint doesn't fire.

Question is, what is it in the implementation that needs it? I don't see anything. Maybe @pq has an idea...

pq:
Right.  We use the pubspec here as a way to identify the root of the project.

https://github.com/dart-lang/linter/blob/1394a375eadf91433e7499e630204b70634742d6/lib/src/ast.dart#L125-L131

which, in turn, is how we identify if we're in `lib/`

https://github.com/dart-lang/linter/blob/1394a375eadf91433e7499e630204b70634742d6/lib/src/rules/avoid_renaming_method_parameters.dart#L83

This is a a kind of funky hack and there are a few other ways to do it.   I'll take a look at alternatives today.

pq:
The current thinking is to update our test to use `workspace`s which are Google3-aware.

Supporting API update in flight here:

https://dart-review.googlesource.com/c/sdk/+/123341

pq:
@davidmorgan: this should be fixed as of the next roll into Google3; please let me know if you see otherwise!

PR #1810, opened 2019-10-31
use workspace package to locate `lib/`

pq:
Fixes: #1786

Should also make `prefer_relative_imports` and `public_member_api_docs` google3-friendly.

Bonus: makes `avoid_renaming_method_parameters` way more performant.

Before:

image

After:

image

🏅 🏅 🏅 

/cc @bwilkerson @scheglov 

/fyi @a14n

### DIFF
--- a/lib/src/ast.dart
+++ b/lib/src/ast.dart
@@ -122,7 +122,7 @@ bool isHashCode(ClassMember element) =>
         getFieldIdentifier(element, 'hashCode') != null);
 
 /// Return true if this compilation unit [node] is declared within the given
-/// [package]'s `lib/` directory.
+/// [package]'s `lib/` directory tree.
 bool isInLibDir(CompilationUnit node, WorkspacePackage package) {
   final cuPath = node.declaredElement.library?.source?.fullName;
   if (cuPath == null) return false;
--- a/lib/src/ast.dart
+++ b/lib/src/ast.dart
@@ -124,9 +124,10 @@ bool isHashCode(ClassMember element) =>
 /// Return true if this compilation unit [node] is declared within the given
 /// [package]'s `lib/` directory.
 bool isInLibDir(CompilationUnit node, WorkspacePackage package) {
-  final libDir = path.join(package.root, 'lib');
   final cuPath = node.declaredElement.library?.source?.fullName;
-  return cuPath?.startsWith(libDir) == true;
+  if (cuPath == null) return false;
+  final libDir = path.join(package.root, 'lib');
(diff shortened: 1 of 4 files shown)


lib/src/ast.dart
```diff
@@ -151,6 +121,14 @@ bool isHashCode(ClassMember element) =>
     (element is FieldDeclaration &&
         getFieldIdentifier(element, 'hashCode') != null);
 
+/// Return true if this compilation unit [node] is declared within the given
+/// [package]'s `lib/` directory.
+bool isInLibDir(CompilationUnit node, WorkspacePackage package) {
+  final libDir = path.join(package.root, 'lib');
+  final cuPath = node.declaredElement.library?.source?.fullName;
+  return cuPath?.startsWith(libDir) == true;
```
bwilkerson: Consider using `path.isWithin`.

pq: Cool.  Will-do!

lib/src/rules/avoid_renaming_method_parameters.dart
```diff
@@ -55,18 +55,29 @@ class AvoidRenamingMethodParameters extends LintRule implements NodeLintRule {
...
+    registry.addCompilationUnit(this, visitor);
   }
 }
 
 class _Visitor extends SimpleAstVisitor<void> {
   final LintRule rule;
+  final LinterContext context;
 
-  _Visitor(this.rule);
+  bool isInLib;
```
srawlins: I think this warrants a comment; just saying that it specifically states whether the _compilation unit_ is in [context]'s package.

lib/src/ast.dart
```diff
@@ -151,6 +121,15 @@ bool isHashCode(ClassMember element) =>
     (element is FieldDeclaration &&
         getFieldIdentifier(element, 'hashCode') != null);
 
+/// Return true if this compilation unit [node] is declared within the given
+/// [package]'s `lib/` directory.
```
srawlins: To differentiate between, say, "public interface" files immediately in lib/, and source files in, e.g. lib/src/, maybe say "declared within the given package's `lib/` directory _tree_."

pq: Good idea.

lib/src/rules/avoid_renaming_method_parameters.dart
pq: Sure thing.  Will do.

pq:
Thanks for the feedback @srawlins!

PR #1812, opened 2019-10-31
👟👟👟 `prefer_relative_imports`: stop parsing pubspecs (query ws package for self-references)

pq:
Speed up `prefer_relative_imports` and make it google3-friendly by querying the workspace package (and forgoing any pubspec parsing).

Before:

image

After:

image

Definitely **_faster_**...  Still need to convince myself that it's correct.  (A few TODOs to consider.)

See also: #1810.

/cc @devoncarew @bwilkerson

### DIFF
--- a/lib/src/rules/prefer_relative_imports.dart
+++ b/lib/src/rules/prefer_relative_imports.dart
@@ -4,8 +4,7 @@
 
 import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
-import 'package:analyzer/file_system/file_system.dart';
-import 'package:yaml/yaml.dart';
+import 'package:path/path.dart' as path;
 
 import '../analyzer.dart';
 import '../ast.dart';
@@ -34,22 +33,6 @@ import 'package:my_package/bar.dart';
 
 ''';
 
-YamlMap _parseYaml(String content) {
-  if (content == null) {
-    return YamlMap();
-  }
-  try {
-    YamlNode doc = loadYamlNode(content);
-    if (doc is YamlMap) {
-      return doc;
-    }
-    return YamlMap();
-    // ignore: avoid_catches_without_on_clauses
-  } catch (_) {
-    return YamlMap();
-  }
-}
-
 class PreferRelativeImports extends LintRule implements NodeLintRule {
   PreferRelativeImports()
       : super(
@@ -72,8 +55,6 @@ class _Visitor extends SimpleAstVisitor<void> {
   final LinterContext context;
 
   bool isInLibFolder;
-  File pubspecFile;
-  YamlMap parsedPubspec;
 
   _Visitor(this.rule, this.context);
 
@@ -85,6 +66,7 @@ class _Visitor extends SimpleAstVisitor<void> {
     String importUri = node?.uri?.stringValue;
     if (importUri == null) return false;
 
+    // todo (pq): is there a better way to do this?
     Uri uri;
     try {
       uri = Uri.parse(importUri);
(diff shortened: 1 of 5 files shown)


lib/src/rules/prefer_relative_imports.dart
```diff
@@ -85,6 +66,7 @@ class _Visitor extends SimpleAstVisitor<void> {
     String importUri = node?.uri?.stringValue;
```
bwilkerson: If this method is only invoked from `visitImportDirective`, then `node` can't be `null` so there's no need for the null check. Consider using `node.uriContent`.

lib/src/rules/prefer_relative_imports.dart
```diff
@@ -85,6 +66,7 @@ class _Visitor extends SimpleAstVisitor<void> {
...
     if (importUri == null) return false;
 
+    // todo (pq): is there a better way to do this?
```
bwilkerson: Consider `node.uriSource.uri`.

devoncarew: Ah, to determine the uri w/o relying on trying to parse and catching exceptions?

lib/src/rules/prefer_relative_imports.dart
```diff
@@ -93,34 +75,19 @@ class _Visitor extends SimpleAstVisitor<void> {
...
-      String content;
-      try {
-        content = pubspecFile.readAsStringSync();
-        // ignore: avoid_catches_without_on_clauses
-      } catch (_) {}
-      parsedPubspec = _parseYaml(content);
-    }
-
-    return parsedPubspec['name'] == segments[0];
+    final source = node.uriElement?.library?.source;
```
bwilkerson: Consider `node.uriSource`.

lib/src/rules/prefer_relative_imports.dart
```diff
@@ -93,34 +75,19 @@ class _Visitor extends SimpleAstVisitor<void> {
...
+    if (source == null) return false;
+    // todo (pq): should context.package.contains(source)work?
```
bwilkerson: Yes, that should work.

lib/src/rules/prefer_relative_imports.dart
```diff
@@ -93,34 +75,19 @@ class _Visitor extends SimpleAstVisitor<void> {
...
+    return path.isWithin(context.package.root, source.fullName);
```
devoncarew: Hmm, I don't follow the logic here. How does this trigger when there's a package self-reference in lib/, but not in bin/? I don't know what `context.package.root` holds.

If you think this logic is correct (which it very well could be), perhaps add more positive and negative tests to make sure it works in several different types of situations?

lib/src/rules/prefer_relative_imports.dart
pq: Thanks!

lib/src/rules/prefer_relative_imports.dart
pq: Oddly it's not...  I'll look a little closer.

/fyi @srawlins

lib/src/rules/prefer_relative_imports.dart
pq: Yeah.  It seems like a bunch of work when I'd expect the URI to already be around.

Moreover, if all we want is to know if it's package-qualified we might simplify further by just testing something like `importUri.startsWith('package:')`.  🤔

lib/src/rules/prefer_relative_imports.dart
pq: The lib check is handled by the `isInLibFolder` bool we set in `visitCompilationUnit` and use to short-circuit this visit which is to say, if we're not in `lib/` we shouldn't even get here.

That said, some more tests are a great idea. 👍

lib/src/rules/prefer_relative_imports.dart
pq: Unfortunately that doesn't seem to work for what we're trying to do (though maybe I'm missing something).  The rub is that:

```dart
import 'dummy.dart';
```    

reports a `package` URI scheme but that confuses our test for whether the import is package qualified.

pq:
Obviously these benchmarks aren't rigorous or the final word but at first blush, tweaking the URI parsing looks promising:

image

🤔

pq:
OK.  Tests added.  After a few runs we're settling at about a `9ms` benchmark.  (Not too shabby having started ~200!)

Unless there are reservations, I'll land and we can iterate as needed.  (Happy to add more tests too.)

PR #1814, opened 2019-11-01
pubspec caching

pq:
Before:

image

### DIFF
--- a/lib/src/rules/avoid_web_libraries_in_flutter.dart
+++ b/lib/src/rules/avoid_web_libraries_in_flutter.dart
@@ -2,13 +2,15 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:io';
+
 import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
 import 'package:analyzer/file_system/file_system.dart';
+import 'package:path/path.dart' as path;
 import 'package:yaml/yaml.dart';
 
 import '../analyzer.dart';
-import '../ast.dart';
 
 const _desc =
     r'Avoid using web-only libraries outside Flutter web plugin packages.';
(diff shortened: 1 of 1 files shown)


lib/src/rules/avoid_web_libraries_in_flutter.dart
```diff
@@ -52,33 +40,82 @@ class AvoidWebLibrariesInFlutter extends LintRule implements NodeLintRule {
...
+}
+
+class Pubspec {
+  static Map<String, Entry> cache = <String, Entry>{};
+
+  static YamlMap forUnit(CompilationUnit unit, LinterContext context) {
+    final filePath = path.join(context.package.root, 'pubspec.yaml');
+    final entry = cache[filePath];
+    if (entry != null) {
+      if (entry.file.modificationStamp != entry.timestamp) {
```
bwilkerson: Getting the modification stamp is not as inexpensive as you might hope, and this gets it twice: once here and once in `update`.

lib/src/rules/avoid_web_libraries_in_flutter.dart
```diff
@@ -52,33 +40,82 @@ class AvoidWebLibrariesInFlutter extends LintRule implements NodeLintRule {
...
+    } catch (_) {
+      // Fall-through.
+    }
+    return YamlMap();
+  }
+
+}
+
+class Pubspec {
+  static Map<String, Entry> cache = <String, Entry>{};
```
bwilkerson: There's no way to flush the cache, so it is effectively a slow memory leak.

lib/src/rules/avoid_web_libraries_in_flutter.dart
```diff
@@ -2,13 +2,15 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:io';
```
bwilkerson: I can't tell where this is being used, but it's somewhat concerning if we really need it.

pq:
Right.  Too many problems with this approach.

Neverminding the leak, a few runs show the benefits are negligible (and under some circumstances actually negative).

Discussion ongoing in #1813 but closing this for now.

PR #1823, opened 2019-11-06
de-linting (keeping ahead of the Joneses)

pq:
More lints that are getting favorable internal attention.

/cc @bwilkerson

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -89,9 +89,10 @@ linter:
 #    - prefer_generic_function_type_aliases # under review (see #1068)
     - prefer_initializing_formals
     - prefer_inlined_adds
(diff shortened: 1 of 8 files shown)


analysis_options.yaml
```diff
@@ -114,7 +115,8 @@ linter:
     - unnecessary_parenthesis
     - unnecessary_statements
     - unnecessary_this
+    - use_function_type_syntax_for_parameters
     - use_setters_to_change_properties
     - use_string_buffers
     - use_to_and_as_if_applicable
-#    - void_checks # under review (see #1068)
+    - void_checks # under review (see #1068)
```
a14n: Perhaps it is not under review anymore :)

pq: Quite right.  Good catch!

PR #1825, opened 2019-11-07
more linting

pq:
More lints that are being considered favorably for `pedantic`.

/cc @a14n @bwilkerson

### DIFF
--- a/test/engine_test.dart
+++ b/test/engine_test.dart
@@ -188,7 +188,7 @@ void defineLinterEngineTests() {
   });
 }
 
-typedef NodeVisitor = void Function(AstNode node);
+typedef NodeVisitor = void Function(Object node);
 
 class MockLinter extends LintRule {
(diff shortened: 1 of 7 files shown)


test/engine_test.dart
```diff
@@ -188,7 +188,7 @@ void defineLinterEngineTests() {
   });
 }
 
-typedef NodeVisitor(node);
+typedef NodeVisitor = void Function(AstNode node);
```
a14n: Regarding its usage, it can be call with `PSEntry` at https://github.com/dart-lang/linter/blob/33a9a0a9df49497bd9bca20c0cb2688e5937db05/test/engine_test.dart#L231

pq: Ah!  Thanks.

pq: (Obviously a bit hasty.)

PR #1827, opened 2019-11-11
Add new lint: unnecessary_final.

davidmorgan:
This lint is more or less the opposite of `prefer_final_locals` plus `prefer_final_in_foreach`, plus checking parameters. (There isn't a `prefer_final_parameters`).

What do you think, please? This seems much closer to common usage. I'm not sure we'll ever get around to enforcing it, but it's nice to have a lint that matches what people do, so that people aren't just citing the existence of those lints as an argument :)

### DIFF
--- a/example/all.yaml
+++ b/example/all.yaml
@@ -136,6 +136,7 @@ linter:
     - unnecessary_await_in_return
     - unnecessary_brace_in_string_interps
     - unnecessary_const
+    - unnecessary_final
     - unnecessary_getters_setters
     - unnecessary_lambdas
     - unnecessary_new
--- a/lib/src/rules.dart
+++ b/lib/src/rules.dart
@@ -139,6 +139,7 @@ import 'rules/unawaited_futures.dart';
 import 'rules/unnecessary_await_in_return.dart';
 import 'rules/unnecessary_brace_in_string_interps.dart';
 import 'rules/unnecessary_const.dart';
+import 'rules/unnecessary_final.dart';
 import 'rules/unnecessary_getters_setters.dart';
 import 'rules/unnecessary_lambdas.dart';
 import 'rules/unnecessary_new.dart';
@@ -298,6 +299,7 @@ void registerLintRules() {
     ..register(UnnecessaryAwaitInReturn())
     ..register(UnnecessaryBraceInStringInterps())
     ..register(UnnecessaryConst())
+    ..register(UnnecessaryFinal())
     ..register(UnnecessaryNew())
     ..register(UnnecessaryNullAwareAssignments())
     ..register(UnnecessaryNullInIfNullOperators())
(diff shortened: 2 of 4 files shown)


a14n:
`final` allows to avoid accidentaly reassign a local variable and prevents bugs.

Personnally I would never enable `unnecessary_final` in one of my project.

See also https://stackoverflow.com/a/316787/634863

davidmorgan:
Thanks @a14n .

We have fans of both approaches--as on that stackoverflow answer. In google3 it looks like >50% prefers `var` to `final`, although it's not overwhelming.

With this lint, the linter can now enforce whichever of the two you like, or neither :)

My personal interest is to try to make the tools good enough that nobody wants `final` for locals. The IDE could highlight local variable reassignments for you; there's no need for you to do it by hand.

pq:
> My personal interest is to try to make the tools good enough that nobody wants final for locals.

I find this idea confusing.  IMO the value of `final` is in communicating intent (that the tools can then use to help keep you on the rails).  I don't understand why we'd ever want to do away with that.

a14n:
Unlike Java `final` can be used to introduce variable in Dart.

It looks like the benefits of using `final` are rarely challenged. The main concern on the Java side is that it adds verbosity:

```java
// in java
String s = "";
// with java>=11
final var s = "";
//--- with final
final String s = "";
// with java>=11
final var s = "";
```

In Dart the difference is smaller (so the main drawback is lowered):

```dart
// in java
var s = '';
// with java>=11
final s = '';
```

davidmorgan:
We've had this discussion a _lot_ of times internally :)

Originally I was on the side of `final`. I'd like it to be the default in the language, but since it isn't, we could at least add it manually everywhere. So, I argued for everyone to use it everywhere.

People didn't want to :) ... they argued that the extra typing (just two characters, but in lots of places) doesn't add enough value. Eventually, after raising this topic repeatedly over a number of years, I figured it's never going to happen: we can't make `final` the standard since Dart was created with `var`.

So I changed my own code to use `var.` e.g. here was the change for `built_value`: https://github.com/google/built_value.dart/commit/7ea8681c0c1a752df1c06c8456f114014071ccca

And my personal experience is that it makes no difference whatsoever. I can't think of a single case where using `var` instead of `final` caused me problems. And it is, indeed, two characters shorter, which is nice.

I hope that one day either enforcement of `final` or enforcement of _not_ using `final` we be added to `package:pedantic`, resolving the issue one way or the other. However, since I have to get a consensus for either change, and the current split is pretty close to fifty fifty, this may unfortunately be an uphill struggle :)

davidmorgan:
One more thought. Enforcing this lint has the nice property that it makes `final` just for fields, and `var` just for locals. That cuts down on confusion about what `final` does/means, and aids readability overall.

Thanks.

PR #1832, opened 2019-11-13
allow overrides that add annotations

pq:
Fixes: #1831.

/cc @bwilkerson

### DIFF
--- a/lib/src/rules/unnecessary_overrides.dart
+++ b/lib/src/rules/unnecessary_overrides.dart
@@ -41,7 +41,7 @@ It's valid to override a member in the following cases:
 
 * if a type (return type or a parameter type) is not the exactly the same as the
 super method,
-* if `covariant` keyword is added to one of the parameters,
+* if the `covariant` keyword is added to one of the parameters,
 * if documentation comments are present on the member,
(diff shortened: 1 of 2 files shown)


lib/src/rules/unnecessary_overrides.dart
```diff
@@ -37,12 +37,13 @@ class A extends B {
...
-It's valid to override method for the following cases:
+It's valid to override a member in the following cases:
 
 * if a type (return type or a parameter type) is not the exactly the same as the
 super method,
-* if `covariant` keyword is added to one of the parameter,
-* if documentation comments are present on the method.
+* if the `covariant` keyword is added to one of the parameters,
+* if documentation comments are present on the member,
+* if annotations are added to the member (e.g., `@nonVirtual`).
```
bwilkerson: nit: "added" isn't quite what you've implemented, but I think the implementation you have is better. It's really more "if the member has annotations other than `override`.". (I don't think the example is necessary, but don't object to it either.)

pq: I'll update the doc.  Thanks!

Issue #1837, opened 2019-11-14
assess (and address) perf issues in pedantic lints


pq:
All lints should be audited but `package:pedantic` lints are a good place to start since they have such reach.

pq:
Stale (and largely done).

PR #1838, opened 2019-11-14
improve perf for `always_require_non_null_named_parameters`

pq:
Before:

image

After:

image

/cc @bwilkerson 

See:  #1837.

### DIFF
--- a/lib/src/rules/always_require_non_null_named_parameters.dart
+++ b/lib/src/rules/always_require_non_null_named_parameters.dart
@@ -13,8 +13,8 @@ const _desc = r'Use @required.';
 
 const _details = r'''
 
-**DO** specify `@required` on named parameter without default value on which an
-assert(param != null) is done.
+**DO** specify `@required` on named parameters without a default value on which 
+an `assert(param != null)` is done.
 
 **GOOD:**
 ```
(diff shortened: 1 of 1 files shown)


lib/src/rules/always_require_non_null_named_parameters.dart
```diff
@@ -68,50 +68,64 @@ class _Visitor extends SimpleAstVisitor<void> {
...
+      _checkInitializerList(getParams(), parent.initializers);
+      _checkParams(getParams(), parent.body);
     } else if (parent is MethodDeclaration) {
-      _checkParams(params, parent.body);
+      _checkParams(getParams(), parent.body);
+    }
+  }
+
+  void _checkAsserts(
+      List<Expression> asserts, List<DefaultFormalParameter> params) {
```
bwilkerson: Consider converting this to check a single assert condition rather than a list of conditions. Invocation sites can then invoke this method directly rather than creating a list that needs to be iterated over.

lib/src/rules/always_require_non_null_named_parameters.dart
```diff
@@ -68,50 +68,64 @@ class _Visitor extends SimpleAstVisitor<void> {
...
+    for (final expression in asserts) {
+      for (final param in params) {
+        if (_hasAssertNotNull(expression, param.identifier.name)) {
+          rule.reportLintForToken(param.identifier.beginToken);
```
bwilkerson: Consider removing the parameter from the list after creating a lint. Otherwise redundant asserts will generate multiple diagnostics. It's a fairly uncommon case, so it might not be worth the work from that perspective, but it also has the performance benefit that the list gets shorter making future iterations over the list faster.

pq: Interesting idea!

Let's discuss in person.  Thanks!

lib/src/rules/always_require_non_null_named_parameters.dart
pq: One less list allocation.  Great idea.  Thanks!

lib/src/rules/always_require_non_null_named_parameters.dart
pq: Updated.  Thanks!

PR #1839, opened 2019-11-14
doc microbenchmark

pq:
/cc @bwilkerson

### DIFF
--- a/tool/benchmark.dart
+++ b/tool/benchmark.dart
@@ -2,6 +2,13 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+/// Micro-benchmarks.
+///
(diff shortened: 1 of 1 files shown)


tool/benchmark.dart
```diff
@@ -2,6 +2,13 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+/// Micro-benchmarks.
+///
+/// More comprehensive benchmarks are run like so:
+///
+///     dart bin/linter.dart --benchmark -q -c example/all.yaml .
+library micro_benchmarks;
```
bwilkerson: You don't need a `library` directive to have a library comment. Dartdoc will take the comment from the first directive in the file.

pq: 🎉 🎉 🎉

PR #1840, opened 2019-11-14
annotate pedantic lint benchmarks

pq:
image

See: #1837 

/cc @bwilkerson

### DIFF
--- a/lib/src/util/score_utils.dart
+++ b/lib/src/util/score_utils.dart
@@ -1,0 +1,41 @@
+// Copyright (c) 2019, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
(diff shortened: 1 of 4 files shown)


lib/src/util/score_utils.dart
```diff
@@ -0,0 +1,41 @@
...
+
+import 'package:analyzer/src/lint/config.dart'; // ignore: implementation_imports
+import 'package:http/http.dart' as http;
+
+const _pedanticOptionsRootUrl =
+    'https://raw.githubusercontent.com/dart-lang/pedantic/master/lib';
+const _pedanticOptionsUrl = '$_pedanticOptionsRootUrl/analysis_options.yaml';
+
+List<String> _pedanticRules;
+Future<List<String>> get pedanticRules async =>
```
bwilkerson: I thought we already had these utilities somewhere else, but I assume we didn't or you wouldn't have copied them.

pq: This PR moves them into `lib/src` so they can be generally used. 👍

PR #1841, opened 2019-11-14
update to new `package:github` API

pq:
/cc @bwilkerson


robbecker-wf:
I'm glad that was an easy upgrade!

PR #1845, opened 2019-11-20
Use TypeSystem class from API.

scheglov:
We would like to remove switch all clients to the API version, and remove `TypeSystem` declared in `analyzer/lib/src/generated/type_system.dart`. This change works with both version of analyzer, `0.39.1` and with my pending changes to update `LinterContext` to return the API version of `TypeSystem`.

### DIFF
--- a/lib/src/analyzer.dart
+++ b/lib/src/analyzer.dart
@@ -28,7 +28,8 @@ export 'package:analyzer/src/dart/error/lint_codes.dart';
 export 'package:analyzer/src/generated/engine.dart'
     show AnalysisContext, AnalysisErrorInfo;
 export 'package:analyzer/src/generated/resolver.dart'
-    show ExitDetector, TypeProvider, TypeSystem;
+    show ExitDetector, TypeProvider;
+export 'package:analyzer/dart/element/type_system.dart' show TypeSystem;
 export 'package:analyzer/src/generated/source.dart' show LineInfo, Source;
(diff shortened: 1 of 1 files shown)


pq:
@scheglov: what's the status of this?

scheglov:
As you can see, there were build errors, and I think we need a new analyzer roll to fix it.

@stereotype441 was going to release new non-breaking `analyzer` version last working Friday, but we had to postpone it because I had a change that we also would like to include, but it was not ready yet.

Once we have a new `analyzer` version, I will review the change, although I don't remember already what specifically I was going to do. I guess I will find out :-)

pq:
😄 

Thanks for the context.  Happy to roll a new linter when this lands.

scheglov:
@pq has a better / wider PR.

stereotype441:
> As you can see, there were build errors, and I think we need a new analyzer roll to fix it.
> 
> @stereotype441 was going to release new non-breaking `analyzer` version last working Friday, but we had to postpone it because I had a change that we also would like to include, but it was not ready yet.
> 
> Once we have a new `analyzer` version, I will review the change, although I don't remember already what specifically I was going to do. I guess I will find out :-)

FYI, analyzer 0.39.2+1 was published yesterday, and it appears to actually be working (unlike analyzer 0.39.2, sorry :grimacing:), so if you had any follow-up work you were waiting to do, it should be unblocked now.

PR #1866, opened 2019-12-03
check for null packages in "isInLibDir" check

pq:
Fixes: #1862 #1863 .

Looking closer at call sites, it's possible for `package` to be null so we need to guard against that.

/cc @bwilkerson 

/fyi @scheglov

### DIFF
--- a/lib/src/ast.dart
+++ b/lib/src/ast.dart
@@ -124,6 +124,7 @@ bool isHashCode(ClassMember element) =>
 /// Return true if this compilation unit [node] is declared within the given
 /// [package]'s `lib/` directory tree.
 bool isInLibDir(CompilationUnit node, WorkspacePackage package) {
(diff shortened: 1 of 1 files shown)


pq:
@scheglov: I just looked a little closer to see why we might be throwing exceptions when visiting the compilation unit.

@bwilkerson: the rub is there is at least one path where a `BazelWorkspace` can return null when being asked to `findPackageFor` a file path.  This may be true of other workspaces too.

bwilkerson:
> the rub is there is at least one path where a BazelWorkspace can return null when being asked to findPackageFor a file path. This may be true of other workspaces too.

Yes, most of the workspace classes can return `null`. There's a method somewhere that runs through all the workspace classes and returns the instance of the right class. If we're only using `BazelWorkspace` then a `null` is expected.

PR #1890, opened 2019-12-13
narrow avoid_as reporting to type annotation node

pq:
/cc @bwilkerson


pq:
Puzzling out why our pana baseline has fallen but that seems unrelated to this  change.

Issue #1891, opened 2019-12-13
[build] code coverage is stalled


pq:
Coverage has been stalled since Sept 11.

Sample failure:

![image](https://user-images.githubusercontent.com/67586/70823332-785dcd00-1d94-11ea-8b69-1850403ffbed.png)

pq:
Coverage is back.  Possibly fixed by #1893.

PR #1892, opened 2019-12-13
debug coverage

pq:
See: #1891.


pq:
Landing TBR since this is just build debugging.

PR #1900, opened 2019-12-18
Use LinterContext.evaluateConstant()

scheglov:


scheglov:
Hm... Not sure why `pana` complains.

pq:
The issue w/ `pana` is unrelated.

PR #1904, opened 2019-12-23
Check doc comments for mixins in slash_for_doc_comments

bwilkerson:
This lint was not checking the doc comments on mixins. It was updated for extensions, but mixins were missed.

It seems unlikely that it will create problems when rolling into Flutter, but I haven't verified that. If you think we should hold off until that's been checked I'm happy to wait.

### DIFF
--- a/lib/src/rules/slash_for_doc_comments.dart
+++ b/lib/src/rules/slash_for_doc_comments.dart
@@ -66,6 +66,7 @@ class SlashForDocComments extends LintRule implements NodeLintRule {
     registry.addFunctionTypeAlias(this, visitor);
     registry.addLibraryDirective(this, visitor);
     registry.addMethodDeclaration(this, visitor);
+    registry.addMixinDeclaration(this, visitor);
     registry.addTopLevelVariableDeclaration(this, visitor);
   }
 }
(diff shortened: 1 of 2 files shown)


bwilkerson:
I just noticed that we're also not checking docs on typedefs that use the generic function type syntax (`typedef F = void Function(int)`). I'll update the PR.

bwilkerson:
PTAL

bwilkerson:
It looks like it's also not catching library documentation comments (those preceding the first directive in the file). I'm not going to try to fix that right now even though the fix should be fairly straight forward.

Also, I have no idea why it failed pana.

pq:
Pana is a know issue.  It's been fixed but there won't be another push until January.

See: https://github.com/dart-lang/pana/issues/591

(I'll go ahead and lower the baseline for now.)

bwilkerson:
I added support for library comments occurring before directives other than a `library` directive.

What do you think about adding support for "doc" comments before local functions? I know that I often comment local functions, but I don't know whether anyone else does. On the other hand, if they don't it won't hurt that we checked, so ...

pq:
> What do you think about adding support for "doc" comments before local functions? 

SGTM!

bwilkerson:
Local function support added. PTAL.

pq:
LGTM

PR #1907, opened 2019-12-23
avoid_redundant_parameter_values

### LINKED ISSUE #1790: Add lint for redundant parameter values
`redundant_parameter_value`: warns when a user explicitly passes the default value for a parameter. 

```
class Foo {
  Foo({bar = ‘baz’});
}
var foo = Foo(bar: ‘baz’); // this should trigger the warning
```

pq:
Fixes #1790

/cc @bwilkerson

### DIFF
--- a/lib/src/rules/avoid_redundant_argument_values.dart
+++ b/lib/src/rules/avoid_redundant_argument_values.dart
@@ -12,7 +12,7 @@ const _desc = r'Avoid redundant argument values.';
 
 const _details = r'''Avoid redundant argument values.
 
-**DON'T** declare argument with values that match their defaults.
+**DON'T** declare arguments with values that match their defaults.
 
 **BAD:**
 ```
@@ -50,6 +50,7 @@ class AvoidRedundantArgumentValues extends LintRule implements NodeLintRule {
   void registerNodeProcessors(NodeLintRegistry registry,
       [LinterContext context]) {
     final visitor = _Visitor(this, context);
+    registry.addInstanceCreationExpression(this, visitor);
     registry.addMethodInvocation(this, visitor);
   }
 }
@@ -60,40 +61,45 @@ class _Visitor extends SimpleAstVisitor {
 
   _Visitor(this.rule, this.context);
 
-  @override
-  void visitMethodInvocation(MethodInvocation node) {
-    if (node.argumentList.arguments.isEmpty) {
+  void check(ArgumentList argumentList, List<ParameterElement> parameters) {
+    if (argumentList.arguments.isEmpty || parameters == null) {
       return;
     }
 
-    final type = node.staticInvokeType;
-    final element = type.element;
-    var parameters;
+    for (var arg in argumentList.arguments) {
+      final param = arg.staticParameterElement;
+      if (param.hasRequired) {
+        continue;
+      }
+      final value = param.constantValue;
+      if (value != null) {
+        if (arg is NamedExpression) {
+          arg = (arg as NamedExpression).expression;
+        }
+        final expressionValue = context.evaluateConstant(arg);
+        if (expressionValue.value == value) {
+          rule.reportLint(arg);
+        }
+      }
+    }
+  }
+
+  @override
+  void visitInstanceCreationExpression(InstanceCreationExpression node) {
+    check(node.argumentList, node.staticElement?.parameters);
+  }
+
+  @override
+  void visitMethodInvocation(MethodInvocation node) {
+    final element = node.staticInvokeType.element;
+    List<ParameterElement> parameters;
     if (element is MethodElement) {
       parameters = element.parameters;
     }
     if (element is FunctionElement) {
       parameters = element.parameters;
     }
-    if (parameters == null) {
-      return;
-    }
 
-    for (var arg in node.argumentList.arguments) {
-      if (arg is NamedExpression) {
-        for (var param in parameters) {
-          if (param.name == arg.name.label.name) {
-            final expression = arg.expression;
-            final value = param.constantValue;
-            if (value != null) {
-              final expressionValue = context.evaluateConstant(expression);
-              if (expressionValue.value == value) {
-                rule.reportLint(arg);
-              }
-            }
-          }
-        }
-      }
-    }
+    check(node.argumentList, parameters);
   }
 }
--- a/lib/src/rules/avoid_redundant_argument_values.dart
+++ b/lib/src/rules/avoid_redundant_argument_values.dart
@@ -21,7 +21,7 @@ void f({bool valWithDefault = true, bool val}) {
 }
 
 void main() {
-  f(valWithDefault = true);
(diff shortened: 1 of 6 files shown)


lib/src/rules/avoid_redundant_parameter_values.dart
```diff
@@ -0,0 +1,99 @@
...
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/element.dart';
+
+import '../analyzer.dart';
+
+const _desc = r'Avoid redundant parameter values.';
+
+const _details = r'''Avoid redundant parameter values.
+
+**DON'T** declare parameters with values that match their defaults.
```
kevmoo: Wouldn't be `arguments`.

Parameter: the thing the function AUTHOR defines.
Argument: the thing the function CALLER provides.

Right? @munificent who learned me this.

bwilkerson: Correct. The rule should probably be `avoid_redundant_argument_values`, and most uses of "parameter" in the comments should be "argument".

lib/src/rules/avoid_redundant_parameter_values.dart
```diff
@@ -0,0 +1,99 @@
...
+
+**BAD:**
+```
+void f({bool valWithDefault = true, bool val}) {
+  ...
+}
+
+void main() {
+  f(valWithDefault = true);
```
bwilkerson: This needs to be `f(valWithDefault: true);`.

lib/src/rules/avoid_redundant_parameter_values.dart
```diff
@@ -0,0 +1,99 @@
...
+```
+
+**GOOD:**
+```
+void f({bool valWithDefault = true, bool val}) {
+  ...
+}
+
+void main() {
+  f(valWithDefault = false);
```
bwilkerson: Ditto.

lib/src/rules/avoid_redundant_parameter_values.dart
```diff
@@ -0,0 +1,99 @@
...
+            name: 'avoid_redundant_parameter_values',
+            description: _desc,
+            details: _details,
+            group: Group.style);
+
+  @override
+  void registerNodeProcessors(NodeLintRegistry registry,
+      [LinterContext context]) {
+    final visitor = _Visitor(this, context);
+    registry.addMethodInvocation(this, visitor);
```
bwilkerson: This should probably also visit `InstanceCreationExpression` and `FunctionExpressionInvocation`.

lib/src/rules/avoid_redundant_parameter_values.dart
```diff
@@ -0,0 +1,99 @@
...
+    if (element is FunctionElement) {
+      parameters = element.parameters;
+    }
+    if (parameters == null) {
+      return;
+    }
+
+    for (var arg in node.argumentList.arguments) {
+      if (arg is NamedExpression) {
+        for (var param in parameters) {
```
bwilkerson: Rather than looping through the parameters for each argument, you could use `Expression.staticParameterElement` to access the parameter directly. If you don't want to do that, you could at least exit this for loop when you've found the matching parameter.

lib/src/rules/avoid_redundant_parameter_values.dart
```diff
@@ -0,0 +1,99 @@
...
+    }
+    if (element is FunctionElement) {
+      parameters = element.parameters;
+    }
+    if (parameters == null) {
+      return;
+    }
+
+    for (var arg in node.argumentList.arguments) {
+      if (arg is NamedExpression) {
```
bwilkerson: Consider generalizing this lint to cover optional positional parameters:

void f(int x, [int y = 0]) {}
int z = f(0, 0);

lib/src/rules/avoid_redundant_parameter_values.dart
```diff
@@ -0,0 +1,99 @@
...
+        for (var param in parameters) {
+          if (param.name == arg.name.label.name) {
```
bwilkerson: Consider checking to see whether the parameter has an `@required` annotation. Under NNBD required parameters won't be allowed to have a default value so the check won't need to cover that case, but with the annotation they can.

lib/src/rules/avoid_redundant_parameter_values.dart
pq: Spot on.  Thanks for the catch!

lib/src/rules/avoid_redundant_parameter_values.dart
pq: Done.  Thanks!

lib/src/rules/avoid_redundant_parameter_values.dart
pq: Done.  👍

lib/src/rules/avoid_redundant_parameter_values.dart
pq: Great idea.  Thanks.

lib/src/rules/avoid_redundant_parameter_values.dart
pq: Good call.  Done!

lib/src/rules/avoid_redundant_parameter_values.dart
pq: Good catch.

pq:
Some improvements. 

@bwilkerson: can you point me to a `FunctionExpressionInvocation` test case?

lib/src/rules/avoid_redundant_argument_values.dart
```diff
@@ -8,11 +8,11 @@ import 'package:analyzer/dart/element/element.dart';
...
 import '../analyzer.dart';
 
-const _desc = r'Avoid redundant parameter values.';
+const _desc = r'Avoid redundant argument values.';
 
-const _details = r'''Avoid redundant parameter values.
+const _details = r'''Avoid redundant argument values.
 
-**DON'T** declare parameters with values that match their defaults.
+**DON'T** declare argument with values that match their defaults.
```
bwilkerson: nit: Perhaps "their defaults" --> "the defaults for the corresponding parameter"?

pq: 👍

bwilkerson:
> can you point me to a `FunctionExpressionInvocation` test case?

Sure. A `FunctionExpressionInvocation` node is used anywhere that a non-identifier expression is used to compute the function to be invoked. For example:
```
class C {
  void Function(int) m() => null;
}
void f(C c) {
  c.m()(3); // Invoke the result of invoking m.
}
```

pq:
As discussed, function types do not allow default values so I think `FunctionExpressionInvocation`s can be taken off the table.

@bwilkerson: how does this look otherwise?

Issue #1896, opened 2019-12-16
consider analysis feedback for enabled conflicting rules


pq:
See https://github.com/dart-lang/linter/issues/1895 for an example (`unnecessary_finals` vs. `prefer_final_locals`).

/fyi @bwilkerson

pq:
(Note to self: we might consider capturing, and at least documenting, these correspondences when we do "the big NNBD audit".)

Implementation bike-shed.

```dart

@incompatibleWith('unnecessary_finals')
class PreferFinalLocals extends LintRule implements NodeLintRule {
  ...
}
```

bwilkerson:
I think that would be a good addition.

pq:
Update: `LintRule`s now allow for specifying `incompatibleRules` which is enforced by DAS.

Next up:

- [ ] add validation to ensure `incompatibleRules` are properly defined in linter (e.g.,  describe actual rules and reference back to each other)
- [ ] audit rules and update incompatible ones

pq:
Closing in favor of specific follow-on issues (#1964 and #1967).

PR #1909, opened 2019-12-26
@IncompatibleWith

pq:
Bike-shedding for #1896.

/cc @bwilkerson

### DIFF
--- a/lib/src/meta.dart
+++ b/lib/src/meta.dart
@@ -3,6 +3,7 @@
 // BSD-style license that can be found in the LICENSE file.
 
 /// Linter metadata annotations.
+library linter_meta;
 
 /// Used to annotate a lint rule `r` whose semantics are incompatible with
 /// a given list of rules `r1`...`rn`.
@@ -27,7 +28,7 @@
 ///     }
 ///
 class IncompatibleWith {
-  /// A list of incompatible rules.
+  /// A list of the ids of incompatible rules.
   final List<String> rules;
 
   /// Initialize a newly created instance to have the given [rules].
--- a/lib/src/meta.dart
+++ b/lib/src/meta.dart
@@ -1,0 +1,35 @@
+// Copyright (c) 2019, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// Linter metadata annotations.
+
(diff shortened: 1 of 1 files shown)


lib/src/meta.dart
```diff
@@ -0,0 +1,35 @@
+// Copyright (c) 2019, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// Linter metadata annotations.
```
bwilkerson: nit: You need a directive in order to differentiate between library docs and docs on a member.

lib/src/meta.dart
```diff
@@ -0,0 +1,35 @@
...
+/// their incompatibility with `r`.  In this case, `'unnecessary_finals'` would
+/// look like this:
+///
+///     @IncompatibleWith(['prefer_local_finals'])
+///     class UnnecessaryFinals extends LintRule implements NodeLintRule {
+///         ...
+///     }
+///
+class IncompatibleWith {
+  /// A list of incompatible rules.
```
bwilkerson: Or, more precisely "of the ids of incompatible rules."

lib/src/meta.dart
```diff
@@ -0,0 +1,35 @@
...
+/// For example, the incompatibility between `prefer_local_finals` and
+/// `unnecessary_finals` could be captured in the `PreferFinalLocals` class
+/// declaration like this:
+///
+///     @IncompatibleWith(['unnecessary_finals'])
+///     class PreferFinalLocals extends LintRule implements NodeLintRule {
+///         ...
+///     }
+///
+/// For consistency of documentation, incompatibility should be declared in both
```
bwilkerson: We should have a test that ensures that this condition is being met.

pq: Absolutely.

lib/src/meta.dart
pq: Yes.   I suppose this is dangling.  (Call for another warning perhaps?  🤔)

I thought I recalled library directives falling out of favor...  Besides cooking up a library name, is there another preferred convention?

bwilkerson: > I thought I recalled library directives falling out of favor ...

They are. Any directive will do, but in a library like this one that doesn't need any directives I don't know what else to do. It seems silly to add a `import 'dart:core';` to avoid adding a library directive.

PR #1911, opened 2019-12-26
IncompatibleWith validation

pq:
Follow-up from #1909.

/cc @bwilkerson

### DIFF
--- a/lib/src/util/lint_cache.dart
+++ b/lib/src/util/lint_cache.dart
@@ -1,0 +1,117 @@
+// Copyright (c) 2019, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:io';
+
+import 'package:analyzer/dart/analysis/analysis_context_collection.dart';
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/element.dart';
+import 'package:analyzer/file_system/physical_file_system.dart';
+
+import '../analyzer.dart';
+import '../rules.dart';
+import '../utils.dart';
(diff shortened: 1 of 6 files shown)


lib/src/util/lint_cache.dart
```diff
@@ -0,0 +1,117 @@
...
+  Future<void> init() async {
+    registerLintRules();
+
+    // Setup details.
+    for (var lint in Analyzer.facade.registeredRules) {
+      details.add(LintRuleDetails(lint.runtimeType.toString(), lint.name));
+    }
+
+    // Process compatibility annotations.
+    final rulePath = File('lib/src/rules').absolute.path;
```
bwilkerson: Alternatively, we could analyze and visit the constructor invocations in `lib/src/rules.dart`. That would probably find all of the lint rule class element faster (although it would miss classes that aren't registered).

pq:
(Hmmmm, looks like I've got some thinking to do to make synchrony work.)

lib/src/util/lint_cache.dart
```diff
@@ -0,0 +1,117 @@
...
+  @override
+  void visitClassDeclaration(ClassDeclaration node) {
+    final classElement = node.declaredElement;
+    if (classElement == null) {
+      return;
+    }
+
+    for (var superType in classElement.allSupertypes) {
+      if (superType.name == _LINT_RULE_CLASS_NAME) {
+        classElements.add(classElement);
```
bwilkerson: If we don't switch to visiting a single file, then consider exiting the loop after the first match is found.

pq: Great idea.

lib/src/util/lint_cache.dart
pq: Let's discuss refining this together.  I have a few other ideas but a whiteboard would make them easier.

I'll land this as a jumping off point.

PR #1913, opened 2019-12-27
add use_key_in_widget_constructors

a14n:


lib/src/rules.dart
```diff
@@ -2,6 +2,7 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+
```
bwilkerson: Consider removing the extra blank line.

lib/src/rules/use_key_in_widget_constructors.dart
```diff
@@ -0,0 +1,95 @@
...
+class _Visitor extends SimpleAstVisitor<void> {
+  final LintRule rule;
+
+  _Visitor(this.rule);
+
+  @override
+  void visitClassDeclaration(ClassDeclaration node) {
+    var classElement = node.declaredElement;
+    if (classElement.isPublic &&
+        !classElement.isAbstract &&
```
bwilkerson: Why are abstract classes exempt? And if it's appropriate to allow an abstract class that has no explicit constructor (this condition), should we also allow an abstract class that has an explicit constructor (the condition starting on line 73)? The two conditions are different and I don't understand the reason for the difference. If the difference is appropriate, consider adding a comment explaining it.

Also, there should be some tests covering abstract classes.

test/rules/use_key_in_widget_constructors.dart
```diff
@@ -0,0 +1,26 @@
...
+
+class NoConstructorWidget extends StatefulWidget { // LINT
+}
+
+class _PrivateWidget extends StatefulWidget { // OK
+}
+
+class MyWidget extends StatelessWidget {
+  MyWidget(); // LINT
+  MyWidget.withKey({Key key}); // OK
```
bwilkerson: Do we want to verify in this case that `key` is being passed to the superclass?

test/rules/use_key_in_widget_constructors.dart
```diff
@@ -0,0 +1,26 @@
...
+  MyWidget.withKeyRequired(Key key); // LINT
+  MyWidget.withKeyOptional([Key key]); // LINT
```
bwilkerson: Out of curiosity, is there a reason (other than convention) why it's considered wrong for `key` to be something other than a named parameter?

bwilkerson:
@devoncarew There was some recent discussion about generating `key` parameters in widget constructors, but I didn't follow the discussion closely enough to understand it. What did we end up deciding and what are your thoughts about this lint?

pq:
fyi @daveshuckerow

lib/src/rules/use_key_in_widget_constructors.dart
a14n: This condition was added to handle [ObstructingPreferredSizeWidget](https://github.com/flutter/flutter/blob/bc0d35c34e2065fd520f8b0bcd959af86affc20a/packages/flutter/lib/src/cupertino/page_scaffold.dart#L203-L209) that is use only once as an interface in flutter repo. I will take a closer look and see if it's worth to keep this exception.

test/rules/use_key_in_widget_constructors.dart
a14n: no other reason than convention

test/rules/use_key_in_widget_constructors.dart
a14n: we could.

lib/src/rules/use_key_in_widget_constructors.dart
```diff
@@ -0,0 +1,95 @@
...
+import '../analyzer.dart';
+import '../util/dart_type_utilities.dart';
+import '../util/flutter_utils.dart';
+
+const _desc = r'Use key in widget constructors.';
+
+const _details = r'''
+**DO** use key in widget constructors.
+
+It's a good pratice to expose the ability to provide a key when creating public
```
devoncarew: sp: practice

devoncarew:
> @devoncarew There was some recent discussion about generating `key` parameters in widget constructors, but I didn't follow the discussion closely enough to understand it. What did we end up deciding and what are your thoughts about this lint?

@a14n, is this lint intended for general use or only for the Flutter framework repo?

From the docs: `It's a good practice to expose the ability to provide a key when creating public widgets`; I think most user widgets will be declared as public - they won't be as worried about visibility in their app code; do we want to warn for most of their widgets if the widget constructors don't have a `key` arg?

I don't know how common or uncommon use of the `key` param is in user widgets. When writing Flutter code myself, I spent some time removing the `key` param automatically added by the `extract widget` refactoring operation (extracting existing build method code into a new widget).

cc @Hixie, @HansMuller, @goderbauer to see if they have opinions about frequency of flutter widgets declaring the `key` arg

Hixie:
I don't have an opinion on whether it's on by default for developers. It's something I would turn on on my own projects, certainly. It's very good practice to always have a key. I'd even put it on private widgets, for completeness (though hopefully the compiler would optimize that out...).

goderbauer:
Yeah, we should turn this on for the flutter repo by default since it is good practice. However, I wouldn't turn this on by default for our users as they are already complaining about "too much boilerplate" and this would probably fall in the same category for them.

lib/src/rules/use_key_in_widget_constructors.dart
pq: @a14n: did you take a look at this?

pq:
@a14n: if you want to address the review comments, I feel good about landing this.

👍

lib/src/rules/use_key_in_widget_constructors.dart
a14n: yes, I change the rule to only look at widgets extending Widget (not implementing)

lib/src/rules/use_key_in_widget_constructors.dart
a14n: done

a14n:
I changed the rule to lint constructor that done use the key provided by super constructor or redirecting constructor.

In practice (on flutter repo) it seems to work as intended.

pq:
It looks like your'e hitting a stackoverflow when benchmarking?


```
#0      _Flutter._isExactWidget (package:linter/src/util/flutter_utils.dart:114:3)
#1      _Flutter.hasWidgetAsAscendant (package:linter/src/util/flutter_utils.dart:105:9)
#2      _Flutter.hasWidgetAsAscendant (package:linter/src/util/flutter_utils.dart:108:12)
```

https://travis-ci.org/dart-lang/linter/jobs/633496482?utm_medium=notification&utm_source=github_status

a14n:
Yep :-( Seems related to a cycle in type hierarchy in some files...

lib/src/util/flutter_utils.dart
```diff
@@ -90,6 +98,19 @@ class _Flutter {
     return false;
   }
 
+  bool hasWidgetAsAscendant(ClassElement element,
+      [List<ClassElement> alreadySeen]) {
```
bwilkerson: Consider using a `Set`. I don't know how deep the widget class hierarchy normally gets, so it might not matter, but containment might be faster to compute using a set.

PR #1928, opened 2020-01-03
check for trailing optional positional args in avoid_redundant_argument_values

pq:
Follow-up from: https://dart-review.googlesource.com/c/sdk/+/130160.

/cc @bwilkerson

### DIFF
--- a/lib/src/rules/avoid_redundant_argument_values.dart
+++ b/lib/src/rules/avoid_redundant_argument_values.dart
@@ -71,10 +71,10 @@ class _Visitor extends SimpleAstVisitor {
     for (var i = arguments.length - 1; i >= 0; --i) {
       var arg = arguments[i];
       final param = arg.staticParameterElement;
-      if (param.isRequiredPositional) {
-        break;
-      } else if (param == null || param.hasRequired || param.isRequiredNamed) {
+      if (param == null || param.hasRequired || param.isRequiredNamed) {
         continue;
+      } else if (param.isRequiredPositional) {
+        break;
       }
       final value = param.constantValue;
       if (value != null) {
--- a/lib/src/rules/avoid_redundant_argument_values.dart
+++ b/lib/src/rules/avoid_redundant_argument_values.dart
@@ -85,7 +85,7 @@ class _Visitor extends SimpleAstVisitor {
           for (var j = i + 1; j < arguments.length && okToRemove; ++j) {
             final param = arg.staticParameterElement;
             // todo (pq): consider instead starting from the end and reporting if the last optional positional is removable
-            if (param.isOptionalPositional) {
+            if (param != null && param.isOptionalPositional) {
               okToRemove = false;
             }
           }
--- a/lib/src/rules/avoid_redundant_argument_values.dart
+++ b/lib/src/rules/avoid_redundant_argument_values.dart
@@ -4,7 +4,6 @@
 
 import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
-import 'package:analyzer/dart/element/element.dart';
 
 import '../analyzer.dart';
 
--- a/lib/src/rules/avoid_redundant_argument_values.dart
+++ b/lib/src/rules/avoid_redundant_argument_values.dart
@@ -84,9 +84,10 @@ class _Visitor extends SimpleAstVisitor {
         final expressionValue = context.evaluateConstant(arg);
         if (expressionValue.value == value) {
           rule.reportLint(arg);
-        } else if (param.isOptionalPositional) {
-          break;
         }
+      }
+      if (param.isOptionalPositional) {
+        break;
       }
     }
   }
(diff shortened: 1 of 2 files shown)


lib/src/rules/avoid_redundant_argument_values.dart
```diff
@@ -79,7 +81,17 @@ class _Visitor extends SimpleAstVisitor {
         }
         final expressionValue = context.evaluateConstant(arg);
         if (expressionValue.value == value) {
-          rule.reportLint(arg);
+          var okToRemove = true;
```
bwilkerson: It isn't entirely clear to me which semantics you're trying to implement here, but I don't think this implements what you wanted. Given the TODO, I think you're trying to simply disable the lint for optional positional arguments. If that's the case, then I think you can just extend the condition on 83 with `&& (!param.isOptionalPositional || i == arguments.length - 1)`, and revert this change other than keeping the TODO.

test/rules/avoid_redundant_argument_values.dart
```diff
@@ -56,4 +58,9 @@ void main() {
   gg(1, 0); //LINT
   gg(1, 1); //OK
   gg(1); //OK
+
+  ggg(
+      1, // OK the  - first argument is required so that we can provide the second argument.
+      3);
+  ggg(1, 2); // LINT
```
bwilkerson: Consider putting the arguments on separate lines so that we can identify which one the lint should be associated with (here and elsewhere).

lib/src/rules/avoid_redundant_argument_values.dart
pq: Thanks.  Updated as per our conversation.

test/rules/avoid_redundant_argument_values.dart
pq: Done.

pq:
@bwilkerson: I think this is fixed now.  PTAL?

lib/src/rules/avoid_redundant_argument_values.dart
```diff
@@ -63,14 +63,18 @@ class _Visitor extends SimpleAstVisitor {
   _Visitor(this.rule, this.context);
 
   void check(ArgumentList argumentList, List<ParameterElement> parameters) {
-    if (argumentList.arguments.isEmpty || parameters == null) {
+    final arguments = argumentList.arguments;
+    if (arguments.isEmpty || parameters == null) {
```
bwilkerson: This code no longer references the parameter `parameters`. Consider removing it.

lib/src/rules/avoid_redundant_argument_values.dart
```diff
@@ -80,6 +84,8 @@ class _Visitor extends SimpleAstVisitor {
         final expressionValue = context.evaluateConstant(arg);
         if (expressionValue.value == value) {
           rule.reportLint(arg);
+        } else if (param.isOptionalPositional) {
+          break;
```
bwilkerson: I think branch this needs to be moved outside the if starting on line 80. Otherwise I think the following case will incorrectly report a lint:
```
void f([int a = 0, int b]) {}
int r = f(0, 1);
```

test/rules/avoid_redundant_argument_values.dart
```diff
@@ -56,4 +58,10 @@ void main() {
   gg(1, 0); //LINT
   gg(1, 1); //OK
   gg(1); //OK
+
+  ggg(
+      1, // OK the  - first argument is required so that we can provide the second argument.
+      3);
+  ggg(1,  // LINT
```
bwilkerson: The code motion change above will cause this to not be a lint, but I think that's better.

lib/src/rules/avoid_redundant_argument_values.dart
pq: Done.

lib/src/rules/avoid_redundant_argument_values.dart
pq: Hmmm.  I'm not sure I'm following.  I've add this test case (and another) and things work as written.

Maybe you can point me to another test I can try?

bwilkerson: Oh, right. Of course that case worked. I just confused myself while writing the test case.

But I thought we'd decided that we should only create a lint for the last argument, and never for multiple positional arguments. The test case you already had creates lints for multiple arguments, and moving this code would fix that case (assuming it's broken; maybe I was confused about that too).

pq: > I thought we'd decided that we should only create a lint for the last argument

We did discuss that but I forgot.

I'll update. 👍

test/rules/avoid_redundant_argument_values.dart
pq: Agreed.

pq:
OK, one more try. I think this has the behavior we discussed. 🤞

lib/src/rules/avoid_redundant_argument_values.dart
```diff
@@ -63,14 +63,18 @@ class _Visitor extends SimpleAstVisitor {
   _Visitor(this.rule, this.context);
 
   void check(ArgumentList argumentList, List<ParameterElement> parameters) {
```
bwilkerson: Consider removing the `parameters` argument, given that it isn't referenced.

PR #1943, opened 2020-01-08
Recommend using runtimeType in hash_and_equals

micimize:
https://github.com/dart-lang/sdk/issues/36004#issuecomment-557958476

I based my early `==` overrides on the `hash_and_equals` docs, so I'm sure others have done the same

### DIFF
--- a/lib/src/rules/hash_and_equals.dart
+++ b/lib/src/rules/hash_and_equals.dart
@@ -39,7 +39,10 @@ class Better {
   Better(this.value);
 
   @override
-  bool operator ==(Object other) => other is Better && other.value == value;
+  bool operator ==(Object other) =>
+      other is Better &&
+      other.runtimeType == runtimeType &&
+      other.value == value;
 
   @override
   int get hashCode => value.hashCode;


micimize:
@googlebot I signed it!

pq:
@bwilkerson

bwilkerson:
I disagree with Hixie on this one. The use of `is` _can_ make an equals method asymmetric, but does not guarantee that it is. Note also https://github.com/dart-lang/sdk/issues/36004#issuecomment-469777523, which correctly states that using `runtimeType` is also not always the the right way to implement equality.

The truth of the matter is that defining equality correctly is hard and there is no single right way to do so. If there were, the equality would be defined by the language rather than being something that developers had to define.

If we want to update the docs to reflect that, then I think the better way to do so would be to have multiple different "good" examples to illustrate that there isn't a one-size-fits-all solution.

As for this specific change, there's no reason to check both `is` and `runtimeType`. If the runtime types are the same then `is` will aways return true.

micimize:
It causes asymmetry if the class is extended ([dartpad](https://dartpad.dartlang.org/1cbfdd048ebc45f2e726f59f74712171), [gist](https://gist.github.com/micimize/1cbfdd048ebc45f2e726f59f74712171)), but I do think it should be the default recommendation, as it is the safest.

`is` was just to cast the value and fit the original example's style. Maybe I could amend the PR to
```dart
class Better {
  final int value;
  Better(this.value);
  @override
  bool operator ==(Object other){
    if (other.runtimeType != runtimeType){
      return false;
    }
    final Better typedOther = other;
    return typedOther.runtimeType == runtimeType &&
      typedOther.value == value;
  }
  @override
  int get hashCode => value.hashCode;
}

/// This class is not extended, and can therefore use `is` rather than `runtimeType`
class _Better {
  final int value;
  _Better(this.value);
  @override
  bool operator ==(Object other) => other is _Better && other.value == value;
  @override
  int get hashCode => value.hashCode;
}
```
Seems a little long winded though

bwilkerson:
> It causes asymmetry if the class is extended (...), but I do think it should be the default recommendation, as it is the safest.

Are all of the "it"s above referring to the use of `is`, or are the second and third referring to `runtimeType`?

micimize:
@bwilkerson edited for clarification: 
> `is` causes asymmetry if the class is extended. I believe extending classes is fairly common, so I think `runtimeType` should be the default recommendation, as it is the safest.

Would my proposed changes be preferable?

micimize:
@bwilkerson cleaning up old PRs – any more thoughts on this?

bwilkerson:
No, no new thoughts. I still don't object to the change, but I think it would be better if the docs had more examples of good styles to point out to users there there isn't any single "right" way to write `equals`.

@pq Do you want to land this as-is?

pq:
I agree the docs could be improved and also that this doesn't need to wait in the meantime.  Thanks for your patience @micimize!

What's a prefered name and email for [`AUTHORS`](https://github.com/dart-lang/linter/blob/master/AUTHORS)?

pq (approved):

micimize:
Thanks @pq! – `Michael Joseph Rosenthal <rosenthalm93@gmail.com>` would be good if you want to add it for something this small.

PR #1947, opened 2020-01-09
update type name accesses

pq:
API updates.

/cc @scheglov @bwilkerson

### DIFF
--- a/lib/src/rules/null_closures.dart
+++ b/lib/src/rules/null_closures.dart
@@ -303,7 +303,7 @@ class _Visitor extends SimpleAstVisitor<void> {
         possibleMethods
             .lookup(NonNullableFunction(library, className, methodName));
 
-    var method = getMethod(type.element.library.name, type.name);
+    var method = getMethod(type.element.library.name, type.element?.name);
     if (method != null) {
       return method;
     }
@@ -313,7 +313,8 @@ class _Visitor extends SimpleAstVisitor<void> {
       return null;
     }
     for (var supertype in element.allSupertypes) {
-      method = getMethod(supertype.element.library.name, supertype.name);
+      method =
+          getMethod(supertype.element.library.name, supertype.element?.name);
       if (method != null) {
         return method;
       }
--- a/lib/src/rules/null_closures.dart
+++ b/lib/src/rules/null_closures.dart
@@ -303,7 +303,7 @@ class _Visitor extends SimpleAstVisitor<void> {
         possibleMethods
             .lookup(NonNullableFunction(library, className, methodName));
 
-    var method = getMethod(type.element.library.name, type.element?.name);
+    var method = getMethod(type.element.library.name, type.element.name);
     if (method != null) {
       return method;
     }
@@ -314,7 +314,7 @@ class _Visitor extends SimpleAstVisitor<void> {
     }
     for (var supertype in element.allSupertypes) {
       method =
-          getMethod(supertype.element.library.name, supertype.element?.name);
+          getMethod(supertype.element.library.name, supertype.element.name);
       if (method != null) {
         return method;
       }
--- a/lib/src/rules/prefer_int_literals.dart
+++ b/lib/src/rules/prefer_int_literals.dart
@@ -71,10 +71,10 @@ class _Visitor extends SimpleAstVisitor<void> {
     if (node is FunctionExpression) {
       var functDeclaration = node.parent;
       if (functDeclaration is FunctionDeclaration) {
-        return functDeclaration.returnType?.type?.name == 'double';
+        return functDeclaration.returnType?.type?.isDartCoreDouble == true;
       }
     } else if (node is MethodDeclaration) {
-      return node.returnType?.type?.name == 'double';
+      return node.returnType?.type?.isDartCoreDouble == true;
     }
     return false;
   }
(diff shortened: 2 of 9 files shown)


lib/src/rules/prefer_int_literals.dart
```diff
@@ -71,26 +71,26 @@ class _Visitor extends SimpleAstVisitor<void> {
...
   bool hasTypeDouble(Expression expression) {
     final parent = expression.parent;
     if (parent is ArgumentList) {
-      return expression.staticParameterElement?.type?.name == 'double';
+      return expression.staticParameterElement?.type?.isDartCoreDouble == true;
     } else if (parent is ListLiteral) {
       final typeArguments = parent.typeArguments?.arguments;
       return typeArguments?.length == 1 &&
-          typeArguments[0]?.type?.name == 'double';
+          typeArguments[0]?.type?.isDartCoreDouble == true;
```
scheglov: `.type` of a type arguments cannot be null, so could be just `typeArguments[0].type.isDartCoreDouble`

lib/src/rules/prefer_int_literals.dart
```diff
@@ -71,26 +71,26 @@ class _Visitor extends SimpleAstVisitor<void> {
...
     } else if (parent is NamedExpression) {
```
scheglov: Maybe call `hasTypeDouble(parent.parent)`

lib/src/rules/unrelated_type_equality_checks.dart
```diff
@@ -146,11 +144,10 @@ bool _hasNonComparableOperands(TypeSystem typeSystem, BinaryExpression node) {
       !(_isFixNumIntX(leftType) && _isCoreInt(rightType));
 }
 
-bool _isCoreInt(DartType type) =>
-    type.name == 'int' && type.element?.library?.name == _dartCoreLibraryName;
+bool _isCoreInt(DartType type) => type.isDartCoreInt;
 
 bool _isFixNumIntX(DartType type) =>
-    (type.name == 'Int32' || type.name == 'Int64') &&
+    (type.element?.name == 'Int32' || type.element?.name == 'Int64') &&
```
scheglov: I'd replaced this with `type is InterfaceType`, and then call `type.element.name` without `null` protection.
Maybe extract `element` into a local variable.

lib/src/rules/use_full_hex_values_for_flutter_colors.dart
```diff
@@ -54,7 +54,7 @@ class _Visitor extends SimpleAstVisitor {
   @override
   void visitInstanceCreationExpression(InstanceCreationExpression node) {
     if (node.staticElement?.library?.name == 'dart.ui' &&
-        node.staticElement?.returnType?.name == 'Color' &&
+        node.staticElement?.returnType?.element?.name == 'Color' &&
         node.staticElement?.name == '') {
```
scheglov: I'd replaced this code with a call to something more generic, like `isConstructorElement(ConstructorElement element, {@required String uriStr, @required String className, @required String constructorName})`.

It is better to check `ConstructorElement.enclosingElement`, which is `ClassElement` than `returnType.element`. Although we probably should override `ConstructorElement.returnType` to return `InterfaceType`.

lib/src/util/dart_type_utilities.dart
```diff
@@ -176,12 +176,12 @@ class DartTypeUtilities {
 
   static bool isClass(DartType type, String className, String library) =>
       type != null &&
```
scheglov: `is InterfaceType`, and then unprotected accesses.

lib/src/util/dart_type_utilities.dart
```diff
@@ -176,12 +176,12 @@ class DartTypeUtilities {
...
-      type.name == className &&
+      type.element?.name == className &&
       type.element?.library?.name == library;
 
   static bool isInterface(
           InterfaceType type, String interface, String library) =>
-      type.name == interface && type.element.library.name == library;
+      type.element?.name == interface && type.element.library.name == library;
```
scheglov: `type.element` cannot be `null`

lib/src/util/dart_type_utilities.dart
```diff
@@ -176,12 +176,12 @@ class DartTypeUtilities {
 
   static bool isClass(DartType type, String className, String library) =>
       type != null &&
-      type.name == className &&
+      type.element?.name == className &&
       type.element?.library?.name == library;
 
   static bool isInterface(
```
scheglov: I see very little difference between this and the previous function.
I'd leave one, called `isInterfaceType()`.

lib/src/rules/null_closures.dart
```diff
@@ -313,7 +313,8 @@ class _Visitor extends SimpleAstVisitor<void> {
       return null;
     }
     for (var supertype in element.allSupertypes) {
-      method = getMethod(supertype.element.library.name, supertype.name);
+      method =
+          getMethod(supertype.element.library.name, supertype.element?.name);
```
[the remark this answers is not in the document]
pq: Great.  Thanks!

lib/src/rules/prefer_int_literals.dart
pq: 👍

lib/src/rules/unrelated_type_equality_checks.dart
pq: 👍

lib/src/util/dart_type_utilities.dart
pq: 👍

lib/src/util/dart_type_utilities.dart
pq: Good catch.

lib/src/rules/prefer_int_literals.dart
```diff
@@ -71,26 +71,26 @@ class _Visitor extends SimpleAstVisitor<void> {
     if (node is FunctionExpression) {
       var functDeclaration = node.parent;
       if (functDeclaration is FunctionDeclaration) {
-        return functDeclaration.returnType?.type?.name == 'double';
+        return functDeclaration.returnType?.type?.isDartCoreDouble == true;
       }
     } else if (node is MethodDeclaration) {
-      return node.returnType?.type?.name == 'double';
+      return node.returnType?.type?.isDartCoreDouble == true;
```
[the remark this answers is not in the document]
pq: Agreed.

lib/src/rules/use_full_hex_values_for_flutter_colors.dart
pq: Good idea(s).  Updated.  Thanks!

lib/src/rules/null_closures.dart
```diff
@@ -303,7 +303,7 @@ class _Visitor extends SimpleAstVisitor<void> {
         possibleMethods
             .lookup(NonNullableFunction(library, className, methodName));
 
-    var method = getMethod(type.element.library.name, type.name);
+    var method = getMethod(type.element.library.name, type.element?.name);
```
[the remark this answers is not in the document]
pq: Yes.  That was overprotective.

Thanks.

PR #1950, opened 2020-01-10
Improve unrelated type logic

### LINKED ISSUE #1899: unrelated_type_equality_checks doesn't trigger on sibling types
Either the description is off, or the lint. Not sure which.

Reading the documentation for #unrelated_type_equality_checks says:

> Comparing references of a type where neither is a subtype of the other most likely will return false and might not reflect programmer's intent.

If we have a simplified version of the protobufEnum world:
```
class C {}

class A extends C {}

class B extends C {}

main() {
  print(A() == B());
}
```

Then A is not a subtype of B nor is B a subtype of A, and from the description it seems I would get a lint, but when I try it I don't.

If sibling types where considered different it would be useful for prottobuf enums, they should never be compared between different enum types, as the answer is always false, even if the underlying value is the same.

Related to internal issue: b/146122892

camsteffen:
If two types are compared where the two types inherit `==` from `Object`, it will trigger unrelated type lints unrelated_type_equality_checks, iterable_contains_unrelated_type and list_remove_unrelated_type.

Fixes #1899 (partially - the rule description could still use improvement)

### DIFF
--- a/lib/src/util/dart_type_utilities.dart
+++ b/lib/src/util/dart_type_utilities.dart
@@ -411,8 +411,12 @@ class DartTypeUtilities {
         // Otherwise, they might be related.
         return false;
       } else {
-        return leftElement.supertype?.isObject == true ||
-            leftElement.supertype != rightElement.supertype;
+        return [leftElement, rightElement].every((element) {
+          var eqClass = element
+              .lookUpConcreteMethod('==', element.library)
+              .enclosingElement;
+          return eqClass is ClassElement && eqClass.isDartCoreObject;
+        });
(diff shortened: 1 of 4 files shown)


bwilkerson:
I need more time to think through the implications of this change before deciding whether it's a positive change (though it's certainly an interesting proposal). I would also want to see what the impact of this change would be on some significant sized packages. Even if it's better but would produce a significant number of new lints for existing users we might need to consider how best to roll it out.

camsteffen:
Closing since this has been in limbo for 2+ years.

PR #1953, opened 2020-01-10
Fix unnecessary_final: don't fire for fields.

### LINKED ISSUE #1952: `unnecessary_final` false positives
`unnecessary_final` fires for final fields, which is not intended. I'll send a fix.

davidmorgan:
Fixes #1952

### DIFF
--- a/lib/src/rules/unnecessary_final.dart
+++ b/lib/src/rules/unnecessary_final.dart
@@ -50,7 +50,7 @@ class UnnecessaryFinal extends LintRule implements NodeLintRule {
     registry
       ..addFormalParameterList(this, visitor)
       ..addForStatement(this, visitor)
-      ..addVariableDeclaration(this, visitor);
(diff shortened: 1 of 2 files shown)


lib/src/rules/unnecessary_final.dart
```diff
@@ -85,9 +85,9 @@ class _Visitor extends SimpleAstVisitor<void> {
   }
 
   @override
-  void visitVariableDeclaration(VariableDeclaration node) {
-    if (node.isFinal) {
-      rule.reportLint(node.name);
+  void visitVariableDeclarationStatement(VariableDeclarationStatement node) {
+    if (node.variables.isFinal) {
+      rule.reportLint(node.variables);
```
bwilkerson: Unlike the previous implementation, this has the potential of highlighting a significantly long range of characters. Consider changing this (and other places the lint is reported for the sake of consistency) to only highlight the keyword `final`.

PR #1957, opened 2020-01-13
Improve prefer_single_quotes message.

### LINKED ISSUE #1954: Suggestion for `prefer_single_quotes` message
Suggestion from an internal user.

The message "Prefer single quotes where they won't require escape sequences." is confusing if your string needs escape sequences for anything other than the single quote character. What is really meant is "Prefer single quotes where they won't require escape sequences for single quotes." ... but this is awkward because of the repetition.

How about: "Only use double quotes if the string contains single quotes."

?

davidmorgan:
Fixes #1954 

Changed slightly from the text on the issue, which was

`Only use double quotes if the string contains single quotes.`

in this PR I use

`Only use double quotes for strings containing single quotes.`

but, happy to tweak it either way :)

Thanks.

### DIFF
--- a/lib/src/rules/prefer_single_quotes.dart
+++ b/lib/src/rules/prefer_single_quotes.dart
@@ -8,8 +8,7 @@ import 'package:meta/meta.dart';
 
 import '../analyzer.dart';
 
-const _desc =
-    r"Only use double quotes for strings containing single quotes.";
+const _desc = r'Only use double quotes for strings containing single quotes.';
 
 const _details = '''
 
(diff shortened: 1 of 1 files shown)


pq:
@davidmorgan: it looks like `dartfmt` needs to be run on this one before committing.

Thanks!

lib/src/rules/prefer_single_quotes.dart
```diff
@@ -9,7 +9,7 @@ import 'package:meta/meta.dart';
 import '../analyzer.dart';
 
 const _desc =
-    r"Prefer single quotes where they won't require escape sequences.";
+    r"Only use double quotes for strings containing single quotes.";
```
pq: Make this single quotes while you're here?

😄

davidmorgan: Done.

PR #1963, opened 2020-01-15
generate scorecard on travis

pq:
/cc @bwilkerson


tool/crawl.dart
```diff
@@ -176,20 +176,27 @@ Future<List<String>> _fetchSdkTags() {
...
+      // Filter on bottom.
+      try {
+        var version = Version.parse(t);
+        return version.compareTo(bottomDartSdk) >= 0;
+      } on FormatException {
+        return false;
+      }
+    }).toList();
+  } on Exception catch (e) {
+    print('exception caught fetching SDK tags');
```
bwilkerson: While I'm sure it's annoying when a build fails for this reason, I don't know whether making this change is the best solution.

How visible is this? Will we notice right away, or is this something that could go on for months without anyone realizing it?

How often does this happen?

pq: It's an unpredictable error that has to do with API throttling on Github.  The only impact is that the scorecard will not correlate newly added lints to the latest SDKs.  (More incentive to keep the caches up to date.)

Another alternative would be to just skip this on travis and generate a less detailed scorecard in the build.

pq:
No sweat.  Thanks for asking! 😄

PR #1966, opened 2020-01-15
tag incompatible rules (finals and quotes)

pq:
/cc @bwilkerson @davidmorgan


davidmorgan:
How about `omit_local_variable_types` and `always_specify_types`?

Thanks :)

PR #1969, opened 2020-01-16
no_runtimeType_toString

a14n:
# Description

Add a lint `no_runtimeType_toString` for Flutter

Related issues:

* flutter/flutter#31695
* flutter/flutter#48607
* flutter/flutter#48892

### DIFF
--- a/lib/src/rules/avoid_runtimeType_toString.dart
+++ b/lib/src/rules/avoid_runtimeType_toString.dart
@@ -1,0 +1,103 @@
+// Copyright (c) 2020, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+
+import '../analyzer.dart';
+
+const _desc = r'Avoid calling toString() on runtimeType.';
+
+const _details = r'''
+
+Calling `toString` on a runtime type is a non-trivial operation that can
+negatively impact performance. It's better to avoid it.
+
+**BAD:**
+```
+class A {
+  String toString() => '$runtimeType()';
+}
+```
+
+**GOOD:**
+```
+class A {
+  String toString() => 'A()';
+}
+```
+
+This lint has some exceptions where performance is not a problem or where real
+type information is more important than performance:
+
+* in assertion
+* in throw expressions
+* in catch clauses
+* in mixin declaration
+* in abstract class
+
+''';
+
+class AvoidRuntimeTypeToString extends LintRule implements NodeLintRule {
+  AvoidRuntimeTypeToString()
+      : super(
+            name: 'avoid_runtimeType_toString',
+            description: _desc,
+            details: _details,
+            group: Group.style);
+
+  @override
+  void registerNodeProcessors(NodeLintRegistry registry,
+      [LinterContext context]) {
+    final visitor = _Visitor(this);
+    registry.addInterpolationExpression(this, visitor);
+    registry.addMethodInvocation(this, visitor);
+  }
+}
+
+class _Visitor extends SimpleAstVisitor<void> {
+  final LintRule rule;
+
+  _Visitor(this.rule);
+
+  @override
+  void visitMethodInvocation(MethodInvocation node) {
+    if (_canSkip(node)) {
+      return;
+    }
+    if (node.methodName.name == 'toString' &&
+        _isRuntimeTypeAccess(node.target)) {
+      rule.reportLint(node.target);
+    }
+    super.visitMethodInvocation(node);
+  }
+
+  @override
+  void visitInterpolationExpression(InterpolationExpression node) {
+    if (_canSkip(node)) {
+      return;
+    }
+    if (_isRuntimeTypeAccess(node.expression)) {
+      rule.reportLint(node.expression);
+    }
+    super.visitInterpolationExpression(node);
+  }
+
+  bool _isRuntimeTypeAccess(Expression target) =>
+      target is PropertyAccess &&
+          target.target is ThisExpression &&
+          target.propertyName.name == 'runtimeType' ||
+      target is SimpleIdentifier && target.name == 'runtimeType';
+
+  bool _canSkip(AstNode node) =>
+      node.thisOrAncestorMatching((n) {
+        if (n is Assertion) return true;
+        if (n is ThrowExpression) return true;
+        if (n is CatchClause) return true;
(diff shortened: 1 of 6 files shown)


a14n:
I do think this lint should live in a Flutter analysis plugin but until flutter/flutter#28327 is addressed it's not possible.

lib/src/rules/avoid_runtimeType_toString.dart
```diff
@@ -0,0 +1,103 @@
...
+  _Visitor(this.rule);
+
+  @override
+  void visitMethodInvocation(MethodInvocation node) {
+    if (_canSkip(node)) {
+      return;
+    }
+    if (node.methodName.name == 'toString' &&
+        _isRuntimeTypeAccess(node.target)) {
+      rule.reportLint(node.target);
```
bwilkerson: I'd prefer having this diagnostic associated with `node.methodName`.

lib/src/rules/avoid_runtimeType_toString.dart
```diff
@@ -0,0 +1,103 @@
...
+
+  _Visitor(this.rule);
+
+  @override
+  void visitMethodInvocation(MethodInvocation node) {
+    if (_canSkip(node)) {
+      return;
+    }
+    if (node.methodName.name == 'toString' &&
+        _isRuntimeTypeAccess(node.target)) {
```
bwilkerson: It probably isn't common enough to worry about, but this should really be `node.realTarget` in order to handle cascades.

lib/src/rules/avoid_runtimeType_toString.dart
```diff
@@ -0,0 +1,103 @@
...
+      rule.reportLint(node.target);
+    }
+    super.visitMethodInvocation(node);
```
bwilkerson: There isn't any reason to invoke the overridden method here (or below).

lib/src/rules/avoid_runtimeType_toString.dart
```diff
@@ -0,0 +1,103 @@
...
+      rule.reportLint(node.expression);
+    }
+    super.visitInterpolationExpression(node);
+  }
+
+  bool _isRuntimeTypeAccess(Expression target) =>
+      target is PropertyAccess &&
+          target.target is ThisExpression &&
+          target.propertyName.name == 'runtimeType' ||
+      target is SimpleIdentifier && target.name == 'runtimeType';
```
bwilkerson: Again, it's likely to be uncommon, but this doesn't check for shadowing:
```
void f() {
  var runtimeType = 'C';
  print('$runtimeType'); // Shouldn't be flagged.
}
```

lib/src/rules/avoid_runtimeType_toString.dart
```diff
@@ -0,0 +1,103 @@
...
+    }
+    if (_isRuntimeTypeAccess(node.expression)) {
+      rule.reportLint(node.expression);
+    }
+    super.visitInterpolationExpression(node);
+  }
+
+  bool _isRuntimeTypeAccess(Expression target) =>
+      target is PropertyAccess &&
+          target.target is ThisExpression &&
```
bwilkerson: Should we also guard against the use of `super`:
```
print('${super.runtimeType}'):
```

test/rules/avoid_runtimeType_toString.dart
```diff
@@ -0,0 +1,39 @@
...
+  }
+}
+
+abstract class B {
+  void f() {
+    final s1 = '$runtimeType'; // OK
+  }
+}
+
+mixin C {
```
bwilkerson: What should the behavior be inside an extension? I'm guessing it depends on the type being extended.

lib/src/rules/avoid_runtimeType_toString.dart
```diff
@@ -0,0 +1,103 @@
...
+* in catch clauses
+* in mixin declaration
+* in abstract class
+
+''';
+
+class AvoidRuntimeTypeToString extends LintRule implements NodeLintRule {
+  AvoidRuntimeTypeToString()
+      : super(
+            name: 'avoid_runtimeType_toString',
```
bwilkerson: In internal discussions I think we've decided to stop using prefixes like "avoid_" or "prefer_" when naming lints (and are thinking about what it would take to rename existing lints). Phil can confirm.

lib/src/rules/avoid_runtimeType_toString.dart
a14n: Actually the issue is with the usage of `runtimeType` and I think it's better to associated this diagnotic with `runtimeType` like in the other case ( `'$runtimeType'`)

lib/src/rules/avoid_runtimeType_toString.dart
a14n: Done

lib/src/rules/avoid_runtimeType_toString.dart
pq: Yeah, where possible it's nice to avoid `avoid` and `prefer`,  etc. in the name moving forward.  `runtimeType_toString` might be a little vague though (maybe prefix w/ `no_`?)

lib/src/rules/avoid_runtimeType_toString.dart
a14n: Done.

In Flutter I didn't see such case. It's quite uncommon in my opinion.

lib/src/rules/avoid_runtimeType_toString.dart
a14n: Done

test/rules/avoid_runtimeType_toString.dart
a14n: Good question... I made the lint trigger for extension on non-abstract classes.

lib/src/rules/avoid_runtimeType_toString.dart
bwilkerson: > Actually the issue is with the usage of `runtimeType` ...

I don't believe that. :-) First, the description says:

> Calling `toString` on a runtime type ...

which clearly implies that the problem is with invoking `toString` on an instance of `Type`. Second, we're explicitly testing for `toString`. If the problem were with `runtimeType` then we wouldn't care how the result was being used.

lib/src/rules/avoid_runtimeType_toString.dart
bwilkerson: No, me either. Might not be worth it in this case. That said, I've seen users do lots of things I didn't imagine they'd do. :-)

lib/src/rules/avoid_runtimeType_toString.dart
a14n: Done

PR #1972, opened 2020-01-20
add missing_whitespace_between_adjacent_strings

a14n:
Related to flutter/flutter#48090

### DIFF
--- a/lib/src/rules/missing_whitespace_between_adjacent_strings.dart
+++ b/lib/src/rules/missing_whitespace_between_adjacent_strings.dart
@@ -14,8 +14,8 @@ const _details = r'''
 Add a trailing whitespace to prevent missing whitespace between adjacent
 strings.
 
-With long text splitted accross adjacent strings it's easy to forget a
-whitespace between strings.
+With long text split accross adjacent strings it's easy to forget a whitespace
+between strings.
 
 **BAD:**
 ```
@@ -94,20 +94,14 @@ class _Visitor extends RecursiveAstVisitor<void> {
         if (e is InterpolationString) return e.value;
         return '';
       }));
-    } else if (string is AdjacentStrings) {
-      return string.strings.any((e) => _visit(e, test));
     }
     throw ArgumentError('${string.runtimeType}: $string');
   }
 
-  bool _hasWhitespace(String value) =>
-      value.contains(eol) || value.contains(space);
-
-  bool _endsWithWhitespace(String string) =>
-      string.endsWith(eol) || string.endsWith(space);
-  bool _startsWithWhitespace(String string) =>
-      string.startsWith(eol) || string.startsWith(space);
+  bool _hasWhitespace(String value) => _whitespaces.any(value.contains);
+  bool _endsWithWhitespace(String value) => _whitespaces.any(value.endsWith);
+  bool _startsWithWhitespace(String value) =>
+      _whitespaces.any(value.startsWith);
 }
 
-const eol = '\u000A';
-const space = '\u0020';
+const _whitespaces = [' ', '\n', '\r', '\t'];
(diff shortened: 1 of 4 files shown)


lib/src/rules/missing_whitespace_between_adjacent_strings.dart
```diff
@@ -0,0 +1,113 @@
...
+import '../analyzer.dart';
+
+const _desc = r'Missing whitespace between adjacent strings.';
+
+const _details = r'''
+
+Add a trailing whitespace to prevent missing whitespace between adjacent
+strings.
+
+With long text splitted accross adjacent strings it's easy to forget a
```
bwilkerson: "splitted" --> "split"

lib/src/rules/missing_whitespace_between_adjacent_strings.dart
```diff
@@ -0,0 +1,113 @@
...
+
+  bool _visit(StringLiteral string, bool Function(Iterable<String>) test) {
+    if (string is SimpleStringLiteral) {
+      return test([string.value]);
+    } else if (string is StringInterpolation) {
+      return test(string.elements.map((e) {
+        if (e is InterpolationString) return e.value;
+        return '';
+      }));
+    } else if (string is AdjacentStrings) {
```
bwilkerson: The individual strings inside an `AdjacentString` should never themselves be adjacent strings, so I believe that this branch can't be reached.

lib/src/rules/missing_whitespace_between_adjacent_strings.dart
```diff
@@ -0,0 +1,113 @@
...
+    throw ArgumentError('${string.runtimeType}: $string');
+  }
+
+  bool _hasWhitespace(String value) =>
+      value.contains(eol) || value.contains(space);
+
+  bool _endsWithWhitespace(String string) =>
+      string.endsWith(eol) || string.endsWith(space);
+  bool _startsWithWhitespace(String string) =>
+      string.startsWith(eol) || string.startsWith(space);
```
bwilkerson: These seem a bit limited. Should we include both carriage return and newline? What about tab?

test/rules/missing_whitespace_between_adjacent_strings.dart
```diff
@@ -0,0 +1,21 @@
...
+  f('long $f line' // LINT
+    'is long');
+  f('long line' ' is long'); // OK
+  f('long line ' 'is long'); // OK
+  f('long $f line ' 'is long'); // OK
+  f('longLineWithoutSpaceCouldBe' 'AnURL'); // OK
+
+  f(RegExp('(\n)+' '(\n)+' '(\n)+')); // OK
+  matches('(\n)+' '(\n)+' '(\n)+'); // OK
+}
```
bwilkerson: I think we should not produce a lint for code like `'your' '$string' 'here'` because we can't know that the value of `string` doesn't start or end with a space.

lib/src/rules/missing_whitespace_between_adjacent_strings.dart
a14n: Added `\r` and `\t`

test/rules/missing_whitespace_between_adjacent_strings.dart
a14n: `'your' '$string' 'here'` will not be linted because the lint is trigger only if the _current string contains a whitespace.

However a lint will be triggered for `'yo ur' '$string' 'here'` and that's fine IMHO. The user can change his code to `yo ur${string}here`.

In practice (on flutter codebase) there were only one place I add a similar case were the variable carried a trailing space. If we hadn't lint those kind of case we had missed around 10 real problems.

bwilkerson: That's reasonable. We can see whether it causes issues in other code bases and adjust it if necessary.

pq:
Cool!  FWIW, I'll take a look at the release bot failures (https://github.com/dart-lang/linter/issues/1974).

PR #1973, opened 2020-01-20
Allow unused parameters when they are deprecated

bwilkerson:
The rule `avoid_unused_constructor_parameters` should not produce a lint for parameters that are deprecated because it's appropriate that they are not referenced.


bwilkerson:
I found an example of this in the Flutter code base. They'd been forced to add an ignore comment.

PR #1983, opened 2020-01-23
unnecessary_string_interpolations

a14n:
# Description

This lint catches `'$stringExpresion'` ; it can be replaced by `stringExpression` directly (modulo null handling)


a14n:
> Does this happen often?

See flutter/flutter#49352

bwilkerson:
Wow, I'm surprised. I wouldn't have expected it to be so prevalent.

PR #1996, opened 2020-02-04
Add unnecessary_escapes_in_string

a14n:
**BAD:**
```
'this string contains 2 \"double quotes\" ';
"this string contains 2 \'single quotes\' ";
```
**GOOD:**
```
'this string contains 2 "double quotes" ';
"this string contains 2 'single quotes' ";
```


lib/src/rules/unnecessary_escapes_in_string.dart
```diff
@@ -0,0 +1,100 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/token.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:meta/meta.dart';
+
+import '../analyzer.dart';
+
+const _desc = r'Remove unnecessary backslash in strings.';
```
bwilkerson: nit: "backslash" --> "backslashes"

a14n:
Interestingly this lint catched some issue in regexs on Flutter codebase.

test/rules/unnecessary_escapes_in_string.dart
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) 2020, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// test w/ `pub run test -N unnecessary_escapes_in_string`
+
+f(o){
```
bwilkerson: It would be good to add at least one example of a raw string and at least one example using a multi-line string.

a14n: I don't know how to mark a `// LINT` with multiline string :-(

a14n: done for raw string

lib/src/rules/unnecessary_escapes_in_string.dart
a14n: Done

test/rules/unnecessary_escapes_in_string.dart
bwilkerson: Multiline strings aren't required to contain end of line markers, so you could do something like
```
f('''_\'_'''); // LINT
f('''_\''''); // OK
f('''\'_'''); // LINT
f('''_\"_'''); // LINT

f("""_\"_"""); // LINT
f("""_\""""); // OK
f("""\"_"""); // LINT
f("""_\'_"""); // LINT
```

a14n: Sure my bad :)

a14n: Actually multiline strings are a little harder to handle.

```dart
f('''_\'_'''); // LINT
f('''_'\''_'''); // OK or it cause error
```

a14n: For now `'''_\'_'''` don't trigger a lint. Only `'''_\"_'''` triggers a lint.

bwilkerson:
```
f('''_'\''_'''); // OK or it cause error
```
That would need to be OK because removing the backslash would result in parse errors.

> For now `'''_\'_'''` don't trigger a lint.

I think that's reasonable, given the complexity of multi-line strings.

bwilkerson: I think it would still be good to add the test cases with comments explaining the desired outcome in cases where the rule doesn't currently produce that outcome. If nothing else we should add the cases that are handled correctly.

a14n: Finally I made the changes for multiline strings.

lib/src/rules/unnecessary_escapes_in_string.dart
```diff
@@ -0,0 +1,149 @@
...
+  @override
+  void visitStringInterpolation(StringInterpolation node) {
+    for (var element in node.elements.whereType<InterpolationString>()) {
+      visitLexeme(
+        element.contents,
+        isSingleQuoted: node.isSingleQuoted,
+        isMultiline: node.isMultiline,
+        // TODO(a14n): should be the following line but the values look buggy
+        // contentsOffset: element.contentsOffset,
+        // contentsEnd: element.contentsEnd,
```
a14n: I faced strange behaviour with the above code. It looks like a bug to me in the returned values.

For instance `"\'$f"`:

```
node:"\'$f"
node.offset:286
element:"\'
element.offset:286
element.contentsOffset:287   // OK
element.contentsEnd:288      // KO should be 289
element.end:289              // OK
element:"
element.offset:291           // OK
element.contentsOffset:292   // KO should be 291
element.contentsEnd:292      // KO should be 291
element.end:292              // OK
```

@bwilkerson WDYT?

bwilkerson: I agree, those values look wrong.

lib/src/rules/unnecessary_escapes_in_string.dart
```diff
@@ -0,0 +1,149 @@
...
+
+    final lexeme = token.lexeme
+        .substring(contentsOffset - token.offset, contentsEnd - token.offset);
+    for (var i = 0; i < lexeme.length; i++) {
+      var current = lexeme[i];
+      var escaped = false;
+      if (current == r'\') {
+        escaped = true;
+        i += 1;
+        current = lexeme[i];
```
bwilkerson: `i` could be past the end of the lexeme at this point.

test/rules/unnecessary_escapes_in_string.dart
bwilkerson: Thanks! I really was only concerned with not forgetting the corner cases, but it's great to have them handled.

lib/src/rules/unnecessary_escapes_in_string.dart
a14n: For non raw string I don't see how this could happen.

bwilkerson: Oh, right. I missed the fact that the last character of the content can't be a backslash because that would cause the terminating quote to be part of the string. Sorry about that.

lib/src/rules/unnecessary_escapes_in_string.dart
pq: @a14n: maybe open a bug on analyzer to track?  (fyi @stereotype441)

a14n:
> Is this good to land?

Yes on my side.

pq:
Thanks!

lib/src/rules/unnecessary_escapes_in_string.dart
a14n: Filed dart-lang/sdk#40495

PR #1999, opened 2020-02-06
add lint replace_escapes_by_raw_strings

a14n:
# Description

A raw string can be used to avoid escaping only backslashes and dollars.

**BAD:**
```dart
var s = 'A string with only \\ and \$';
```

**GOOD:**
```dart
var s = r'A string with only \ and $';
```

### DIFF
--- a/example/all.yaml
+++ b/example/all.yaml
@@ -127,6 +127,7 @@ linter:
     - provide_deprecation_message
     - public_member_api_docs
     - recursive_getters
+    - replace_escapes_by_raw_strings
     - slash_for_doc_comments
     - sort_child_properties_last
     - sort_constructors_first
--- a/example/all.yaml
+++ b/example/all.yaml
@@ -127,7 +127,6 @@ linter:
     - provide_deprecation_message
     - public_member_api_docs
     - recursive_getters
-    - replace_escapes_by_raw_strings
     - slash_for_doc_comments
     - sort_child_properties_last
     - sort_constructors_first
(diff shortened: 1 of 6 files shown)


example/all.yaml
```diff
@@ -127,6 +127,7 @@ linter:
     - provide_deprecation_message
     - public_member_api_docs
     - recursive_getters
+    - replace_escapes_by_raw_strings
```
pq: Wondering about the name...  Maybe something like `prefer_raw_strings`?  Or `unnecessary_escaped_strings`?

lib/src/rules/replace_escapes_by_raw_strings.dart
```diff
@@ -0,0 +1,72 @@
+// Copyright (c) 2020, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+
+import '../analyzer.dart';
+
+const _desc = r'Replace escaping by a raw string.';
```
pq: by => with

example/all.yaml
bwilkerson: I think we wanted to move away from using "prefer", etc. in the name. Perhaps `use_raw_strings`?

pq: Yeah.  I like `use_raw_strings` better anyway!

How does that sound @a14n ?

a14n: SGTM. Renamed

pq:
Hmmm, we can ignore the `score` bot failure.

Thanks for putting this together!

Issue #883, opened 2018-02-09
Allow lints to accept configurations


a14n:
It would be useful and nice to be able to customize some lints with parameter/configuration.

For instance:

```yaml
linter:
  rules:
    - file_headers:
        pattern: '// Copyright 2015 The Chromium Authors. All rights reserved.\n// Use of this source code isgoverned by a BSD-style license that can be\n// found in the LICENSE file.'
    - prefer_const_declarations:
        local_variables: false
```

a14n:
@bwilkerson @pq wdyt ?

bwilkerson:
Honestly, I have mixed feelings. On the one hand, this can be really useful and allow a much richer UX. But it's easy to abuse, and to end up with rules that are difficult to configure.

I worked on a similar product at a previous company, and we made the mistake of adding configuration support. We didn't have outside contributions, so we had no one to blame but ourselves, but we ended up with a product that users couldn't use until they had spent multiple hours going through the list of lints to decide whether they wanted to use the lint and if so how to configure it. The task was so daunting that many of our customers never realized the full potential value of the product.

So, yeah, it would be great to be able to specify the file header to a lint so that the quick fix could insert it when it's missing. On the other hand, I'm not convinced that it isn't better, for something like "include local variables in the lint" to just make them two separate lints and allow users to enable the functionality they want by choosing the lints they enable (something they have to do anyway).

pq:
I always thought we'd do this (see for example this old tracking issue: https://github.com/dart-lang/linter/issues/7) but as time has gone on I'm not as keen (for the reasons that Brian cites).

Very open to reconsidering though!

a14n:
I agree configuration should be an exception. And for a `file_header` lint I pointed above configuration is definitively mandatory.

jibiel:
@bwilkerson 
> But it's easy to abuse, and to end up with rules that are difficult to configure.

So the way to go as I see it is to introduce just a few super basic options to a linter rule which will not allow overconfiguring by design.

A good example is the [Ruby linter](https://docs.rubocop.org/rubocop/1.6/cops_layout.html) which introduces a few basic options for each rule: `Exclude`, `Enabled` and `EnforcedStyle`, and maybe 1 or 2 specific (but still basic) options depending on the rule.

Here's how it looks:

```yaml
Layout/AccessModifierIndentation:
  EnforcedStyle: outdent
  IndentationWidth: 0

Layout/LineLength:
  Exclude:
    - config/**/*
[...]

Issue #498, opened 2017-03-16
Make unnecessary_lambdas configurable to exclude some cases


a14n:
On the Flutter codebase [there are some cases](https://github.com/flutter/flutter/pull/8810#issue-214534304) we'd like to exclude [for readability reason](https://github.com/flutter/flutter/pull/8595#discussion_r104559527). It'd be great to make `unnecessary_lambdas` configurable to be able to exclude those methods/functions.

bwilkerson:
I'm not sure what configuration data you want to be able to supply. Could you give me an example?

a14n:
In [the following example](https://github.com/flutter/flutter/pull/8595#discussion_r104559527) :

```diff
-    setState(() {
-      popCalcExpression();
-    });
+    setState(popCalcExpression);
```

@Hixie argues:

>Please don't do this to setState. We're trying to make setState look like a block statement, where it's "obvious" that the statements inside the block are executed immediately. This change hides what's going on, it makes it look like we're taking a reference to call later or something.

So I'd like to tell to the lint `unnecessary_lambdas` to not provide lints for [`State.setState`](https://github.com/flutter/flutter/blob/e0f26b44f202b198a4057ea1e39dc07df3cbe1f1/packages/flutter/lib/src/widgets/framework.dart#L949). I donn't really know how to provide this reference. Perhaps something like that:

```yaml
  - unnecessary_lambdas:
      exclude:
      - library: package:flutter/src/widgets/framework.dart
        ref: State.setState
```

alexeieleusis:
Probably we need a more comprehensive recommendation in the Effective Dart guideline. One thing we can do is whitelist certain functions/methods from being linted as a temporary workaround.

I understand Flutter model is such that you want a localized place where mutation happens and most things outside it are immutable. So probably more annotations and analysis around things that are explicitly mutable or immutable would make static analysis easier and above all code easier to understand.

As we have recently reviewed rules implementation and analysis results we have found cases that although are like what the rule describes most likely don't need to be changed but there is no easy way to discard them.

I wouldn't go as far a suggesting to annotate to ease static analysis, which is not a goal by itself, but certainly it can help to make intent explicit.

bwilkerson:
My preference would be, if possible, to refine the rule so that it understands without further input which cases should be flagged and which shouldn't. This might be difficult, but if it's possible it's worth the effort.

[...]

PR #2000, opened 2020-02-07
add avoid_escaping_inner_quotes

a14n:
# Description

Change quotes to avoid escaping inner quotes.

**BAD:**

```dart
var s = 'It\'s not fun';
```

**GOOD:**

```dart
var s = "It's not fun";
```

### DIFF
--- a/example/all.yaml
+++ b/example/all.yaml
@@ -17,6 +17,7 @@ linter:
     - avoid_double_and_int_checks
     - avoid_empty_else
     - avoid_equals_and_hash_code_on_mutable_classes
+    - avoid_escaping_inner_quotes
     - avoid_field_initializers_in_const_classes
     - avoid_function_literals_in_foreach_calls
     - avoid_implementing_value_types
@@ -49,7 +50,6 @@ linter:
     - camel_case_types
     - cancel_subscriptions
     - cascade_invocations
-    - change_quotes
     - close_sinks
     - comment_references
     - constant_identifier_names
--- a/example/all.yaml
+++ b/example/all.yaml
@@ -49,6 +49,7 @@ linter:
     - camel_case_types
     - cancel_subscriptions
     - cascade_invocations
+    - change_quotes
     - close_sinks
     - comment_references
     - constant_identifier_names
--- a/lib/src/rules.dart
+++ b/lib/src/rules.dart
@@ -18,6 +18,7 @@ import 'rules/avoid_classes_with_only_static_members.dart';
 import 'rules/avoid_double_and_int_checks.dart';
 import 'rules/avoid_empty_else.dart';
 import 'rules/avoid_equals_and_hash_code_on_mutable_classes.dart';
+import 'rules/avoid_escaping_inner_quotes.dart';
 import 'rules/avoid_field_initializers_in_const_classes.dart';
 import 'rules/avoid_function_literals_in_foreach_calls.dart';
 import 'rules/avoid_implementing_value_types.dart';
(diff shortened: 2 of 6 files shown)


a14n:
I started with a single lint but splitted it for several reasons:

* the region highlighted is not the same: `\'` in single quoted strings hightlights the whole string to be converted to double string and `\e` highlight only the backslash.
* different fixes (convert string vs. remove backslash)
* simplier implementation
* finer lint tuning

PS: I also have a `unnecessary_raw_strings` lint coming.

pq:
> I have to wonder whether it's beneficial for users that we have so many lints dealing with escapes in strings

I feel the same way.

I think the best justification for multiple lints is when it's likely that folks might only agree with some of the advice.  On the surface, these feel like a bunch of variations on a theme and it's hard to imagine not wanting to enable them all.  Moreover,  I worry that having so many will actually hurt adoption (but that's just speculation).

> PS: I also have a unnecessary_raw_strings lint coming.

😎 This. one seems a little different I think and maybe should be it's own but how do you feel about one `unnecessary_escapes_in_string` for the others?

As for implementation, there are a few composite-like lints that could serve as inspiration (you probably even wrote some! 😉)   I have some ideas for what we can do on the fix end but will have to look a little closer before suggesting anything.

As always, thanks for the thoughtful back and forth!

bwilkerson:
The fixes are easy to deal with as long as the lint produces different codes for the different kinds of violations (which also allows for producing different messages).

a14n:
It's hard to imagine what users want but having small lints will cover more use cases that a single lint. IIRC there were a similar issue with `always_specify_types` that some users would like to split. 

Actually we could also have a meta-lint or tags to compose several lints into a single activation key (in analysis_options.yaml).

One could also argument that the agregation will be done by pedantic :)

bwilkerson:
> > PS: I also have a `unnecessary_raw_strings` lint coming.
>
> This one seems a little different I think and maybe should be it's own ...

Possibly. I tend to see them as all being related to trying to keep string literals as simple as possible:
- there shouldn't be backslashes that aren't needed, whether intrinsically or by changing the quote style, and
- strings shouldn't be raw when a non-raw string would be equivalent.

Perhaps `use_simple_string_literals` or `keep_string_literals_simple` or `complex_string_literals`?

a14n:
I initially had `simplifiable_strings` before splitting it. But there are many cases in addition to the above cases that could be lints on strings (unnecessary multiline string, mergeable adjacent strings that fit on the same line, consistant strings in adjacent strings... some of them can be incompatible)

>I tend to see them as all being related

I already had feedbacks that some people don't want unnecessary_raw_string because they like to always have raw for regexp (which is not my preference)

pq:
> One could also argument that the agregation will be done by pedantic :)

This is a good point.  Rulesets like this can help.

Having said that, I still think there's something to be said for limiting the cognitive load of lots of micro lints.

The more I think about it the more I'm inclined to consolidate these into a mega-lint as @bwilkerson proposed above.


Maybe this whole conversation is just further motivation to explore configurability in  general (as prompted by you in https://github.com/dart-lang/linter/issues/883 and https://github.com/dart-lang/linter/issues/498)?

pq:
> I already had feedbacks that some people don't want unnecessary_raw_string because they like to always have raw for regexp (which is not my preference)

😬 

OK, this is good to know.  Maybe this one _should be_ separate.  I'd still like explore consolidating the others?

a14n:
Renamed to `avoid_escaping_inner_quotes` according to https://github.com/dart-lang/linter/issues/2002#issuecomment-588253987

PR #2001, opened 2020-02-11
unnecessary_raw_strings

a14n:
# Description

Use raw string only when needed.

**BAD:**
```
var s1 = r'a';
```

**GOOD:**
```
var s1 = 'a';
var s2 = r'$a';
var s3 = r'\a';
```


a14n:
ping :)

PR #2003, opened 2020-02-13
Use DiagnosticableMixin instead of Diagnosticable

gspencergoog:
# Description

Flutter is planning to deprecate `Diagnosticable` in favor of `DiagnosticableMixin`, so this is just a preemptive change to make that deprecation a non-issue for the linter. This change will also make it possible for Flutter to change all of the classes that use "extends Diagnosticable" to switch to "with DiagnosticableMixin" without triggering the lint.

The implementation of `Diagnosticable` in Flutter is already using `DiagnosticableMixin`, even on the stable channel.

### DIFF
--- a/lib/src/rules/diagnostic_describe_all_properties.dart
+++ b/lib/src/rules/diagnostic_describe_all_properties.dart
@@ -17,7 +17,7 @@ const _desc = r'DO reference all public properties in debug methods.';
 const _details = r'''
 **DO** reference all public properties in `debug` method implementations.
 
-Implementers of `Diagnosticable` should reference all public properties in
+Implementers of `DiagnosticableMixin` should reference all public properties in
 a `debugFillProperties(...)` or `debugDescribeChildren(...)` method
 implementation to improve debuggability at runtime.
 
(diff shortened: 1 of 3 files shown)


bwilkerson:
Do we know when the switchover will take place? The one concern I have here is timing. If we merge this change and it gets rolled into a release of Flutter before clients have converted over it will cause their older code to no longer be linted. I don't know the severity of that event, or the likelihood, but wanted to make sure you're aware.

The alternative would be to lint classes that implement either interface until the old one has been removed.

gspencergoog:
This shouldn't be a problem: the stable channel for Flutter (and since 5/30/2019) already has `Diagnosticable` defined as:

```dart
abstract class Diagnosticable with DiagnosticableMixin {
  const Diagnosticable();
}
```

So the change shouldn't break anyone, it just broadens the lint to things that should have been linted already (classes that just use `with DiagnosticableMixin` without implementing `Diagnosticable` will now get the lint too).

bwilkerson:
> ... the stable channel for Flutter (and since 5/30/2019) already has Diagnosticable defined as ...

Right. You said that but I missed the implication. Sorry for the noise.

PR #2007, opened 2020-02-20
allow LINT to reference a relative line

a14n:
This PR allow to mark lints in test file on another line:

```dart
// LINT [+1] lint on the following line
var x = y;
// LINT [+3] lint 3 lines after this line
var x = '''


''';

var x = y;
// LINT [-1] lint on the above line
```

The previous pattern `// LINT [column:length] message` still works and you can combine them with `// LINT [lineDelta,column:length] message`

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -9,7 +9,7 @@ repository: https://github.com/dart-lang/linter
 documentation: https://dart-lang.github.io/linter/lints
 
 environment:
-  sdk: '>=2.2.2 <3.0.0'
+  sdk: '>=2.6.0 <3.0.0'
 
 dependencies:
   analyzer: ^0.39.3
--- a/test/rule_test.dart
+++ b/test/rule_test.dart
@@ -247,9 +247,8 @@ void testRule(String ruleName, File file,
 
     var lineNumber = 1;
     for (var line in file.readAsLinesSync()) {
-      var annotation = extractAnnotation(line);
+      var annotation = extractAnnotation(lineNumber, line);
       if (annotation != null) {
-        annotation.lineNumber = lineNumber;
         expected.add(AnnotationMatcher(annotation));
       }
       ++lineNumber;
--- a/test/util/annotation.dart
+++ b/test/util/annotation.dart
@@ -6,49 +6,28 @@ import 'package:analyzer/error/error.dart';
 import 'package:analyzer/src/generated/source.dart';
 import 'package:linter/src/analyzer.dart';
 
-Annotation extractAnnotation(String line) {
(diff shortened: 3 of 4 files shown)


pq:
This feels like it adds a lot of noise to support the uncommon case. I don't object to supporting it, but would prefer things stay simple where they can.

I wonder if your proposed support for `// LINT [lineDelta,column:length]` alone is sufficient?

a14n:
> I wonder if your proposed support for `// LINT [lineDelta,column:length]` alone is sufficient?

That would imply:

* to change 19 occurances from `// LINT [column:length]` to `// LINT [0,column:length]`
* having to specify column and number even if we only want to check a lint is on the following line (`// LINT [+1]` is enough)

pq:
Sorry, I though `lineDelta` was an optional specification and that the default behavior would work as it currently does.  I think that would be my preference.  But maybe we can pop up and discuss the motivation.  Could you remind me why you're wanting this?  A warning embedded in a multi-line string?  Do you anticipate lots of uses?

a14n:
>I though lineDelta was an optional specification

I'm not sure to understand what you mean so let me add some examples allowed with this PRs:

```dart
var x = ''; // LINT
var x = ''; // LINT with a message
var x = ''; // LINT [5:1]  old syntax allowed
// LINT [+1] lint in following line
var x = ''; 
// LINT [-1] lint in previous line
// LINT [+1,5:1] lint in following line at a given column with a length
var x = ''; 
// LINT [-1,5:1] lint in previous line at a given column with a length
```

>A warning embedded in a multi-line string?

Yes :)

An upcoming lint is to prefer newline to start multiline strings:

```dart
    // BAD
    var json = '''{
  "a": 1
}''';

    // GOOD
    var json = '''
{
  "a": 1
}''';
```

To test it today you have to create an integration test and it's not a simple task.

pq:
Sorry to have let this one lag.  If I understand this right, it's purely additive and will require *no* changes to existing tests?  That's cool.

I'm curious if @bwilkerson has any additional thoughts about API but it works for me.

(We should also probably document this somewhere...)

bwilkerson:
I'm ok with this change, but my personal preference, as I've said before, would be to change the testing framework to mirror what we do in `analyzer` and `analysis server`. Doing so would allow the tests to
- be finer grained (testing one thing in each test rather than everything in a single test),
- be more specific about where the lint should be reported, and
- be faster by avoiding reads from disk.

In addition, it would avoid the bug in analyzer that when you open a test file it gets analyzed and those errors never go away despite the test files being excluded from analysis.

PR #2022, opened 2020-03-02
add newline_to_start_multiline_strings

a14n:


lib/src/rules/newline_to_start_multiline_strings.dart
```diff
@@ -0,0 +1,91 @@
...
+const _desc = r'Start multiline strings with a newline.';
+
+const _details = r'''
+
+Multiline strings are easier to read when they start with a newline (a newline
+starting a multiline string is ignored).
+
+**BAD:**
+```
+var s1 = """{
```
bwilkerson: Consider using single quotes for the examples.

lib/src/rules/newline_to_start_multiline_strings.dart
```diff
@@ -0,0 +1,91 @@
...
+
+  @override
+  void registerNodeProcessors(NodeLintRegistry registry,
+      [LinterContext context]) {
+    final visitor = _Visitor(this);
+    registry.addCompilationUnit(this, visitor);
+  }
+}
+
+class _Visitor extends RecursiveAstVisitor<void> {
```
bwilkerson: This should extend `SimpleAstVisitor` and register for the other visit methods that need to be called. The way it's currently written we'll end up visiting the AST structure twice, which is not very efficient.

lib/src/rules/newline_to_start_multiline_strings.dart
```diff
@@ -0,0 +1,91 @@
...
+    super.visitStringInterpolation(node);
+  }
+
+  void _visitSingleStringLiteral(SingleStringLiteral node, String lexeme) {
+    if (node.isMultiline &&
+        lineInfo.getLocation(node.offset).lineNumber !=
+            lineInfo.getLocation(node.end).lineNumber) {
+      var index = 3;
+      if (node.isRaw) index += 1;
+      if (['\n', '\r'].every((e) => e != lexeme[index])) {
```
bwilkerson: Not a required change, but rather than creating a new list for every string literal, consider something like
```
bool startWithNewLine(int index) => lexeme.startsWith('\n', index) || lexeme.startsWith('\r', index);
if (!startWithNewLine(node.isRaw ? 4 : 3)) {
```

test/rules/newline_to_start_multiline_strings.dart
```diff
@@ -0,0 +1,24 @@
...
+  f('''this
+ is a multiline string''');
+
+  f('''this is a multiline string $f'''); // OK
+  // OK
+  f('''
+this is a multiline string $f''');
+  // LINT [+1]
+  f('''this
+ is a multiline string$f''');
```
bwilkerson: A couple of additional tests to consider:
```
f('''$o''');
f("""uses double quotes""");
```

pq:
I haven't dug in but you may have noticed this failure on the benchmark bot?

```
Running the linter benchmark...
Exception while using a NewlineToStartMultilineStrings to visit a CompilationUnitImpl
RangeError (index): Invalid value: Not in range 0..2, inclusive: 3
#0      _StringBase.[] (dart:core-patch/string_patch.dart:247:55)
#1      _Visitor._visitSingleStringLiteral.<anonymous closure> (package:linter/src/rules/newline_to_start_multiline_strings.dart:86:48)
#2      ListMixin.every (dart:collection/list.dart:121:16)
#3      _Visitor._visitSingleStringLiteral (package:linter/src/rules/newline_to_start_multiline_strings.dart:86:24)
#4      _Visitor.visitStringInterpolation (package:linter/src/rules/newline_to_start_multiline_strings.dart:75:5)
```

Otherwise, I wonder about the name.  I think we've converged on plurals in general.  So maybe:

`newlines_to_start_multiline_strings`

or (bike-shedding), I might prefer:

`leading_newlines_in_multiline_strings`

pq:
Whoops.  Looks like you need to reformat before landing...

pq:
Super. Thanks!

PR #2023, opened 2020-03-03
add duplicated_keys

### LINKED ISSUE #2004: Warn for duplicate keys in map literal
When I write
```dart
final uvIndexes = {
  1: 'low',
  1: 'moderate',
  3: 'high',
};
```

I would like a lint that warns me that I have a duplicate key in my map because it is most likely a bug.
The code is legal but chances are very low that is this intentional.

Note: it does warn already for `const` map.

a14n:
# Description

Warn for duplicate keys in set/map literals

Fixes #2004

### DIFF
--- a/lib/src/rules/duplicated_keys.dart
+++ b/lib/src/rules/duplicated_keys.dart
@@ -4,14 +4,16 @@
 
 import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/constant/value.dart';
 
 import '../analyzer.dart';
 
-const _desc = r"Two keys in a constant set/map shouldn't be equal.";
+const _desc =
+    r"Two constant-valued elements / keys in a set / map literal shouldn't be equal.";
 
 const _details = r'''
 
-Two keys in a constant set/map shouldn't be equal.
+Two constant-valued elements / keys in a set / map literal shouldn't be equal.
 
 **BAD:**
 ```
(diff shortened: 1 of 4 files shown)


lib/src/rules/duplicated_keys.dart
```diff
@@ -0,0 +1,83 @@
+// Copyright (c) 2020, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+
+import '../analyzer.dart';
+
+const _desc = r"Two keys in a constant set/map shouldn't be equal.";
```
bwilkerson: The description says "constant", but neither the examples nor the tests use `const`. So I suspect that you mean "Two constant-valued elements / keys in a set / map literal shouldn't be equal.".

lib/src/rules/duplicated_keys.dart
```diff
@@ -0,0 +1,83 @@
...
+  void visitSetOrMapLiteral(SetOrMapLiteral node) {
+    final expressions = node.isSet
+        ? node.elements.whereType<Expression>()
+        : node.elements.whereType<MapLiteralEntry>().map((e) => e.key);
+    final alreadySeen = [];
+    for (final expression in expressions) {
+      final constEvaluation = context.evaluateConstant(expression);
+      if (constEvaluation.errors.isNotEmpty) continue;
+      final dartObject = constEvaluation.value;
+      if (alreadySeen.contains(dartObject)) {
```
bwilkerson: If you made `alreadySeen` a set, then (a) lookup would probably be faster with large literals and (b) you could use `add` here which would also be more efficient.

pq:
I'm w/ @bwilkerson: this feels like a great candidate for a hint.

Thanks, as always, @a14n!

bwilkerson:
Unless you have an argument for why it shouldn't be a hint, we'd like to make it a hint. If we do, do you want to do that work, or do you want one of us to do that?

a14n:
Please do it!

PR #2036, opened 2020-03-17
allow unused `_` params

### LINKED ISSUE #1793: avoid_unused_constructor_parameters false positives
An unused constructor parameter is occasionally needed to force something to be instantiated, for example when using dependency injection.

We could use an `ignore` comment for these cases, but it would be nice to have a prettier / more descriptive way to do it.

Some other languages, e.g. python, allow prefixing the variable name with `_` to mark it as something that is intentionally unused.

What do you think, please? I'd be happy to add this if it sounds good.

Thanks!

pq:
Fixes #1793

/cc @bwilkerson @davidmorgan

### DIFF
--- a/lib/src/rules/avoid_unused_constructor_parameters.dart
+++ b/lib/src/rules/avoid_unused_constructor_parameters.dart
@@ -54,11 +54,12 @@ class _ConstructorVisitor extends RecursiveAstVisitor {
   final Set<FormalParameter> unusedParameters;
 
   _ConstructorVisitor(this.rule, this.element)
-      : unusedParameters = element.parameters.parameters
-            .where((p) =>
-                p.declaredElement is! FieldFormalParameterElement &&
-                !p.declaredElement.hasDeprecated)
-            .toSet();
+      : unusedParameters = element.parameters.parameters.where((p) {
+          final element = p.declaredElement;
+          return element is! FieldFormalParameterElement &&
+              !element.hasDeprecated &&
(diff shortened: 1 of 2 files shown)


lib/src/rules/avoid_unused_constructor_parameters.dart
```diff
@@ -54,11 +54,12 @@ class _ConstructorVisitor extends RecursiveAstVisitor {
...
-      : unusedParameters = element.parameters.parameters
-            .where((p) =>
-                p.declaredElement is! FieldFormalParameterElement &&
-                !p.declaredElement.hasDeprecated)
-            .toSet();
+      : unusedParameters = element.parameters.parameters.where((p) {
+          final element = p.declaredElement;
+          return element is! FieldFormalParameterElement &&
+              !element.hasDeprecated &&
+              element.name != '_';
```
bwilkerson: Don't we want to allow any name that consists entirely of underscores? Otherwise we're limiting users to at most one ignored unused parameter.

pq: Absolutely. Good catch.

srawlins: Analyzer's `lib/src/lint/util.dart` has `isJustUnderscores` for this... deprecated though...

pq:
@bwilkerson: updated.

davidmorgan:
Thanks!

PR #2049, opened 2020-03-31
Add new lint, sized_box_for_whitespace

domesticmouse:
# Description

Adding a new Flutter lint, encouraging developers to use `SizedBox` instead of a `Container` for
creating whitespace in a layout.

### DIFF
--- a/lib/src/rules/prefer_sizedbox_over_container.dart
+++ b/lib/src/rules/prefer_sizedbox_over_container.dart
@@ -1,0 +1,100 @@
+// Copyright (c) 2020, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+
+import '../analyzer.dart';
+import '../util/flutter_utils.dart';
+
+const _desc = r'Prefer SizedBox over Container.';
+
+const _details = r'''Prefer SizedBox over Container for introducing blank space.
+
+A `Container` is a heavier Widget than a `SizedBox`, and as bonus, `SizedBox` 
+has a `const` constructor.
+
+**BAD:**
+```
+Widget buildRow() {
+  return Row(
+    children: <Widget>[
+      const MyLogo(),
+      Container(width: 4),
+      const Expanded(
+        child: Text('...'),
+      ),
+    ],
+  );
+}
+```
+
+**GOOD:**
+```
+Widget buildRow() {
+  return Row(
+    children: const <Widget>[
+      MyLogo(),
+      SizedBox(width: 4),
+      Expanded(
+        child: Text('...'),
+      ),
+    ],
+  );
+}
+```
+''';
+
+class PreferSizedBoxOverContainer extends LintRule implements NodeLintRule {
+  PreferSizedBoxOverContainer()
+      : super(
+            name: 'prefer_sizedbox_over_container',
+            description: _desc,
+            details: _details,
+            group: Group.style);
+
+  @override
+  void registerNodeProcessors(NodeLintRegistry registry,
+      [LinterContext context]) {
+    final visitor = _Visitor(this);
+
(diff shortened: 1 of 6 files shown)


lib/src/rules/prefer_sizedbox_over_container.dart
```diff
@@ -0,0 +1,100 @@
...
+    ],
+  );
+}
+```
+''';
+
+class PreferSizedBoxOverContainer extends LintRule implements NodeLintRule {
+  PreferSizedBoxOverContainer()
+      : super(
+            name: 'prefer_sizedbox_over_container',
```
bwilkerson: We have decided to move away from the "always", "prefer", etc. terminology, at least for rules related to the style guide (which this isn't) because those designations tend to change over time and several of our rules are now misnamed. I don't know whether we want to apply the same reasoning to non-style-guide rules. @pq Thoughts?

Also, a minor nit, I think "sizedbox" wants to be "sized_box".

Other than those questions, lgtm.

domesticmouse: How about naming it `sized_box_for_whitespace`?

pq: Brian is right.  Where possible, it's nice to avoid names that suggest severities.  To that end, `sized_box_for_whitespace` sounds great to me!

domesticmouse:
With regards to the [test failure](https://travis-ci.org/github/dart-lang/linter/jobs/669476834?utm_medium=notification&utm_source=github_status), I need some help understanding how to debug a failing linter test run.

I've run the test suite under the debugger in my IDE, and I can see the `rule.reportLint` line being executed, but I'm still getting a failing test. I'm confused as to how our lint tests work...

test/rules/sized_box_for_whitespace.dart
```diff
@@ -0,0 +1,61 @@
+// Copyright (c) 2020, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// test w/ `pub run test -N sized_box_for_whitespace`
+
+// ignore_for_file: sized_box_for_whitespace
```
a14n, line 7: This line should be removed to fix CI

domesticmouse: Thanks @a14n =)

a14n:
I did a similar lint some times ago. Applying it to flutter codebase I saw some false positives (especially in tests IIRC) because `Container` has a special handling if it's unconstrained.

a14n:
That is. Most of the time this lint is really helpful :)

pq:
RE: @a14n's comment:

> Applying it to flutter codebase I saw some false positives (especially in tests IIRC) because Container has a special handling when it's unconstrained.

It'd be great to capture these cases as they come up.  Ideally we can  make the lint more robust; minimally we can document.

Needless to say, cheers for this @domesticmouse!

a14n:
I may have been wrong or confused with another thing... IIRC there was differences when height or width are not set both https://github.com/flutter/flutter/blob/19e7db585dc33491cacce78571da119a1dd9a094/packages/flutter/lib/src/widgets/container.dart#L418-L424

Hixie:
We should consider the SizedBox.shrink and SizedBox.expand constructors too

domesticmouse:
@Hixie can you please expand out on what rules you are seeing? Is there a specific set of `Container` arguments that should map to `SizedBox.shrink` and `SizedBox.expand`?

a14n:
IMHO there should be 2 new lints:

- `SizedBox.shrink` instead of `SizedBox(height: 0, width: 0 ...)`
- `SizedBox.expand` instead of `SizedBox(height: double.infinity, width: double.infinity ...)`

But I'm not sure those cases arise enough to justify the cost of a lint.

a14n:
FWIW in the flutter codebase there would be around 30 lints for `SizedBox.shrink` and 0 for `SizedBox.expand`

Hixie:
Yeah people feel bad when they type "double.infinity" so they type "expand" instead. But it's common to type the 0 because that doesn't feel as bad. :-)

PR #2051, opened 2020-04-01
Fix sort_child_properties_last example

rokob:
The GOOD example both violated the lint and did not correspond to the same layout as the BAD example. This makes the GOOD example match the existing BAD example, and moves the floatingActionButton above child so that it passes the lint.

### DIFF
--- a/lib/src/rules/sort_child_properties_last.dart
+++ b/lib/src/rules/sort_child_properties_last.dart
@@ -36,6 +36,7 @@ return Scaffold(
       ],
       mainAxisAlignment: MainAxisAlignment.center,
     ),
+    widthFactor: 0.5,
   ),
   floatingActionButton: FloatingActionButton(
     child: Icon(Icons.add),
@@ -52,6 +53,7 @@ return Scaffold(
     title: Text(widget.title),
   ),
   body: Center(
+    widthFactor: 0.5,
     child: Column(
       mainAxisAlignment: MainAxisAlignment.center,
       children: <Widget>[
--- a/lib/src/rules/sort_child_properties_last.dart
+++ b/lib/src/rules/sort_child_properties_last.dart
@@ -36,11 +36,11 @@ return Scaffold(
       ],
       mainAxisAlignment: MainAxisAlignment.center,
     ),
-    floatingActionButton: FloatingActionButton(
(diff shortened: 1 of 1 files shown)


lib/src/rules/sort_child_properties_last.dart
```diff
@@ -52,8 +52,13 @@ return Scaffold(
     title: Text(widget.title),
   ),
   body: Center(
-    mainAxisAlignment: MainAxisAlignment.center,
+    floatingActionButton: FloatingActionButton(
```
mehmetf, line 55: Thanks for doing this.

Unfortunately, both examples are wrong :-/. floatingActionButton is not a property of `Center` widget. It is a property of `Scaffold`.

Here's the proper bad example:

```dart
return Scaffold(
  appBar: AppBar(
    title: Text(widget.title),
  ),
  body: Center(
    child: Column(
      children: <Widget>[
        Text(
          'You have pushed the button this many times:',
         ),
        Text(
          '$_counter',
          style: Theme.of(context).textTheme.display1,
         ),
      ],
      mainAxisAlignment: MainAxisAlignment.center,
    ),
    widthFactor: 0.5,
  ),
  floatingActionButton: FloatingActionButton(
    child: Icon(Icons.add),
    onPressed: _incrementCounter,
    tooltip: 'Increment',
  ),
);
```

here's the good example:

```dart
return Scaffold(
  appBar: AppBar(
    title: Text(widget.title),
  ),
  body: Center(
    widthFactor: 0.5,
    child: Column(
      mainAxisAlignment: MainAxisAlignment.center,
      children: <Widget>[
        Text(
          'You have pushed the button this many times:',
         ),
        Text(
          '$_counter',
          style: Theme.of(context).textTheme.display1,
         ),
      ],
    ),
  ),
  floatingActionButton: FloatingActionButton(
    child: Icon(Icons.add),
    onPressed: _incrementCounter,
    tooltip: 'Increment',
  ),
);
```

rokob: Gotcha, even worse then. I updated to a correct set of examples.

pq:
Thanks a million @rokob and cheers @mehmetf for the review!

PR #2055, opened 2020-04-07
Add three more new unsafe APIs to the unsafe_html rule

srawlins:
# Description

Three new unsafe APIs from dart:html have been identified:

* `Window.open`
* `new Element.html`
* `new DocumentFragment.html`


pubspec.yaml
```diff
@@ -12,7 +12,7 @@ environment:
   sdk: '>=2.6.0 <3.0.0'
 
 dependencies:
-  analyzer: ^0.39.3
+  analyzer: ^0.39.6
```
srawlins, line 15: Analyzer 0.39.6 has a few new classes in the Mock SDK needed for testing these changes.

lib/src/rules/unsafe_html.dart
```diff
@@ -62,8 +66,14 @@ class _Visitor extends SimpleAstVisitor<void> {
       LintCode('unsafe_html', '$_descPrefix (assigning "srcdoc" attribute).');
   static const createFragmentMethodCode = LintCode('unsafe_html',
       '$_descPrefix (calling the "createFragment" method of Element).');
+  static const openMethodCode = LintCode(
+      'unsafe_html', '$_descPrefix (calling the "open" method of Window).');
```
bwilkerson, line 70: These appear to be falling into three groups ("assigning '{0}' attribute", "calling the '{0}' method of '{1}'" and "calling the '{0}' constructor of '{1}'"). Consider making three parameterized lint codes for these three cases.

And although it's kind of a nit, please use single quotes around source code for consistency with the rest of the diagnostics that analyzer, front end and linter produce.

srawlins: The reason that there is a code per attribute/method name/constructor name is so that we can make separate allow-lists for approved usages. The solution isn't perfect, because each allow-list only has file granularity. So approving `Element.createFragment` in file `foo.dart` means that it can be used any number of times in that file :(, but at least that approval does not mean that `Element.setInnerHtml` can be used within `foo.dart`; separate allow-lists.

bwilkerson: I don't understand how the allow-lists work. Could you point me to some documentation or code related to that? (Probably via a different channel.)

srawlins: Sorry, your suggestion actually makes perfect sense, since the interpolated message strings are still unique. I've updated the code to use such message templates as you suggest.

lib/src/util/unrelated_types_visitor.dartBKUP
```diff
@@ -0,0 +1,205 @@
+// Copyright (c) 2016, the Dart project authors. Please see the AUTHORS file
```
bwilkerson, line 1: This file has a strange suffix. Was it included accidentally?

srawlins: Gah! It sure was. Removed.

PR #2057, opened 2020-04-08
Check unsafe attributes and methods on extensions

### LINKED ISSUE #2056: unsafe_html does not account for extensions
Take this example:

```dart
extension on ScriptElement {
  void sneakySetSrc(String url) => src = url;
}
```

`unsafe_html` doesn't currently account for an implicit `this` receiver. Boo. I'm working on a fix. I have to imagine this applies to other lint rules as well. ☹️

srawlins:
# Description

Account for implicit targets on assignments and method invocations.

Fixes #2056

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -8,7 +8,7 @@ analyzer:
     - .history/**
     - test/_data/**
     - test/mock_packages/**
-    #- test/rules/**
+    - test/rules/**
 linter:
   rules:
 #    - always_put_control_body_on_new_line
@@ -102,7 +102,6 @@ linter:
     - unnecessary_overrides
     - unnecessary_parenthesis
     - unnecessary_statements
-    - unsafe_html
     - use_setters_to_change_properties
     - use_string_buffers
     - use_to_and_as_if_applicable
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -8,7 +8,7 @@ analyzer:
     - .history/**
     - test/_data/**
     - test/mock_packages/**
-    - test/rules/**
+    #- test/rules/**
 linter:
   rules:
 #    - always_put_control_body_on_new_line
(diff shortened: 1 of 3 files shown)


analysis_options.yaml
```diff
@@ -8,7 +8,7 @@ analyzer:
     - .history/**
     - test/_data/**
     - test/mock_packages/**
-    - test/rules/**
+    #- test/rules/**
```
bwilkerson, line 11: I suspect that you intended to uncomment this.

analysis_options.yaml
```diff
@@ -102,6 +102,7 @@ linter:
     - unnecessary_overrides
     - unnecessary_parenthesis
     - unnecessary_statements
+    - unsafe_html
```
bwilkerson, line 105: I'm not sure this adds value because the linter package doesn't use the html library.

analysis_options.yaml
srawlins: Apologies.

analysis_options.yaml
srawlins: Apologies.

analysis_options.yaml
bwilkerson: No problem. Still LGTM.

lib/src/rules/unsafe_html.dart
```diff
@@ -134,12 +139,24 @@ class _Visitor extends SimpleAstVisitor<void> {
...
-    if (type == null) return;
-
     var methodName = node.methodName?.name;
     if (methodName == null) return;
 
+    // The static type of the target.
+    DartType type;
+    if (node.realTarget == null) {
+      // Implicit `this` target.
+      var enclosingElement = node.methodName.staticElement.enclosingElement;
```
pq, line 149: It looks like you need to be null-safe here:

https://travis-ci.org/github/dart-lang/linter/jobs/672562418

so, maybe:

`var enclosingElement = node.methodName.staticElement?.enclosingElement`

srawlins: yeah, fixed. Thanks!

PR #2060, opened 2020-04-09
check catch clauses in `avoid_types_as_parameter_names`

### LINKED ISSUE #1493: avoid_types_as_parameter_names should trigger on catch functions
Right now a user can write the following, which is not caught by [avoid_types_as_parameter_names](https://github.com/dart-lang/linter/blob/master/test/rules/avoid_types_as_parameter_names.dart):

```dart
try {
  // code;
} catch (ArgumentError) {
  // code;
}
```

but I think it should be caught.

pq:
Fixes #1493

/cc @bwilkerson @srawlins

### DIFF
--- a/lib/src/rules/avoid_types_as_parameter_names.dart
+++ b/lib/src/rules/avoid_types_as_parameter_names.dart
@@ -38,6 +38,7 @@ class AvoidTypesAsParameterNames extends LintRule implements NodeLintRule {
       NodeLintRegistry registry, LinterContext context) {
     final visitor = _Visitor(this);
     registry.addFormalParameterList(this, visitor);
+    registry.addCatchClause(this, visitor);
   }
 }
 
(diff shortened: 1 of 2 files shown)


lib/src/rules/avoid_types_as_parameter_names.dart
```diff
@@ -46,17 +47,28 @@ class _Visitor extends SimpleAstVisitor<void> {
 
   _Visitor(this.rule);
 
+  bool hasTypeName(String name) =>
```
bwilkerson, line 50: 'hasTypeName' --> 'isTypeName'

This seems very prone to false positives. I'd like to see us add a method to `LinterContext` (or someplace better) that would allow us to perform a name lookup at a given point in the AST. This method should then be rewritten to perform that lookup (in the parent of the declaration site so we don't get back the existing declaration) and see whether the name is bound to a type.

pq: > 'hasTypeName' --> 'isTypeName'

👍 

As for false positives, totally agreed.  The current heuristic is a kludge.  Adding a comment for the `LinterContext` idea.

fyi @scheglov: who might have ideas/reservations.

scheglov: There is already `LinterContext.resolveNameInScope()`, which probably will help here.
I will need to try to switch this lint to it.

pq: Ah, cool!

pq:
> assuming that this doesn't cause failures when run over important code bases

FWIW: I did an internal test (where this is enabled) and all clear. 👍

lib/src/rules/avoid_types_as_parameter_names.dart
scheglov: https://github.com/dart-lang/linter/pull/2062

PR #2076, opened 2020-04-21
[Lint] Prefer intValue.isEven check

### LINKED ISSUE #2065: Use intValue.isOdd/isEven instead of intValue % 2 == 1/0
Guessing this is pretty common!

huynhducloi00:
# Description
Add a lint check to advice users to replace the use of "%2==0" with intValue.isEven and intValue.isOdd.
This also takes into account "!=" and all the comparator operator.

Fixes #2065

### DIFF
--- a/example/all.yaml
+++ b/example/all.yaml
@@ -160,6 +160,7 @@ linter:
     - unsafe_html
     - use_full_hex_values_for_flutter_colors
     - use_function_type_syntax_for_parameters
+    - use_is_even_rather_than_modulo
     - use_key_in_widget_constructors
     - use_raw_strings
     - use_rethrow_when_possible
--- a/lib/src/rules.dart
+++ b/lib/src/rules.dart
@@ -115,6 +115,7 @@ import 'rules/prefer_if_null_operators.dart';
 import 'rules/prefer_initializing_formals.dart';
 import 'rules/prefer_inlined_adds.dart';
 import 'rules/prefer_int_literals.dart';
+import 'rules/use_is_even_rather_than_modulo.dart';
 import 'rules/prefer_interpolation_to_compose_strings.dart';
 import 'rules/prefer_is_empty.dart';
 import 'rules/prefer_is_not_empty.dart';
@@ -337,6 +338,7 @@ void registerLintRules() {
     ..register(UnsafeHtml())
     ..register(UseFullHexValuesForFlutterColors())
     ..register(UseFunctionTypeSyntaxForParameters())
+    ..register(UseIsEvenRatherThanModuloCheck())
     ..register(UseKeyInWidgetConstructors())
     ..register(UseRethrowWhenPossible())
     ..register(UseRawStrings())
--- a/lib/src/rules/use_is_even_rather_than_modulo.dart
+++ b/lib/src/rules/use_is_even_rather_than_modulo.dart
@@ -1,0 +1,80 @@
+// Copyright (c) 2020, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/token.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+
+import '../analyzer.dart';
+
+const _desc =
+    r'Prefer intValue.isOdd/isEven instead of checking the result of % 2.';
+
+const _details = r'''
+
+**PREFER** the use of intValue.isOdd/isEven to check for evenness.
+
+**BAD:**
+```
+bool isEven = 1 % 2 == 0;
+bool isOdd = 13 % 2 == 1;
+```
+
+**GOOD:**
+```
+bool isEven = 1.isEven;
+bool isOdd = 13.isOdd;
+```
+
+''';
+
+class UseIsEvenRatherThanModuloCheck extends LintRule implements NodeLintRule {
+  UseIsEvenRatherThanModuloCheck()
+      : super(
+            name: 'use_is_even_rather_than_modulo',
+            description: _desc,
+            details: _details,
+            group: Group.style);
+
+  @override
+  void registerNodeProcessors(NodeLintRegistry registry,
+      [LinterContext context]) {
+    final visitor = _Visitor(this);
+    registry.addBinaryExpression(this, visitor);
+  }
+}
+
+class _Visitor extends SimpleAstVisitor<void> {
+  final LintRule rule;
+  _Visitor(this.rule);
+
+  @override
+  void visitBinaryExpression(BinaryExpression node) {
+    // This lint error only happens when the operator is equality.
+    if (node.operator.type != TokenType.EQ_EQ) {
+      return;
+    }
+    var left = node.leftOperand;
+    var leftType = left.staticType;
+    var right = node.rightOperand;
+    var rightType = right.staticType;
+    // Both sides have to have static type of int
+    if (!(right is IntegerLiteral &&
+        leftType.isDartCoreInt == true &&
+        rightType.isDartCoreInt == true)) {
+      return;
+    }
+    // The left side expression has to be modulo by 2 type.
+    if (left is BinaryExpression) {
+      var rightChild = left.rightOperand;
+      var rightChildType = rightChild.staticType;
+      if (left.operator.type == TokenType.PERCENT &&
+          rightChild is IntegerLiteral &&
+          rightChild.value == 2 &&
+          rightChildType.isDartCoreInt == true) {
+        rule.reportLint(node);
+      }
+    }
+  }
+}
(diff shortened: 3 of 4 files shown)


huynhducloi00:
@googlebot I signed it!

huynhducloi00:
i will run "pub run grinder docs --dir=doc_location" after a +2

lib/src/rules/prefer_int_value_is_even_check.dart
```diff
@@ -0,0 +1,77 @@
...
+const _desc =
+    r'Prefer intValue.isOdd/isEven instead of checking the result of % 2.';
+
+const _details = r'''
+
+**PREFER** the use of intValue.isOdd/isEven to check for evenness.
+
+**BAD:**
+```
+bool isEven = 1 % 2==0;
```
bwilkerson, line 20: Consider cleaning up the spacing in the examples to improve readability.

lib/src/rules/prefer_int_value_is_even_check.dart
```diff
@@ -0,0 +1,77 @@
...
+      return;
+    }
+    if (left is BinaryExpression) {
+      var rightChild = left.rightOperand;
+      var rightChildType = left.staticType;
+      if (rightChildType?.isDartCoreInt == true &&
+          rightChild is IntegerLiteral &&
+          rightChild.value == 2 &&
+          BooleanExpressionUtilities.ALL_COMPARISONS_OPERATIONS
+              .contains(node.operator.type)) {
```
bwilkerson, line 72: We should check the `node`'s operator at the beginning of the method because it will be more efficient than waiting until this point.

Also, we should check the `left` operand's operator to ensure that it's `%`. As written I believe that the lint rule will produce a diagnostic for `a + 2 == 0` (which should be added as a test case either way).

lib/src/util/boolean_expression_utilities.dart
```diff
@@ -10,6 +10,15 @@ class BooleanExpressionUtilities {
   static HashSet<TokenType> BOOLEAN_OPERATIONS =
       HashSet.from(const [TokenType.AMPERSAND_AMPERSAND, TokenType.BAR_BAR]);
 
+  static HashSet<TokenType> ALL_COMPARISONS_OPERATIONS = HashSet.from(const [
```
bwilkerson, line 13: I'd drop the prefix "ALL_" from the name. I don't think it clarifies the meaning of the name, and the form without it is more consistent with the names of the other constants in this class.

Speaking of which, I'd argue that it would be better to be inconsistent and use a set literal here, which would allow this field to be marked as a `const`. I suspect that the other fields here were created before set literals existed. They should be converted at some point, but I'm not asking you to do that as part of this PR.

test/rules/prefer_int_value_is_even_check.dart
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) 2020, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// test w/ `pub run test -N prefer_int_value_is_even_check`
+
+bool isEven = 1 % 2 == 0; //LINT
+bool isOdd = 13 % 2 == 1; //LINT
+
+bool a = 1 % 2 >= 0; //LINT
```
bwilkerson, line 10: This example is concerning. The modulo operator will always return a non-negative value, so the condition `>= 0` is always `true`. Replacing this with an invocation of either `isEven` or `isOdd` would not preserve semantics, so we shouldn't produce a diagnostic recommending that change.

We either need to add logic to the lint rule to ensure that the diagnostic is only produced when the inequality can validly be replaced by one of those invocations, or we need to change the lint rule so that it only uses the equality operators.

But I'll note that even with the equality operators we can have the same situation. Consider `a % 2 == 3`, which is always `false` and similarly shouldn't produce a diagnostic.

lib/src/rules/prefer_int_value_is_even_check.dart
```diff
@@ -0,0 +1,77 @@
...
+bool isEven =1.isEven;
+bool isOdd = 13.isOdd;
+```
+
+''';
+
+class PreferIntValueIsEvenCheck extends LintRule implements NodeLintRule {
+  PreferIntValueIsEvenCheck()
+      : super(
+            name: 'prefer_int_value_is_even_check',
```
bwilkerson, line 35: You had no way of knowing, but we've decided that it was a mistake to include prefixes like "prefer" in lint names. The reason is that the style guide can change over time, and things that were once discouraged might later be forbidden. Consider something like "use_is_even_rather_than_modulo" (alternative suggestions are welcome).

lib/src/rules/prefer_int_value_is_even_check.dart
pq: Yes please!

lib/src/rules/prefer_int_value_is_even_check.dart
pq: > use_is_even_rather_than_modulo

Great name.  I was grasping for one and this tops it.  👍

pq:
Thanks a million for the contribution!  👍 to Brian's review.

> i will run "pub run grinder docs --dir=doc_location" after a +2

Don't worry about running docs.  Those get applied on another branch and I do that whenever I publish.

lib/src/rules/prefer_int_value_is_even_check.dart
```diff
@@ -0,0 +1,77 @@
...
+  _Visitor(this.rule);
+
+  @override
+  void visitBinaryExpression(BinaryExpression node) {
+    var left = node.leftOperand;
+    var leftType = left.staticType;
+    var right = node.rightOperand;
+    var rightType = right.staticType;
+    // Both sides have to have static type of int
+    if (!(leftType?.isDartCoreInt == true &&
```
scheglov, line 60: Left and right expression cannot have `null` types unless there is a serious bug in analyzer.
So, remove `?` here and below.

lib/src/rules/prefer_int_value_is_even_check.dart
```diff
@@ -0,0 +1,77 @@
...
+        rightType?.isDartCoreInt == true &&
+        right is IntegerLiteral)) {
```
scheglov, line 62: I think `right is IntegerLiteral` is faster than invoking `isDartCoreInt`, so it should be before them.

And the check for the operator token (Brian mentioned it below) should be the top one. If it is not `%` - exit from this method.

test/rules/prefer_int_value_is_even_check.dart
```diff
@@ -0,0 +1,25 @@
...
+int number=3;
+bool c = number % 2 == 0; //LINT
+bool d = number % 2 != 0; //LINT
+
+// Modulo by any other number than 2 is okay
+d = number % 3 ==1;
```
scheglov, line 16: Space.

test/rules/prefer_int_value_is_even_check.dart
```diff
@@ -0,0 +1,25 @@
...
+
+// Compare to not an IntegerLiteral is okay.
+bool b = 1 % 2 < 3-3;
```
scheglov, line 19: Space.

lib/src/rules/prefer_int_value_is_even_check.dart
```diff
@@ -0,0 +1,77 @@
...
+            name: 'prefer_int_value_is_even_check',
+            description: _desc,
+            details: _details,
+            group: Group.style);
+
+  @override
+  void registerNodeProcessors(NodeLintRegistry registry,
+      [LinterContext context]) {
+    final visitor = _Visitor(this);
+    registry.addSimpleIdentifier(this, visitor);
```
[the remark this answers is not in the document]
huynhducloi00: Done

lib/src/rules/prefer_int_value_is_even_check.dart
huynhducloi00: Done

lib/src/rules/prefer_int_value_is_even_check.dart
huynhducloi00: Good catch. Done

lib/src/util/boolean_expression_utilities.dart
huynhducloi00: obsolete, because i am now following bwilkerson to only check for equality == 0/1 operator for this lint.

test/rules/prefer_int_value_is_even_check.dart
huynhducloi00: sound good to me. yes, so i simplify this lint check to only warn %2 ==0 and %2 ==1. other cases would not yield a diagnostic.

lib/src/rules/prefer_int_value_is_even_check.dart
huynhducloi00: thanks for the name suggestion. Done

lib/src/rules/prefer_int_value_is_even_check.dart
huynhducloi00: Done

lib/src/rules/prefer_int_value_is_even_check.dart
huynhducloi00: Done

test/rules/prefer_int_value_is_even_check.dart
huynhducloi00: Done

pq:
🎉🎉🎉

Yes, thanks!

Issue #2082, opened 2020-04-30
`prefer_mixin` should whitelist (some) library-defined abstract class mixins


pq:
The lint fires, for example, here:

```dart
class AnalysisResults with IterableMixin<AnalysisResult> {

}
```

as `IterableMixin` looks like this in the SDK:

```dart
/// This [Iterable] mixin implements all [Iterable] members except `iterator`.
///
/// All other methods are implemented in terms of `iterator`.
abstract class IterableMixin<E> implements Iterable<E> {

}
```

IMO, we shouldn't flag this use.

bwilkerson:
@lrhn Will mixins like `IterableMixin` be converted in the upgrade to Null Safety?

lrhn:
There are no plans to make that change in the current SDK iterations.

I hope we'll eventually get around to deprecating classes-used-as-mixins, but it is a breaking change to change `class` to `mixin` in case someone does `extends ClassName` somewhere.
So we'll need a migration step (or allow `extends MixinName` as a shorthand for `extends Object with MixinName` (aka `with MixinName`).

bwilkerson:
In that case, yes, we probably ought to special case the classes in the SDK that are intended to be used as mixins.

pq:
Thanks for the follow-ups.  A quick scan  comes up with these initial candidates:

* `IterableMixin`
* `ListMixin`
* `MapMixin`
* `SetMixin`
* `StringConversionSinkMixin` (dart:convert)

Chime in if there are  any I missed?  👍

PR #2088, opened 2020-05-03
add allowed “legacy” SDK class mixins

pq:
Fixes: https://github.com/dart-lang/linter/issues/2082

/cc @bwilkerson

### DIFF
--- a/lib/src/rules/prefer_mixin.dart
+++ b/lib/src/rules/prefer_mixin.dart
@@ -5,6 +5,8 @@
 import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
 import 'package:analyzer/dart/element/element.dart';
+import 'package:analyzer/dart/element/type.dart';
+import 'package:linter/src/util/dart_type_utilities.dart';
 
 import '../analyzer.dart';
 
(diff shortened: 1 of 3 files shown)


lib/src/rules/prefer_mixin.dart
```diff
@@ -56,9 +58,22 @@ class _Visitor extends SimpleAstVisitor<void> {
...
         rule.reportLint(type);
       }
     }
   }
+
+  /// Check for "legacy"  classes that cannot easily be made `mixin`s for
+  /// compatibility reasons.
+  /// (See: https://github.com/dart-lang/linter/issues/2082)
+  static bool isAllowed(DartType type) =>
+      DartTypeUtilities.isClass(type, 'IterableMixin', 'dart.core') ||
```
bwilkerson, line 73: For efficiency's sake, consider adding a new utility:
```
static bool isClassElement(ClassElement element, String className, String library) =>
    element.name == className &&
    element.library?.name == library;
```
and using it directly, rather than creating a type just to extract the element from it.

test/rules/prefer_mixin.dart
```diff
@@ -4,8 +4,103 @@
...
+
 class A {}
+
 class B extends Object with A {} // LINT
 
 mixin M {}
+
 class C with M {} // OK
+
+class I with IterableMixin //OK
```
bwilkerson, line 19: I don't have strong feelings about it, but you could significantly shorten the test code by writing `abstract class I with IterableMixin {} // OK`. If it's abstract it doesn't need to implement the abstract methods from the mixin.

lib/src/rules/prefer_mixin.dart
pq: Good catch.  Done!

pq:
Thanks!

PR #2090, opened 2020-05-04
new lint: exhaustive_cases

### LINKED ISSUE #2083: lint to ensure exhaustive enum-like class switches
**Exhaustive enum-like class switch**

For classes that define private constructors, and a number of static const instances, switching on an instance of that class should be exhaustive. This should cover e.g. proto enums and other cases.

Open question: is inferring "enum-like" characteristics sufficient, or do we need a formal annotation?

pq:
Fixes #2083

/cc @bwilkerson

### DIFF
--- a/lib/src/rules/exhaustive_cases.dart
+++ b/lib/src/rules/exhaustive_cases.dart
@@ -10,7 +10,7 @@ import 'package:analyzer/dart/element/type.dart';
 import '../analyzer.dart';
 import '../util/dart_type_utilities.dart';
 
-const _desc = r'Define case clauses for all constants in enum-like classes';
+const _desc = r'Define case clauses for all constants in enum-like classes.';
 
 const _details = r'''
 Switching on instances of enum-like classes should be exhaustive.
--- a/lib/src/rules/exhaustive_cases.dart
+++ b/lib/src/rules/exhaustive_cases.dart
@@ -119,6 +119,7 @@ class _Visitor extends SimpleAstVisitor {
       for (var member in statement.members) {
         if (member is SwitchCase) {
           var expression = member.expression;
+          // todo (pq): add a test to ensure that this handles prefixed identifiers.
           if (expression is Identifier) {
             var element = expression.staticElement;
             if (element is PropertyAccessorElement) {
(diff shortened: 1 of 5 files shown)


lib/src/rules/exhaustive_cases.dart
```diff
@@ -0,0 +1,145 @@
...
+      var enumDescription = DartTypeUtilities.asEnumLikeClass(classElement);
+      if (enumDescription == null) {
+        return;
+      }
+
+      var enumConstantNames = enumDescription.enumConstantNames;
+      for (var member in statement.members) {
+        if (member is SwitchCase) {
+          var expression = member.expression;
+          if (expression is Identifier) {
```
bwilkerson, line 122: I suspect that this will miss the case where the enum-like class is imported using a prefix (in which case the expression will be a `PropertyAccess`).

lib/src/util/dart_type_utilities.dart
```diff
@@ -12,6 +12,55 @@ import 'package:meta/meta.dart';
...
+        continue;
+      }
+      enumConstantNames.add(field.name);
+    }
+    if (enumConstantNames.length < 2) {
+      return null;
+    }
+
+    // And no subclasses in the defining library.
+    var type = classElement.thisType;
```
bwilkerson, line 49: Consider moving this before the loop (on line 33) and using it inside the loop.

lib/src/util/dart_type_utilities.dart
```diff
@@ -12,6 +12,55 @@ import 'package:meta/meta.dart';
...
+    var compilationUnit = classElement.library.definingCompilationUnit;
+    for (var cls in compilationUnit.types) {
+      var classType = cls.thisType;
+      do {
+        classType = classType.superclass;
+        if (classType == type) {
+          return null;
+        }
+      } while (classType != null && !classType.isDartCoreObject);
```
bwilkerson, line 58: It isn't necessary to have both conditions (the supertype of `Object` is `null`).

lib/src/rules/exhaustive_cases.dart
pq: Ah.  I hadn't considered that.  Thanks.

I'll follow-up w/ an integration test.

pq: OK, yeah, I've got this reproduced.  I'll follow-up with a fix shortly.

pq:
Integration test to follow.

PR #2092, opened 2020-05-05
new lint: no_default_cases

### LINKED ISSUE #2084: lint to flag default clauses in enum-like switch statements
**Avoid default clauses in enum-like switch statements**

Good:

```dart
  switch (testEnum) {
    case TestEnum.A:
      return '123';
    case TestEnum.B:
      return 'abc';
  }
  // Default here.
  return null;
```

Bad:

```dart
  switch (testEnum) {
    case TestEnum.A:
      return '123';
    case TestEnum.B:
      return 'abc';
    default:
      return null;
  }
```

Related to #2083, `default` should be forbidden in enum and enum-like case switches.

pq:
Fixes #2084

/cc @bwilkerson

### DIFF
--- a/lib/src/rules/no_default_cases.dart
+++ b/lib/src/rules/no_default_cases.dart
@@ -4,10 +4,11 @@
 
 import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/element.dart';
 import 'package:analyzer/dart/element/type.dart';
-import 'package:linter/src/util/dart_type_utilities.dart';
 
 import '../analyzer.dart';
+import '../util/dart_type_utilities.dart';
 
 const _desc = r'No default cases.';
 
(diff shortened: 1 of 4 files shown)


lib/src/rules/no_default_cases.dart
```diff
@@ -0,0 +1,90 @@
...
+      var classElement = expressionType.element;
+      if (!classElement.isEnum) {
+        var enumDescription = DartTypeUtilities.asEnumLikeClass(classElement);
+        if (enumDescription == null) {
+          return;
+        }
+      }
+
+      for (var member in statement.members) {
+        if (member is SwitchDefault) {
```
bwilkerson, line 84: I believe that it would be more efficient to check for a default case _before_ checking the type of the switch expression. If you do that, you might consider an even more radical change of making this visit `SwitchDefault` nodes rather than `SwitchStatement`s. That would also avoid the extra loop through the members.

pq: Great idea.  Done!

PR #2093, opened 2020-05-05
0.1.116

pq:
# 0.1.116

* new lint: `no_default_cases` (experimental)
* new lint: `exhaustive_cases`
* updated `type_annotate_public_apis` to allow inferred types in final field assignments
* updated `prefer_mixin` to allow "legacy" SDK abstract class mixins
* new lint: `use_is_even_rather_than_modulo`
* update `unsafe_html` to use a `SecurityLintCode` (making it un-ignorable)
* improved `sized_box_for_whitespace` to address false-positives

/cc @bwilkerson 

/fyi @srawlins

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,13 @@
+# 0.1.116
+
+* new lint: `no_default_cases`
+* new lint: `exhaustive_cases`
+* updated `type_annotate_public_apis.dart` to allow inferred types in final field assignments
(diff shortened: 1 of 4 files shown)


pq:
Good point.

Two ideas:

1. back `no_default_case_clause` out of the release, or
2. mark it as experimental and note that it's subject to change

Thoughts?

(I'm keen to get this out so @jefflim-google can play with `exhaustive_cases`.)

bwilkerson:
Marking it as experimental sounds like a reasonable approach. That way they can play with both.

PR #2100, opened 2020-05-18
+ do_not_use_environment

### LINKED ISSUE #2096: New lint: avoid_compile_constants
Avoid using compile time constants

The lint would trigger upon the use of the three Flutter fromEnvironment constructors:
`int.fromEnvironment`
`string.fromEnvironment`
`bool.fromEnvironment` 

`bool.hasEnvironment` should also trigger the lint.

Using a compile constant creates a hidden global state in the program, with no defined method to determine the state or understand its use when reading the code  (unlike flags a compile time constant might only be referenced during the fromEnvironment constructor call). If used heavily it makes the program hard to understand.

pq:
Fixes #2096

/cc @bwilkerson

### DIFF
--- a/lib/src/rules/do_not_use_environment.dart
+++ b/lib/src/rules/do_not_use_environment.dart
@@ -7,7 +7,7 @@ import 'package:analyzer/dart/ast/visitor.dart';
 
 import '../analyzer.dart';
 
-const _desc = r'Do not use environment variables.';
+const _desc = r'Do not use environment declared variables.';
 
 const _details = r'''
 Using values derived from the environment at compile-time, creates
@@ -45,7 +45,7 @@ class _Visitor extends SimpleAstVisitor {
 
   @override
   void visitInstanceCreationExpression(InstanceCreationExpression node) {
-    if (!node.staticElement.isFactory) {
+    if (node.staticElement?.isFactory != true) {
       return;
     }
 
--- a/lib/src/rules/do_not_use_environment.dart
+++ b/lib/src/rules/do_not_use_environment.dart
@@ -1,0 +1,69 @@
+// Copyright (c) 2020, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+
+import '../analyzer.dart';
+
+const _desc = r'Do not use environment variables.';
+
+const _details = r'''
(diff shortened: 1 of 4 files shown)


lib/src/rules/do_not_use_environment.dart
```diff
@@ -0,0 +1,69 @@
+// Copyright (c) 2020, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+
+import '../analyzer.dart';
+
+const _desc = r'Do not use environment variables.';
```
bwilkerson, line 10: One small nit, which you can choose to ignore: the term "environment variable" usually refers to variables inherited by a process from the parent process. In Dart, those are accessed using `Platform.environment`. In the analyzer code base we refer to the `fromEnvironment` keys as "declared variables", but that term isn't used anywhere else that I'm aware of. I don't know of a good alternative, but "environment variable" might be confusing.

pq: I'm leaning towards "environment declared variables".

pq: Thanks!

Gustl22:
@pq for me it is not clear what to use as alternative? Can we state that too inside the "GOOD" or "DO" description?

pq:
@Gustl22: any contributions to the docs here would be most welcome! 

/fyi @tomlenehan-google

Gustl22:
@pq thank you for your quick answer :) I would add it, but I honestly don't know what to do as alternative:
- `Platform.fromEnvironment()` only works for server or Desktop as far as I know
- using different main methods with config as files is very limiting as one would need to create one file for each possibility of combination

I think of a test use case, where can execute a flutter client/app for different scenarios via environment variables (e.g. set Network latency via a mocked network client, set api endpoint or some kind of app / firebase keys).

pq:
Yeah.  I think this is tricky and a lint that may only situationally apply.  To avoid further confusion, it'd be great to describe that situation better in the docs.  (FWIW I think this is generally true of our docs which tend to be a bit light on context.)

Curious if @goderbauer has anything to add from the flutter perspective or @rakudrama who had some interesting pushback to the idea of this lint in the first place...

Thanks for nudging this along @Gustl22!

goderbauer:
I wouldn't recommend this lint for Flutter. `fromEnvironment` can be a useful tool to build variants of an app while ensuring that unused variants are properly tree-shaken out to reduce binary size. I can see how the lint can be useful in other circumstances, though.

Issue #2104, opened 2020-05-26
Export test utilities


pq:
Test utilities

    test/util/annotation.dart
    test/util/test_resource_provider.dart

are depended on internally.  We should export them (so they don't need to be copied).

pq:
(This is part of #2105.)

bwilkerson:
Prior art: In the `analyzer` package, we created a directory under `lib` (`lib/src/test_utilities`) in which to put such things. The alternative we discussed was to create a separate package to hold testing support that other packages could have a dev dependency on.

pq:
Thanks! `lib/src/test_utilities` is exactly the direction I'm leaning.

PR #2109, opened 2020-05-27
move test utilities

pq:
Fixes: #2104 

/cc @bwilkerson @davidmorgan

### DIFF
--- a/lib/src/test_utilities/annotation.dart
+++ b/lib/src/test_utilities/annotation.dart
@@ -3,8 +3,7 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'package:analyzer/error/error.dart';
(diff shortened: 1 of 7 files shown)


davidmorgan:
Thanks!

Would it be possible to leave the old symbols around for one release, please? That'll make it easier to land the next linter roll ;) as we can roll -> fix google3 -> remove -> roll.

pq:
@davidmorgan: not a problem.  I've added some (re)`exports` like so:

```dart
@Deprecated(
    'Prefer importing `package:linter/src/test_utilities/test_resource_provider.dart` directly')
export 'package:linter/src/test_utilities/test_resource_provider.dart';
```

So we should be good to go?

davidmorgan:
LG, thanks!

PR #2114, opened 2020-05-28
Updated doc gen for proposed new site design

parlough:
Depended on by: https://github.com/dart-lang/linter/pull/2115

Read the linked PR for more information, but this implements some minor cleanup and organizational changes for the proposed new site design. 

As I mentioned on the other PR, please feel free to to present heavy feedback and large changes.

PR #2115, opened 2020-05-28
Propose a new, but familiar site design

parlough:
Depends on: https://github.com/dart-lang/linter/pull/2114

Proposes(and creates) a new site design working to follow the general idea, shape, and organization of the original, as it was formatted rather well and got the information across quite well. I went ahead and worked on this, despite not being super familiar with design or CSS, to better align the site with some of the other Dart websites and clean up a few things. 

Please feel free to to present heavy feedback and large changes.

Rough list of changes:
- Same general layout
- Cleaned up design which better aligns with other Dart styles
- More consistency between links and pages
- Slightly larger fonts
- Fixed various link interaction bounding
- Cleaned up messages, such as footer
- Reactivity with links for better UX
- Less bright
- Updated Dart icon
- More shades of blue :)
- Adds some fun transitions and hover colors
- Removed old, seemingly unused resources
- Removed extra support for long outdated browser editions

image


pq:
This looks fantastic.  Thank you for taking it on!

I'll take a closer look at the source but  to be honest the initial implementation was a bit of a hack job so I'm expecting nothing but improvements. 😉 

As for styling, do you think we ought to update the  badges as well?

image

parlough:
@pq Do you mean change them from the preexisting ones or the new ones I have here? 

If you mean from the new ones, yeah I wasn't exactly sure what to do with the badges. I tried a few different designs initially using solely CSS, but haven't yet came up with anything great. As a result I stuck with the shields, making new ones which attempted to match the new style better. I'll keep trying to think of a new design for them. If you or anyone have any suggestions, I can work on implementing that change. 

If you mean from the preexisting ones, I do think for the new design it was necessary to have something new, the existing ones didn't really match so well due to the gradients and text shadows they had.

pq:
I didn't realize they were new, sorry!  It sounds like we agree there's still a little room for improvement but that's way down the list.

Thanks!

PR #2122, opened 2020-06-02
special case assert initializers and  const contexts in `prefer_is_empty`

pq:
Fixes https://github.com/dart-lang/linter/issues/1719

/cc @bwilkerson

### DIFF
--- a/lib/src/rules/prefer_is_empty.dart
+++ b/lib/src/rules/prefer_is_empty.dart
@@ -3,6 +3,8 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'package:analyzer/dart/ast/ast.dart';
+// ignore: implementation_imports
+import 'package:analyzer/src/dart/ast/ast.dart' show ExpressionImpl;
 import 'package:analyzer/dart/ast/token.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
 import 'package:analyzer/dart/element/type.dart';
@@ -128,6 +130,20 @@ class _Visitor extends SimpleAstVisitor<void> {
       return;
     }
     final binaryExpression = search as BinaryExpression;
+
+    // Don't lint if we're in an assert initializer.
+    final assertInitializer =
+        search.parent.thisOrAncestorOfType<AssertInitializer>();
+    if (assertInitializer != null) {
+      return;
+    }
+
+    // Or in a const context.
+    // See: https://github.com/dart-lang/linter/issues/1719
+    final impl = binaryExpression as ExpressionImpl;
+    if (impl.inConstantContext) {
+      return;
+    }
 
     final operator = binaryExpression.operator;
 
--- a/lib/src/rules/prefer_is_empty.dart
+++ b/lib/src/rules/prefer_is_empty.dart
@@ -131,11 +131,15 @@ class _Visitor extends SimpleAstVisitor<void> {
     }
     final binaryExpression = search as BinaryExpression;
 
-    // Don't lint if we're in an assert initializer.
-    final assertInitializer =
-        search.parent.thisOrAncestorOfType<AssertInitializer>();
-    if (assertInitializer != null) {
-      return;
+    // Don't lint if we're in a const constructor initializer.
+    final constructorInitializer =
+        search.parent.thisOrAncestorOfType<ConstructorInitializer>();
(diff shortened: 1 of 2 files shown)


lib/src/rules/prefer_is_empty.dart
```diff
@@ -129,6 +131,20 @@ class _Visitor extends SimpleAstVisitor<void> {
     }
     final binaryExpression = search as BinaryExpression;
 
+    // Don't lint if we're in an assert initializer.
+    final assertInitializer =
+        search.parent.thisOrAncestorOfType<AssertInitializer>();
```
bwilkerson, line 136: This isn't quite right. I think we need to ignore anything inside any `ConstructorInitializer`, but only when the constructor has been marked as `const`.

On the other hand, it's possible that we don't need to test for these cases at all, because `inConstantContext` might return `true` in the initializer list of const constructors (and given that one of the test cases is for a non-assert initializer and it passes, I suspect that's true).

test/rules/prefer_is_empty.dart
```diff
@@ -4,17 +4,33 @@
 
 // test w/ `pub run test -N prefer_is_empty`
 
+const l = '';
+const bool empty = l.length == 0; //OK
+
+class A {
+  final List<String> a;
+  const A(this.a) : assert(a.length > 0); //OK
```
bwilkerson, line 12: It would be good to have some tests using a non-const constructor to show that we still generate the lint in those cases.

lib/src/rules/prefer_is_empty.dart
pq: I was hoping to get rid of this check too but `inConstantContext` is not returning true in the initializer test case 🤔.  Checking for assert initializers and testing `inConstantContext` felt like maybe a sweet spot.  (Though kludgey.)

Maybe we can discuss offline?

test/rules/prefer_is_empty.dart
pq: Yes.  Good idea.

lib/src/rules/prefer_is_empty.dart
bwilkerson: Does `inConstantContext` return true in the non-assert initializer cases:
```
class C {
  final bool empty;
  const C(dynamic l) : empty = l.length == 0; // here
  const C.a(this.empty);
  const C.b(dynamic l) : this.a(l.length == 0); // here
}
class E {
  const E(bool b);
}
class F extends E {
  const F(dynamic l) : super(l.length == 0); // here
}
```
If so, then it sounds like a bug that it doesn't within assert initializers.

pq: It does.  I've added some tests too.

@scheglov:  thoughts on why this is not working in the assert initializer case?

scheglov: I don't see anything about constructor initializers in the spec.

image

image

pq: Ah, super.  Thank you!

bwilkerson: That begs the question of why it's returning `true` in other constructor initializers. According to the spec it seems like it shouldn't be. I think that the inconsistency is pointing out a bug, but I'm not sure what the bug is.

But if `inConstantContext` should not be returning `true` in these cases, then we just need to not generate the lint in a constructor initializer where the constructor is marked as `const`.

scheglov: There is a difference between a constant context, and "constantness requirement". Any expression in constant constructors must be a constant expression, but does not introduce a constant context itself. But once you said `const C() : x = const MyClass(...)`, everything in `...` is in the constant context.

It does not look to me that `inConstantContext` returns `true` for constructor initializers.
```dart
  solo_test_inConstantContext_XXX() {
    parse('''
class C {
  final bool empty;
  const C(dynamic l) : empty = l.length == 0; // 0
  const C.a(this.empty);
  const C.b(dynamic l) : this.a(l.length == 0); // 1
}
class E {
  const E(bool b);
}
class F extends E {
  const F(dynamic l) : super(l.length == 0); // 2
}''');
    assertInContext("length == 0; // 0", false);
    assertInContext("length == 0); // 1", false);
    assertInContext("length == 0); // 2", false);
  }
```

PR #2126, opened 2020-06-06
Linter: add 'avoid_relative_imports'

### LINKED ISSUE #1601: Add 'avoid_relative_imports'

2ZeroSix:
# Description

This is a clone of #1602 with fixes addressed in review comments and rebased onto master.

Fixes #1601


lib/src/rules/avoid_relative_imports.dart
```diff
@@ -0,0 +1,97 @@
+// Copyright (c) 2018, the Dart project authors. Please see the AUTHORS file
```
bwilkerson, line 1: The year should be updated to 2020.

lib/src/rules/avoid_relative_imports.dart
```diff
@@ -0,0 +1,97 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+
+import '../analyzer.dart';
+import '../ast.dart';
+
+const _desc = r'Avoid relative imports for all files`.';
```
bwilkerson, line 11: There's an extra backquote after "files".

lib/src/rules/avoid_relative_lib_imports.dart
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2018, the Dart project authors. Please see the AUTHORS file
+// Copyright (c) 2020, the Dart project authors. Please see the AUTHORS file
```
bwilkerson, line 1: I'm guessing that you updated the wrong copyright notice. This one should be reverted.

lib/src/rules/avoid_relative_imports.dart
```diff
@@ -0,0 +1,97 @@
...
+
+...
+```
+
+''';
+
+class AvoidRelativeImports extends LintRule implements NodeLintRule {
+  AvoidRelativeImports()
+      : super(
+            name: 'avoid_relative_imports',
```
bwilkerson, line 50: You had no way of knowing, but we have decided that it was a mistake to use "avoid", "prefer" and other terms from the style guide in the names of lints. Perhaps we could name this something like "always_use_package_imports" (I'm open to alternatives).

lib/src/rules/avoid_relative_imports.dart
```diff
@@ -0,0 +1,97 @@
...
+  final LintRule rule;
+  final LinterContext context;
+
+  bool isInLibFolder;
+
+  _Visitor(this.rule, this.context);
+
+  bool isRelativeImport(ImportDirective node) {
+    // Ignore this compilation unit if it's not in the lib/ folder.
+    if (!isInLibFolder) return false;
```
bwilkerson, line 74: I believe that this PR has the opposite problem from the previous PR in that this will have false negatives in the case where a library outside of `lib` is using a relative path to reference a library inside of `lib` (which should also be disallowed).

lib/src/rules/avoid_relative_lib_imports.dart
2ZeroSix: updated

lib/src/rules/avoid_relative_imports.dart
2ZeroSix: updated

lib/src/rules/avoid_relative_imports.dart
2ZeroSix: updated

lib/src/rules/avoid_relative_imports.dart
2ZeroSix: renamed to `always_use_package_imports`

lib/src/rules/avoid_relative_imports.dart
2ZeroSix: I added some clarification in description:
```
const _desc = r'Avoid relative imports for files in `lib/`.';
```
and details:
```
Might be used with 'avoid_relative_lib_imports' to avoid relative imports of
files within `lib/` directory outside of it. (for example `test/`)
```


It seems strange to duplicate logic of another lint rule, so I added a link to lint that solves problem for files outside of `lib/`.


Does this seem reasonable to you?

pq:
LGTM.  Thanks!

PR #2158, opened 2020-06-19
Fix prefer_relative_imports when using path dependencies

### LINKED ISSUE #2005: `prefer_relative_imports` conflicts with path dependencies
**Describe the issue**
With `prefer_relative_imports` enabled, I see warnings in the following situation:

I have a package that is being pulled into a third-party library. As the first step, all of the related code has been moved from its locations within `lib` (e.g. `myproject/lib`) to a package within `vendor` (e.g. `myproject/vendor/mylibrary`). To be a complete package, _that_ folder, in turn, has its own `lib` directory (e.g. `myproject/vendor/mylibrary/lib/mylibrary.dart`).

In my `pubspec.yaml`, I have that package defined accordingly:

```
name: myproject
dependencies:
  mylibrary:
    path: vendor/mylibrary
```

When that package is imported (e.g. as `import "package:mylibrary/mylibrary.dart"`), I get the following warning:

```
Prefer relative imports for files in `lib/`.
```

Put another way, while the included file is within _a_ `lib/`, it's not in the _same_ `lib/`.

**To Reproduce**
Reproducing the above involves creating a complete Dart package in a directory parallel to `lib`, but not _within_ `lib`, and linking to it from both your `pubspec.yaml` and your application code.

**Expected behavior**
The above situation should not generate warnings.

beroso:
Fixes #2005

### DIFF
--- a/lib/src/rules/prefer_relative_imports.dart
+++ b/lib/src/rules/prefer_relative_imports.dart
@@ -8,6 +8,7 @@ import 'package:path/path.dart' as path;
 
 import '../analyzer.dart';
 import '../ast.dart';
+import 'implementation_imports.dart' show samePackage;
 
 const _desc = r'Prefer relative imports for files in `lib/`.';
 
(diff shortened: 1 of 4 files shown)


beroso:
@googlebot I signed it!

PR #2161, opened 2020-06-23
Reference the function unawaited from the meta package

bwilkerson:
The function is being moved from `package:pedantic` to `package:meta`. There do not appear to be any required coding changes (the lint doesn't care which function the future is passed to), but there are a couple of minor doc changes.

This PR should not be merged until after a version of `package:meta` has been published that contains the new function.

@davidmorgan


davidmorgan:
LG, thanks!

PR #2166, opened 2020-06-25
Don’t lint as expressions.

### LINKED ISSUE #2163: Analyzer gives incorrect "unnecessary statement" hint.
Example:
```dart
foo(Object o) {
  o as int;
  return o.toRadixString(16);
}
```
The analyzer hints (checked in nullsafety.dartpad.dev):
```
Avoid using unnecessary statements - line 2
```
However, if I remove the statement, the next line becomes an error because the statement introduces a promotion of `o`, so it's actually quite necessary. 

A statement which causes a promotion does have a side-effect.

pq:
Fixes #2163

I considered some fancy  footwork to  check for really side-effecting expressions but opted instead for the simpler  implementation which I think will ultimately serve us better in the long run (esp. post  NNBD migration).

/cc @bwilkerson @MichaelRFairhurst

### DIFF
--- a/test/rules/unnecessary_statements.dart
+++ b/test/rules/unnecessary_statements.dart
@@ -4,6 +4,12 @@
 
 // test w/ `pub run test -N unnecessary_statements`
 
+String f(Object o) {
+  // See: https://github.com/dart-lang/linter/issues/2163
+  o as int; // OK
+  return /* o.toRadixString(16); */ null;
+}
+
 notReturned() {
   1; // LINT
   1 + 1; // LINT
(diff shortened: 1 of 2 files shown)


test/rules/unnecessary_statements.dart
```diff
@@ -4,6 +4,12 @@
 
 // test w/ `pub run test -N unnecessary_statements`
 
+String f(Object o) {
+  // See: https://github.com/dart-lang/linter/issues/2163
+  o as int; // OK
+  return /* o.toRadixString(16); */ null;
```
bwilkerson, line 10: Unless the commented out code adds value, consider removing it from the test.

PR #2172, opened 2020-06-29
skip deferred functions in unnecessary_lambdas

### LINKED ISSUE #2164: unnecessary_lambdas should ignore functions whose import is deferred
**Describe the issue**
Functions whose imports is deferred cannot be torn-off. Those shouldn't be linted for `unnecessary_lambdas`.

**To Reproduce**
```
import 'package:a.b.c.foo/foo.dart' deferred as foo;

SomeDeferredLoaderModel(
  loader: () => foo.loadLibrary(),
  barProvider: () => foo.provideBar(),
  quxProvider: () => foo.provideQux(),
);
```

**Expected behavior**
`() => foo.provideBar()` and `() => foo.provideQux()` shouldn't be linted as those

pq:
Fixes #2164

/cc @bwilkerson

### DIFF
--- a/lib/src/rules/unnecessary_lambdas.dart
+++ b/lib/src/rules/unnecessary_lambdas.dart
@@ -168,6 +168,12 @@ class _Visitor extends SimpleAstVisitor<void> {
         rule.reportLint(nodeToLint);
       }
     } else if (node is MethodInvocation) {
+      var target = node.target;
+      if (target is SimpleIdentifier &&
+          target?.staticElement is PrefixElement) {
+        return;
+      }
+
       var parent = nodeToLint.parent;
       if (parent is NamedExpression) {
         var argType = parent.staticType;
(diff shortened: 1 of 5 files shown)


lib/src/rules/unnecessary_lambdas.dart
```diff
@@ -168,6 +168,12 @@ class _Visitor extends SimpleAstVisitor<void> {
         rule.reportLint(nodeToLint);
       }
     } else if (node is MethodInvocation) {
+      var target = node.target;
+      if (target is SimpleIdentifier &&
+          target?.staticElement is PrefixElement) {
```
tijoforyou, line 173: Should all prefixed imports be ignored? Or just those that are deferred? For example I could have

```dart
import 'package:quiver/strings.dart' as strings;

int countBlanks(List<String> list) {
  return list.where((s) => strings.isBlank(s)).length;
}
```

This can (and should) be linted? `return list.where(strings.isBlank).length;`

And a nit: No need for the null-aware operator. If we get past the is check, it won't be null anymore.

pq: Good catch(es).  Thanks!

pq: https://github.com/dart-lang/linter/pull/2178

PR #2175, opened 2020-07-06
move annotation_matcher to test_utilities

pq:
Follow-on from https://github.com/dart-lang/linter/issues/2104

/cc @davidmorgan @bwilkerson


pq:
Oh, this is interesting.  Doing this would require making `package:test` or `package:matcher` a non-dev dependency.  (Mulling over possible issues... 🤔 )

lib/src/test_utilities/annotation_matcher.dart
```diff
@@ -0,0 +1,43 @@
+// Copyright (c) 2019, the Dart project authors. Please see the AUTHORS file
```
bwilkerson, line 1: nit: year

pq:
Backing out of this for now.  Adding a non-dev dependency on `package:matcher` feels a little risky...

PR #2180, opened 2020-07-10
Check the variable type in prefer_typing_uninitialized_variables to account for type inference

bwilkerson:
Under null safety it will sometimes be possible for local variables without a type annotation to have a type inferred for them even if they don't have an initializer. This updates the rule to not flag such variables. I believe that this is consistent with the intent of the rule, which is to prevent variables whose type is `dynamic`.

### DIFF
--- a/lib/src/rules/prefer_typing_uninitialized_variables.dart
+++ b/lib/src/rules/prefer_typing_uninitialized_variables.dart
@@ -85,7 +85,7 @@ class _Visitor extends SimpleAstVisitor<void> {
     }
 
     node.variables.forEach((v) {
-      if (v.initializer == null) {
+      if (v.initializer == null && v.declaredElement.type.isDynamic) {
         rule.reportLint(v);
       }
     });


kevmoo:
Would be nice to have a test for this – so we don't regress in the future!

scheglov:
Well, the issue is that it this feature not only has not been implemented in analyzer yet, its [specification](https://github.com/dart-lang/language/pull/991) has not been finished yet :-)

eernstg:
PS: dart-lang/language#1009 is the most recent work on the specification of definite assignment, and the language team decided on going forward with the proposal without support for `var x = null;` as a way to make `x` nullable.

srawlins:
@bwilkerson any updates here? I'm not sure the premise is correct... I think `var foo;` is `dynamic`, even if it is definitely assigned later.

bwilkerson:
I believe that you're right, that `var` without an initializer means `dynamic`, in which case this PR isn't needed.

PR #2183, opened 2020-07-14
add unique codes for unsafe_html variations

pq:
/cc @bwilkerson @srawlins

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -12,7 +12,7 @@ environment:
   sdk: '>=2.6.0 <3.0.0'
 
 dependencies:
-  analyzer: ^0.39.9
+  analyzer: ^0.39.13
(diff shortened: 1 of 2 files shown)


pubspec.yaml
```diff
@@ -12,7 +12,7 @@ environment:
   sdk: '>=2.6.0 <3.0.0'
 
 dependencies:
-  analyzer: ^0.39.9
+  analyzer: ^0.39.13
```
bwilkerson, line 15: Why such a big jump in version numbers?

pq: I believe that's when `SecurityLintCodeWithUniqueName` was introduced?

bwilkerson: Sorry. I was confused about what I was looking at. LGTM.

PR #2197, opened 2020-07-22
Fix RangeError in unnecessary_string_escapes

MichaelRFairhurst:
Previous versions of pkg:analyzer didn't give the correct offsets for
even simple interpolation strings such as "$foo". This prompted the
workaround code here. However, the workaround code here doesn't handle
unclosed quotes:

```dart
  "foo$bar
```

The analyzer API has been fixed, and supports this error recovery, so we
can drop the workaround and that will fix the RangeError we've been
seeing.


pq:
Awesome.  Thanks!

PR #2199, opened 2020-07-23
Fix an NPE in prefer_asserts_in_initializer_lists

bwilkerson:
This is to fix the bug reported in #2198.

There isn't a test for it, though I suspect it happens when there's code of the form
```dart
class C {
  C(this.x);
}
```
(a field initializer for a non-existent field).

I'm happy to add a test, but I'm not sure of the best approach.

### DIFF
--- a/lib/src/rules/prefer_asserts_in_initializer_lists.dart
+++ b/lib/src/rules/prefer_asserts_in_initializer_lists.dart
@@ -72,7 +72,7 @@ class _AssertVisitor extends RecursiveAstVisitor {
             _hasAccessor(element) &&
             !constructorElement.parameters
                 .whereType<FieldFormalParameterElement>()
-                .any((p) => p.field.getter == element);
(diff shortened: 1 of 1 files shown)


pq:
@bwilkerson: I note there are other places where we're accessing this getter.  Should we be guarding them all?

bwilkerson:
> Should we be guarding them all?

Unless we know that they're already guarded, yes. Did you mean other places in this class, or places in other lint rules?

pq:
> other lint rules?

This.

bwilkerson:
I only see one use of `getter` and only one use of `field`, so I'm confused. Can you tell me where you're seeing these references?

pq:
Ah!  Apologies, I was looking at `fields` (not `field`).  Sorry about that!

bwilkerson:
No problem. I'm just glad you were being diligent!

PR #2208, opened 2020-07-31
add lint unnecessary_nullable_for_final_variable_declarations

### LINKED ISSUE #2203: final variable with nullable type initialized with non-nullable type
```
int m() => 2;
final int? a = m(); // LINT to use int instead of int?
```

a14n:
# Description

Use non-nullable type for final variable initialized with non-nullable value.

**BAD:**
```
final int? i = 1;
```
**GOOD:**
```
final int i = 1;
```

Fixes #2203

### DIFF
--- a/lib/src/rules/unnecessary_nullable_for_final_variable_declarations.dart
+++ b/lib/src/rules/unnecessary_nullable_for_final_variable_declarations.dart
@@ -8,12 +8,13 @@ import 'package:analyzer/dart/ast/visitor.dart';
 
 import '../analyzer.dart';
 
-const _desc = r'Use non-nullable type '
-    'for final variable initialized with non-nullable value.';
+const _desc = r'Use a non-nullable type for a final variable initialized '
+    'with a non-nullable value.';
 
 const _details = r'''
 
-Use non-nullable type for final variable initialized with non-nullable value.
+Use a non-nullable type for a final variable initialized with a non-nullable
+value.
 
 **BAD:**
 ```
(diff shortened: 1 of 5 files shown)


lib/src/rules/unnecessary_nullable_for_final_variable_declarations.dart
```diff
@@ -0,0 +1,108 @@
...
+import 'package:analyzer/dart/ast/visitor.dart';
+
+import '../analyzer.dart';
+
+const _desc = r'Use non-nullable type '
+    'for final variable initialized with non-nullable value.';
+
+const _details = r'''
+
+Use non-nullable type for final variable initialized with non-nullable value.
```
bwilkerson, line 16: "Use a non-nullable type for a final variable initialized with a non-nullable value."

lib/src/rules/unnecessary_nullable_for_final_variable_declarations.dart
```diff
@@ -0,0 +1,108 @@
...
+
+  @override
+  void visitFieldDeclaration(FieldDeclaration node) {
+    if (!_isNonNullableEnabled) return;
+
+    for (var variable in node.fields.variables) {
+      // We could also include public members in private classes but to do that
+      // we'd need to ensure that there are no instances of either the
+      // enclosing class or any subclass of the enclosing class that are ever
+      // accessible outside this library.
```
bwilkerson, line 72: I don't think that's actually the issue here. The only reason I can think of for not linting public fields is because either a superclass or subclass might require the nullable type because of inheritance semantics. We could check the signature against superclasses, but we can't know find all subclasses unless the class is private (because then all of the subclasses will be in the same library).

We should update the comment, but I'm _not_ recommending that we do the extra work to lint public members in private classes. I don't think that case will come up often enough to be worth the effort, and we can always add it later if I'm wrong.

lib/src/rules/unnecessary_nullable_for_final_variable_declarations.dart
a14n: Done

lib/src/rules/unnecessary_nullable_for_final_variable_declarations.dart
a14n: Comment removed.

PR #2209, opened 2020-08-03
Lint proposal: avoid_type_to_string

nscobie:
# Description

This rule would suggest users avoid calling `toString()` on objects of type `Type` (declared in `dart:core`), since the result may be minified by some compilers in release mode.

# Context

This was inspired by a recent post on an internal mailing list. No official verdict on whether this is a good idea has been reached at the time of writing, but it seemed useful so I decided to try implementing it for fun. If people want to move forward with this lint rule, then this PR could be a good starting point.

Credit for the idea and reasoning goes to @matanlurey 

**Should this be further discussed on this PR, a new issue, or the mailing list?**

# Reasoning

> \<Type\>.toString() does not contractually return the user-defined name of the Type (or underlying class ). What that means in practice is that development mode compilers (such as DDC, or Dart JIT) where code-size is not a concern use the full name, but release-mode compilers (such as Dart2JS or Dart AOT) often choose to minify these symbols.

# ~~Known issues~~

~~A false positive is thrown when attempting to call a local lambda function declared inside another function, if the outer class extends `Type`.~~

<details>

```dart
class GoodWithType extends Type {
  void good() {
    String toString() => null;
    toString(); // Should be OK, but the rule currently does not see the local toString lambda which hides Object.toString
  }
}
```

</details>

**Edit: this is now fixed.**

### DIFF
--- a/lib/src/rules/avoid_type_to_string.dart
+++ b/lib/src/rules/avoid_type_to_string.dart
@@ -74,6 +74,8 @@ class AvoidTypeToString extends LintRule implements NodeLintRule {
 
     // These nodes delegate to general visitArgumentList,
     // since SimpleAstVisitor only calls visits for concrete node subtypes.
+    // TODO: replace these with addArgumentList(...) once it's added to the registry API.
+    // Context: https://github.com/dart-lang/linter/pull/2209#discussion_r465196745
     registry.addFunctionExpressionInvocation(this, visitor);
     registry.addInstanceCreationExpression(this, visitor);
     registry.addRedirectingConstructorInvocation(this, visitor);
--- a/lib/src/rules/avoid_type_to_string.dart
+++ b/lib/src/rules/avoid_type_to_string.dart
@@ -57,7 +57,7 @@ class AvoidTypeToString extends LintRule implements NodeLintRule {
             name: 'avoid_type_to_string',
             description: _desc,
             details: _details,
-            group: Group.style);
+            group: Group.errors);
 
   @override
   void registerNodeProcessors(NodeLintRegistry registry,
--- a/lib/src/rules/avoid_type_to_string.dart
+++ b/lib/src/rules/avoid_type_to_string.dart
@@ -141,7 +141,7 @@ class _Visitor extends SimpleAstVisitor {
     final targetType = (node.realTarget?.staticType is InterfaceType)
         ? node.realTarget.staticType as InterfaceType
         : thisType;
-    _reportIfToStringOnCoreTypeClass(node, targetType, node.methodName);
+    _reportIfToStringOnCoreTypeClass(targetType, node.methodName);
   }
 
   @override
@@ -154,23 +154,21 @@ class _Visitor extends SimpleAstVisitor {
       final targetType = (expression.realTarget?.staticType is InterfaceType)
           ? expression.realTarget?.staticType as InterfaceType
           : thisType;
-      _reportIfToStringOnCoreTypeClass(
-          expression, targetType, expression.propertyName);
+      _reportIfToStringOnCoreTypeClass(targetType, expression.propertyName);
     } else if (expression is PrefixedIdentifier) {
       final prefixType = expression.prefix.staticType;
       if (prefixType is InterfaceType) {
-        _reportIfToStringOnCoreTypeClass(
-            expression, prefixType, expression.identifier);
+        _reportIfToStringOnCoreTypeClass(prefixType, expression.identifier);
       }
     } else if (expression is SimpleIdentifier) {
-      _reportIfToStringOnCoreTypeClass(expression, thisType, expression);
+      _reportIfToStringOnCoreTypeClass(thisType, expression);
     }
   }
 
-  void _reportIfToStringOnCoreTypeClass(AstNode offendingNode,
+  void _reportIfToStringOnCoreTypeClass(
       InterfaceType targetType, SimpleIdentifier methodIdentifier) {
     if (_isToStringOnCoreTypeClass(targetType, methodIdentifier)) {
-      rule.reportLint(offendingNode);
+      rule.reportLint(methodIdentifier);
     }
   }
 
--- a/lib/src/rules/avoid_type_to_string.dart
+++ b/lib/src/rules/avoid_type_to_string.dart
@@ -9,7 +9,8 @@ import 'package:analyzer/dart/element/type.dart';
 
 import '../analyzer.dart';
 
-const _desc = r'Avoid <Type>.toString() in production code since results may be minified.';
+const _desc =
+    r'Avoid <Type>.toString() in production code since results may be minified.';
 
 const _details = r'''
 
@@ -50,9 +51,8 @@ Object baz(Thing myThing) {
 
 ''';
 
-// TODO: handle local toString lambda declared in scope
 class AvoidTypeToString extends LintRule implements NodeLintRule {
-  AvoidTypeToString() 
+  AvoidTypeToString()
       : super(
             name: 'avoid_type_to_string',
             description: _desc,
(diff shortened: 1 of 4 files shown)


lib/src/rules/avoid_type_to_string.dart
```diff
@@ -0,0 +1,151 @@
...
+  @override
+  void visitMixinDeclaration(MixinDeclaration node) {
+    // Nodes visited in DFS, so this will be set before each visitMethodInvocation.
+    thisType = node.declaredElement.thisType;
+  }
+
+  @override
+  void visitExtensionDeclaration(ExtensionDeclaration node) {
+    // Nodes visited in DFS, so this will be set before each visitMethodInvocation.
+    thisType = node.declaredElement.extendedType as InterfaceType;
```
bwilkerson, line 94: This isn't safe. The extended type isn't required to be an interface type (it could, for example, be a function type). You'll need to test to see whether it is an interface type and set `thisType` to `null` if it isn't (to clear out any previously stored value).

lib/src/rules/avoid_type_to_string.dart
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+  @override
+  void visitMethodInvocation(MethodInvocation node) {
+    final methodName = node.methodName.name;
+    final targetType = (node.realTarget?.staticType is InterfaceType)
+        ?  node.realTarget.staticType as InterfaceType : thisType;
```
bwilkerson, line 101: Contrary to its name, a `MethodInvocation` doesn't always represent the invocation of a method. It is also used to represent function invocations because syntactically they can look just like a method invocation. In such a case you don't want to default to using `thisType`.

You can tell the difference by looking at `node.methodName.staticElement`. That element could also be used to determine whether the method being invoked is from `Object` (see `_rootMethodIsCoreToString`) by asking `method.enclosingElement is ClassElement && method.enclosingElement.isDartCoreObject`.

lib/src/rules/avoid_type_to_string.dart
```diff
@@ -0,0 +1,153 @@
...
+
+  @override
+  void visitExtensionDeclaration(ExtensionDeclaration node) {
+    // Nodes visited in DFS, so this will be set before each visitMethodInvocation.
+    thisType = node.declaredElement.extendedType as InterfaceType;
+  }
+
+  @override
+  void visitMethodInvocation(MethodInvocation node) {
+    final methodName = node.methodName.name;
```
bwilkerson, line 99: It would be more efficient to return here if `methodName != 'toString'`.

test/rules/avoid_type_to_string.dart
```diff
@@ -0,0 +1,110 @@
...
+    TypeGrandChildNoOverride().toString(); // LINT
+  }
+}
+
+class BadWithType extends Type {
+  void doBad() {
+    toString(); // LINT
+    this.toString(); // LINT
+
+    Future.value('hello').whenComplete(toString); // LINT
```
bwilkerson, line 43: I believe that the issue here is that these expressions appear as arguments and not that they are arguments to `whenComplete`. Consider writing a dummy function like
```
void takesFunction(Function f) {}
```
that's used to test that these expressions are caught.

lib/src/rules/avoid_type_to_string.dart
```diff
@@ -0,0 +1,153 @@
...
+    final targetType = (node.realTarget?.staticType is InterfaceType)
+        ?  node.realTarget.staticType as InterfaceType : thisType;
+    final library = node.methodName.staticElement?.library;
+    if (_isToStringOnCoreTypeClass(methodName, targetType, library)) {
+      rule.reportLint(node);
+    }
+
+    node.argumentList.arguments.forEach(_validateArgument);
```
bwilkerson, line 107: `MethodInvocation` is not the only structure that can include an argument list; both `InstanceCreationExpression` and `FunctionExpressionInvocation` can as well. A better approach would be to remove this line and add a `visitArgumentList` method to catch all explicit argument lists. (That still won't catch implicit arguments such as the right operand in a `BinaryExpression` or the index in an `IndexExpression`, but I suspect that those are cases that don't occur often enough to worry about.)

lib/src/rules/avoid_type_to_string.dart
```diff
@@ -0,0 +1,153 @@
...
+    String methodName;
+    InterfaceType targetType;
+    LibraryElement library;
+    if (expression is PropertyAccess) {
+      methodName = expression.propertyName.name;
+      targetType = (expression.realTarget?.staticType is InterfaceType) 
+          ? expression.realTarget?.staticType as InterfaceType : thisType;
+      library = expression.propertyName.staticElement?.library;
+    }
+    else if (expression is SimpleIdentifier) {
```
bwilkerson, line 120: Please run the dart formatter over the code.

lib/src/rules/avoid_type_to_string.dart
```diff
@@ -0,0 +1,153 @@
...
+      rule.reportLint(expression);
+    }
+  }
+
+  bool _isToStringOnCoreTypeClass(String methodName, InterfaceType targetType, LibraryElement library) =>
+      methodName == 'toString'
+      && _coreTypeClassInAncestors(targetType)
+      && _rootMethodIsCoreToString(targetType, library);
+
+  bool _coreTypeClassInAncestors(InterfaceType targetType) {
```
bwilkerson, line 139: Consider using the linter context to access the type `Type` (`context.typeProvider.typeType`) and then using the type system (`context.typeSystem`) to ask whether the `targetType` is a subtype of `Type`. That might be slightly cleaner.

lib/src/rules/avoid_type_to_string.dart
nscobie: Good catch!

lib/src/rules/avoid_type_to_string.dart
nscobie: Ah makes sense. Thanks for the sample code, it helped a lot!

lib/src/rules/avoid_type_to_string.dart
nscobie: Good point. After refactoring things I still don't check this immediately, since I want to determine `targetType` first before calling the shared `_isToStringOnCoreTypeClass` function. However, the method name is checked much sooner now.

test/rules/avoid_type_to_string.dart
nscobie: Very good point, that feels much better.

lib/src/rules/avoid_type_to_string.dart
nscobie: Excellent point! I've added checks to `InstanceCreationExpression`, `FunctionExpressionInvocation`, `RedirectingConstructorInvocation`, and `visitSuperConstructorInvocation`.

lib/src/rules/avoid_type_to_string.dart
nscobie: Oopsies. :)

lib/src/rules/avoid_type_to_string.dart
nscobie: This was very good advice, thank you.

nscobie:
@bwilkerson Thank you for the helpful feedback! I've refactored and improved things quite a bit by addressing your suggestions, but there are probably still edge cases I'm missing.

I believe Travis CI is failing for this PR due to a pre-existing ordering issue with `example/all.yaml` documented by #2210

lib/src/rules/avoid_type_to_string.dart
```diff
@@ -0,0 +1,189 @@
...
+        _Visitor(this, context.typeSystem, context.typeProvider.typeType);
+    // Gathering meta information at these nodes.
+    // Nodes visited in DFS, so this will be called before
+    // each visitMethodInvocation.
+    registry.addClassDeclaration(this, visitor);
+    registry.addMixinDeclaration(this, visitor);
+    registry.addExtensionDeclaration(this, visitor);
+
+    // These nodes delegate to general visitArgumentList,
+    // since SimpleAstVisitor only calls visits for concrete node subtypes.
```
bwilkerson, line 76: Sorry, I should have been more explicit. What I was thinking is that this code could register to visit argument lists:
```
registry.addArgumentList(this, visitor);
```
and you could process all argument lists by overriding `visitArgumentList` (which you are now already doing). In other words, that one line could replace lines 77-80. This has the advantage that if a new language construct is added and the AST gets a new node containing an argument list, this lint will continue to work without needing to be updated.

lib/src/rules/avoid_type_to_string.dart
```diff
@@ -0,0 +1,189 @@
...
+      }
+    } else if (expression is SimpleIdentifier) {
+      _reportIfToStringOnCoreTypeClass(expression, thisType, expression);
+    }
+  }
+
+  void _reportIfToStringOnCoreTypeClass(AstNode offendingNode,
+      InterfaceType targetType, SimpleIdentifier methodIdentifier) {
+    if (_isToStringOnCoreTypeClass(targetType, methodIdentifier)) {
+      rule.reportLint(offendingNode);
```
bwilkerson, line 173: We generally associate diagnostics with a single token where possible because it improves the UX. In this case, it would be better to report the lint on `methodIdentifier` and drop the `offendingNode` parameter.

lib/src/rules/avoid_type_to_string.dart
nscobie: > This has the advantage that if a new language construct is added and the AST gets a new node containing an argument list, this lint will continue to work without needing to be updated.

I had the same concern, but I wasn't able to find an `addArgumentList` method on `registry`, or [anywhere else in dart-lang/sdk](https://github.com/dart-lang/sdk/search?q=addArgumentList&unscoped_q=addArgumentList). The closest option seems to be `addTypeArgumentList`, but that seems to be for things like function declarations?

lib/src/rules/avoid_type_to_string.dart
nscobie: Makes sense. 👍

lib/src/rules/avoid_type_to_string.dart
bwilkerson: You're right, that node appears to have been missed when writing the registry API. I'll get that fixed, but it will be a while before it's published and available for the linter package to use. In the meantime, this code is the best we can do, but a TODO comment suggesting this change would be good.

nscobie: 👍

nscobie:
Heads up, I've been adding a lot of successive commits rather than amending the first. If this ends up getting merged it should probably be squashed first. I would do that now, but I'm afraid of messing up the revision history of PR feedback.

nscobie:
> LGTM. I don't have any opinion about squashing.

I don't have permission to merge, but other repos I am familiar with have a "Squash and merge" option in the merge button dropdown. If that option is available for this repo, I think that would be ideal since my commit messages might not make sense outside the context of this PR, but I don't mind too much either way.

Thank you so much for all your thoughtful feedback and helpful advice!

pq:
Yep, we can squash and merge no problem.

@nscobie, I'd like to add something to `AUTHORS`.  How would you like to be attributed?

PR #2213, opened 2020-08-06
add lint unnecessary_null_check

a14n:
Don't apply null check when nullable value is accepted.

**BAD:**

```
f(int? i);
m() {
  int? j;
  f(j!);
}
```

**GOOD:**

```
f(int? i);
m() {
  int? j;
  f(j);
}
```

### DIFF
--- a/lib/src/rules/unnecessary_null_check.dart
+++ b/lib/src/rules/unnecessary_null_check.dart
@@ -5,7 +5,6 @@
 import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/token.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
-import 'package:analyzer/dart/element/element.dart';
 import 'package:analyzer/dart/element/type.dart';
 
 import '../analyzer.dart';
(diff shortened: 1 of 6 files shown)


lib/src/rules/unnecessary_null_check.dart
```diff
@@ -0,0 +1,114 @@
...
+  _Visitor(this.rule, this.context);
+
+  final LintRule rule;
+  final LinterContext context;
+
+  @override
+  void visitPostfixExpression(PostfixExpression node) {
+    if (node.operator.type != TokenType.BANG) return;
+
+    final parent = node.unParenthesized.parent;
```
bwilkerson, line 67: `node.unParenthesized` is equivalent to `node` here because `node` will never be a `ParenthesizedExpression` (also on line 96). I think you're trying to handle having parentheses around the postfix expression, but to do that you need to write
```
var unwappedNode = node;
while (unwrappedNode.parent is ParenthesizedExpression) {
  unwrappedNode = unwrappedNode.parent;
}
```

Note that there aren't any tests for this code path.

lib/src/rules/unnecessary_null_check.dart
```diff
@@ -0,0 +1,114 @@
...
+    DartType expectedType;
```
bwilkerson, line 68: Consider extracting lines 100-102 into a separate method. Then we can remove the local variable and clean up the control flow below. For example, line 71 could become
```
reportIfNullable(node, parent.declaredElement.type);
return;
```
It would also be more efficient.

lib/src/rules/unnecessary_null_check.dart
```diff
@@ -0,0 +1,114 @@
...
+    // in variable declaration
+    if (parent is VariableDeclaration) {
+      expectedType = parent.declaredElement.type;
+    }
+    // as named parameter of function
+    if (parent is NamedExpression && parent.parent is ArgumentList) {
```
bwilkerson, line 74: I think lines 74-93 could be simplified to
```
if (parent is NamedExpression) {
  unwrappedNode = parent;
  parent = parent.parent;
}
if (parent is ArgumentList) {
  reportIfNullable(node, unwrappedNode.staticParameterElement.type);
  return;
}

File: lib/src/rules/unnecessary_null_check.dart (hunk 1 above)
Comment (a14n) [reply lib/src/rules/unnecessary_null_check.dart]: Tests for parenthesis added and I used :

```
    var realNode = node
        .thisOrAncestorMatching((e) => e.parent is! ParenthesizedExpression);
```

File: lib/src/rules/unnecessary_null_check.dart (hunk 2 above)
Comment (a14n) [reply lib/src/rules/unnecessary_null_check.dart]: done

File: lib/src/rules/unnecessary_null_check.dart (hunk 3 above)
Comment (a14n) [reply lib/src/rules/unnecessary_null_check.dart]: So much shorter! :+1:

**Comment by a14n** [comment]
PTAL

File: lib/src/rules/unnecessary_null_check.dart [hunk 4]
```diff
@@ -0,0 +1,97 @@
...
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/type.dart';
+
+import '../analyzer.dart';
+
+const _desc = r'Unnecessary null check.';
+
+const _details = r'''
+
+Don't apply null check when nullable value is accepted.
```
pq, line 16: =>

    Don't apply a null check when a nullable value is accepted.

?

a14n:
ping

a14n:
@pq is there something blocking my PR?

pq:
@a14n: apologies for the delay!  I was OOO and slowly making my way through a backlog.

Sorry to have left you hanging.  Thanks for the contribution and nudge!

PR #2216, opened 2020-08-06
do lint unnecessary_this on classes and mixins

pq:
/cc @bwilkerson


pq:
There are actually diagnostics in these cases so not something we want to lint as well.

PR #2224, opened 2020-08-18
check for Future subtypes

### LINKED ISSUE #2211: `unawaited_futures` should work w/ sub-types of `Future`
Unawaited  sub-types of `Future` are not flagged but should be.

pq:
Fixes #2211

/cc @bwilkerson

### DIFF
--- a/lib/src/rules/unawaited_futures.dart
+++ b/lib/src/rules/unawaited_futures.dart
@@ -5,6 +5,7 @@
 import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
 import 'package:analyzer/dart/element/element.dart';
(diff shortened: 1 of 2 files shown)


lib/src/rules/unawaited_futures.dart
```diff
@@ -81,7 +82,11 @@ class _Visitor extends SimpleAstVisitor<void> {
     if (expr is AssignmentExpression) return;
 
     var type = expr?.staticType;
-    if (type?.isDartAsyncFuture == true) {
+    if (type == null) {
+      return;
+    }
+    if (type.isDartAsyncFuture ||
```
bwilkerson, line 88: It looks like this test will be duplicated when `implementsInterface` is invoked, so we can probably remove the first condition.

pq: Nice.  Thanks!

PR #2229, opened 2020-09-02
update to preferred analyzer API

pq:
/cc @scheglov @bwilkerson

### DIFF
--- a/lib/src/rules/prefer_final_fields.dart
+++ b/lib/src/rules/prefer_final_fields.dart
@@ -190,7 +190,8 @@ class _Visitor extends SimpleAstVisitor<void> {
           if (isFieldInAllConstructors) {
             rule.reportLint(variable);
           }
-        } else if (element.initializer != null) {
+        } else if (element is PropertyInducingElement &&
+            element.hasInitializer) {
           rule.reportLint(variable);
         }
       }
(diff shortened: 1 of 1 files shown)


lib/src/rules/prefer_final_fields.dart
```diff
@@ -190,7 +190,8 @@ class _Visitor extends SimpleAstVisitor<void> {
           if (isFieldInAllConstructors) {
             rule.reportLint(variable);
           }
-        } else if (element.initializer != null) {
+        } else if (element is PropertyInducingElement &&
```
scheglov, line 193: This is semantically correct, but a tiny bit confusing the reader.
It is not that there is an *option* that the `element` has any other type.
As such, it might be more explicit to cast it in the `final element` declaration.

scheglov (approved):

lib/src/rules/prefer_final_fields.dart
pq: If that's safe/preferred, I'm all for it and I updated.  Take another look to make sure I got it right?

scheglov (approved):
LGTM

bwilkerson (approved):

pq:
Thanks!

PR #2240, opened 2020-09-16
add tighten_type_of_initializing_formals

### LINKED ISSUE #2237: Lint to annotate parameter with non-nullable if assert !=null
```dart
class A {
  A.c1(this.p) : assert(p != null);
  A.c2(this.p);
  final String? p;
}
```
the code should be:
```dart
class A {
  A.c1(String this.p) : assert(p != null);
  A.c2(this.p);
  final String? p;
}
```

a14n:
# Description

Tighten type of initializing formal if an non null assert exists.
**BAD:**
```dart
class A {
  A.c1(this.p) : assert(p != null);
  A.c2(this.p);
  final String? p;
}
```
**GOOD:**
```dart
class A {
  A.c1(String this.p) : assert(p != null);
  A.c2(this.p);
  final String? p;
}
```

Fixes #2237


lib/src/rules/tighten_type_of_initializing_formals.dart
```diff
@@ -0,0 +1,92 @@
...
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/element.dart';
+
+import '../analyzer.dart';
+
+const _desc = r'Tighten type of initializing formal.';
+
+const _details = r'''
+
+Tighten type of initializing formal if an non null assert exists.
```
bwilkerson, line 17: "an non null" --> "a non-null"

lib/src/rules/tighten_type_of_initializing_formals.dart
```diff
@@ -0,0 +1,92 @@
...
+  _Visitor(this.rule, this.context);
+
+  final LintRule rule;
+  final LinterContext context;
+
+  bool _isNonNullableEnabled;
+
+  @override
+  void visitCompilationUnit(CompilationUnit node) {
+    _isNonNullableEnabled = node.featureSet.isEnabled(Feature.non_nullable);
```
bwilkerson, line 68: Not a request for a change in this PR, just an observation. We should just add this to the `LinterContext` and then we could simply not register in `registerNodeProcessors`. That would be more efficient.

lib/src/rules/tighten_type_of_initializing_formals.dart
```diff
@@ -0,0 +1,92 @@
...
+
+    node.initializers
+        .whereType<AssertInitializer>()
+        .map((e) => e.condition)
+        .whereType<BinaryExpression>()
+        .where((element) => element.operator.type == TokenType.BANG_EQ)
+        .map((e) => e.rightOperand is NullLiteral
+            ? e.leftOperand
+            : e.leftOperand is NullLiteral ? e.rightOperand : null)
+        .where((e) => context.typeSystem.isNullable(e.staticType))
```
bwilkerson, line 83: As far as I can tell, `e` can be `null`, which will throw in `.staticType`.

I think this will happen when neither operand is `null`:
```
C(this.p, this.q): assert(p != q);
```

lib/src/rules/tighten_type_of_initializing_formals.dart
```diff
@@ -0,0 +1,92 @@
...
+import 'package:analyzer/dart/element/element.dart';
+
+import '../analyzer.dart';
+
+const _desc = r'Tighten type of initializing formal.';
+
+const _details = r'''
+
+Tighten type of initializing formal if an non null assert exists.
+
```
pq, line 18: A little more rationale would be great here if possible.  Why are we recommending this?

test/rules/experiments/nnbd/rules/tighten_type_of_initializing_formals.dart
```diff
@@ -0,0 +1,29 @@
...
+       assert(p2 != null),
+       assert(p4 != null);
+
+  String? p;
+  String? p1;
+  String? p2;
+  String? p3;
+  String? p4;
+  String? p5;
+}
```
pq, line 29: Trailing newline?

lib/src/rules/tighten_type_of_initializing_formals.dart
a14n: Is the following better:

> Tighten type of initializing formal if a non-null assert exists. Using a
non-nullable type will provide static warnings in the analyzer instead of
runtime errors.

lib/src/rules/tighten_type_of_initializing_formals.dart
a14n: Fixed.

test/rules/experiments/nnbd/rules/tighten_type_of_initializing_formals.dart
a14n: Fixed

lib/src/rules/tighten_type_of_initializing_formals.dart
bwilkerson: To allow the type system to catch problems rather than have them only be caught at run-time. See #2237 for more discussion.

lib/src/rules/tighten_type_of_initializing_formals.dart
a14n: Fixed

lib/src/rules/tighten_type_of_initializing_formals.dart
a14n: Updated with Brian comment :)

pq: Thanks!

bwilkerson (approved):

lib/src/rules/tighten_type_of_initializing_formals.dart
pq: /fyi @scheglov

lib/src/rules/tighten_type_of_initializing_formals.dart
```diff
@@ -0,0 +1,94 @@
...
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/element.dart';
+
+import '../analyzer.dart';
+
+const _desc = r'Tighten type of initializing formal.';
+
+const _details = r'''
+
+Tighten type of initializing formal if a non-null assert exists. This allow the
```
pq, line 17: =>

This allows

a14n: fixed

pq: Thanks!

pq:
Regarding:

https://travis-ci.org/github/dart-lang/linter/jobs/727830934#L601

I'm surprised to see a formatting failure since we have `test/validate_format_test.dart`...  🤔

pq:
(Not surprised enough to hold this up though.)

PR #2244, opened 2020-09-18
add cast_nullable_to_non_nullable

### LINKED ISSUE #2235: Consider linting on casts from nullable to non-nullable types.
When casting to a type in null safe Dart, it's easy to forget to account for null.  For example, casting directly from `dynamic` to `int` doesn't account for `null`: the user should either have handle the null case before casting, or else should cast to `int?` and then handle null from there.  Some users may prefer to enforce this via a lint. See https://github.com/dart-lang/language/issues/1170 for further discussion.

cc @Hixie

a14n:
# Description

Don't cast a nullable value to a non nullable subtype. This hide a null check
and most of the time it is not what is expected.
**BAD:**
```
class A {}
class B extends A {}
A? a;
var v = a as B;
```
**GOOD:**
```
class A {}
class B extends A {}
A? a;
var v = a! as B;
var v = a as B?;
```

Fixes #2235

### DIFF
--- a/example/all.yaml
+++ b/example/all.yaml
@@ -52,6 +52,7 @@ linter:
     - camel_case_types
     - cancel_subscriptions
     - cascade_invocations
+    - cast_to_non_nullable_child
     - close_sinks
     - comment_references
     - constant_identifier_names
--- a/example/all.yaml
+++ b/example/all.yaml
@@ -52,7 +52,7 @@ linter:
     - camel_case_types
     - cancel_subscriptions
     - cascade_invocations
-    - cast_to_non_nullable_child
+    - cast_nullable_to_non_nullable
     - close_sinks
     - comment_references
     - constant_identifier_names
--- a/lib/src/rules.dart
+++ b/lib/src/rules.dart
@@ -53,6 +53,7 @@ import 'rules/camel_case_extensions.dart';
 import 'rules/camel_case_types.dart';
 import 'rules/cancel_subscriptions.dart';
 import 'rules/cascade_invocations.dart';
+import 'rules/cast_to_non_nullable_child.dart';
 import 'rules/close_sinks.dart';
 import 'rules/comment_references.dart';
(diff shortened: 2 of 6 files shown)


bwilkerson (commented):
This PR differs in a couple of interesting ways from what the original issue discussed:
- it explicitly allows casts from `dynamic`, and
- it explicitly allows casts from `T?` to `T`.

I'm curious to understand why. Note that the description also makes it sound like these cases will be covered, so if there's a good reason for the exclusions we'll need to update the description to explain what's really covered here.

a14n:
> it explicitly allows casts from dynamic

Trying the lint on flutter code base I saw a several case where the expression type was dynamic and I'm not sure it's worth to replace `myDyn.value as T` by `myDyn.value! as T`.

> it explicitly allows casts from T? to T.

That's what I understood from [@hixie's comments](https://github.com/dart-lang/linter/issues/2235#issuecomment-690912865)

bwilkerson:
I think it would be worth discussing the specific cases where the expression type is `dynamic` on the original issue. I'm interested to know whether ignoring these cases is actually the right approach.

Also, I interpreted those comments to mean that he's more concerned about one case than the other, not that he's unconcerned with the other, so we should clarify.

a14n:
Re-reading the related issues I may have misunderstood. I'll remove the restriction (and rename the lint)

bwilkerson:
I'd still prefer to get clarification on the other issue. I'm not actually claiming to know the answer to the question of what this lint should do, just wanting to understand so that I can check the implementation against the intent. :-)

a14n:
PTAL

I rename the lint to `cast_nullable_to_non_nullable` according to https://github.com/dart-lang/linter/issues/2235#issuecomment-696823493

bwilkerson (approved):

pq (approved):
🎉

PR #2245, opened 2020-09-18
add null_check_on_nullable_type_parameter

### LINKED ISSUE #2235: Consider linting on casts from nullable to non-nullable types.

### LINKED ISSUE #2243: Consider a lint warning on uses of the null check operator on variables of type `T?`
Given a generic type parameter `T` which has a nullable bound (e.g. the default bound of `Object?`), it is very easy to introduce erroneous null checks when working with a variable of type `T?`.  Specifically, it is not uncommon to have `T? x;` and want to assert that `x` has been set to a valid value of type `T`.  A common mistake is to do so using `x!`.  This is almost always incorrect, since if `T` is a nullable type, `x` may validly hold `null` as a value of type `T`.  Example:

```dart
T run<T>(T callback()) {
  T? result;
   (() { result = callback(); })();
  return result!;
}

void test() {
   assert(run<int?>(() => null) == null);  // Erroneously throws an error
}
```

A recommended pattern for writing this code is to instead to cast to `T`:

```dart
T run<T>(T callback()) {
  T? result;
   (() { result = callback(); })();
  return result as T;
}

void test() {
   assert(run<int?>(() => null) == null);  // succeeds
}
```

I think the conditions under which the lint would fire would be:
 - If the null check operator is applied to something of type `T?`
 - And `T` is a generic type parameter
 - And the bound of `T` is potentially nullable

cc @Hixie @matanlurey @lrhn @munificent @eernstg @natebosch @jakemac53 for any thoughts on this.

a14n:
# Description

Don't use null check on a potentially nullable type parameter.
**BAD:**
```dart
T run<T>(T callback()) {
  T? result;
   (() { result = callback(); })();
  return result!;
}
```
**GOOD:**
```dart
T run<T>(T callback()) {
  T? result;
   (() { result = callback(); })();
  return result as T;
}
```

Fixes #2243

In this PR I also extended `unnecessary_null_checks` to return cases.

### DIFF
--- a/lib/src/rules/null_check_on_nullable_type_parameter.dart
+++ b/lib/src/rules/null_check_on_nullable_type_parameter.dart
@@ -1,0 +1,91 @@
+// Copyright (c) 2020, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/analysis/features.dart';
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/token.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+
+import '../analyzer.dart';
+import 'unnecessary_null_checks.dart';
+
+const _desc = r"Don't use null check on a potentially nullable type parameter.";
+
+const _details = r'''
+
+Don't use null check on a potentially nullable type parameter.
+
+Given a generic type parameter `T` which has a nullable bound (e.g. the default
+bound of `Object?`), it is very easy to introduce erroneous null checks when
+working with a variable of type `T?`. Specifically, it is not uncommon to have
+`T? x;` and want to assert that `x` has been set to a valid value of type `T`.
+A common mistake is to do so using `x!`. This is almost always incorrect, since
+if `T` is a nullable type, `x` may validly hold `null` as a value of type `T`.
+
+**BAD:**
+```
+T run<T>(T callback()) {
+  T? result;
+   (() { result = callback(); })();
+  return result!;
+}
+```
+
+**GOOD:**
+```
+T run<T>(T callback()) {
+  T? result;
+   (() { result = callback(); })();
+  return result as T;
+}
+```
+
+''';
+
+class NullCheckOnNullableTypeParameter extends LintRule
+    implements NodeLintRule {
+  NullCheckOnNullableTypeParameter()
+      : super(
+          name: 'null_check_on_nullable_type_parameter',
+          description: _desc,
+          details: _details,
+          maturity: Maturity.experimental,
+          group: Group.style,
+        );
+
+  @override
+  void registerNodeProcessors(NodeLintRegistry registry,
+      [LinterContext context]) {
+    final visitor = _Visitor(this, context);
+    registry.addCompilationUnit(this, visitor);
+    registry.addPostfixExpression(this, visitor);
+  }
+}
+
+class _Visitor extends SimpleAstVisitor<void> {
+  _Visitor(this.rule, this.context);
+
+  final LintRule rule;
+  final LinterContext context;
+
+  bool _isNonNullableEnabled;
+
+  @override
+  void visitCompilationUnit(CompilationUnit node) {
+    _isNonNullableEnabled = node.featureSet.isEnabled(Feature.non_nullable);
+  }
+
+  @override
+  void visitPostfixExpression(PostfixExpression node) {
+    if (!_isNonNullableEnabled) return;
+    if (node.operator.type != TokenType.BANG) return;
+
+    final expectedType = getExpectedType(node);
+    if (context.typeSystem.isPotentiallyNullable(node.operand.staticType) &&
+        expectedType != null &&
+        context.typeSystem.isPotentiallyNullable(expectedType)) {
+      rule.reportLintForToken(node.operator);
+    }
+  }
+}
--- a/lib/src/rules/null_check_on_nullable_type_parameter.dart
+++ b/lib/src/rules/null_check_on_nullable_type_parameter.dart
@@ -83,11 +83,12 @@ class _Visitor extends SimpleAstVisitor<void> {
     if (node.operator.type != TokenType.BANG) return;
 
     final expectedType = getExpectedType(node);
-    if (node.operand.staticType is TypeParameterType &&
-        context.typeSystem.isNullable(node.operand.staticType) &&
+    final type = node.operand.staticType;
+    if (type is TypeParameterType &&
+        context.typeSystem.isNullable(type) &&
         expectedType != null &&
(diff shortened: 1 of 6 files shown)


lib/src/rules/null_check_on_nullable_type_parameter.dart
```diff
@@ -0,0 +1,91 @@
...
+    _isNonNullableEnabled = node.featureSet.isEnabled(Feature.non_nullable);
+  }
+
+  @override
+  void visitPostfixExpression(PostfixExpression node) {
+    if (!_isNonNullableEnabled) return;
+    if (node.operator.type != TokenType.BANG) return;
+
+    final expectedType = getExpectedType(node);
+    if (context.typeSystem.isPotentiallyNullable(node.operand.staticType) &&
```
bwilkerson, line 85: I don't think this quite captures the original request. The original request specified an operand whose type is explicitly nullable (`T?`), where that type is a `TypeParameterType` that is potentially nullable. In particular, I think the test case `m1<T>(T p) => p!;` should not create a lint because in this case the operator isn't being used to remove nullability from the type (which would be more obvious if the method had an explicit return type of `T`).

It might be worthwhile asking on the issue whether that test case ought to produce a lint. It's possible that it was an oversight that the original description doesn't match that case.

lib/src/rules/null_check_on_nullable_type_parameter.dart
```diff
@@ -0,0 +1,91 @@
...
+        expectedType != null &&
+        context.typeSystem.isPotentiallyNullable(expectedType)) {
```
bwilkerson, line 87: This doesn't validate that the expected type is `T`. In particular, I think the following would produce a lint when it's outside the original request:
```
R m<P, R>(P p) => p!;
```

It might be worth asking on the issue to make suer we really don't want to catch the case above.

lib/src/rules/unnecessary_null_checks.dart
```diff
@@ -95,3 +75,33 @@ class _Visitor extends SimpleAstVisitor<void> {
     }
   }
 }
+
+DartType getExpectedType(PostfixExpression node) {
```
bwilkerson, line 79: Consider moving this to one of the utility libraries.

test/rules/experiments/nnbd/rules/null_check_on_nullable_type_parameter.dart
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) 2020, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// test w/ `pub run test -N null_check_on_nullable_type_parameter`
+
+m1<T>(T p) => p!; // LINT
```
bwilkerson, line 7: I think these methods should have an explicit return type. As written, the return type is `dynamic`, which is potentially nullable, so the last part of the condition isn't actually being tested.

lib/src/rules/unnecessary_null_checks.dart
a14n: I put this function here because it is really related to the lints (that is where _null check_ checks happen). If we move it in  utility library we should rename it to clearly indicate that it not only returns the expected type but it also only works where `null_check_on_nullable_type_parameter` and `unnecessary_null_checks` should apply.

bwilkerson: sgtm. It just feels like a code smell to have code being shared between lints that isn't in a utility library that's intended to be shared.

a14n: To be clear, is it ok to let this function here?

bwilkerson: It doesn't impact the correctness of the lint, so I suppose so.

a14n: So I think I addressed all your comments and this PR is ready for another look :)

test/rules/experiments/nnbd/rules/null_check_on_nullable_type_parameter.dart
```diff
@@ -0,0 +1,19 @@
...
+// test w/ `pub run test -N null_check_on_nullable_type_parameter`
+
+T m1<T>(T p) => p!; // OK
+T m2a<T>(T? p) => p!; // LINT
+dynamic m2b<T>(T? p) => p!; // OK
+T m3<T extends Object>(T? p) => p!; // OK
+T m4<T extends Object?>(T? p) => p!; // LINT
+T m5<T extends dynamic>(T? p) => p!; // LINT
+T m6<T>(T p) => p!.a; // OK
+T m7<T>(T p) => p!.m(); // OK
```
bwilkerson, line 14: It might be worthwhile to test cascades (`p!..m()`).

test/rules/experiments/nnbd/rules/null_check_on_nullable_type_parameter.dart
```diff
@@ -0,0 +1,19 @@
...
+
+T m10<T>(T? p) { return p!; } // LINT
+T m20<T>(T? p) { T t = p!; } // LINT
```
bwilkerson, line 17: Should this also catch assignments such as
```
T m30<T>(T? p) {
  T t;
  t = p!;
}
```
or even
```
class C<T> {
  T t;
  m(T? p) {
    t = p!;
  }
}
```

lib/src/rules/null_check_on_nullable_type_parameter.dart
```diff
@@ -0,0 +1,95 @@
...
+    _isNonNullableEnabled = node.featureSet.isEnabled(Feature.non_nullable);
+  }
+
+  @override
+  void visitPostfixExpression(PostfixExpression node) {
+    if (!_isNonNullableEnabled) return;
+    if (node.operator.type != TokenType.BANG) return;
+
+    final expectedType = getExpectedType(node);
+    if (node.operand.staticType is TypeParameterType &&
```
bwilkerson, line 86: Consider assigning `node.operand.staticType` to a local variable (such as `operandType`). I think this condition would be easier to read.

pq:
This looks great.

The pana failure is distracting but I think ignorable.

image

I'm guessing this is a version skew issue since `dartfmt` is clean for me and also in our checks on travis.


/fyi @jonasfj

pq (approved):

devoncarew:
We're seeing a (low volume) NPE from this lint internally, on or around line https://github.com/dart-lang/linter/blob/master/lib/src/rules/null_check_on_nullable_type_parameter.dart#L81:

https://github.com/dart-lang/linter/issues/2306

PR #2254, opened 2020-09-30
guard against null target types

pq:
Fixes benchmark bot breakage: https://travis-ci.org/github/dart-lang/linter/jobs/731431248

/cc @bwilkerson @scheglov

### DIFF
--- a/lib/src/rules/avoid_type_to_string.dart
+++ b/lib/src/rules/avoid_type_to_string.dart
@@ -174,10 +174,9 @@ class _Visitor extends SimpleAstVisitor {
     }
   }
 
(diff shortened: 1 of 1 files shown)


bwilkerson (approved):
I think we should have a regression test to ensure that this doesn't happen again. As far as I can see, the only way for `targetType` to be `null` is if `thisType` is `null`, which can happen if we're not inside a class, mixin or extension, or if we're inside an extension whose `on` type is not an interface type. I expect that the first cases are handled because the invocation of `toString` won't ever be the method defined in `Object`, so the extension case is almost certainly what caused the bug.

PR #2260, opened 2020-09-30
Use analyzer 0.40.4 and CompoundAssignmentExpression.

scheglov:
I have a version of analyzer that stops setting `SimpleIdentifier.staticElement/staticType` when the identifier is the final identifier of the LHS of an assignment (or increment/decrement). This will happen in a new breaking changes version of `analyzer`. This PR prepares the `linter` for using `CompoundAssignmentExpression` instead.


bwilkerson (approved):

pq (approved):
👍

Issue #2259, opened 2020-09-30
short-circuit AST visits in registerNodeProcessors(where possible)


pq:
See for example #2258 which demonstrates how error-prone our current approach is.  Ideally we wouldn't be visiting at all in this case and the visitor not registered in `registerNodeProcessors`.

pq:
**Forward-pointer.**  Consider something like:

```dart

  @override
  void registerNodeProcessors(
      NodeLintRegistry registry, LinterContext context) {
    if (!isInLibDir(context.currentUnit.unit, context.package)) {
      return;
    }
   ...
}

```

(Untested.)

pq:
Fixed w/ #2263, #2264, and #2265.

PR #2263, opened 2020-10-01
re-work lib dir checking

pq:
See: #2259

### DIFF
--- a/lib/src/rules/public_member_api_docs.dart
+++ b/lib/src/rules/public_member_api_docs.dart
@@ -78,6 +78,10 @@ class PublicMemberApiDocs extends LintRule implements NodeLintRule {
   @override
   void registerNodeProcessors(
       NodeLintRegistry registry, LinterContext context) {
+    if (!isInLibDir(context.currentUnit.unit, context.package)) {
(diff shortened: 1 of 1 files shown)


bwilkerson (approved):

lib/src/rules/public_member_api_docs.dart
```diff
@@ -78,6 +78,10 @@ class PublicMemberApiDocs extends LintRule implements NodeLintRule {
   @override
   void registerNodeProcessors(
       NodeLintRegistry registry, LinterContext context) {
+    if (!isInLibDir(context.currentUnit.unit, context.package)) {
```
pq, line 81: @scheglov : is there ever a chance that `context.currentUnit` would be null?  (Or the corresponding  `unit` be null?)  Should I be null-checking anywhere  here?

scheglov: No, `context.currentUnit` cannot be `null`. And its `unit` also cannot be `null`.

pq: thank you!

PR #2265, opened 2020-10-02
short circuit node processor registry outside NNBD

pq:
See: #2259.

/cc @scheglov @bwilkerson @a14n

### DIFF
--- a/lib/src/rules/cast_nullable_to_non_nullable.dart
+++ b/lib/src/rules/cast_nullable_to_non_nullable.dart
@@ -50,7 +50,7 @@ class CastNullableToNonNullable extends LintRule implements NodeLintRule {
   @override
   void registerNodeProcessors(
       NodeLintRegistry registry, LinterContext context) {
-    if (!context.currentUnit.unit.featureSet.isEnabled(Feature.non_nullable)) {
+    if (!context.isEnabled(Feature.non_nullable)) {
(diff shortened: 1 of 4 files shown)


lib/src/rules/cast_nullable_to_non_nullable.dart
```diff
@@ -48,10 +48,13 @@ class CastNullableToNonNullable extends LintRule implements NodeLintRule {
         );
 
   @override
-  void registerNodeProcessors(NodeLintRegistry registry,
-      [LinterContext context]) {
+  void registerNodeProcessors(
+      NodeLintRegistry registry, LinterContext context) {
+    if (!context.currentUnit.unit.featureSet.isEnabled(Feature.non_nullable)) {
```
bwilkerson, line 53: I added an `isEnabled` getter to the `context` to make this a bit more readable. Basically, you can drop `currentUnit.unit.featureSet`.

bwilkerson (approved):

scheglov (approved):

a14n (approved):

lib/src/rules/cast_nullable_to_non_nullable.dart
pq: Oh, that's great.  Thanks!

PR #2272, opened 2020-10-08
prefer_constructors_over_static_methods: take generics into account

### LINKED ISSUE #2149: Incorrect prefer_constructors_over_static_methods warning
Consider:

```dart
class Foo {
  Foo._();

  static Foo bar<T>(T value) => Foo._();

  Foo.bar2<T>(T value);
}
```

The static method results in:
`Prefer defining constructors instead of static methods to create instances. (prefer_constructors_over_static_methods`

However, it is not possible to create a constructor with a type parameter.

Dart 2.7.0

srawlins:
# Description

Compare the method's return _type_ with it's parent class's _type_ rather than
comparing _elements_ to account for generics (don't report a method which
returns `A<int>`).

And don't report any generic method, because it might use the type parameters
in its parameters, which a constructor cannot do.

Fixes #2149

### DIFF
--- a/lib/src/rules/prefer_constructors_over_static_methods.dart
+++ b/lib/src/rules/prefer_constructors_over_static_methods.dart
@@ -78,12 +78,11 @@ class _Visitor extends SimpleAstVisitor<void> {
   void visitMethodDeclaration(MethodDeclaration node) {
     final returnType = node.returnType?.type;
     final parent = node.parent;
-    final typeParameters = node.typeParameters?.typeParameters ?? [];
     if (node.isStatic &&
         parent is ClassOrMixinDeclaration &&
         returnType is InterfaceType &&
-        returnType == parent.declaredElement.thisType &&
-        typeParameters.isEmpty &&
+        parent.typeParameters == null &&
+        node.typeParameters == null &&
         _hasNewInvocation(returnType, node.body)) {
       rule.reportLint(node.name);
     }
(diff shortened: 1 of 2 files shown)


lib/src/rules/prefer_constructors_over_static_methods.dart
```diff
@@ -78,10 +78,12 @@ class _Visitor extends SimpleAstVisitor<void> {
   void visitMethodDeclaration(MethodDeclaration node) {
     final returnType = node.returnType?.type;
     final parent = node.parent;
+    final typeParameters = node.typeParameters?.typeParameters ?? [];
     if (node.isStatic &&
         parent is ClassOrMixinDeclaration &&
         returnType is InterfaceType &&
-        returnType.element == parent.declaredElement &&
+        returnType == parent.declaredElement.thisType &&
```
bwilkerson, line 85: If I'm understanding this correctly, this line will be true only when `parent` has no type parameters (isn't generic), so this effectively disables the lint for generic classes. I have to wonder, though whether it might be more clear to future readers to simply ask whether the class declaration has type parameters (`parent.typeParameters != null`).

lib/src/rules/prefer_constructors_over_static_methods.dart
```diff
@@ -78,10 +78,12 @@ class _Visitor extends SimpleAstVisitor<void> {
...
+        typeParameters.isEmpty &&
```
bwilkerson, line 86: Rather than create an empty list for every non-generic method, consider dropping line 81 and using `node.typeParameters != null` here instead. (I don't think we need to guard against a malformed declaration in which there is an empty list of type parameters.)

bwilkerson (approved):

lib/src/rules/prefer_constructors_over_static_methods.dart
srawlins: Haha definitely simpler. Thanks

lib/src/rules/prefer_constructors_over_static_methods.dart
srawlins: Done

PR #2274, opened 2020-10-08
void_checks: Specify individual types which a FutureOr<void> target can accept

### LINKED ISSUE #2185: void_checks false positive with FutureOr<void> and NNBD
```dart
// @dart = 2.9
import 'dart:async';

void m(FutureOr<void> Function() f) {}

void main() {
  m(() {}); // void_checks false positive
}
```

srawlins:
# Description

The type of `() {}` changed to `Null Function()` (from `void Function()` in Null Safety. The check performed in void_checks must then compare assignability with `Future<dynamic>?`.

Fixes #2185

### DIFF
--- a/lib/src/rules/void_checks.dart
+++ b/lib/src/rules/void_checks.dart
@@ -42,7 +42,6 @@ class VoidChecks extends LintRule implements NodeLintRule {
   void registerNodeProcessors(
       NodeLintRegistry registry, LinterContext context) {
     final visitor = _Visitor(this, context);
-    registry.addCompilationUnit(this, visitor);
     registry.addMethodInvocation(this, visitor);
     registry.addInstanceCreationExpression(this, visitor);
     registry.addAssignmentExpression(this, visitor);
@@ -55,8 +54,6 @@ class _Visitor extends SimpleAstVisitor<void> {
 
   final LinterContext context;
   final TypeSystem typeSystem;
-
-  InterfaceType _futureDynamicType;
 
   _Visitor(this.rule, this.context) : typeSystem = context.typeSystem;
 
@@ -88,13 +85,6 @@ class _Visitor extends SimpleAstVisitor<void> {
     final type = node.writeType;
     _check(type, node.rightHandSide?.staticType, node,
         checkedNode: node.rightHandSide);
-  }
-
-  @override
-  void visitCompilationUnit(CompilationUnit node) {
-    _futureDynamicType = context.typeSystem.leastUpperBound(
-        context.typeProvider.futureDynamicType,
-        context.typeProvider.nullType) as InterfaceType;
   }
 
   @override
--- a/lib/src/rules/void_checks.dart
+++ b/lib/src/rules/void_checks.dart
@@ -65,8 +65,7 @@ class _Visitor extends SimpleAstVisitor<void> {
     if (type.isDartCoreNull) return true;
     if (type.isDartAsyncFuture &&
         type is InterfaceType &&
-        (type.typeArguments.first.isVoid ||
-            type.typeArguments.first.isDartCoreNull)) {
+        isTypeAcceptableWhenExpectingVoid(type.typeArguments.first)) {
       return true;
     }
     return false;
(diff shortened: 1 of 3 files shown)


lib/src/rules/void_checks.dart
```diff
@@ -81,7 +81,9 @@ class _Visitor extends SimpleAstVisitor<void> {
 
   @override
   void visitCompilationUnit(CompilationUnit node) {
-    _futureDynamicType = context.typeProvider.futureDynamicType;
+    _futureDynamicType = context.typeSystem.leastUpperBound(
+        context.typeProvider.futureDynamicType,
+        context.typeProvider.nullType) as InterfaceType;
```
scheglov, line 86: Ah, if you need this only for `FutureOr`, instead use `typeProvider.futureOrElement.instantiate` and provide `nullabilitySuffix`. `LUB` is more expensive.

lib/src/rules/void_checks.dart
```diff
@@ -131,10 +133,12 @@ class _Visitor extends SimpleAstVisitor<void> {
...
-            !isTypeAcceptableWhenExpectingVoid(type) ||
-        expectedType.isDartAsyncFutureOr &&
-            (expectedType as InterfaceType).typeArguments.first.isVoid &&
+    }
+
+    var futureOrVoidType =
+        context.typeProvider.futureOrType2(context.typeProvider.voidType);
+    if (expectedType.isVoid && !isTypeAcceptableWhenExpectingVoid(type) ||
+        expectedType == futureOrVoidType &&
             !typeSystem.isAssignableTo(type, _futureDynamicType)) {
```
scheglov, line 142: Can you explain the portion of expecting `FutureOr<void>` and `typeSystem.isAssignableTo(type, _futureDynamicType)`?

lib/src/rules/void_checks.dart
```diff
@@ -131,10 +133,12 @@ class _Visitor extends SimpleAstVisitor<void> {
...
-    } else if (expectedType.isVoid &&
-            !isTypeAcceptableWhenExpectingVoid(type) ||
-        expectedType.isDartAsyncFutureOr &&
-            (expectedType as InterfaceType).typeArguments.first.isVoid &&
+    }
+
+    var futureOrVoidType =
+        context.typeProvider.futureOrType2(context.typeProvider.voidType);
+    if (expectedType.isVoid && !isTypeAcceptableWhenExpectingVoid(type) ||
+        expectedType == futureOrVoidType &&
```
scheglov, line 141: It seems to me that we are doing it the opposite way here - we should not construct `FutureOr<void>` and then use `==` so do force the analyzer to deconstruct it again. Instead check `isDartAsyncFutureOr` and that the type argument `isVoid`.

scheglov: I also looked at `isTypeAcceptableWhenExpectingVoid` and it seems that it can be improved by recursion instead of repeating the checks for `void` and `Null`.

lib/src/rules/void_checks.dart
srawlins: I'm not sure why futureDynamicType was chosen originally. Seems like a strange choice to me. This fix just makes it so that when Null is used where the expected type is `FutureOr<void>`, then the lint rule will not complain "Hey don't assign a Null to `FutureOr<void>`.

If I changed this check to "is the type assignable to Future<void>", for example, then it no longer catches a case like:

```dart
FutureOr<void> a = 1;
```

I think because technically this is legal, `int` is assignable to `FutureOr<void>`, but the lint wishes to call it out as "Don't assign to void."

lib/src/rules/void_checks.dart
srawlins: Ah I understand. OK fixed both issues.

lib/src/rules/void_checks.dart
srawlins: If the return type of `f` in your example were `FutureOr<void>`, I think we want to allow:

* `Future<dynamic>` (tested on [line 58](https://github.com/dart-lang/linter/blob/master/test/rules/void_checks.dart#L58)),
* `Future<void>` (not tested?),
* `Future<Null>` (not tested?),
* `FutureOr<dynamic>` (not tested?),
* `FutureOr<void>` (line 71),
* `FutureOr<Null>` (newly tested in this PR, [line 10](https://github.com/dart-lang/linter/pull/2274/files#diff-e6fb8c48dbd9c0d925e3d68ea6bf3b1b429a4c3f77a105afbb36e3e50dfc07e5))
* `dynamic` (line 72),
* `Null` (line 73),

and nothing else, I think. Right now the language allows, e.g. an int to be returned where `FutureOr<void>` is expected.

scheglov: If we can have this nice exhaustive list of types, I think it would be better to check for it explicitly, like we did for `void`, instead of using a more heavy `isAssignableTo`.

srawlins: Ah ha, good point. I've used isTypeAcceptableWhenExpectingVoid to write a isTypeAcceptableWhenExpectingFutureOrVoid.

scheglov (approved):

PR #2275, opened 2020-10-08
type_init_formals: Allow types not equal to the field type

### LINKED ISSUE #2192: prefer_initializing_formals false positive in NNBD
In the following code `p` param is non-nullable but `p` field is nullable so we can not use initializing formal.

```dart
// @dart = 2.9
class A {
  A({required String p}) : p = p;
  String? p;
}
```

srawlins:
# Description

If the type of the field formal parameter is different from the field type, it has implications, and should be left alone.

Fixes #2192

### DIFF
--- a/lib/src/rules/type_init_formals.dart
+++ b/lib/src/rules/type_init_formals.dart
@@ -59,12 +59,16 @@ class _Visitor extends SimpleAstVisitor<void> {
 
   @override
   void visitFieldFormalParameter(FieldFormalParameter node) {
-    if (node.type != null) {
+    var nodeType = node.type;
+    if (nodeType != null) {
       var class_ =
           node.thisOrAncestorOfType<ClassDeclaration>()?.declaredElement;
       var field = class_.getField(node.identifier.name);
-      if (node.type?.type == field.type) {
-        rule.reportLint(node.type);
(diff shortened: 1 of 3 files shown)


lib/src/rules/type_init_formals.dart
```diff
@@ -60,7 +60,12 @@ class _Visitor extends SimpleAstVisitor<void> {
   @override
   void visitFieldFormalParameter(FieldFormalParameter node) {
     if (node.type != null) {
-      rule.reportLint(node.type);
+      var class_ =
+          node.thisOrAncestorOfType<ClassDeclaration>()?.declaredElement;
+      var field = class_.getField(node.identifier.name);
+      if (node.type?.type == field.type) {
```
bwilkerson, line 66: Consider assigning `node.type` to a local variable so that we don't need to do the extra null check here.

On the other hand, invalid code might (I haven't tested this) result in `field` being `null` if the field formal parameter references a non-existent field:
```
class C {
  C(int this.x);
}
```

bwilkerson (approved):

lib/src/rules/type_init_formals.dart
srawlins: Good catch! Caught a crash. Added a test and fixed.

PR #2276, opened 2020-10-09
unrelated_type_equality_checks: ignore nullability

### LINKED ISSUE #2196: unrelated_type_equality_checks false positive in NNBD
In the following code `p` field is nullable but `p` param is non-nullable.

```dart
// @dart = 2.9
class A {
  int? p;
  bool m(num p) => p == this.p; // false positive here
}
```

If I make the `p` field non-nullable the lint disappear.

srawlins:
# Description

Nullability should be ignored when considering "relatedness". It is valid to compare `num` and `int?` objects.

Fixes #2196

### DIFF
--- a/lib/src/util/dart_type_utilities.dart
+++ b/lib/src/util/dart_type_utilities.dart
@@ -396,40 +396,41 @@ class DartTypeUtilities {
     return nodes;
   }
 
-  /// Return whether [leftType] and [rightType] are _definitely_ unrelated.
+  /// Returns whether [leftType] and [rightType] are _definitely_ unrelated.
   ///
   /// For the purposes of this function, here are some "relation" rules:
   /// * `dynamic` and `Null` are considered related to any other type.
-  /// * Two equal types are considered related, e.g. classes `int` and `int`,
-  ///   classes `List<String>` and `List<String>`,
-  ///   classes `List<T>` and `List<T>`, and type variables `A` and `A`.
-  /// * Two types such that one is a subtype of the other, such as classes
-  ///   `List<dynamic>` and `Iterable<dynamic>`, and type variables `A` and `B`
-  ///   where `A extends B`.
-  /// * Two types, each representing a class:
+  /// * Two types which are equal modulo nullability are considered related,
+  ///   e.g. `int` and `int`, `String` and `String?`, `List<String>` and
+  ///   `List<String>`, `List<T>` and `List<T>`, and type variables `A` and `A`.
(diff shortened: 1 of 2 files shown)


lib/src/util/dart_type_utilities.dart
```diff
@@ -396,40 +396,41 @@ class DartTypeUtilities {
...
-  /// * Two equal types are considered related, e.g. classes `int` and `int`,
-  ///   classes `List<String>` and `List<String>`,
-  ///   classes `List<T>` and `List<T>`, and type variables `A` and `A`.
-  /// * Two types such that one is a subtype of the other, such as classes
-  ///   `List<dynamic>` and `Iterable<dynamic>`, and type variables `A` and `B`
-  ///   where `A extends B`.
-  /// * Two types, each representing a class:
+  /// * Two types which are equal modulo nullability are considered related,
+  ///   e.g. `int` and `int`, `String` and `String?`, `List<String>` and
+  ///   `List<String>`, `List<T>` and `List<T>`, and type variables `A` and `A`.
```
bwilkerson, line 405: Was one of these `List<String>`s suppose to have a question mark?

lib/src/util/dart_type_utilities.dart
```diff
@@ -396,40 +396,41 @@ class DartTypeUtilities {
...
   ///   * are related if their supertypes are equal, e.g. `List<dynamic>` and
   ///     `Set<dynamic>`.
-  /// * Two types, each representing a type variable, are related if their
-  ///   bounds are related.
+  /// * Two type variables are related if their bounds are related.
   /// * Otherwise, the types are related.
   // TODO(srawlins): typedefs and functions in general.
   static bool unrelatedTypes(
-      TypeSystem typeSystem, DartType leftType, DartType rightType) {
+      TypeSystem typeSystem, DartType leftType_, DartType rightType_) {
```
bwilkerson, line 421: Personally I'd prefer to not use underscore as a suffix, but to use names like `promotedLeftType` (or even just `left`) below, but that's clearly a matter of personal taste, so take it or leave it as you see fit.

bwilkerson (approved):

lib/src/util/dart_type_utilities.dart
srawlins: No. It didn't before.I just wanted to add one example where they are equal except for nullability.

lib/src/util/dart_type_utilities.dart
srawlins: I like it. Done.

PR #2281, opened 2020-10-09
always_require_non_null_named_parameters: skip in Null Safety

### LINKED ISSUE #2236: always_require_non_null_named_parameters is obsolete in null-safety
```dart
class Example {
  final int number;

  // LINT: always_require_non_null_named_parameters
  Example({required this.content});
}
```

I believe the lint should probably just turn itself off in null-safety mode?

srawlins:
# Description

This lint should not run in a Null Safety library; the "required" rules cover this in static analysis done by the Analyzer.

Without this new check, the new test results in an incorrect lint report.

Fixes #2236

### DIFF
--- a/lib/src/rules/always_require_non_null_named_parameters.dart
+++ b/lib/src/rules/always_require_non_null_named_parameters.dart
@@ -5,6 +5,7 @@
 import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/token.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:meta/meta.dart';
 
 import '../analyzer.dart';
 import '../util/dart_type_utilities.dart';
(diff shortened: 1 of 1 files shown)


lib/src/rules/always_require_non_null_named_parameters.dart
```diff
@@ -50,24 +51,26 @@ class AlwaysRequireNonNullNamedParameters extends LintRule
   @override
   void registerNodeProcessors(
       NodeLintRegistry registry, LinterContext context) {
-    final visitor = _Visitor(this);
+    final visitor = _Visitor(this,
+        isNullSafety: context
+            .currentUnit.unit.declaredElement.library.isNonNullableByDefault);
```
bwilkerson, line 56: We've actually been using a slightly different pattern. First, we added `isEnabled` to `LinterContext` to make it a little shorter. Second, we're using a pattern like
```
if (is not non-nullable) {
  registry.add...
}
```
By not registering in cases where the rule shouldn't run, the visitor methods won't even be called.

srawlins: Ah, excellent. Done.

bwilkerson (approved):

pq (approved):
👍

pq:
It occurs to me that it would be interesting to document that this lint (and others) are not compatible w/ NNBD. Down the road maybe we could even consider diagnostics in cases where we know the lint won't be run?

PR #2285, opened 2020-10-12
allow overriding abstract fields in NNBD

### LINKED ISSUE #2269: overridden_fields should not report if the parent field is abstract
With NNBD field can be abstract and it should be ok to override an abstract field.

```dart
class A { abstract String s; }
class B extends A {
  @override // should be ok
  String s; 
}

### PR DESCRIPTION
**Pull request by pq**
Fixes #2269

### DIFF
--- a/lib/src/rules/overridden_fields.dart
+++ b/lib/src/rules/overridden_fields.dart
@@ -2,6 +2,7 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'package:analyzer/dart/analysis/features.dart';
 import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
 import 'package:analyzer/dart/element/element.dart';
@@ -107,15 +108,17 @@ class OverriddenFields extends LintRule implements NodeLintRule {
   @override
   void registerNodeProcessors(
       NodeLintRegistry registry, LinterContext context) {
-    final visitor = _Visitor(this);
+    final visitor = _Visitor(this, context);
     registry.addFieldDeclaration(this, visitor);
   }
 }
 
 class _Visitor extends SimpleAstVisitor<void> {
   final LintRule rule;
+  bool nnbdEnabled;
 
-  _Visitor(this.rule);
+  _Visitor(this.rule, LinterContext context)
+      : nnbdEnabled = context.isEnabled(Feature.non_nullable);
 
   @override
   void visitFieldDeclaration(FieldDeclaration node) {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by a14n — APPROVED** [review]
LGTM

File: lib/src/rules/overridden_fields.dart [hunk 1]
```diff
@@ -125,7 +128,7 @@ class _Visitor extends SimpleAstVisitor<void> {
 
     node.fields.variables.forEach((VariableDeclaration variable) {
       final field = _getOverriddenMember(variable.declaredElement);
-      if (field != null) {
+      if (field != null && (!nnbdEnabled || !field.isAbstract)) {
```
bwilkerson, line 131: I don't believe that the condition needs to be this complex. If null safety is not enabled then the field can't be abstract, so testing the abstractness of the field should be sufficient. Let's drop the feature check.

test/rules/experiments/nnbd/rules/overridden_fields.dart
```diff
@@ -0,0 +1,181 @@
+// Copyright (c) 2016, the Dart project authors. Please see the AUTHORS file
```
bwilkerson, line 1: nit: year

test/rules/experiments/nnbd/rules/overridden_fields.dart
```diff
@@ -0,0 +1,181 @@
...
+}
+
+mixin M on A1 {
+  @override
+  int f; // LINT
+
+  int g; // OK
+}
+
+class BB {
```
bwilkerson, line 173: Abstract fields can only be declared in abstract classes, so you need to add the modifier `abstract` to this class declaration. (We might want to enhance the test framework to report if there are any diagnostics in the file other than the ones that are marked so that issues like this will be caught.)

There are a couple of other cases that might be worth testing:
- a subclass that mixes in this class,
- a subclass that implements the abstract field by adding both a getter and a setter,
- a base class that defines an `abstract final` field with a subclass extends the base.

bwilkerson (approved):

lib/src/rules/overridden_fields.dart
pq: Good catch.

test/rules/experiments/nnbd/rules/overridden_fields.dart
pq: We've been preserving the years for copies of existing tests.  Happy to rethink though?

test/rules/experiments/nnbd/rules/overridden_fields.dart
pq: Thanks!  Adding some more tests.

PR #2297, opened 2020-10-15
Fix RangeError in unnecessary_string_escapes

scheglov:
This happens for
```dart
/// Note that putting `''` on the new line is important to get a token
/// with `'\` with no closing quote.
String unclosedQuote() => '\
    '';
```
  Reported internally [in here](https://crash.corp.google.com/browse?q=product_name%3D%27Dart_analysis_server%27+AND+product.Version%3D%272.11.337030498-edge%2Bgoogle3-v2%27&stbtiq=&reportid=9a128a0e6f2659b4&index=9#0) for example.

  I did not put it into the test because it is a parse error, and there are subsequence crashes during reporting a logical failure (OK vs. LINT) when the file also has parse errors. I'd put this test case into a separate file, but there is not support for multiple files (yet). Should we have it?

### DIFF
--- a/lib/src/rules/unnecessary_string_escapes.dart
+++ b/lib/src/rules/unnecessary_string_escapes.dart
@@ -105,7 +105,7 @@ class _Visitor extends SimpleAstVisitor<void> {
     for (var i = 0; i < lexeme.length; i++) {
       var current = lexeme[i];
       var escaped = false;
-      if (current == r'\') {
+      if (current == r'\' && i < lexeme.length - 1) {
         escaped = true;
         i += 1;
         current = lexeme[i];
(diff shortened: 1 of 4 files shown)


pq:
>  there is not support for multiple files (yet). Should we have it?

This could be handled in an integration test.

https://github.com/dart-lang/linter/blob/b2b99eb654b8617d1bf0106c91736446b168b120/test/integration_test.dart#L22

but you're right, we don't support multiple files for the standard rule tests.

pq (approved):
👍

pq:
Feel free to add a test if you like or just land as is.  Up to you!

pq (approved):

pq:
Interestingly, I see a new failure in the benchmarks:

https://travis-ci.org/github/dart-lang/linter/jobs/736175484

Maybe triggered by your new test case?

bwilkerson (approved):

scheglov:
@pq Yes, strange. Does it run lints with new fixes, or a previous version?

pq:
It runs lints from the source in the PR.  My guess is that your new test code is just triggering an existing bug in `UseRawStrings`.

pq:
Thank you!

PR #2299, opened 2020-10-16
refactor integration tests

pq:
Running individual integration tests is tedious and the integration test library unwieldy.  To help remedy, this defines dedicated test groups for core and rule tests and further extracts rule tests into their own individually runnable libraries.

/cc @bwilkerson


bwilkerson (approved):
Much better.

PR #2303, opened 2020-10-26
Add weak_mode_unnecessary_null_checks

a14n:
# Description

Unnecessary null check for non-nullable value in a opted-out library.
**BAD:**
```
nonNullable?.property;
if (nonNullable != null) {}
```
**GOOD:**
```
nonNullable.property;
```


lib/src/rules/weak_mode_unnecessary_null_checks.dart
```diff
@@ -0,0 +1,104 @@
...
+    if (node.featureSet.isEnabled(Feature.non_nullable)) {
+      return;
+    }
+    super.visitCompilationUnit(node);
+  }
+
+  @override
+  void visitPropertyAccess(PropertyAccess node) {
+    if (node.operator.type == TokenType.QUESTION_PERIOD &&
+        context.typeSystem.isNonNullable(node.target.staticType)) {
```
a14n, line 64: @bwilkerson I don't know why but the type of an element in opted-in library is always with the * suffix and so `context.typeSystem.isNonNullable(node.target.staticType)` is never true. I thought it will be `String` instead of `String*`. Am I missing something?

lib/src/rules/weak_mode_unnecessary_null_checks.dart
```diff
@@ -0,0 +1,104 @@
...
+
+class _Visitor extends RecursiveAstVisitor<void> {
+  _Visitor(this.rule, this.context);
+
+  final LintRule rule;
+  final LinterContext context;
+
+  @override
+  void visitCompilationUnit(CompilationUnit node) {
+    if (node.featureSet.isEnabled(Feature.non_nullable)) {
```
bwilkerson, line 55: The current convention is to use `LinterContext.isEnabled` to test this, and to put the test in `registerNodeProcessors` if the whole rule is conditional. By not registering the visitor at all in those cases we improve overall performance.

lib/src/rules/weak_mode_unnecessary_null_checks.dart
bwilkerson: As per the semantics from the spec, in an opted out library all types (whether they come from opted-in or opted-out libraries) are converted to have a * suffix.

bwilkerson (commented):
I'm not certain that we can get the information this lint would require, but even if we could I'd have some concerns about false positives. Consider the following. In an opted-in library we have
```
int identity(int i) => i;
```
and in an opted-out library we have
```
import 'optedOut.dart';

void main() {
  print(identity(null) != null);
}
```
According to the mixed-mode semantics, I believe (though I haven't tried running it), that this would print "false" without crashing. In other words, the non-null return type on `identity` doesn't guarantee that it can't return `null` when called from an opted-out library.

a14n:
This lint could/should be used by maintainer that don't want to migrate now but like to have feedback about bad semantics in their opted-out libraries until they do the migration. I'm not really concerned about false positives as their number should be really small.

This lint was suggested by @Alexv525 to catch `if (pixels != null)` in opted-out libraries (a breaking change made `pixels` non-nullable and `if (pixels != null)` should be replaced by `if (hasPixels)`)

>I'm not certain that we can get the information this lint would require

:( IMHO it would have helped during the weak mode transition. 

Finally I really don't know why travis isn't failling because I see failures locally. @pq any idea?

lib/src/rules/weak_mode_unnecessary_null_checks.dart
a14n: Thanks for the explanation. I thought * was only used for types coming from opted-out into opted-in.

a14n:
PTAL: I implemented a workaround to get the nullability of the target.

lib/src/rules/weak_mode_unnecessary_null_checks.dart
```diff
@@ -0,0 +1,132 @@
...
+  void visitMethodInvocation(MethodInvocation node) {
+    if (node.operator?.type == TokenType.QUESTION_PERIOD &&
+        isNonNullable(node.target)) {
+      rule.reportLintForToken(node.operator);
+    }
+  }
+
+  @override
+  void visitPostfixExpression(PostfixExpression node) {
+    if (node.operator.type == TokenType.QUESTION &&
```
bwilkerson, line 79: I don't believe that the operator of a `PostfixExpression` can ever be a question mark.

lib/src/rules/weak_mode_unnecessary_null_checks.dart
```diff
@@ -0,0 +1,132 @@
...
+        isNonNullable(node.operand)) {
+      rule.reportLintForToken(node.operator);
+    }
+  }
+
+  @override
+  void visitBinaryExpression(BinaryExpression node) {
```
bwilkerson, line 86: Should this rule catch uses of `??`? What about `??=`?

lib/src/rules/weak_mode_unnecessary_null_checks.dart
```diff
@@ -0,0 +1,132 @@
...
+    if (element == null ||
+        !element.library.featureSet.isEnabled(Feature.non_nullable)) {
+      return false;
+    }
+    final node = getNode(element);
+    TypeAnnotation type;
+    if (node is FunctionDeclaration) type = node.returnType;
+    if (node is MethodDeclaration) type = node.returnType;
+    if (element is PropertyAccessorElement && element.isSynthetic) {
+      type = (getNode(element.variable).parent as VariableDeclarationList).type;
```
bwilkerson, line 122: The method `getNode` is fairly expensive, so consider instead doing something like
```
if (element is PropertyAccessorElement && element.isSynthetic) {
  element = element.variable;
}
```
before line 117 and replacing this branch with something like
```
if (node is FieldDeclaration) ...
```

Also consider using an `else if` structure here to make it clear that only one of the conditions is expected to be true.

bwilkerson (commented):
I'm still not convinced that we should support this rule, and I think we should resolve that question before you put more effort into it.

@leafpetersen I'd be interested in your take on this question: Should we be making it easy for users to remove null checks in code in an opted-out library that's referencing code in an opted-in library? I thought that in such mixed mode cases we can't guarantee that declarations with a non-null type won't evaluate to `null`.

AlexV525:
> I'm not certain that we can get the information this lint would require, but even if we could I'd have some concerns about false positives. Consider the following. In an opted-in library we have
> 
> ```
> int identity(int i) => i;
> ```
> 
> and in an opted-out library we have
> 
> ```
> import 'optedOut.dart';
> 
> void main() {
>   print(identity(null) != null);
> }
> ```

That's an interesting case, but shouldn't we prevent users from sending null into opted-in library? which might address to another linter rules.

For now because the opt-out library could get null when a subclass override the getter to return null after first called or whatever, but more common situation is users make a nullable check on an opted-in non-nullable getter. I think we should somehow sort out the whole sequence about what exactly the nullability passed through opted-in and out libraries. Since Flutter already fully migrated (almost maybe), that might affects a lot of developers.

bwilkerson:
> I think we should somehow sort out the whole sequence about what exactly the nullability passed through opted-in and out libraries.

That has already been sorted out, and is included in the specification of Null Safety. See https://github.com/dart-lang/language/blob/master/accepted/future-releases/nnbd/feature-specification.md#runtime-semantics for details.

leafpetersen:
> @leafpetersen I'd be interested in your take on this question: Should we be making it easy for users to remove null checks in code in an opted-out library that's referencing code in an opted-in library? I thought that in such mixed mode cases we can't guarantee that declarations with a non-null type won't evaluate to `null`.

I don't have strong objections to this lint existing.  It is absolutely the case that in mixed mode opted out code can abuse APIs in ways that will cause null to be returned from a non-nullable return-typed method.  On the other hand, if someone writes a non-nullable return type on a method, that is arguably at least as strong a signal as documenting it with `/// This function never returns null`.  There are functions so documented that can return null if you abuse them, but I think I would be supportive of a lint for functions documented (annotated) to not return null, even if it wasn't guaranteed.  Is there an accompanying lint to warn when `null` is passed in to opted in code where it shouldn't be allowed?

In general, I don't see this lint as all that valuable for existing code - you should migrate it!  But I do see the value for people who are writing new opted out code against opted in APIs.  In my ideal world, all new code would be written null safe, but providing value to users who can't migrate yet seems good to me.

cc @lrhn @natebosch @jakemac53 @munificent for thoughts?

AlexV525:
> In general, I don't see this lint as all that valuable for existing code - you should migrate it! But I do see the value for people who are writing new opted out code against opted in APIs. In my ideal world, all new code would be written null safe, but providing value to users who can't migrate yet seems good to me.

We cannot guarantee that 100% of users will migrate to nnbd, even if we set a constraint to a newer SDK, they might still writing mixed mode code. When it comes to packages, they'll be in mixed mode if any packages haven't migrated yet, and unable to write opted-in code.

Thus we should prevent opted-in code be overridden to nullable from opted-out code, and tell users not to check a nullability with opted-in non-null value. Does it make sense to you?

a14n:
> Is there an accompanying lint to warn when null is passed in to opted in code where it shouldn't be allowed?

I planned to add such a lint if the current one (weak_mode_unnecessary_null_checks) makes sense and lands.

bwilkerson:
Ok, if Leaf is comfortable with such a lint, then I am too. Let me know when you've addressed my specific comments and I should take another look.

natebosch:
> It is absolutely the case that in mixed mode opted out code can abuse APIs in ways that will cause null to be returned from a non-nullable return-typed method. On the other hand, if someone writes a non-nullable return type on a method, that is arguably at least as strong a signal as documenting it with `/// This function never returns null`.

I'm pretty sure we have at least a method or two which were intentionally implemented to allow a `null` to flow through a non-nullable return type in order to use the null safety migration as a place to improve the API without breaking opt-out code. @jakemac53 do you remember any specific places we did this?

jakemac53:
> I'm pretty sure we have at least a method or two which were intentionally implemented to allow a `null` to flow through a non-nullable return type in order to use the null safety migration as a place to improve the API without breaking opt-out code. @jakemac53 do you remember any specific places we did this?

I don't remember specifically anything, although I would not be surprised if there are places that would return `null` if you give them `null` etc. For the most part the behavior here is somewhat undefined.

In general when in an opted out library I would probably not start removing null checks that existed previously, even if the opted in api says it doesn't return null.

natebosch:
I also would generally not recommend using a lint like this.

srawlins:
I'll close this out as stale, since most Dart in the world is now null safe, and not run in mixed mode.

PR #2312, opened 2020-10-30
add support for github actions

devoncarew:
- add support for github actions

This add support for github actions. This is more or less a straight port of the existing travis.yaml file (https://github.com/dart-lang/linter/blob/master/.travis.yml). We may need to land this PR to see the actions run as there aren't any existing actions for the repo.

It would be great to evolve dart support for actions a bit so that a per-repo install-dart.sh isn't required. Perhaps there's some creative use of the std. dart docker container which would make the install script not needed.

### DIFF
--- a/.github/install-dart.sh
+++ b/.github/install-dart.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+
+# Fast fail the script on failures.
+set -ex
+
+echo Installing Dart ${DART_VERSION}
+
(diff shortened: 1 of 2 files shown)


devoncarew:
Looks like something like this might work:

```
jobs:
  build:
    container:
      image:  google/dart:latest
```

I'll explore some simplification after this PR lands.

pq (approved):
💎

pq:
Awesome!

PR #2316, opened 2020-10-31
fix failing lines_longer_than_80_chars integration test

pq:
~With analyzer 0.40.5, comma-separated ignores appear to no longer allow for spaces. I'm not sure if this is a big (@bwilkerson?) but this fix updates our newly failing test:~

https://travis-ci.org/github/dart-lang/linter/jobs/740287956#L438-L462

**EDIT:** real issue was new diagnostics.

/cc @a14n @bwilkerson

### DIFF
--- a/test/_data/lines_longer_than_80_chars/a.dart
+++ b/test/_data/lines_longer_than_80_chars/a.dart
@@ -29,5 +29,5 @@
  * line with a CRLF at its end                                                 a
  */
 
-// ignore: lines_longer_than_80_chars,lines_longer_than_80_chars,lines_longer_than_80_chars
+// ignore: lines_longer_than_80_chars, lines_longer_than_80_chars, lines_longer_than_80_chars
 var a = 'abcde abcde abcde abcde abcde abcde abcde abcde abcde abcde abcde abcde ';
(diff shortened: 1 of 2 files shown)


bwilkerson:
It would definitely be a bug if spaces are no longer allowed, but I'm not convinced that that's what's going on. Looking at the output you referenced I don't see any diagnostic suggesting that this is the case. The unexpected diagnostic I see is:
```
/home/travis/build/dart-lang/linter/test/_data/lines_longer_than_80_chars/a.dart 28:40 [hint] The diagnostic 'lines_longer_than_80_chars' does not need to be ignored here because it is already being ignored.
```
This is a new diagnostic that was introduced after 0.40.4 that is triggered by having `lines_longer_than_80_chars` appear multiple times in the same `ignore` comment. It has nothing to do with spaces after commas.

I would suggest that you fix that new diagnostic first and see whether the test continues to fail. My guess is that it won't. But I believe that the reason this PR "fixes" the test is that you copied the diagnostic above into the expectations, not because you deleted the spaces after the commas.

pq:
Good catch!  Red herrings all around. 

Fixed!

bwilkerson (approved):

Issue #2315, opened 2020-10-31
gh actions build not running tests?  


pq:
Looking at a recent run, I'm  not seeing the tests run:

![image](https://user-images.githubusercontent.com/67586/97781057-fad6b300-1b45-11eb-8d71-f1a182e577a5.png)

https://github.com/dart-lang/linter/runs/1333844495?check_suite_focus=true

(**EDIT**: or actually maybe they are but we're not waiting for them to complete?)

Interesting aside, on travis tests are running but not failing as expected.  See for example:

https://travis-ci.org/github/dart-lang/linter/jobs/740287956

(I suspect I messed  this up when I  got coverage working again.)

Also, to investigate: is coverage working properly with gh-actions?

/fyi @devoncarew

pq:
I think maybe this is it:

https://github.com/dart-lang/linter/blob/c02ae2c457951ea680fda3bae932acc679ebfef0/tool/travis.sh#L68

I'm guessing we need to install a coveralls token or maybe use the dedicated GH action?

https://github.com/marketplace/actions/coveralls-github-action

devoncarew:
Ah, yeah, very likely there's a bit more setup to do for actions.

Additionally, perhaps we should have the script run the tests even if there's no coverage token.

PR #2320, opened 2020-11-02
run tests w/o coverage token

pq:
/cc  @devoncarew @bwilkerson 

Fixes: #2315

### DIFF
--- a/tool/travis.sh
+++ b/tool/travis.sh
@@ -55,17 +55,17 @@ else
 
   echo ""
 
-  OBS_PORT=9292
-
-  # Run the tests.
-  dart --disable-service-auth-codes \
-    --disable-analytics \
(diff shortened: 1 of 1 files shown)


tool/travis.sh
```diff
@@ -94,5 +94,10 @@ else
     coveralls-lcov var/lcov.info
 
     exit $status
+  else
+    # Run the tests w/o coverage
+    dart --disable-service-auth-codes \
```
devoncarew, line 99: We may not need `--disable-service-auth-codes` (but I don't think having it hurts anything).

devoncarew (approved):

tool/travis.sh
pq: Ah, right.  Good catch.  I'll remove.  Thanks!

PR #2328, opened 2020-11-04
ignore prefer_constructors_over_static_methods

pq:
Address lint.

Tempting to move this to `EnumLikeClassDescription` but  that'd be technically API breaking 😬.

/cc @bwilkerson

### DIFF
--- a/lib/src/util/dart_type_utilities.dart
+++ b/lib/src/util/dart_type_utilities.dart
@@ -14,6 +14,7 @@ import '../ast.dart';
 typedef AstNodePredicate = bool Function(AstNode node);
 
 class DartTypeUtilities {
+  // ignore: prefer_constructors_over_static_methods
   static EnumLikeClassDescription asEnumLikeClass(ClassElement classElement) {
     // See discussion: https://github.com/dart-lang/linter/issues/2083
     //
--- a/lib/src/util/dart_type_utilities.dart
+++ b/lib/src/util/dart_type_utilities.dart
@@ -14,6 +14,7 @@ import '../ast.dart';
 typedef AstNodePredicate = bool Function(AstNode node);
 
 class DartTypeUtilities {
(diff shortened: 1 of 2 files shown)


lib/src/util/dart_type_utilities.dart
```diff
@@ -14,6 +14,7 @@ import '../ast.dart';
 typedef AstNodePredicate = bool Function(AstNode node);
 
 class DartTypeUtilities {
+  // ignore: prefer_constructors_over_static_methods
```
bwilkerson, line 17: I don't understand why this lint is being reported here. Shouldn't it only report static methods whose return type is the same as the enclosing class (modulo type parameters)? There isn't any way to write a constructor in `DartTypeUtilities` that returns an `EnumLikeClassDescription`. I suspect that the lint has a bug; it checks that the return type is an interface type but not that it's the same as the enclosing class.

Actually, I think there's a second bug. The lint needs to not flag static methods from which `null` can be returned if null safety is enabled. Under null safety it is no longer possible to return `null` from a factory constructor.

You should merge this either way to get rid of the diagnostic, but we should think about whether the lint is broken and needs to be fixed.

bwilkerson (approved):
We can definitely make this a static method on `EnumLikeClassDescription`, we just need to stage it by marking this method as deprecated until we're ready to publish a breaking change version.

lib/src/util/dart_type_utilities.dart
pq: Thanks for the reflections.  I was making a note to a similar effect.  It's interesting that this is just firing now but I haven't looked closer.  The NNBD issue I hadn't thought of, so that's a good catch! 👍

pq: https://github.com/dart-lang/linter/issues/2330

Issue #2280, opened 2020-10-09
false positive: prefer_constructors_over_static_methods should not report on method which can return null


srawlins:
**Describe the issue**

Under Null safety, factory constructors can no longer return null. So prefer_constructors_over_static_methods should not suggest that a static method which can return null should be converted to a constructor.

**Examples**

```dart

class A {
  static A x() => null;

  static A y() {
    if (1==1) return A();
    else return null;
  }

  static A? z() {
    return aOrNull;
  }
}

A? get aOrNull => 1 == 2 ? A() : null;

```

**To Reproduce**

The examples above

**Expected behavior**

No lint reported on any of the methods.

**Additional context**

I think the implementation would be split in two:

* In a pre-null safety library, this lint should check for literal null returns
* In a null safety library, we could just skip methods which return a nullable type.

PR #2335, opened 2020-11-06
nullable return type test case

pq:
The NNBD case appears to have been addressed by the `typeSystem.isAssignableTo` check added in #2333.

@bwilkerson: correct me if I've missed any  cases?

Closes: #2280.

### DIFF
--- a/test/rules/experiments/nnbd/rules/prefer_constructors_over_static_methods.dart
+++ b/test/rules/experiments/nnbd/rules/prefer_constructors_over_static_methods.dart
@@ -1,4 +1,4 @@
-// Copyright (c) 2017, the Dart project authors. Please see the AUTHORS file
+// Copyright (c) 2020, the Dart project authors. Please see the AUTHORS file
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
--- a/test/rules/experiments/nnbd/rules/prefer_constructors_over_static_methods.dart
+++ b/test/rules/experiments/nnbd/rules/prefer_constructors_over_static_methods.dart
@@ -1,0 +1,58 @@
+// Copyright (c) 2017, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// test w/ `pub run test -N prefer_constructors_over_static_methods`
+
+class A {
+  static final array = <A>[];
+
+  A.internal();
+
+  static A bad1() => // LINT
+      new A.internal();
+
+  static A get newA => // LINT
+      new A.internal();
+
+  static A bad2() { // LINT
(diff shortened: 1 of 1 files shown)


test/rules/experiments/nnbd/rules/prefer_constructors_over_static_methods.dart
```diff
@@ -0,0 +1,58 @@
...
+
+  static A get newA => // LINT
+      new A.internal();
+
+  static A bad2() { // LINT
+    final a = new A.internal();
+    return a;
+  }
+
+  static A good1(int i) { // OK
```
bwilkerson, line 23: I'm not sure why this one is OK; it seems like it could easily be a factory constructor:

```
factory A(int i) => array[i];
```

On the other hand, given a static field:

```
static final Map<String, A> = {};
```

the following method can't be converted because it can return `null`:

```
static A bad3(String key) {
  return map[key];
}
```

test/rules/experiments/nnbd/rules/prefer_constructors_over_static_methods.dart
```diff
@@ -0,0 +1,58 @@
...
+    return array[i];
+  }
+
+  factory A.good2() { // OK
```
bwilkerson, line 27: I would have assumed that this lint never flags constructors, so I'm not sure what value this (or the next) test adds.

bwilkerson (approved):

test/rules/experiments/nnbd/rules/prefer_constructors_over_static_methods.dart
```diff
@@ -0,0 +1,58 @@
...
+  factory A.good3() { // OK
+    return new A.internal();
+  }
+
+  static A generic<T>() => // OK
+      A.internal();
+
+  static Object ok() => Object(); // OK
+
+  static A? ok() => 1==1 ? null : A(); // OK
```
pq, line 40: FWIW: this is the delta w/ the non-NNBD version...

test/rules/experiments/nnbd/rules/prefer_constructors_over_static_methods.dart
pq: Agreed.  These tests were written loooonnnnggg ago and contributed by Juan.  Not sure what the rationale was?

test/rules/experiments/nnbd/rules/prefer_constructors_over_static_methods.dart
pq: 👍

test/rules/experiments/nnbd/rules/prefer_constructors_over_static_methods.dart
bwilkerson: Me either. But that shouldn't stop us from cleaning it up.

Issue #1649, opened 2019-07-15
Make prefer_collection_literals less suggestive for LinkedHashMap and LinkedHashSet


jamesderlin:
I have code where insertion order is important, and even though `Map`'s default implementation is a `LinkedHashMap`, I'd like to emphasize this in code by explicitly using `LinkedHashMap`:

```dart
final lruCache = LinkedHashMap<String, Stuff>();
```

This triggers the `prefer_collection_literals` lint.

I thought maybe I instead could do:

```dart
final LinkedHashMap<String, Stuff> lruCache = {};
```

but that runs afoul of `--no-implicit-dynamic`.  I thought that I could give in and be redundant:

```dart
final LinkedHashMap<String, Stuff> lruCache = <String, Stuff>{};
```

but that doesn't work either:

>   error • A value of type 'Map<String, Stuff>' can't be assigned to a variable of type 'LinkedHashMap<String, Stuff>' at ... • invalid_assignment

So I think that there's no way for me to actually initialize a type declared as a `LinkedHashMap` while appeasing the linter.

I think that `prefer_collection_literals` should ignore `LinkedHashMap` (and `LinkedHashSet`).  In https://github.com/dart-lang/linter/issues/1425#issuecomment-462200784, @bwilkerson asked:

> > I'd like to drop the flagging of:
> > 
> > 1. `LinkedHashSet` constructors altogether, and
> 
> I'm not sure why. As I understand it, a set literal creates an instance of `LinkedHashSet`, so the two should be consistent.

I claim that if someone is explicitly using `LinkedHashMap`/`LinkedHashSet` instead of just `Map` or `Set`, then they care about being explicit. (And if we don't think that's important, then I'd argue that `LinkedHashMap` and `LinkedHashSet` shouldn't even exist as separate classes.)

jamesderlin:
> I thought maybe I instead could do:
> 
> ```dart
> final LinkedHashMap<String, Stuff> lruCache = {};
> ```
> 
> but that runs afoul of `--no-implicit-dynamic`. I thought that I could give in and be redundant:

And that doesn't work because `{}` (a `Map`) isn't assignable to `LinkedHashMap` (the `invalid_assignment` error from the third attempt) and therefore the type isn't inferred.

So the fact that `Map` isn't assignable to `LinkedHashMap` suggests to me that the linter should *not* be suggesting that map literals replace a default construction of `LinkedHashMap`.

bwilkerson:
[...]

PR #2340, opened 2020-11-07
Fix false positives in prefer_collection_literals

natebosch:
Towards #1649

Allow using `LinkedHashSet` and `LinkedHashMap` for named arguments,
assignment to a variable, and when used in a binary expression where a
static type is pushed down to the arguments.

Refactor `_shouldSkipLinkedHashLint`. Extract a method to determine the
type that is enforced statically for the expression rather than
duplicate the subsequent check for each type of parent node.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 0.1.125-dev
+
+* fixed false positives in `prefer_colleciton_literals`.
+
 # 0.1.124
 
 * fixed false positives in `prefer_constructors_over_static_methods`
--- a/lib/src/rules/prefer_collection_literals.dart
+++ b/lib/src/rules/prefer_collection_literals.dart
@@ -142,35 +142,34 @@ class _Visitor extends SimpleAstVisitor<void> {
 
   bool _shouldSkipLinkedHashLint(
       InstanceCreationExpression node, bool Function(DartType node) typeCheck) {
-    if (_isHashMap(node) || _isHashSet(node)) {
-      // Skip: LinkedHashSet<int> s =  ...; or LinkedHashMap<int> s =  ...;
-      var parent = node.parent;
-      if (parent is VariableDeclaration) {
-        var parent2 = parent.parent;
-        if (parent2 is VariableDeclarationList) {
-          var assignmentType = parent2.type?.type;
-          if (assignmentType != null && !typeCheck(assignmentType)) {
-            return true;
-          }
-        }
-      }
-      // Skip: function(LinkedHashSet()); when function(LinkedHashSet mySet) or
-      // function(LinkedHashMap()); when function(LinkedHashMap myMap)
-      if (parent is ArgumentList) {
-        final paramType = parent.arguments.first.staticParameterElement?.type;
-        if (paramType != null && !typeCheck(paramType)) {
-          return true;
-        }
-      }
-      // Skip: <int, LinkedHashSet>{}.putIfAbsent(3, () => LinkedHashSet());
-      // or <int, LinkedHashMap>{}.putIfAbsent(3, () => LinkedHashMap());
-      if (parent is ExpressionFunctionBody) {
-        var expressionType = parent.expression.staticType;
-        if (expressionType != null && !typeCheck(expressionType)) {
-          return true;
-        }
-      }
+    if (!_isHashMap(node) && !_isHashSet(node)) return false;
+    var parent = node.parent;
+    var staticType = _enforcedType(parent);
+    return staticType != null && !typeCheck(staticType);
+  }
+}
+
+/// Returns the static type which is pushed into an expression by it's parent.
+DartType _enforcedType(AstNode parent) {
+  if (parent is VariableDeclaration) {
+    var parent2 = parent.parent;
+    if (parent2 is VariableDeclarationList) {
+      return parent2.type?.type;
     }
-    return false;
+  }
+  if (parent is ArgumentList) {
+    return parent.arguments.first.staticParameterElement?.type;
+  }
+  if (parent is NamedExpression) {
+    return parent.staticType;
+  }
+  if (parent is ExpressionFunctionBody) {
+    return parent.expression.staticType;
+  }
+  if (parent is AssignmentExpression) {
+    return parent.leftHandSide.staticType;
+  }
+  if (parent is BinaryExpression) {
+    return parent.staticType;
   }
 }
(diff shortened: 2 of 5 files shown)


lib/src/rules/prefer_collection_literals.dart
```diff
@@ -142,35 +142,35 @@ class _Visitor extends SimpleAstVisitor<void> {
...
+DartType /*?*/ _enforcedType(AstNode parent) {
+  if (parent is VariableDeclaration) {
+    var parent2 = parent.parent;
+    if (parent2 is VariableDeclarationList) {
+      return parent2.type?.type;
     }
-    return false;
   }
+  if (parent is ArgumentList) {
+    return parent.arguments.first.staticParameterElement?.type;
```
bwilkerson, line 161: The use of `first` here looks suspicious. I'm guessing that you need to pass the `InstanceCreationExpression` into this method rather than its parent, and ask that node for its `staticParameterElement`.

You can test this by defining and using a method like `printIndentedSet(int indent, Set<int> ids)` in the test code. I believe that `first.staticParameterElement` would return an element whose type is `int`.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# 0.1.125-dev
+
+* fixed false positives in `prefer_colleciton_literals`.
```
kevmoo, line 3:
```suggestion
* Fixed false positives in `prefer_colleciton_literals`.
```

lib/src/rules/prefer_collection_literals.dart
```diff
@@ -142,35 +142,35 @@ class _Visitor extends SimpleAstVisitor<void> {
...
+  }
+  if (parent is NamedExpression) {
+    return parent.staticType;
```
bwilkerson, line 164: This doesn't look right either. I believe that the static type of a `NamedExpression` is the static type of the expression being named (which is the instance creation expression). I suspect that you need `parent.staticParameterElement?.type` in order to get the type of the parameter associated with the argument.

lib/src/rules/prefer_collection_literals.dart
```diff
@@ -142,35 +142,35 @@ class _Visitor extends SimpleAstVisitor<void> {
...
+  }
+  if (parent is ExpressionFunctionBody) {
+    return parent.expression.staticType;
```
bwilkerson, line 167: This also looks wrong. `parent.expression` will always return the instance creation expression from which we started. I think what you want to do here is recurse to see whether you can find a type that would be inferred as the return type of the closure (assuming that this is the body of a closure).

lib/src/rules/prefer_collection_literals.dart
```diff
@@ -142,35 +142,35 @@ class _Visitor extends SimpleAstVisitor<void> {
...
+  }
+  if (parent is AssignmentExpression) {
+    return parent.leftHandSide.staticType;
+  }
+  if (parent is BinaryExpression) {
+    return parent.staticType;
```
bwilkerson, line 173: This also looks wrong. If the instance creation is the right operand then you want the type of the parameter. If it's the left operand then there is no context type.

lib/src/rules/prefer_collection_literals.dart
```diff
@@ -142,35 +142,35 @@ class _Visitor extends SimpleAstVisitor<void> {
...
-      }
+    if (!_isHashMap(node) && !_isHashSet(node)) return false;
+    var parent = node.parent;
+    var staticType = _enforcedType(parent);
+    return staticType != null && !typeCheck(staticType);
+  }
+}
+
+/// Returns the static type which is pushed into an expression by it's parent.
+DartType /*?*/ _enforcedType(AstNode parent) {
```
bwilkerson, line 153: I might be completely misunderstanding what this method is trying to accomplish, but it looks like it's trying to find the type that's required by the context in which the `InstanceCreationExpression` appears. If so, then I think what we really need here is the algorithm implemented by the `_ContextTypeVisitor` in the `analysis_server` package, which is both more complete and more correct.

If I'm not understanding the purpose of this method then my comments below are probably also wrong.

CHANGELOG.md
natebosch: Trying to match style with the rest of the file.

lib/src/rules/prefer_collection_literals.dart
natebosch: yes, that's exactly the goal. Is there a reasonable way we could expose that functionality here?

Specifically, the problem this was attempting to solve is that in some places if we _don't_ specifically use `LinkedHashMap` over `{}` then we'll get some other analysis error. We want to suppress the lint in only those places where honoring the lint would cause a static error.

lib/src/rules/prefer_collection_literals.dart
natebosch: This was the existing behavior, you are correct that it has a bug. Let's fix that separately rather than have it wait to figure out the rest of this PR. https://github.com/dart-lang/linter/pull/2347 should fix it.

@pq - is there a typical process I should be following when fixing these false negatives? I have another category of false negative with this lint that I'd like to fix, but I don't know how much cleanup it will require.

lib/src/rules/prefer_collection_literals.dart
bwilkerson: I suspect that we should just add a TODO and tackle that later. It will require moving the `_ContextTypeVisitor` to the `analyzer` package and exposing an API through which it can be used. That's a bigger task than I'm asking you to tackle here.

lib/src/rules/prefer_collection_literals.dart
pq: Fixing false negatives is a commitment since it will likely lead to breakage on the next  Google3 roll.  Ideally issues get  fixed proactively before a roll is tried.  Happy to share my approach.

lib/src/rules/prefer_collection_literals.dart
natebosch: This is another existing bug with false negatives.

Is this something you might be able to work on? I can easily add some test cases showing the problem but I'm having a hard time figuring out the right things to look for in the analyzer APIs.

bwilkerson: If you can send me the test case(s), I'll be happy to at least look at what it would take to implement them.

srawlins:
@natebosch are you still interested in landing this? It looks like there are some unresolved comments. No worries if we should just close it for now and pick it up later.

lib/src/rules/prefer_collection_literals.dart
natebosch: https://github.com/dart-lang/linter/pull/3705/files adds a line in the test data library that should have a lint but doesn't due to this bug.

natebosch:
I can't find anything worth landing here other than bringing more attention to the false negative we discovered. https://github.com/dart-lang/linter/pull/3705/files

It's been a while since I looked at this, but IIRC I found some more cases that this code doesn't cover and this version likely isn't enough of an improvement to land on it's own. It's probably more worthwhile to focus on exposing the context type calculation that analyzer does to the linter as discussed in https://github.com/dart-lang/linter/pull/2340#discussion_r525627256

PR #2347, opened 2020-11-18
Fix false negative for collections as parameters

natebosch:
Check the parameter element type for the specific collection
instantiation argument rather than check the parameter type of the first
argument.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# 0.1.126-dev
+
+* fixed false negatives for `prefer_collection_literals` when a Set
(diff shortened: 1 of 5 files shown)


bwilkerson (approved):

natebosch:
Surprisingly there don't appear to be any existing violations that were masked by the bug. This one should be feasible to land. @pq can  you help me with that?

pq:
Wow.  Amazing. Thanks for doing the legwork!

I'll ping you about timing a release.

pq (approved):
💎

PR #2350, opened 2020-11-24
Fix false negatives for collection assignments

natebosch:
Previously the lint would allow passing a `LinkedHashSet` instantiation
if the static type wasn't `Set`, and similarly for `LinkedHashMap`. This
allowed cases where the literal would also work when passed or assigned
to something requiring any supertype like `Object`, `dynamic`, or
`Iterable`.

Invert the condition used for the `typeCheck` callback and add checks
looking specifically for the `LinkedHash` variants of the static type.
Add tests demonstrating the new places the lint will trigger.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,11 @@
 # 0.1.126-dev
 
-* fixed false negatives for `prefer_collection_literals` when a Set
-  instantiation is passed as the argument to a function in any position other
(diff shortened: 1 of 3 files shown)


bwilkerson (approved):

bwilkerson (approved):

pq (approved):
Awesome.  Thanks!

PR #2354, opened 2020-11-30
The first step to migrate to ErrorCode constructor with named parameters

scheglov:
This is a step to do https://dart-review.googlesource.com/c/sdk/+/174322


bwilkerson (approved):

pq (commented):
👍

PR #2356, opened 2020-12-01
cache updates

pq:
/cc @bwilkerson


bwilkerson (approved):
I've been amazed recently at how much data is added by each of these cache updates. I wonder whether we might not want to find a more compact form, such as by specifying the first sdk version associated with each linter version. Certainly not high priority, but just a thought.

pq:
I've been mulling over exactly the same thought.  No concrete plan yet but I fully agree.  This format worked OK when there were fewer releases but at this cadence it's a bit much.  Thanks for the nudge!

PR #2358, opened 2020-12-03
simple machine readable rule listing

pq:
Sample (snipped) output:

```json
[
  {
    "name": "always_use_package_imports",
    "description": "Avoid relative imports for files in `lib/`.",
    "group": "errors",
    "details": "*DO* avoid relative imports for files in `lib/`.\n\nWhen mixing relative and absolute imports it's possible to create confusion\nwhere the same member gets imported in two different ways. One way to avoid\nthat is to ensure you consistently use absolute imports for files withing the\n`lib/` directory.\n\nThis is the opposite of 'prefer_relative_imports'.\nMight be used with 'avoid_relative_lib_imports' to avoid relative imports of\nfiles within `lib/` directory outside of it. (for example `test/`)\n\n**GOOD:**\n\n```\nimport 'package:foo/bar.dart';\n\nimport 'package:foo/baz.dart';\n\nimport 'package:foo/src/baz.dart';\n...\n```\n\n**BAD:**\n\n```\nimport 'baz.dart';\n\nimport 'src/bag.dart'\n\nimport '../lib/baz.dart';\n\n...\n```\n\n"
  },
  {
    "name": "avoid_empty_else",
    "description": "Avoid empty else statements.",
    "group": "errors",
    "details": "\n**AVOID** empty else statements.\n\n**BAD:**\n```\nif (x > y)\n  print(\"1\");\nelse ;\n  print(\"2\");\n```\n\n"
  },
  {
    "name": "avoid_print",
    "description": "Avoid `print` calls in production code.",
    "group": "errors",
    "details": "**DO** avoid `print` calls in production code.\n\n**BAD:**\n```\nvoid f(int x) {\n  print('debug: $x');\n  ...\n}\n```\n"
  }
]
```

/cc @csells @devoncarew

### DIFF
--- a/tool/machine.dart
+++ b/tool/machine.dart
@@ -1,0 +1,37 @@
+// Copyright (c) 2015, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:convert';
+
+import 'package:analyzer/src/lint/registry.dart';
+import 'package:args/args.dart';
+import 'package:linter/src/analyzer.dart';
+import 'package:linter/src/rules.dart';
+
+/// Generates a list of lint rules in machine format suitable for consumption by
+/// other tools.
+void main([List<String> args]) {
+  var parser = ArgParser()
+    ..addFlag('pretty',
+        abbr: 'p', help: 'Pretty-print output.', defaultsTo: true);
+  var options = parser.parse(args);
+
+  registerLintRules();
+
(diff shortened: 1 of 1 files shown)


devoncarew (approved):
lgtm

Do we want this tool to be run by specific developers? To be committed to the repo periodically, like when we re-gen docs?

pq:
I'm not sure?  Committing this data is an interesting idea and I hadn't considered it.  We might also add other bits too (e.g., badging data or "is this in pedantic?")

I opened https://github.com/dart-lang/linter/issues/2359 to discuss.

PR #2361, opened 2020-12-07
prefer `dart analyze`

pq:
Migrate from `dartanalyzer` in build.

Sample of a breakage:

image

/cc @bwilkerson @devoncarew

### DIFF
--- a/tool/travis.sh
+++ b/tool/travis.sh
@@ -48,10 +48,7 @@ else
   echo "Running main linter bot"
 
   # Verify that the libraries are error free.
-  dartanalyzer --fatal-warnings \
-    bin/linter.dart \
-    lib/src/rules.dart \
-    test/all.dart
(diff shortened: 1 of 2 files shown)


tool/travis.sh
```diff
@@ -48,10 +48,7 @@ else
   echo "Running main linter bot"
 
   # Verify that the libraries are error free.
-  dartanalyzer --fatal-warnings \
-    bin/linter.dart \
-    lib/src/rules.dart \
-    test/all.dart
+  dart analyze --fatal-warnings --fatal-infos .
```
devoncarew, line 51: `--fatal-warnings` defaults to true; consider just using `--fatal-infos`

devoncarew (approved):

tool/travis.sh
pq: Good catch.

PR #2363, opened 2020-12-07
drop unnecessary `new`

pq:
/cc @bwilkerson


bwilkerson (approved):
I will point out that, while the `new` isn't necessary, it does make it clear that this is a constructor invocation. You could consider using `const` in place of `new` rather than just deleting it if you think it might be more clear to readers.

pq:
Thanks!

PR #2383, opened 2020-12-17
Add font-weight 700 for Roboto Mono

parlough:
A small change, but since we are using it for inline code text, the lack of 700 makes it slightly blurry.

The change is minor and I tried to include screenshots but compression makes the difference hard to tell.


pq:
Excellent.  Thanks!

Issue #2368, opened 2020-12-09
consider enforcing sort order in the build


pq:
See churn in https://github.com/dart-lang/linter/pull/2362.  Also, discussion in #2367.

pq:
Now that we're in the SDK, we can do this easily. 🎉

pq:
Implemented w/ https://github.com/dart-lang/sdk/commit/349e6266ff67f645dcf24155c2b3cf3334f0fdc0.

parlough:
One less thing to consider 🥳

pq:
💯

pq:
We've wanted this for a while. Nice reward for the long haul of migrating the linter into the SDK!

parlough:
I wonder why your commits aren't closing the linked issues automatically? Is it because you're not actually committing, but the Commit Queue is? I've had the same problem with closing site-www issues from the SDK before.

pq:
I'm not sure to be honest.

@whesse @athomas?

athomas:
I believe it's because copybara (the tool that pushes the code from googlesource.com didn't have push access to the linter repo). I made an attempt at fixing that. Let me know if it works!

Background: to close issues in another repo the user pushing the commit has to have access to the other repo. I couldn't find documentation on what access exactly, though. GitHub scans the pushed commits for keywords and then applies its own rules to decide what to do.

pq:
Thanks for looking!

Not sure it worked unfortunately. I had to make a manual edit here:

athomas:
> Not sure it worked unfortunately. I had to make a manual edit here:

Unfortunately, I don't know... 🤷

This is basically what we're doing but I'm not sure what permission we need to grant the bot:
https://docs.github.com/en/repositories/creating-and-managing-repositories/creating-an-issues-only-repository

 It's an "app" so I can't grant it write access to linter specifically, and it already has write access to every repo.

pq:
Ah well. Thanks for looking!

On the bright side, the plan is to move the linter issues over to the SDK sooner than later so the issue won't be long lived.

Again, thanks!

PR #2392, opened 2020-12-22
Cleanup and organize some code

parlough:
There isn't any functional changes here, mostly just cleanup and standardization:

- Switching to const when appropriate
- Switching to final if field isn't reassigned
- Switching from a list to a set if just used for a `contains` check
- Reorganizes some constructors and fields to be consistent
- Minor miscellaneous changes

### DIFF
--- a/lib/src/rules/avoid_equals_and_hash_code_on_mutable_classes.dart
+++ b/lib/src/rules/avoid_equals_and_hash_code_on_mutable_classes.dart
@@ -65,10 +65,10 @@ class C {
 ''';
 
 /// The name of the top-level variable used to mark a immutable class.
-String _immutableVarName = 'immutable';
+const _immutableVarName = 'immutable';
 
 /// The name of `meta` library, used to define analysis annotations.
-String _metaLibName = 'meta';
+const _metaLibName = 'meta';
 
 bool _isImmutable(Element element) =>
     element is PropertyAccessorElement &&
--- a/lib/src/rules/avoid_field_initializers_in_const_classes.dart
+++ b/lib/src/rules/avoid_field_initializers_in_const_classes.dart
@@ -55,7 +55,7 @@ class AvoidFieldInitializersInConstClasses extends LintRule
 }
 
 class HasParameterReferenceVisitor extends RecursiveAstVisitor {
-  Iterable<ParameterElement> parameters;
+  final Iterable<ParameterElement> parameters;
 
   bool useParameter = false;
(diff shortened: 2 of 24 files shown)


pq:
Thanks for this!

I need to simmer on a few suggestions and cycle back w/ some other contributors who are out over the holiday but a few quick thoughts.

> - Switching to const when appropriate
> - Switching to final if field isn't reassigned

+1

> - Switching from a list to a set if just used for a `contains` check

I'm having deja vu but can't remember the substance of a conversation I had a while back with @bwilkerson about this.  Brian: do you recall?

> - Reorganizes some constructors and fields to be consistent

Though it's not enforced, we do use the analyzer's "Sort Members" command.  I'd like to see this enforced in the build (https://github.com/dart-lang/linter/issues/2368) but that's tricky since the bits we'd want to repurpose are in the analysis server package (which is not published).  Could you characterize the consistency you're after?  (And thanks for the care!)

> - Minor miscellaneous changes

parlough:
> I'm having deja vu but can't remember the substance of a conversation I had a while back with @bwilkerson about this. Brian: do you recall?

I am also curious about this, so I'll wait for the input from Brian. In the end, if so desired, I can switch these back to lists.

> Though it's not enforced, we do use the analyzer's "Sort Members" command. I'd like to see this enforced in the build (#2368) but that's tricky since the bits we'd want to repurpose are in the analysis server package (which is not published). Could you characterize the consistency you're after? (And thanks for the care!)

That would definitely be nice to have. I simply moved a few constructors for visitors below the field declarations as that seemed to be the common standard elsewhere in the code, and it is what makes sense to me.

> Minor miscellaneous changes

If you were curious what these were, they were mostly to enable some of the final changes. For example I switched this line:
`environment = environment.where((e) => e.target != node).toList();`
to this:
`environment.removeWhere((e) => e.target == node);`

---

I also added the `int lineNumber` optional parameter to the `Annotation.forLint` constructor as the only time the constructor was used, set the `lineNumber` with a cascade immediately afterwards.

pq:
Thanks for your continued patience on this one.  I'm working on migrating to NNBD (#2399) and would prefer to wait for that to land.  If that stalls we can revisit?

pq:
sure thing.  My guess is the first one may be imperfect and we can tighten some things over a few passes.  Input will be welcome!

parlough:
Will revisit this after the nnbd migration if necessary.

PR #2394, opened 2020-12-22
Switch to addArgumentList and remove TODO

parlough:
# Description

The version of `analyzer` at the time did not yet have `addArgumentList` so this TODO was left. I've gone ahead and switched to the new method and removed the TODO.

Refer to https://github.com/dart-lang/linter/pull/2209#discussion_r465171453 for context.

Cc @bwilkerson

### DIFF
--- a/lib/src/rules/avoid_type_to_string.dart
+++ b/lib/src/rules/avoid_type_to_string.dart
@@ -72,8 +72,6 @@ class AvoidTypeToString extends LintRule implements NodeLintRule {
     registry.addMixinDeclaration(this, visitor);
     registry.addExtensionDeclaration(this, visitor);
 
-    // These nodes delegate to general visitArgumentList,
-    // since SimpleAstVisitor only calls visits for concrete node subtypes.
     registry.addArgumentList(this, visitor);
 
     // Actually checking things at these nodes.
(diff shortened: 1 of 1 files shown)


pq:
This looks good to me, but @bwilkerson, I think you have more context...

lib/src/rules/avoid_type_to_string.dart
```diff
@@ -74,12 +74,7 @@ class AvoidTypeToString extends LintRule implements NodeLintRule {
 
     // These nodes delegate to general visitArgumentList,
     // since SimpleAstVisitor only calls visits for concrete node subtypes.
-    // TODO: replace these with addArgumentList(...) once it's added to the registry API.
-    // Context: https://github.com/dart-lang/linter/pull/2209#discussion_r465196745
-    registry.addFunctionExpressionInvocation(this, visitor);
-    registry.addInstanceCreationExpression(this, visitor);
-    registry.addRedirectingConstructorInvocation(this, visitor);
-    registry.addSuperConstructorInvocation(this, visitor);
+    registry.addArgumentList(this, visitor);
```
bwilkerson, line 77: The methods that are no longer being registered should be removed from the visitor because they are no longer being invoked and the comment on lines 75-76 should be removed as it no longer applies.

bwilkerson (approved):

pq (approved):
💎

Issue #2359, opened 2020-12-03
tool for machine-format lint listing


pq:
https://github.com/dart-lang/linter/pull/2358 introduces a simple tool to generate a JSON dump of rules.  Now what?

Should this output be committed?  Should this tool get run as part of the CI? Should we add more data (e.g., rule set inclusions, like "pedantic"?)

/cc @csells @devoncarew @bwilkerson

pq:
Sample (abbreviated) output:

```json
[
  {
    "name": "always_use_package_imports",
    "description": "Avoid relative imports for files in `lib/`.",
    "group": "errors",
    "details": "*DO* avoid relative imports for files in `lib/`.\n\nWhen mixing relative and absolute imports it's possible to create confusion\nwhere the same member gets imported in two different ways. One way to avoid\nthat is to ensure you consistently use absolute imports for files withing the\n`lib/` directory.\n\nThis is the opposite of 'prefer_relative_imports'.\nMight be used with 'avoid_relative_lib_imports' to avoid relative imports of\nfiles within `lib/` directory outside of it. (for example `test/`)\n\n**GOOD:**\n\n```\nimport 'package:foo/bar.dart';\n\nimport 'package:foo/baz.dart';\n\nimport 'package:foo/src/baz.dart';\n...\n```\n\n**BAD:**\n\n```\nimport 'baz.dart';\n\nimport 'src/bag.dart'\n\nimport '../lib/baz.dart';\n\n...\n```\n\n"
  },
  {
    "name": "avoid_empty_else",
    "description": "Avoid empty else statements.",
    "group": "errors",
    "details": "\n**AVOID** empty else statements.\n\n**BAD:**\n```\nif (x > y)\n  print(\"1\");\nelse ;\n  print(\"2\");\n```\n\n"
  },
  {
    "name": "avoid_print",
    "description": "Avoid `print` calls in production code.",
    "group": "errors",
    "details": "**DO** avoid `print` calls in production code.\n\n**BAD:**\n```\nvoid f(int x) {\n  print('debug: $x');\n  ...\n}\n```\n"
  }
]```

bwilkerson:
> Now what?

That depends. What is the purpose or motivation for this work?

pq:
A first cut of a listing is now being produced as part of our docgen and lives here:

https://dart-lang.github.io/linter/lints/machine/rules.json

pq:
This is finished.  Refinements tracked in new issues.

PR #2396, opened 2020-12-23
add maturity and rule set info to machine json output

pq:
Adds maturity and rule set info to machine json.

```json
  {
    "name": "avoid_empty_else",
    "description": "Avoid empty else statements.",
    "group": "errors",
    "maturity": "stable",
    "sets": [
      "flutter",
      "pedantic"
    ],
    "details": "\n**AVOID** empty else statements.\n\n**BAD:**\n```\nif (x > y)\n  print(\"1\");\nelse ;\n  print(\"2\");\n```\n\n"
  },
```

See: #2359.

/cc @bwilkerson

### DIFF
--- a/tool/machine.dart
+++ b/tool/machine.dart
@@ -9,6 +9,8 @@ import 'package:args/args.dart';
 import 'package:linter/src/analyzer.dart';
 import 'package:linter/src/rules.dart';
 
+import 'doc.dart';
+
 /// Generates a list of lint rules in machine format suitable for consumption by
 /// other tools.
 void main([List<String> args]) {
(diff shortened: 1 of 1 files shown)


bwilkerson (approved):

parlough:
@pq Was it intentional to leave `effective_dart` out of the `sets` field? Perhaps as the `lints` sets will likely supplant them? No worries if so, just wanted to verify.

pq:
Hrmm.  That's a good question.  I can't think of a very good reason to exclude that set.  We can make sure it gets added when `core` and `recommended` get added in.

Thanks!

PR #2411, opened 2021-01-09
Increase break point for margin-0 transition

parlough:
Due to the wrapper width being increased a little while ago, the website can appear outside the browser window as the transition to the 100% contained version happened too late. This causes the transition to occur before the wrapper runs out of space.

This screenshot illustrates the issue:
image


pq:
Thanks!

pq (approved):

PR #2413, opened 2021-01-09
Update misc testing documentation

parlough:
Switches test run instructions from `pub run test` to simply `dart test`.

### DIFF
--- a/doc/WritingLints.MD
+++ b/doc/WritingLints.MD
@@ -91,9 +91,9 @@ The test suite run during the linter's CI, can be run locally like so:
 
     $ dart test/all.dart
 
-alternatively, tests can be run using `pub`:
+alternatively, tests can be run using `dart test`:
 
-    $ pub run test
+    $ dart test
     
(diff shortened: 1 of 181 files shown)


pq:
This is a little tricky.  Since the linter source is used internally at Google (and possibly elsewhere) for testing, changes like this are breaking and need to be carefully considered.  Consider opening an issue proposing the change so we can pickup (and promote) conversation there?

Thanks!

parlough:
@pq Thanks for pointing that out, I didn't fully consider that. I've gone ahead and undid the removal of the test util so the `pub run test -> dart test` updates can still be merged if desired. 

After some more testing, there seems to still be some issues with VM debugging anyway that I will investigate more and followup with an issue to discuss.

parlough:
Will revisit this in the future.

pq:
Thanks a million @parlough!  (And sorry for the slow resolution on  things generally.  Should speed up once we square away a nullability release...)

PR #2417, opened 2021-01-13
Initial commit of avoid_dynamic_calls.

matanlurey:
Closes https://github.com/dart-lang/linter/issues/1782.

---

This is a pretty difficult lint to get "right". I tried my best with the following principals:

1. Simple _existence_ of `dynamic` or a dynamic-ly typed expression should not cause a lint. 
    There are many places where a dynamic expression can be used in a way that does not "cause" a dynamic call:

    - Some types of expressions, such as `as`, `is`, `??`, or `x!` are non-virtual, and are not dynamic calls.
    - Some trickier cases like members that exist on `Object?` (`==`, `hashCode`, `runtimeType`, `toString`).

2. Implicit downcasts from `dynamic` to non-`dynamic`, while bad, is _not_ a dynamic call (it's a runtime cast).
    There are other lints and analysis modes that can lint these.

3. While not explicitly typed `dynamic`, a static type of `Function` being invoked _is_ a dynamic call.

---

Open to any and all suggestions. I'm also happy landing this is an experimental state for others to give more feedback. One thing I'd like to avoid is splitting this into like, half-a-dozen _different_ lints or too many ways to opt-out yet; I think `// ignore: avoid_dynamic_call` isn't great, but we can look at the UX of opting-out independent of landing this lint.

### DIFF
--- a/lib/src/rules/avoid_dynamic_calls.dart
+++ b/lib/src/rules/avoid_dynamic_calls.dart
@@ -139,7 +139,7 @@ class _Visitor extends SimpleAstVisitor<void> {
   @override
   void visitAssignmentExpression(AssignmentExpression node) {
     if (node.readType?.isDynamic != true) {
-      // An assignment expression can only be a dynamid call if it is a
+      // An assignment expression can only be a dynamic call if it is a
       // "compound assignment" (i.e. such as `x += 1`); so if `readType` is not
       // dynamic, we don't need to check further.
       return;
--- a/lib/src/rules/avoid_dynamic_calls.dart
+++ b/lib/src/rules/avoid_dynamic_calls.dart
@@ -1,0 +1,221 @@
+// Copyright (c) 2021, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/token.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import '../analyzer.dart';
+
+const _desc = r'Avoid method calls and property access on a "dynamic" target.';
+
+const _details = r'''
+
+**DO** avoid method calls or property accessors on an object that is either
+explicitly or implicitly statically typed "dynamic". Dynamic calls are treated
+slightly different in every runtime environment and compiler, but most
+production modes (and even some development modes) have both compile size and
+runtime performance penalties associated with dynamic calls.
+
+Additionally, targets typed "dynamic" disabled most static analysis, meaning it
+is easier to lead to a runtime "NoSuchMethodError" or "NullError" than properly
+statically typed Dart code.
+
+Note, that despite "Function" being a type, the semantics are close to identical
+to "dynamic", and calls to an object that are typed "Function" will also trigger
+this lint.
+
+**BAD:**
+```
+void explicitDynamicType(dynamic object) {
+  print(object.foo());
+}
+
+void implicitDynamicType(object) {
+  print(object.foo());
+}
+
+abstract class SomeWrapper {
+  T doSomething<T>();
+}
+
+void inferredDynamicType(SomeWrapper wrapper) {
+  var object = wrapper.doSomething();
+  print(field.foo());
+}
+
+void functionType(Function function) {
+  function();
+}
+```
+
+**GOOD:**
+```
+void explicitType(Fooable object) {
+  object.foo();
+}
+
+void castedType(dynamic object) {
+  (object as Fooable).foo();
+}
+
+abstract class SomeWrapper {
+  T doSomething<T>();
+}
+
+void inferredType(SomeWrapper wrapper) {
+  var object = wrapper.doSomething<Fooable>();
+  object.foo();
+}
+```
+
+''';
+
+class AvoidDynamicCalls extends LintRule implements NodeLintRule {
+  AvoidDynamicCalls()
+      : super(
+          name: 'avoid_dynamic_calls',
+          description: _desc,
+          details: _details,
+          group: Group.errors,
+        );
+
+  @override
+  void registerNodeProcessors(
+    NodeLintRegistry registry,
+    LinterContext context,
+  ) {
+    assert(context != null);
+    final visitor = _Visitor(this);
+    registry
+      ..addAssignmentExpression(this, visitor)
+      ..addBinaryExpression(this, visitor)
+      ..addFunctionExpressionInvocation(this, visitor)
+      ..addIndexExpression(this, visitor)
+      ..addMethodInvocation(this, visitor)
+      ..addPostfixExpression(this, visitor)
+      ..addPrefixExpression(this, visitor)
+      ..addPrefixedIdentifier(this, visitor)
+      ..addPropertyAccess(this, visitor);
+  }
+}
+
+class _Visitor extends SimpleAstVisitor<void> {
+  final LintRule rule;
+
+  _Visitor(this.rule);
+
+  bool _lintIfDynamic(Expression node) {
+    if (node?.staticType?.isDynamic == true) {
+      rule.reportLint(node);
+      return true;
+    } else {
+      return false;
+    }
+  }
+
+  void _lintIfDynamicOrFunction(Expression node) {
+    final staticType = node?.staticType;
+    if (staticType == null) {
+      return;
+    }
+    if (staticType.isDynamic || staticType.isDartCoreFunction) {
+      rule.reportLint(node);
+    }
+  }
+
+  @override
+  void visitAssignmentExpression(AssignmentExpression node) {
+    if (node.readType?.isDynamic != true) {
+      return;
+    }
+    if (node.operator.type == TokenType.QUESTION_QUESTION_EQ) {
+      // x ??= foo is not a dynamic call.
+      return;
+    }
+    rule.reportLint(node);
+  }
+
+  @override
+  void visitBinaryExpression(BinaryExpression node) {
+    switch (node.operator.type) {
+      case TokenType.QUESTION_QUESTION:
+        return;
+      case TokenType.EQ_EQ:
+      case TokenType.BANG_EQ:
+        if (node.rightOperand is NullLiteral) {
+          return;
+        }
+        break;
+    }
+    _lintIfDynamic(node.leftOperand);
+    // We don't check node.rightOperand, because that is an implicit cast, not a
+    // dynamic call (the call itself is based on leftOperand). While it would be
+    // useful to do so, it is better solved by other more specific lints to
+    // disallow implicit casts from dynamic.
+  }
+
+  @override
+  void visitFunctionExpressionInvocation(FunctionExpressionInvocation node) {
+    if (node.function != null) {
+      _lintIfDynamicOrFunction(node.function);
+    }
+  }
+
+  @override
+  void visitIndexExpression(IndexExpression node) {
+    _lintIfDynamic(node.realTarget);
+  }
+
+  @override
+  void visitMethodInvocation(MethodInvocation node) {
+    final receiverWasDynamic = _lintIfDynamic(node.realTarget);
+    if (!receiverWasDynamic && node.target == null) {
+      _lintIfDynamicOrFunction(node.function);
+    }
+  }
+
+  @override
+  void visitPrefixExpression(PrefixExpression node) {
+    if (_lintIfDynamic(node.operand)) {
+      return;
+    }
+    switch (node.operator.type) {
+      case TokenType.PLUS_PLUS:
+      case TokenType.MINUS_MINUS:
+        // The ++ and -- operator expressions don't resolve a static type.
+        if (node.staticType?.isDynamic == true) {
+          rule.reportLint(node.operand);
+        }
+    }
+  }
+
+  @override
+  void visitPrefixedIdentifier(PrefixedIdentifier node) {
+    _lintIfDynamic(node.prefix);
+  }
+
+  @override
+  void visitPostfixExpression(PostfixExpression node) {
+    if (node.operator.type == TokenType.BANG) {
+      // x! is not a dynamic call, even if "x" is dynamic.
+      return;
+    }
+    if (_lintIfDynamic(node.operand)) {
+      return;
+    }
+    switch (node.operator.type) {
+      case TokenType.PLUS_PLUS:
+      case TokenType.MINUS_MINUS:
+        // The ++ and -- operator expressions don't resolve a static type.
+        if (node.staticType?.isDynamic == true) {
+          rule.reportLint(node.operand);
+        }
+    }
+  }
+
+  @override
+  void visitPropertyAccess(PropertyAccess astNode) {
+    _lintIfDynamic(astNode.realTarget);
+  }
+}
--- a/lib/src/rules/avoid_dynamic_calls.dart
+++ b/lib/src/rules/avoid_dynamic_calls.dart
@@ -11,18 +11,18 @@ const _desc = r'Avoid method calls and property access on a "dynamic" target.';
 
 const _details = r'''
 
-**DO** avoid method calls or property accessors on an object that is either
+**DO** avoid method calls or accessing properties on an object that is either
 explicitly or implicitly statically typed "dynamic". Dynamic calls are treated
 slightly different in every runtime environment and compiler, but most
 production modes (and even some development modes) have both compile size and
 runtime performance penalties associated with dynamic calls.
 
-Additionally, targets typed "dynamic" disabled most static analysis, meaning it
+Additionally, targets typed "dynamic" disables most static analysis, meaning it
 is easier to lead to a runtime "NoSuchMethodError" or "NullError" than properly
 statically typed Dart code.
 
 Note, that despite "Function" being a type, the semantics are close to identical
-to "dynamic", and calls to an object that are typed "Function" will also trigger
+to "dynamic", and calls to an object that is typed "Function" will also trigger
 this lint.
 
 **BAD:**
@@ -41,7 +41,11 @@ abstract class SomeWrapper {
 
 void inferredDynamicType(SomeWrapper wrapper) {
   var object = wrapper.doSomething();
-  print(field.foo());
+  print(object.foo());
+}
+
+void callDynamic(dynamic function) {
+  function();
 }
 
 void functionType(Function function) {
@@ -67,6 +71,10 @@ void inferredType(SomeWrapper wrapper) {
   var object = wrapper.doSomething<Fooable>();
   object.foo();
 }
+
+void functionTypeWithParameters(Function() function) {
+  function();
+}
 ```
 
 ''';
@@ -78,6 +86,7 @@ class AvoidDynamicCalls extends LintRule implements NodeLintRule {
           description: _desc,
           details: _details,
           group: Group.errors,
+          maturity: Maturity.experimental,
         );
 
   @override
@@ -119,7 +128,10 @@ class _Visitor extends SimpleAstVisitor<void> {
     if (staticType == null) {
       return;
     }
-    if (staticType.isDynamic || staticType.isDartCoreFunction) {
+    if (staticType.isDynamic) {
+      rule.reportLint(node);
+    }
+    if (staticType.isDartCoreFunction) {
       rule.reportLint(node);
     }
   }
@@ -127,6 +139,9 @@ class _Visitor extends SimpleAstVisitor<void> {
   @override
   void visitAssignmentExpression(AssignmentExpression node) {
     if (node.readType?.isDynamic != true) {
+      // An assignment expression can only be a dynamid call if it is a
+      // "compound assignment" (i.e. such as `x += 1`); so if `readType` is not
+      // dynamic, we don't need to check further.
       return;
     }
     if (node.operator.type == TokenType.QUESTION_QUESTION_EQ) {
--- a/lib/src/rules/avoid_dynamic_calls.dart
+++ b/lib/src/rules/avoid_dynamic_calls.dart
@@ -7,7 +7,7 @@ import 'package:analyzer/dart/ast/token.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
 import '../analyzer.dart';
 
-const _desc = r'Avoid method calls and property access on a "dynamic" target.';
+const _desc = r'Avoid method calls or property accesses on a "dynamic" target.';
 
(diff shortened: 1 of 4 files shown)


test/rules/experiments/nnbd/rules/avoid_dynamic_calls.dart
```diff
@@ -0,0 +1,128 @@
...
+void otherPropertyAccessOrCalls(dynamic a) {
+  a(); // LINT
+  a?.b; // LINT
+  a!.b; // LINT
+  a?.b(); // LINT
+  a!.b(); // LINT
+}
+
+void functionExpressionInvocations(dynamic a(), Function b()) {
+  a(); // OK
```
srawlins, line 56: How about an OK check for a Function-typed expression. like a parameter, `void Function() c`, and `c()`.

srawlins (commented):
I think this looks great.

srawlins:
I'm passing the review to @pq; implementation looks great; this will be much hailed inside Google.

test/rules/experiments/nnbd/rules/avoid_dynamic_calls.dart
```diff
@@ -0,0 +1,128 @@
...
+class Wrapper<T> {
+  final T field;
+  Wrapper(this.field);
+}
+
+void fieldDynamicType(Wrapper<dynamic> wrapper) {
+  final field = wrapper.field;
+  field.foo(); // LINT
+  field.bar; // LINT
+  wrapper.field(); // LINT
```
askeksa-google, line 35: Maybe add a parenthesized version:
```dart
  (wrapper.field)(); // LINT
```

test/rules/experiments/nnbd/rules/avoid_dynamic_calls.dart
```diff
@@ -0,0 +1,128 @@
...
+void functionExpressionInvocations(dynamic a(), Function b()) {
+  a(); // OK
+  a()(); // LINT
+  b(); // OK
+  b()(); // LINT
+}
+
+void typedFunctionButBasicallyDynamic(Function a, Wrapper<Function> b) {
+  a(); // LINT
+  b.field(); // LINT
```
askeksa-google, line 64: Maybe add a parenthesized version:
```dart
  (b.field)(); // LINT
```

askeksa-google (approved):
This is great!

lib/src/rules/avoid_dynamic_calls.dart
```diff
@@ -0,0 +1,226 @@
...
+
+''';
+
+class AvoidDynamicCalls extends LintRule implements NodeLintRule {
+  AvoidDynamicCalls()
+      : super(
+          name: 'avoid_dynamic_calls',
+          description: _desc,
+          details: _details,
+          group: Group.errors,
```
bwilkerson, line 80: If this is experimental, then we should mark it as such by adding an argument of the form `maturity: Maturity.experimental`.

lib/src/rules/avoid_dynamic_calls.dart
```diff
@@ -0,0 +1,226 @@
...
+    if (node.operator.type == TokenType.QUESTION_QUESTION_EQ) {
+      // x ??= foo is not a dynamic call.
+      return;
+    }
+    rule.reportLint(node);
+  }
+
+  @override
+  void visitBinaryExpression(BinaryExpression node) {
+    switch (node.operator.type) {
```
bwilkerson, line 141: We should first check `node.operator.isUserDefinableOperator`. That might even remove the need for the additional checks below, but I'd have to double check to be certain.

lib/src/rules/avoid_dynamic_calls.dart
```diff
@@ -0,0 +1,226 @@
...
+      case TokenType.MINUS_MINUS:
+        // The ++ and -- operator expressions don't resolve a static type.
+        if (node.staticType?.isDynamic == true) {
+          rule.reportLint(node.operand);
+        }
+    }
+  }
+
+  @override
+  void visitPrefixedIdentifier(PrefixedIdentifier node) {
```
bwilkerson, line 199: The primary use case for `PrefixedIdentifier` is to represent an identifier that's preceded by an import prefix. In those cases we don't want to trigger the lint. It's quite possible that `_lintIfDynamic` will correctly not report a lint in those cases, but we should add a test to ensure that there isn't a false positive here. Perhaps something like
```
import 'dart:math' as math;

var p = math.pi;
```

lib/src/rules/avoid_dynamic_calls.dart
```diff
@@ -0,0 +1,226 @@
...
+      return;
+    }
+    if (staticType.isDynamic || staticType.isDartCoreFunction) {
+      rule.reportLint(node);
+    }
+  }
+
+  @override
+  void visitAssignmentExpression(AssignmentExpression node) {
+    if (node.readType?.isDynamic != true) {
```
bwilkerson, line 129: Consider adding a comment indicating that the condition will always be false if the assignment is not a compound assignment. That isn't immediately obvious.

lib/src/rules/avoid_dynamic_calls.dart
```diff
@@ -0,0 +1,226 @@
...
+
+const _desc = r'Avoid method calls and property access on a "dynamic" target.';
+
+const _details = r'''
+
+**DO** avoid method calls or property accessors on an object that is either
+explicitly or implicitly statically typed "dynamic". Dynamic calls are treated
+slightly different in every runtime environment and compiler, but most
+production modes (and even some development modes) have both compile size and
+runtime performance penalties associated with dynamic calls.
```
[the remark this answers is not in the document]
matanlurey: Thanks1

lib/src/rules/avoid_dynamic_calls.dart
matanlurey: I think that doesn't work; consider:

```dart
void example(dynamic a) {
  a != 1;
}
```

The `!=` "operator" is not user-definable, but invokes a user-definable operator (`==`).

lib/src/rules/avoid_dynamic_calls.dart
matanlurey: Woof. This was tricky, @srawlins helped a lot. The (mock?) sdk is not setup to import `dart:math` in these test cases, so I went with `dart:core` and `print`. Hope that's fine

matanlurey:
PTAL, and thanks for all the motivating +1s and great ideas to expand testing and correctness.

lib/src/rules/avoid_dynamic_calls.dart
bwilkerson: Unless there are other examples, it might still be worthwhile to check the general case and the one exception. My concern is a false positive when there's a non-user-definable operator other than `??`, `==`, `!=`, `as` and `is`. For example, does this produce a lint in the following case:
```
void f(dynamic o) {
  o && false;
}
```

lib/src/rules/avoid_dynamic_calls.dart
bwilkerson: I didn't realize that `dart:math` wasn't supported, sorry. But yes, any valid import works. I just wanted to ensure that prefixes didn't look like they had the type `dynamic`.

lib/src/rules/avoid_dynamic_calls.dart
matanlurey: Good suggestions. PTAL!

bwilkerson (approved):

matanlurey:
Thanks! @pq, as lint 👑 , I want to give you a chance to do a once-over before merging. Thanks in advance!

test/rules/experiments/nnbd/rules/avoid_dynamic_calls.dart
```diff
@@ -0,0 +1,128 @@
...
+  a && b; // LINT
+  b + a; // OK; this is an implicit downcast, not a dynamic call
+  a ?? b; // OK; this is a null comparison, not a dynamic call.
+  a is int; // OK
+  a is! int; // OK
+  a as int; // OK
+}
+
+void equalityExpressions(dynamic a, dynamic b) {
+  a == b; // LINT
```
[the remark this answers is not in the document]
askeksa-google: Actually, members of `Object` are a special case. See overall comment.

askeksa-google (changes requested):
I dug into the `==` issue. As it turns out, the CFE treats invocations of `Object` members as non-dynamic even on receivers of type `dynamic`. This means that all of the following are OK:
```dart
void foo(dynamic a, dynamic b) {
  a == b;
  a.hashCode;
  a.noSuchMethod;
  a.noSuchMethod(Invocation());
  a.runtimeType;
  a.toString;
  a.toString();
}
```
For the methods (`noSuchMethod` and `toString`) this only holds when the argument list matches the declaration in `Object`. Thus, the following should lint:
```dart
void foo(dynamic a, dynamic b) {
  a.noSuchMethod();
  a.noSuchMethod(Invocation(), "extra");
  a.toString(radix: 16);
}
```
This is also in line with the language specification, which treats `Object` members specially w.r.t. the type of the invocation.

lib/src/rules/avoid_dynamic_calls.dart
```diff
@@ -0,0 +1,246 @@
+// Copyright (c) 2021, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/token.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import '../analyzer.dart';
+
+const _desc = r'Avoid method calls and property access on a "dynamic" target.';
```
pq, line 10: => property accesses?

or =>

on "dynamic" targets.

?

lib/src/rules/avoid_dynamic_calls.dart
```diff
@@ -0,0 +1,246 @@
...
+
+const _details = r'''
+
+**DO** avoid method calls or accessing properties on an object that is either
+explicitly or implicitly statically typed "dynamic". Dynamic calls are treated
+slightly different in every runtime environment and compiler, but most
```
pq, line 16: => differently?

lib/src/rules/avoid_dynamic_calls.dart
```diff
@@ -0,0 +1,246 @@
...
+production modes (and even some development modes) have both compile size and
+runtime performance penalties associated with dynamic calls.
+
+Additionally, targets typed "dynamic" disables most static analysis, meaning it
+is easier to lead to a runtime "NoSuchMethodError" or "NullError" than properly
+statically typed Dart code.
+
+Note, that despite "Function" being a type, the semantics are close to identical
```
pq, line 24: => Note that despite

?

lib/src/rules/avoid_dynamic_calls.dart
```diff
@@ -0,0 +1,246 @@
...
+to "dynamic", and calls to an object that is typed "Function" will also trigger
+this lint.
+
```
pq, line 27: Super awesome docs.

Would that they were all this good!

💎

lib/src/rules/avoid_dynamic_calls.dart
```diff
@@ -0,0 +1,246 @@
...
+          group: Group.errors,
+          maturity: Maturity.experimental,
+        );
+
+  @override
+  void registerNodeProcessors(
+    NodeLintRegistry registry,
+    LinterContext context,
+  ) {
+    assert(context != null);
```
pq, line 97: Curious about this assert.  I think I noticed this snuck in elsewhere but can't remember the rationale...  Anyway, I don't think it's necessary?

cc @scheglov

pq (approved):
🎉 

Super awesome!

lib/src/rules/avoid_dynamic_calls.dart
bwilkerson: "method calls" and "property access" should either both be plural or neither be plural.

bwilkerson:
> As it turns out, the CFE treats invocations of `Object` members as non-dynamic even on receivers of type `dynamic`.

I'm guessing that by "non-dynamic" you mean that they don't incur the cost of "dynamic" calls (because they do still perform a lookup to find the right method to invoke).

If that's true, then the question becomes: Do we want to flag all invocations on an expression whose type is `dynamic` as stated in the docs, or do we only want to flag invocations that have an associated cost using todays compiler? The former is more consistent and requires less maintenance (with the second answer we need to change this lint every time the implementation of these cases changes), so I would lean in that direction.

matanlurey:
> ```dart
> void foo(dynamic a, dynamic b) {
>   a == b;
>   a.hashCode;
>   a.noSuchMethod;
>   a.noSuchMethod(Invocation());
>   a.runtimeType;
>   a.toString;
>   a.toString();
> }
> ```

@askeksa-google:

I think the CFE might have a different definition of "dynamic call" than what this lint is trying to address. Consider:

```
void foo(dynamic a, dynamic b) {
  a == b;
  a.hashCode;
  a.noSuchMethod;
  a.runtimeType;
  a.toString;
  a.toString();
}

void main() {
  foo(Example(), {});
}

class Example {
  operator==(_) {
    print('>>> A.operator==');
    return true;
  }
}
```

... this prints `>>> A.operator==`, which to me, means it dispatched, on an otherwise untyped object, the `==` operator, because it could be user-definable. I'm not very aware of how maybe the CFE and VM implements this, but at least for Dart2JS and DDC, this is a non-zero cost and requires interceptors, something we are trying to avoid with this lint.

lib/src/rules/avoid_dynamic_calls.dart
matanlurey: Thanks @pq!

lib/src/rules/avoid_dynamic_calls.dart
matanlurey: I'll remove it, just cargo-cult from other lints.

test/rules/experiments/nnbd/rules/avoid_dynamic_calls.dart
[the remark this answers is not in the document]
matanlurey: I think it's a special case, but only for the VM, and not worth special, special-casing :P

askeksa-google:
> I'm guessing that by "non-dynamic" you mean that they don't incur the cost of "dynamic" calls (because they do still perform a lookup to find the right method to invoke).

Yes, what I mean is that they are treated exactly the same as if the receiver had type `Object?` rather than `dynamic`.

I would definitely say we should not treat these calls as dynamic for the purpose of the lint. It is not so much a matter of what the current compilers are able to optimize (there are lots of other cases where a dynamic call can be eliminated in practice). It is a situation which specifically called out in the spec as an exception to the rules about member access on `dynamic`.

The spec does not define what is a dynamic call or not (since that is implementation terminology), but it distinguishes between member access which is statically typed and member access which is not. And in this distinction, accessing an `Object` member on a `dynamic` receiver falls on the "statically typed" side.

matanlurey:
I could see dropping the `Object.<method>` defined bits before landing this lint, especially if means adoption will be higher, so feel free to push-back if you think `==`, `.runtimeType`, `.hashCode`, and `.toString` should be special-cased and I'll make an adjustment.

lib/src/rules/avoid_dynamic_calls.dart
```diff
@@ -0,0 +1,245 @@
...
+  @override
+  void visitPrefixExpression(PrefixExpression node) {
+    if (_lintIfDynamic(node.operand)) {
+      return;
+    }
+    switch (node.operator.type) {
+      case TokenType.PLUS_PLUS:
+      case TokenType.MINUS_MINUS:
+        // The ++ and -- operator expressions don't resolve a static type.
+        if (node.staticType?.isDynamic == true) {
```
scheglov, line 211: Could you clarify what kind of a code we are trying to check here?

askeksa-google:
> ... this prints `>>> A.operator==`, which to me, means it dispatched, on an otherwise untyped object, the `==` operator, because it could be user-definable. I'm not very aware of how maybe the CFE and VM implements this, but at least for Dart2JS and DDC, this is a non-zero cost and requires interceptors, something we are trying to avoid with this lint.

That dispatch happens for a receiver of type `Object` (or any other concrete type) as well (after checking both operands against `null`), so this is not the distinction we are after, as I see it.

> I could see dropping the Object.<method> defined bits before landing this lint, especially if means adoption will be higher, so feel free to push-back if you think ==, .runtimeType, .hashCode, and .toString should be special-cased and I'll make an adjustment.

Yes, I think that would make the lint both more practically useful and more in line with how the spec describes `dynamic`.

Another way to see it is that for a `dynamic x` the spec basically defines `x.objectAccessor` to be treated as `(x as Object?).objectAccessor` where `as Object?` is a no-op at runtime (but has static type implications).

lib/src/rules/avoid_dynamic_calls.dart
bwilkerson: Cases where we're applying `++` or `--` to something with type `dynamic`, such as
```
void f(dynamic x) {
  x++;
}
```

matanlurey: I clarified it @scheglov!

matanlurey:
I loosened checks around members that exist on `Object?` and tightened up some docs.

lib/src/rules/avoid_dynamic_calls.dart
```diff
@@ -193,6 +195,11 @@ class _Visitor extends SimpleAstVisitor<void> {
 
   @override
   void visitMethodInvocation(MethodInvocation node) {
+    final methodName = node.methodName.name;
+    if (methodName == 'noSuchMethod' || methodName == 'toString') {
```
askeksa-google, line 199: Also check the number of arguments. In order to be ignored by the lint, there should be 1 positional parameter for `noSuchMethod` and 0 for `toString`, and in both cases no named parameters.

lib/src/rules/avoid_dynamic_calls.dart
```diff
@@ -0,0 +1,269 @@
...
+    final receiverWasDynamic = _lintIfDynamic(node.realTarget);
+    if (!receiverWasDynamic && node.target == null) {
+      _lintIfDynamicOrFunction(node.function);
+    }
+  }
+
+  @override
+  void visitPrefixExpression(PrefixExpression node) {
+    // Implemented as: node?.staticType?.isDynamic == true
+    if (_lintIfDynamic(node.operand)) {
```
scheglov, line 212: Thank you for the clarification.

I think we want to to such switch on the operator to explicitly express what we are checking and when. Doing one generic `_lintIfDynamic` when for `++x` the type is `null` does not make this code as easy to read as it could be.

And for `++` and `--` we want to check `CompoundAssignmentExpression.readType` directly. Because this is the type on which we will call `operator+` or `operator-`. Just like we do for assignments.

Similarly for postfix expressions.

lib/src/rules/avoid_dynamic_calls.dart
```diff
@@ -216,6 +235,11 @@ class _Visitor extends SimpleAstVisitor<void> {
 
   @override
   void visitPrefixedIdentifier(PrefixedIdentifier node) {
+    final property = node.identifier.name;
+    if (property == 'hashCode' || property == 'runtimeType') {
```
askeksa-google, line 233: A tearoff of `noSuchMethod` or `toString` is also not considered dynamic.

test/rules/experiments/nnbd/rules/avoid_dynamic_calls.dart
```diff
@@ -120,12 +120,19 @@ void binaryExpressions(dynamic a, int b, bool c) {
...
+  a == null; // OK.
+  a != b; // OK
   a != null; // OK.
 }
 
+void membersThatExistOnObject(dynamic a) {
+  a.hashCode; // OK
+  a.runtimeType; // OK
+  a.noSuchMethod(null as Invocation); // OK
+  a.toString(); // OK
```
askeksa-google, line 133: Also test tearoffs of `noSuchMethod` and `toString` (OK) and calling these with the wrong number of arguments, or with named arguments (LINT).

lib/src/rules/avoid_dynamic_calls.dart
matanlurey: Thanks, that's a solid idea. I've consolidated and de-duplicated the code into `_lintPrefixOrPostfixExpression(node, node.operand);`. PTAL.

scheglov (approved):

lib/src/rules/avoid_dynamic_calls.dart
```diff
@@ -21,6 +21,16 @@ Additionally, targets typed "dynamic" disables most static analysis, meaning it
...
 
+There is an exception to methods and properties that exist on "Object?":
+- a.hashCode
+- a.runtimeType
+- a.noSuchMethod(someInvocation)
+- a.toString()
+
+... these members are dynamically dispatched in the web-based runtimes, but not
+in the VM-based ones. Additionally, they are so common that it would be very
+punishing to disallow `any.toString()` or `any == true`, for example.
```
askeksa-google, line 32: I think this description is misleading. AFAIU, dart2js does not distinguish between dynamic and non-dynamic calls at all (except for possibly inserting extra type checks). In DDC, the calls in
```dart
String foo1(dynamic x) => x.toString();
String foo2(Object? x) => x.toString();
String foo3(Object x) => x.toString();
```
are all translated identically, but actual dynamic calls, like
```dart
String bar1(dynamic x) => x.toString(color: "fuchsia");
String bar2(dynamic x) => x.nothing();
```
are translated differently from that (using `dart.dsend`, giving the name of the method as a string).

So the categorization of these calls as non-dynamic holds equally well across all of our compilers.

askeksa-google (approved):
With comments addressed, this looks good.

test/rules/experiments/nnbd/rules/avoid_dynamic_calls.dart
matanlurey: I did the most I could. I think being more comprehensive than this is likely infeasible, as the analyzer issues compile-time errors already for the wrong number of arguments on these methods, including tear-offs, and I'm not able to lint them as a result.

lib/src/rules/avoid_dynamic_calls.dart
matanlurey: That can't possibly be true, because `undefined.toString()` and`null.toString()` in JS is a runtime error.

matanlurey: In either case I think we can tweak the description later. Merging!

PR #2429, opened 2021-01-25
add lint use_named_const

a14n:
Use already defined const value.

**BAD:**

```
const Duration(seconds: 0);
```

**GOOD:**

```
Duration.zero;
```

### DIFF
--- a/lib/src/rules/use_named_const.dart
+++ b/lib/src/rules/use_named_const.dart
@@ -1,0 +1,81 @@
+// Copyright (c) 2021, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/element.dart';
+
+import '../analyzer.dart';
+
+const _desc = r'Use predefined const.';
+
+const _details = r'''
+
+Use already defined const value.
+
+**BAD:**
+```
+const Duration(seconds: 0);
+```
+
+**GOOD:**
+```
+Duration.zero;
+```
+
+''';
+const lintName = 'use_named_const';
+
+class UseNamedConst extends LintRule implements NodeLintRule {
+  UseNamedConst()
+      : super(
+          name: lintName,
+          description: _desc,
+          details: _details,
+          group: Group.style,
+        );
+
+  @override
+  void registerNodeProcessors(
+      NodeLintRegistry registry, LinterContext context) {
+    final visitor = _Visitor(this, context);
+    registry.addInstanceCreationExpression(this, visitor);
+  }
+}
+
+class _Visitor extends SimpleAstVisitor<void> {
+  _Visitor(this.rule, this.context);
+
+  final LintRule rule;
+  final LinterContext context;
+
+  @override
+  void visitInstanceCreationExpression(InstanceCreationExpression node) {
+    if (node.isConst) {
+      final nodeField =
+          node.thisOrAncestorOfType<VariableDeclaration>()?.declaredElement;
+
+      final value = context.evaluateConstant(node).value;
+      final element = node.staticType.element;
+      if (element is ClassElement) {
+        for (final field
+            in element.fields.where((e) => e.isStatic && e.isConst)) {
+          if (nodeField is FieldElement &&
+              (nodeField.enclosingElement as ClassElement).name ==
+                  element.name &&
+              nodeField.name == field.name) {
+            continue;
+          }
+          if (field.computeConstantValue() == value) {
+            rule.reportLint(node,
+                arguments: ['${element.name}.${field.name}'],
+                errorCode:
+                    const LintCode(lintName, 'Use predefined const {0}.'));
+            return;
+          }
+        }
+      }
+    }
+  }
+}
--- a/lib/src/rules/use_named_const.dart
+++ b/lib/src/rules/use_named_const.dart
@@ -53,6 +53,7 @@ class _Visitor extends SimpleAstVisitor<void> {
   @override
   void visitInstanceCreationExpression(InstanceCreationExpression node) {
     if (node.isConst) {
+      final library = (node.root as CompilationUnit).declaredElement.library;
       final nodeField =
           node.thisOrAncestorOfType<VariableDeclaration>()?.declaredElement;
 
@@ -61,17 +62,13 @@ class _Visitor extends SimpleAstVisitor<void> {
       if (element is ClassElement) {
         for (final field
             in element.fields.where((e) => e.isStatic && e.isConst)) {
-          if (nodeField is FieldElement &&
-              (nodeField.enclosingElement as ClassElement).name ==
-                  element.name &&
-              nodeField.name == field.name) {
-            continue;
-          }
-          if (field.computeConstantValue() == value) {
+          if (field != nodeField &&
+              field.computeConstantValue() == value &&
+              (field.isPublic || field.library == library)) {
             rule.reportLint(node,
                 arguments: ['${element.name}.${field.name}'],
-                errorCode:
-                    const LintCode(lintName, 'Use predefined const {0}.'));
+                errorCode: const LintCode(
+                    lintName, "Try using the predefined constant '{0}'."));
             return;
           }
         }
--- a/lib/src/rules/use_named_constants.dart
+++ b/lib/src/rules/use_named_constants.dart
@@ -70,8 +70,7 @@ class _Visitor extends SimpleAstVisitor<void> {
 
         for (final field
             in element.fields.where((e) => e.isStatic && e.isConst)) {
-          if (field != nodeField &&
-              field.isAccessibleIn(library) &&
+          if (field.isAccessibleIn(library) &&
               field.computeConstantValue() == value) {
             rule.reportLint(node,
                 arguments: ['${element.name}.${field.name}'],
--- a/lib/src/rules/use_named_constants.dart
+++ b/lib/src/rules/use_named_constants.dart
@@ -8,11 +8,11 @@ import 'package:analyzer/dart/element/element.dart';
 
 import '../analyzer.dart';
 
-const _desc = r'Use predefined const.';
+const _desc = r'Use predefined named constants.';
 
 const _details = r'''
 
-Use already defined const value.
+Where possible, use already defined const values.
 
 **BAD:**
 ```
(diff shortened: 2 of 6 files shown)


lib/src/rules/use_named_const.dart
```diff
@@ -0,0 +1,81 @@
...
+
+      final value = context.evaluateConstant(node).value;
+      final element = node.staticType.element;
+      if (element is ClassElement) {
+        for (final field
+            in element.fields.where((e) => e.isStatic && e.isConst)) {
+          if (nodeField is FieldElement &&
+              (nodeField.enclosingElement as ClassElement).name ==
+                  element.name &&
+              nodeField.name == field.name) {
```
bwilkerson, line 67: I think this could be simplified. Two elements representing the same field (or any declaration) will always be equal as long as they were generated in the same analysis session. So I think we ought to be able to reduce this to `nodeField == field`. (In which case we could fold the two conditions together and avoid the `continue`.

lib/src/rules/use_named_const.dart
```diff
@@ -0,0 +1,81 @@
...
+            continue;
+          }
+          if (field.computeConstantValue() == value) {
+            rule.reportLint(node,
+                arguments: ['${element.name}.${field.name}'],
+                errorCode:
+                    const LintCode(lintName, 'Use predefined const {0}.'));
```
bwilkerson, line 74: The style for messages is to enclose Dart code in single quotes and to use complete sentences, to the message ought to be `"Use the predefined constant '{0}'."`. Although that's really more of a correction message (in which case it ought to start with "Try using"). The problem is that the code is duplicating the value of a constant field on the class.

lib/src/rules/use_named_const.dart
```diff
@@ -0,0 +1,81 @@
...
+  void visitInstanceCreationExpression(InstanceCreationExpression node) {
+    if (node.isConst) {
+      final nodeField =
+          node.thisOrAncestorOfType<VariableDeclaration>()?.declaredElement;
+
+      final value = context.evaluateConstant(node).value;
+      final element = node.staticType.element;
+      if (element is ClassElement) {
+        for (final field
+            in element.fields.where((e) => e.isStatic && e.isConst)) {
```
bwilkerson, line 63: I think this also needs to check that the field is visible in the library containing the `node`.

bwilkerson (commented):
The only other source of false positives that I can think of is if the values of the two constants is only coincidentally the same and might diverge in the future. For example:
```
class Version {
  static const Version fooVersion = Version('2.3.4');
  static const Version barVersion = Version('2.3.4');
  const Version();
}
```
I suspect that that's rare enough that it's not a big deal, but wanted to mention it so that if others think it's more common that I expect we could discuss it.

lib/src/rules/use_named_const.dart
a14n: Done

lib/src/rules/use_named_const.dart
a14n: Done

lib/src/rules/use_named_const.dart
a14n: Done

lib/src/rules/use_named_const.dart
```diff
@@ -0,0 +1,78 @@
...
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/element.dart';
+
+import '../analyzer.dart';
+
+const _desc = r'Use predefined const.';
+
+const _details = r'''
+
+Use already defined const value.
```
pq, line 15: Maybe something like?:

"Where possible, use already defined const values."

lib/src/rules/use_named_const.dart
```diff
@@ -0,0 +1,78 @@
...
+      final nodeField =
+          node.thisOrAncestorOfType<VariableDeclaration>()?.declaredElement;
+
+      final value = context.evaluateConstant(node).value;
+      final element = node.staticType.element;
+      if (element is ClassElement) {
+        for (final field
+            in element.fields.where((e) => e.isStatic && e.isConst)) {
+          if (field != nodeField &&
+              field.computeConstantValue() == value &&
```
pq, line 66: @scheglov, @bwilkerson : might `computeConstantValue()` be expensive?  Should we consider moving this check to last?

lib/src/rules/use_named_const.dart
```diff
@@ -61,17 +62,13 @@ class _Visitor extends SimpleAstVisitor<void> {
...
-          if (field.computeConstantValue() == value) {
+          if (field != nodeField &&
+              field.computeConstantValue() == value &&
+              (field.isPublic || field.library == library)) {
             rule.reportLint(node,
                 arguments: ['${element.name}.${field.name}'],
-                errorCode:
-                    const LintCode(lintName, 'Use predefined const {0}.'));
+                errorCode: const LintCode(
+                    lintName, "Try using the predefined constant '{0}'."));
```
bwilkerson, line 71: Sorry, I wasn't very clear before. Every error code has a required problem `message` and can optionally have a `correction` message. The problem message should state what's wrong with the code, such as "Constants should be referenced instead of duplicating their value." or "Don't inline constant fields." (or hopefully something better that you come up with). The correction message should suggest a possible way to fix the code, such as "Try using the predefined constant '{0}'.". It might look something like:
```
const LintCode(lintName,
    "Constants should be referenced instead of duplicating their value.",
    correction: "Try using the predefined constant '{0}'.")
```

lib/src/rules/use_named_const.dart
bwilkerson: That depends on (a) whether the value has already been computed and (b) how complex the expression is. It generally won't be too expensive, but it certainly wouldn't hurt to test the known to be inexpensive tests first.

lib/src/rules/use_named_const.dart
```diff
@@ -0,0 +1,78 @@
...
+const Duration(seconds: 0);
+```
+
+**GOOD:**
+```
+Duration.zero;
+```
+
+''';
+const lintName = 'use_named_const';
```
bwilkerson, line 28: Perhaps `use_named_constants'?

pq: Nice.  +1.

lib/src/rules/use_named_const.dart
```diff
@@ -0,0 +1,78 @@
...
+          node.thisOrAncestorOfType<VariableDeclaration>()?.declaredElement;
+
+      final value = context.evaluateConstant(node).value;
+      final element = node.staticType.element;
+      if (element is ClassElement) {
+        for (final field
+            in element.fields.where((e) => e.isStatic && e.isConst)) {
+          if (field != nodeField &&
+              field.computeConstantValue() == value &&
+              (field.isPublic || field.library == library)) {
```
scheglov, line 67: Why not `field.isAccessibleIn(library)`?

lib/src/rules/use_named_const.dart
scheglov: As Brian said - we will compute the value only once, so the order mostly does not matter.

a14n:
Thanks for your comments. PTAL

lib/src/rules/use_named_const.dart
a14n: Unfortunately this new format doesn't display the expected constant in the output of dartanalyze.

lib/src/rules/use_named_constants.dart
```diff
@@ -0,0 +1,79 @@
...
+      final library = (node.root as CompilationUnit).declaredElement.library;
+      final nodeField =
+          node.thisOrAncestorOfType<VariableDeclaration>()?.declaredElement;
+
+      final value = context.evaluateConstant(node).value;
+      final element = node.staticType.element;
+      if (element is ClassElement) {
+        for (final field
+            in element.fields.where((e) => e.isStatic && e.isConst)) {
+          if (field != nodeField &&
```
bwilkerson, line 65: We might want to allow multiple constants in a single class to have the same value without suggesting that one be defined in terms of the other. For example:
```
class Insets {
  static const buttonInsets = Insets(2); // OK
  static const textInsets = Insets(2); // OK
```
If so, then we probably just want a single test outside the loop that the class containing the `nodeField` is different than `element`. What do you think?

lib/src/rules/use_named_const.dart
bwilkerson: Yes, `dartanalyze` doesn't display correction messages by default. I can't remember whether `dart analyze` does. I suppose we could duplicate the information, using a message something like

"The constant '{0}' should be referenced instead of duplicating its value."

lib/src/rules/use_named_constants.dart
```diff
@@ -0,0 +1,87 @@
...
+        // class A {
+        //   const A();
+        //   static const a = A();
+        //   static const b = A();
+        // }
+        if (nodeField?.enclosingElement == element) return;
+
+        for (final field
+            in element.fields.where((e) => e.isStatic && e.isConst)) {
+          if (field != nodeField &&
```
bwilkerson, line 73: Given the test on line 69, I don't think we need to compate `field != nodeField`.

bwilkerson (approved):

lib/src/rules/use_named_constants.dart
```diff
@@ -60,15 +60,23 @@ class _Visitor extends SimpleAstVisitor<void> {
       final value = context.evaluateConstant(node).value;
```
pq, line 60: consider moving `value`, `library` and `nodeField` into the `if (element is ClassElement)` block?

pq (commented):
So awesome.

And timely!  See: https://github.com/flutter/devtools/pull/2633#discussion_r565832558

👍

a14n:
>And timely! See: flutter/devtools#2633 (comment)

Unfortunately this lint would not have caught this. The lint only checks named constants on the instantiated class (to limit false positives).

You can see its usage on flutter at https://github.com/flutter/flutter/pull/74788/files

bwilkerson:
> I wonder if there would be value in having a registry of known common constants?

I'm skeptical about being able to do a reasonable job of this for primitive types. If I write `const Color(0, 0, 255)`, and the class `Color` defines a constant like `static const Color blue = Color(0, 0, 255)``, then there's a reasonable chance I'd want to use the already defined constant. But if I write `widget.inset = 8` (yes, I know that's not how it's done), and the class `HardwareConstants` defines a constant like `static const int bitsPerByte = 8;`, it's highly unlikely that I'd want to use it. Reusing objects is much easier to get right than reusing values.

pq:
It's a good point about primitives and I completely agree.  Objects are more promising for sure but even they might need more context than a simple mapping.  Food for thought anyway!

a14n:
What @bwilkerson said. The current behaviour of this lint look good to me and allowing it to check the instantiation against all constants in the scope risks to have perf impact and higher chance of false positives.

pq:
SGTM.  Thanks!

PR #2435, opened 2021-01-27
Use package:test support for coverage

kevmoo:


kevmoo:
Coverage went down, but I feel this is the right way to do it.

CC @grouma

pq:
I *love* the idea of this but notice that we're getting a lot of duplicated output.

image

I notice in the `all.dart` suite that you removed that I set up mock IO to redirect output which I'm guessing addresses this issue though  I can't for the life of me I can't recall what I was thinking...

Happy to take a look tomorrow.  I agree that this seems cleaner.  Thanks!

kevmoo:
@natebosch @jakemac53 @grouma – we talked about having a "GitHub Actions" auto-detect mode in pkg:test.

I swear we opened an issue, but i can't find it.

natebosch:
> I swear we opened an issue, but i can't find it.

I can't either. File https://github.com/dart-lang/test/issues/1415 so that we don't lose track.

kevmoo:
Closing this out.

Issue #2206, opened 2020-07-30
new lint to ensure proper use of `BuildContexts` (use_build_context_synchronously?)


pq:
**New lint:** Do not store BuildContext into variables, or use across async calls.

This lint will flag source where

- a `BuildContext` is ever assigned to anything other than a local variable, or

- a `BuildContext` is ever passed as an argument to any async function or whose corresponding parameter doesn't have the type `BuildContext` (where an async method is one that returns either `Future` or `Stream`).

Proposed name: `do_not_store_build_contexts` ?

/cc @jefflim-google @bwilkerson

jefflim-google:
> * a BuildContext is ever assigned to anything other than a local variable, or

This sounds correct.

> * a BuildContext is ever passed as an argument to any async function

This is too restrictive. A context can be used in a async function before await is used.

```
Future<bool> doSomething(BuildContext context) async {
  // At this stage, using `context` is fine.
  final navigator = Navigator.of(context);
  final result = await shouldDoSomething();

  // After the await, using `context` is NOT ok.
  navigator.pushNamed('blah');  // This should be permitted.
  Navigator.of(context).pushNamed('blah');  // This should trigger the lint.
}
```

jefflim-google:
Re lint naming: I'm not familiar with with how lints are generally named. Should they be named according to what is permitted, or what is forbidden? `use_build_context_synchronously` ? vs `do_not_store_build_context`

bwilkerson:
> A context can be used in a async function before await is used.

There is currently a technical limitation that would make that difficult: the linter looks at a single library at a time. That means that when it is analyzing the library containing the invocation of the async function it does not have visibility of whether the parameter is referenced after the first `await`.

How many issues would we fail to catch if it ignored passing a context to an async function?

> Should they be named according to what is permitted, or what is forbidden?

We have not been very consistent about that. I suppose it depends on what you're trying to emphasize. In this case I like your suggestion of `use_build_context_synchronously` because it leaves room to expand the lint to find other ways in which that could be violated, whereas `do_not_store_build_context` already fails to imply that they shouldn't be passed as arguments to async functions.

jefflim-google:
I was thinking that the linter would allow the function call without issue:
```
  myMethod(context);
```

But would analyze the function that was called instead: 
```
Future<void> myMethod(BuildContext context) async {
```

This can work by observing a single library at a time still?

jefflim-google:
Expanding on that,

```
[...]

PR #2437, opened 2021-01-28
use build context synchronously (WIP)

pq:
Bumping for initial review.

Marked experimental and ready for broader internal testing.

/cc @bwilkerson 

---

For testing you'll need to drop in `lib/src/rules/use_build_context_synchronously.dart` and merge the changes to `lib/src/rules.dart`. That should be it. 👍 

See: #2206.

/fyi @jefflim-google

### DIFF
--- a/lib/src/rules/use_build_context_synchronously.dart
+++ b/lib/src/rules/use_build_context_synchronously.dart
@@ -13,42 +13,40 @@ import '../analyzer.dart';
 const _desc = r'Do not use BuildContexts across async calls.';
 
 const _details = r'''
-**DO NOT** use BuildContexts across async calls.
+**DO NOT** use BuildContext across asynchronous gaps.
 
-TODO: add rationale.
+Storing `BuildContext` for later usage can easily lead to difficult to diagnose
+crashes. Asynchronous gaps are implicitly storing `BuildContext` and are some of
+the easiest to overlook when writing code.
 
-TODO: describe mechanics.
+When a `BuildContext` is used from a `StatefulWidget`, the `mounted` property
+must be checked after an asynchronous gap.
 
+**GOOD:**
+```
+void onButtonTapped(BuildContext context) {
+  Navigator.of(context).pop();
+}
+```
 
 **BAD:**
 ```
-class MyState extends State<MyWidget> {
-  void m() async {
-    // Uses context from State.
-    Navigator.of(context).pushNamed('routeName');
-
-    await Future<void>.delayed(Duration());
-    // ^--- async gap.
-
-    // Without a mounted check, this is unsafe.
-    Navigator.of(context).pushNamed('routeName'); // LINT
-  }
+void onButtonTapped(BuildContext context) async {
+  await Future.delayed(const Duration(seconds: 1));
+  Navigator.of(context).pop();
 }
 ```
 
 **GOOD:**
 ```
-class MyState extends State<MyWidget> {
-  void m() async {
-    // Uses context from State.
-    Navigator.of(context).pushNamed('routeName');
+class _MyWidgetState extends State<MyWidget> {
+  ...
 
-    await Future<void>.delayed(Duration());
+  void onButtonTapped() async {
+    await Future.delayed(const Duration(seconds: 1));
 
     if (!mounted) return;
-
-    // OK. We checked mounted first.
-    Navigator.of(context).pushNamed('routeName'); // OK
+    Navigator.of(context).pop();
   }
 }
 ```
@@ -87,7 +85,7 @@ class UseBuildContextSynchronously extends LintRule implements NodeLintRule {
   }
 
   static bool inTestDir(CompilationUnit unit) {
-    var path = unit.declaredElement?.source?.fullName;
+    var path = unit.declaredElement?.source.fullName;
     return path != null && _testDirectories.any(path.contains);
   }
 }
@@ -123,7 +121,7 @@ class _Visitor extends SimpleAstVisitor {
   void check(AstNode node) {
     // Walk back and look for an async gap that is not guarded by a mounted
     // property check.
-    var child = node;
+    AstNode? child = node;
     while (child != null && child is! FunctionBody) {
       var parent = child.parent;
       if (parent is Block) {
@@ -151,13 +149,12 @@ class _Visitor extends SimpleAstVisitor {
   }
 
   /// todo (pq): replace in favor of flutter_utils.isBuildContext
-  bool isBuildContext(DartType type) {
+  bool isBuildContext(DartType? type) {
     if (type is! InterfaceType) {
       return false;
     }
     var element = type.element;
-    return element != null &&
-        element.name == _nameBuildContext &&
+    return element.name == _nameBuildContext &&
         element.source.uri == _uriFramework;
   }
 
@@ -173,7 +170,7 @@ class _Visitor extends SimpleAstVisitor {
           var operand = condition.operand;
           // stateContext.mounted => mounted
           while (operand is PrefixedIdentifier) {
-            operand = (operand as PrefixedIdentifier).identifier;
+            operand = operand.identifier;
           }
           if (operand is SimpleIdentifier) {
             if (operand.name == 'mounted') {
--- a/lib/src/rules/use_build_context_synchronously.dart
+++ b/lib/src/rules/use_build_context_synchronously.dart
@@ -140,6 +140,11 @@ class _Visitor extends SimpleAstVisitor {
             return;
           }
         }
+      } else if (parent is IfStatement) {
+        // if (mounted) { ... do ... }
+        if (isMountedCheck(parent, positiveCheck: true)) {
+          return;
+        }
       }
 
       child = parent;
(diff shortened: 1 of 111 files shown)


lib/src/rules/use_build_context_synchronously.dart
```diff
@@ -0,0 +1,199 @@
...
+class MyState extends State<MyWidget> {
+  void m() async {
+    // Uses context from State.
+    Navigator.of(context).pushNamed('routeName');
+
+    await Future<void>.delayed(Duration());
+
+    if (!mounted) return;
+
+    // OK. We checked mounted checked first.
```
bwilkerson, line 51: nit: "We checked mounted checked first." --> "We checked mounted first."

lib/src/rules/use_build_context_synchronously.dart
```diff
@@ -0,0 +1,199 @@
...
+    registry.addInstanceCreationExpression(this, visitor);
+    registry.addFunctionExpressionInvocation(this, visitor);
+  }
+}
+
+class _Visitor extends SimpleAstVisitor {
+  static const _nameBuildContext = 'BuildContext';
+  final Uri _uriFramework;
+
+  // todo (pq): replace in favor of flutter_utils.isBuildContext
```
bwilkerson, line 81: This comment probably became detached after a sort.

lib/src/rules/use_build_context_synchronously.dart
```diff
@@ -0,0 +1,199 @@
...
+    return false;
+  }
+
+  void check(AstNode node) {
+    // Walk back and look for an async gap that is not guarded by a mounted
+    // property check.
+    var child = node;
+    while (child != null && child is! FunctionBody) {
+      var parent = child.parent;
+      if (parent is Block) {
```
bwilkerson, line 103: This is missing checks for the other node types that we discussed this morning.

lib/src/rules/use_build_context_synchronously.dart
```diff
@@ -0,0 +1,199 @@
...
+            return;
+          }
+        }
+      }
+
+      child = child.parent;
+    }
+  }
+
+  bool isAsync(Statement statement) {
```
bwilkerson, line 121: The idea is right, but I think you actually need to use a full visitor here to visit every part of the statement. The `await` expression doesn't need to be the immediate child of an expression statement. Consider, for example, `print(await f());`. This method will miss that case.

lib/src/rules/use_build_context_synchronously.dart
```diff
@@ -0,0 +1,199 @@
...
+    // their own mounted checks.  The cost of this generality is the possibility
+    // of false negatives.
+    if (statement is IfStatement) {
+      var condition = statement.condition;
+      if (condition is PrefixExpression) {
+        if (condition.operator.type == TokenType.BANG) {
+          var operand = condition.operand;
+          if (operand is SimpleIdentifier) {
+            if (operand.name == 'mounted') {
+              return true;
```
bwilkerson, line 169: This doesn't check for a return statement. The code `if (!mounted) print('Oops');` would also pass this test.

bwilkerson (commented):
I'm guessing that you weren't really ready for a review and that you know most of what I found to comment on. I didn't realize 'till most of the way through, so I decided to comment anyway just in case any of it's useful.

pq:
Thanks for the feedback @bwilkerson.  It's always welcome! 🙏

lib/src/rules/use_build_context_synchronously.dart
pq: yep.  We should really fix this some day!

lib/src/rules/use_build_context_synchronously.dart
pq: Absolutely.  Ironically, that's much simpler too.  Done.

lib/src/rules/use_build_context_synchronously.dart
pq: Exactly right (and hilarious).  Fixed.

lib/src/rules/use_build_context_synchronously.dart
```diff
@@ -0,0 +1,220 @@
...
+          if (isMountedCheck(s)) {
+            return;
+          } else if (isAsync(s)) {
+            rule.reportLint(node);
+            return;
+          }
+        }
+      }
+
+      child = child.parent;
```
bwilkerson, line 145: Consider just `child = parent;`.

lib/src/rules/use_build_context_synchronously.dart
```diff
@@ -0,0 +1,220 @@
...
+    // their own mounted checks.  The cost of this generality is the possibility
+    // of false negatives.
+    if (statement is IfStatement) {
+      var condition = statement.condition;
+      if (condition is PrefixExpression) {
+        if (condition.operator.type == TokenType.BANG) {
+          var operand = condition.operand;
+          // stateContext.mounted => mounted
+          while (operand is PrefixedIdentifier) {
+            operand = operand.identifier;
```
bwilkerson, line 180: `operand.identifier` will always return a `SimpleIdentifier`, so there's no need for a loop, just use an `if`.

lib/src/rules/use_build_context_synchronously.dart
```diff
@@ -0,0 +1,220 @@
...
+
+  bool isMountedCheck(Statement statement) {
+    // This is intentionally naive.  Using a simple 'mounted' property check
+    // as a signal plays nicely w/ unanticipated framework classes that provide
+    // their own mounted checks.  The cost of this generality is the possibility
+    // of false negatives.
+    if (statement is IfStatement) {
+      var condition = statement.condition;
+      if (condition is PrefixExpression) {
+        if (condition.operator.type == TokenType.BANG) {
```
bwilkerson, line 176: This covers the case of
```
if (!mounted) return;
```
but not
```
if (mounted) {
  context.something();
}
```
Do we need to support the latter case?

test/rules/use_build_context_synchronously.dart
```diff
@@ -0,0 +1,269 @@
...
+  // OR: should that be disallowed in another lint?
+  if (true) {
+    Navigator.of(context).pushNamed('routeName'); // LINT
+  }
+}
+
+void topLevel3(BuildContext context) async {
+  while (true) {
+    // OK the first time only!
+    Navigator.of(context).pushNamed('routeName'); // TODO: LINT
```
bwilkerson, line 257: To better support loops do we need to support `if (!mounted) break;` and `if (!mounted) continue;` (and/or the `if (mounted) ...` pattern I suggested above)?

test/rules/use_build_context_synchronously.dart
```diff
@@ -0,0 +1,269 @@
...
+  await Future<void>.delayed(Duration());
+  Navigator.of(context).pushNamed('routeName'); // LINT
+}
+
+void topLevel2(BuildContext context) async {
+  Navigator.of(context).pushNamed('routeName'); // OK
+
+  await Future<void>.delayed(Duration());
+  // todo (pq): validate other conditionals (for, while, do, ...)
+  // OR: should that be disallowed in another lint?
```
bwilkerson, line 248: Even if it's disallowed in another lint we won't know that both lints are always enabled together, so we need to decide what the right behavior is without depending on another lint.

lib/src/rules/use_build_context_synchronously.dart
pq: Good catch.

test/rules/use_build_context_synchronously.dart
pq: TBH, I'm not sure.  I'm inclined to support switch cases and have added a TODO but would like to get some input from @jefflim-google to get a sense for what standard practice looks like before going too far.  The `if (mounted) ... ` case seems straight-forward so I added it.

lib/src/rules/use_build_context_synchronously.dart
pq: I think yeah.  Added.

lib/src/rules/use_build_context_synchronously.dart
```diff
@@ -0,0 +1,239 @@
...
+          }
+          var then = statement.thenStatement;
+          if (then is ReturnStatement) {
+            return true;
+          }
+          if (then is BreakStatement) {
+            return true;
+          }
+          if (then is Block) {
+            return then.statements.last is ReturnStatement;
```
bwilkerson, line 210: Or `break`? Or `continue`? Consider creating a utility method for "terminates control".

lib/src/rules/use_build_context_synchronously.dart
```diff
@@ -0,0 +1,239 @@
...
+        if (isMountedCheck(parent, positiveCheck: true)) {
+          return;
+        }
+      }
+
+      child = parent;
+    }
+  }
+
+  bool isAsync(Statement statement) {
```
bwilkerson, line 154: Do we need to worry about the case where a method or function is invoked that returns a `Future` but where the returned `Future` isn't awaited?

bwilkerson (approved):

lib/src/rules/use_build_context_synchronously.dart
pq: Good idea.  Added.

lib/src/rules/use_build_context_synchronously.dart
pq: It's a good question.  In general I think this will get caught by `unawaited_futures` but something to think about.

passsy:
That's a fantastic addition, thanks!

pq:
Feedback is welcome.  Thank you @passsy!

PR #2445, opened 2021-02-07
dependency updates

pq:
/cc @bwilkerson

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -13,25 +13,25 @@ environment:
 
 dependencies:
   analyzer: ^0.41.1
-  args: '>=1.4.0 <2.0.0'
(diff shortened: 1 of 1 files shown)


bwilkerson (approved):
I don't know what the current advice is, but I would have expected that we wanted to restrict these versions to being null safe versions, and not allow pub to solve with a non-null safe version of any packages. But maybe that's wrong. You might want to loop in Nate or Jake on this one.

Of course, it's also the case that in practice the version numbers are actually pinned to the versions in the SDK whenever this package is used, so maybe it doesn't really matter much.

pq:
This was just a babsytep to migration but in the end probably not worth bothering so I'll close it out in favor of doing it for real once `analyzer` is published.

PR #2452, opened 2021-02-12
nnbd migration

pq:
First cut at migration using the migration tool.

As discussed, I'd like to land this (if green) and iterate on improvements.

See: https://github.com/dart-lang/linter/issues/2453

### DIFF
--- a/lib/src/ast.dart
+++ b/lib/src/ast.dart
@@ -122,7 +122,6 @@ bool isHashCode(ClassMember element) =>
 bool isInLibDir(CompilationUnit node, WorkspacePackage? package) {
   if (package == null) return false;
   final cuPath = node.declaredElement!.library.source.fullName;
-  if (cuPath == null) return false;
   final libDir = path.join(package.root, 'lib');
   return path.isWithin(libDir, cuPath);
 }
--- a/lib/src/ast.dart
+++ b/lib/src/ast.dart
@@ -15,7 +15,7 @@ import 'analyzer.dart';
 import 'utils.dart';
 
 /// Returns direct children of [parent].
-List<Element> getChildren(Element parent, [String name]) {
+List<Element> getChildren(Element parent, [String? name]) {
   final children = <Element>[];
   visitChildren(parent, (Element element) {
     if (name == null || element.displayName == name) {
@@ -27,12 +27,12 @@ List<Element> getChildren(Element parent, [String name]) {
 }
 
 /// Return the compilation unit of a node
-CompilationUnit getCompilationUnit(AstNode node) =>
+CompilationUnit? getCompilationUnit(AstNode node) =>
     node.thisOrAncestorOfType<CompilationUnit>();
 
 /// Returns a field identifier with the given [name] in the given [decl]'s
 /// variable declaration list or `null` if none is found.
-SimpleIdentifier getFieldIdentifier(FieldDeclaration decl, String name) {
+SimpleIdentifier? getFieldIdentifier(FieldDeclaration decl, String name) {
   for (var v in decl.fields.variables) {
     if (v.name?.name == name) {
       return v.name;
@@ -50,7 +50,7 @@ AstNode getNodeToAnnotate(Declaration node) {
 /// If the [node] is the finishing identifier of an assignment, return its
 /// "writeElement", otherwise return its "staticElement", which might be
 /// thought as the "readElement".
-Element getWriteOrReadElement(SimpleIdentifier node) {
+Element? getWriteOrReadElement(SimpleIdentifier node) {
   var writeElement = _getWriteElement(node);
   if (writeElement != null) {
     return writeElement;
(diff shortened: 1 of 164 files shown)


lib/src/ast.dart
```diff
@@ -268,7 +268,7 @@ bool _checkForSimpleSetter(MethodDeclaration setter, Expression expression) {
   if (expression is! AssignmentExpression) {
     return false;
   }
-  final assignment = expression as AssignmentExpression;
+  final assignment = expression;
```
bwilkerson, line 270: This assignment is unlikely to be necessary now.

lib/src/formatter.dart
```diff
@@ -14,15 +14,15 @@ import 'util/score_utils.dart';
 // Number of times to perform linting to get stable benchmarks.
 const benchmarkRuns = 10;
 
-String getLineContents(int lineNumber, AnalysisError error) {
+String getLineContents(int? lineNumber, AnalysisError error) {
```
bwilkerson, line 17: `lineNumber` should not be nullable.

lib/src/formatter.dart
```diff
@@ -14,15 +14,15 @@ import 'util/score_utils.dart';
...
   final path = error.source.fullName;
   final file = File(path);
   String failureDetails;
   if (!file.existsSync()) {
```
bwilkerson, line 21: It isn't part of the migration, but there's no value in testing whether the file exists. Just read the content and catch the exception if the file doesn't exist.

lib/src/formatter.dart
```diff
@@ -166,17 +166,17 @@ abstract class ReportFormatter {
...
-  final int elapsedMs;
-  final String fileRoot;
+  final int? fileCount;
+  final int? elapsedMs;
+  final String? fileRoot;
   final bool showStatistics;
-  final bool machineOutput;
-  final bool quiet;
+  final bool? machineOutput;
+  final bool? quiet;
```
bwilkerson, line 179: `bool?` is a strong signal that there's a bug in the code. Consider making these non-nullable and clean up the code.

lib/src/formatter.dart
```diff
@@ -322,10 +322,10 @@ class SimpleFormatter implements ReportFormatter {
 
 class _Stat implements Comparable<_Stat> {
   final String name;
-  final int elapsed;
+  final int? elapsed;
```
bwilkerson, line 325: This should be non-nullable.

lib/src/rules/always_put_control_body_on_new_line.dart
```diff
@@ -93,14 +93,14 @@ class _Visitor extends SimpleAstVisitor<void> {
     _checkNodeOnNextLine(node.body, node.rightParenthesis.end);
   }
 
-  void _checkNodeOnNextLine(AstNode node, int controlEnd) {
+  void _checkNodeOnNextLine(AstNode? node, int controlEnd) {
```
bwilkerson, line 96: `node` should be non-nullable.

lib/src/rules/avoid_renaming_method_parameters.dart
```diff
@@ -82,7 +82,7 @@ class _Visitor extends SimpleAstVisitor<void> {
       return;
     }
 
-    final classElement = parentElement as ClassElement;
+    final classElement = parentElement;
```
bwilkerson, line 85: This variable is unnecessary now.

lib/src/rules/avoid_returning_null.dart
```diff
@@ -75,25 +75,25 @@ class _Visitor extends SimpleAstVisitor<void> {
...
   @override
   void visitMethodDeclaration(MethodDeclaration node) {
-    if (_isPrimitiveType(node.declaredElement.returnType)) {
+    if (_isPrimitiveType(node.declaredElement!.returnType)) {
       _visitFunctionBody(node.body);
     }
   }
 
-  void _visitFunctionBody(FunctionBody node) {
+  void _visitFunctionBody(FunctionBody? node) {
```
bwilkerson, line 90: `node` should be non-nullable. The analyzer APIs have been updated to disallow any function from having a null body. The compiler won't catch the change when you update to the next version of analyzer.

bwilkerson (approved):
Sorry, I thought you were ready for a review, but I've had to refresh several times since I started, and now I have no idea what I have and haven't looked at.

It probably doesn't matter, though, because there's way too much to do a decent review, so I'll just have to trust that you'll clean it up after you commit it. I made a couple of specific requests near the beginning, but in general here's a list of the kinds of issues I saw that I think should be cleaned up:

1. Every use of the null check operator (`!`) should be removed. There are lots of different ways to do that depending on the context, every single one should be removed, or there should be a comment explaining why it isn't possible to get rid of it.

2. Every question mark that can be removed should be. There are lots of places where there are nullable types that I think we can get rid of. One common pattern before was to check for null argument values in every method, but the code is much cleaner in a null safe world if we check at every call site.

3. Every local variable that was created in order to hold a casted value should be removed.

pubspec.yaml
```diff
@@ -30,7 +30,8 @@ dev_dependencies:
     path: test/mock_packages/fixnum
   github: '>=6.0.5 <8.0.0'
   grinder: ^0.9.0-nullsafety
-  #http: ^0.13.0
+  # http: ^0.13.0
```
bwilkerson, line 33: Remove the commented out line?

pq: Done.

pq:
As described in #2453, landing this as-is with tidying to follow.

Issue #2453, opened 2021-02-12
`package:linter` nnbd migration plan


pq:
The `package:linter` migration plan:

- [x] 1. land an initial migration that is as tool-driven as possible that runs tests, etc in unsound mode (@pq)
- [x] 2. follow-up with changes to tidy up `lib` (e.g., remove unnecessary null operators) (@pq) (see: https://github.com/dart-lang/linter/pull/2452#pullrequestreview-589866184)
- [x] 3. cleanup `tool` (gated somewhat by `dev_dependencies` such as `github`) (@pq)
- [x] 4. fix-up build scripts and remove `unsound` flags (https://github.com/dart-lang/linter/pull/2467)

1. and 2. will happen before a release (3 is stretch).

Related issues:

https://github.com/SpinlockLabs/github.dart/issues/236

In conversation, @bwilkerson and I agreed (again) how nice it would be to have some lints to disallow ~`?`s and~ `!`s as this would help in tidying.  (I think @scheglov is wanting something similar.)

bwilkerson:
To be clear, I don't think we want a lint to disallow `?`s. There are times when they're necessary. I just want one to disallow null checks because I think they're almost always wrong.

scheglov:
I thought so too, that `Type?` and `?.` are useful and are not unsafe.
But `!` null checks are unsafe.

pq:
Sorry yes, I wrote in haste.  I 100% agree.

PR #2454, opened 2021-02-13
nnbd cleanup

pq:
Some nullability cleanup (see #2453).

/cc @bwilkerson

### DIFF
--- a/lib/src/ast.dart
+++ b/lib/src/ast.dart
@@ -255,7 +255,7 @@ bool _checkForSimpleGetter(MethodDeclaration getter, Expression? expression) {
       Element? getterElement = getter.declaredElement;
       // Skipping library level getters, test that the enclosing element is
       // the same
-      if (staticElement.enclosingElement == getterElement?.enclosingElement) {
+      if (staticElement.enclosingElement == getterElement!.enclosingElement) {
         return staticElement.isSynthetic && staticElement.variable.isPrivate;
       }
     }
--- a/lib/src/ast.dart
+++ b/lib/src/ast.dart
@@ -255,7 +255,7 @@ bool _checkForSimpleGetter(MethodDeclaration getter, Expression? expression) {
       Element? getterElement = getter.declaredElement;
       // Skipping library level getters, test that the enclosing element is
       // the same
-      if (staticElement.enclosingElement == getterElement!.enclosingElement) {
+      if (staticElement.enclosingElement == getterElement?.enclosingElement) {
         return staticElement.isSynthetic && staticElement.variable.isPrivate;
       }
     }
--- a/lib/src/ast.dart
+++ b/lib/src/ast.dart
@@ -252,11 +252,10 @@ bool _checkForSimpleGetter(MethodDeclaration getter, Expression? expression) {
   if (expression is SimpleIdentifier) {
     var staticElement = expression.staticElement;
     if (staticElement is PropertyAccessorElement) {
-      var getterElement = getter.declaredElement?.enclosingElement;
+      var enclosingElement = getter.declaredElement?.enclosingElement;
       // Skipping library level getters, test that the enclosing element is
       // the same
-      if (getterElement != null &&
-          staticElement.enclosingElement == getterElement) {
+      if (staticElement.enclosingElement == enclosingElement) {
         return staticElement.isSynthetic && staticElement.variable.isPrivate;
       }
     }
--- a/lib/src/ast.dart
+++ b/lib/src/ast.dart
@@ -252,10 +252,11 @@ bool _checkForSimpleGetter(MethodDeclaration getter, Expression? expression) {
   if (expression is SimpleIdentifier) {
     var staticElement = expression.staticElement;
     if (staticElement is PropertyAccessorElement) {
-      Element? getterElement = getter.declaredElement;
+      var getterElement = getter.declaredElement?.enclosingElement;
       // Skipping library level getters, test that the enclosing element is
       // the same
-      if (staticElement.enclosingElement == getterElement!.enclosingElement) {
+      if (getterElement != null &&
+          staticElement.enclosingElement == getterElement) {
         return staticElement.isSynthetic && staticElement.variable.isPrivate;
       }
     }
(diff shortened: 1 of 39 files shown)


lib/src/rules/prefer_is_empty.dart
```diff
@@ -133,7 +133,7 @@ class _Visitor extends SimpleAstVisitor<void> {
 
     // Don't lint if we're in a const constructor initializer.
     final constructorInitializer =
-        search.parent!.thisOrAncestorOfType<ConstructorInitializer>();
+        search.parent?.thisOrAncestorOfType<ConstructorInitializer>();
```
bwilkerson, line 136: Another place where `parent?.` can safely be removed.

lib/src/rules/prefer_iterable_whereType.dart
```diff
@@ -81,7 +81,7 @@ class _Visitor extends SimpleAstVisitor<void> {
       if (expression is IsExpression && expression.notOperator == null) {
         final target = expression.expression;
         if (target is SimpleIdentifier &&
-            target.name == arg.parameters!.parameters.first.identifier!.name) {
+            target.name == arg.parameters?.parameters.first.identifier!.name) {
```
bwilkerson, line 84: Consider assigning `arg.parameters?.parameters.first.identifier` to a local variable to remove the other null check operator.

bwilkerson (approved):

lib/src/ast.dart
```diff
@@ -252,10 +252,11 @@ bool _checkForSimpleGetter(MethodDeclaration getter, Expression? expression) {
   if (expression is SimpleIdentifier) {
     var staticElement = expression.staticElement;
     if (staticElement is PropertyAccessorElement) {
-      Element? getterElement = getter.declaredElement;
+      var getterElement = getter.declaredElement?.enclosingElement;
```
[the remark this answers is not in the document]
pq: Right.  Good catch.

lib/src/ast.dart
```diff
@@ -252,10 +252,11 @@ bool _checkForSimpleGetter(MethodDeclaration getter, Expression? expression) {
...
       // Skipping library level getters, test that the enclosing element is
       // the same
-      if (staticElement.enclosingElement == getterElement!.enclosingElement) {
+      if (getterElement != null &&
+          staticElement.enclosingElement == getterElement) {
```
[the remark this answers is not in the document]
pq: 👍

lib/src/rules/avoid_equals_and_hash_code_on_mutable_classes.dart
```diff
@@ -110,7 +110,7 @@ class _Visitor extends SimpleAstVisitor<void> {
   }
 
   ClassElement? _getClassForMethod(MethodDeclaration node) =>
-      node.parent!.thisOrAncestorOfType<ClassDeclaration>()?.declaredElement;
+      node.parent?.thisOrAncestorOfType<ClassDeclaration>()?.declaredElement;
```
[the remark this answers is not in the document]
pq: Great question.  I'll add a TODO to follow-up.

lib/src/rules/avoid_init_to_null.dart
```diff
@@ -87,18 +87,26 @@ class _Visitor extends SimpleAstVisitor<void> {
 
   @override
   void visitDefaultFormalParameter(DefaultFormalParameter node) {
+    var declaredElement = node.declaredElement;
+    if (declaredElement == null) {
```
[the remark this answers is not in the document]
pq: Yeah.  I generally prefer that but have wondered whether the pile up of conditions is harder for folks to follow.  (at least you agree that it isn't! 😉)

Updating.

lib/src/rules/avoid_shadowing_type_parameters.dart
```diff
@@ -63,10 +63,11 @@ class _Visitor extends SimpleAstVisitor<void> {
 
   @override
   void visitGenericTypeAlias(GenericTypeAlias node) {
-    if (node.functionType?.typeParameters == null) {
+    var typeParameters = node.functionType?.typeParameters;
+    if (typeParameters == null) {
       return;
     }
-    _checkForShadowing(node.functionType!.typeParameters, node.typeParameters);
+    _checkForShadowing(typeParameters, node.typeParameters);
```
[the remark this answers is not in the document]
pq: Agreed.  This is a case of me striving for a kind of consistency (inconsistently!)  Anyway, I put it right and it feel more natural to me too.  Thanks!

lib/src/rules/prefer_is_empty.dart
pq: 👍

lib/src/rules/prefer_iterable_whereType.dart
pq: Great catch.  I missed that!

pq (commented):
Lots of good catches. Thanks!

lib/src/rules/prefer_iterable_whereType.dart
```diff
@@ -80,8 +80,10 @@ class _Visitor extends SimpleAstVisitor<void> {
       expression = expression?.unParenthesized;
       if (expression is IsExpression && expression.notOperator == null) {
         final target = expression.expression;
+        var identifier = arg.parameters?.parameters.first.identifier;
         if (target is SimpleIdentifier &&
-            target.name == arg.parameters?.parameters.first.identifier!.name) {
+            identifier != null &&
+            target.name == identifier.name) {
```
parlough: I think this would be cleaner as:

```dart
final target = expression.expression;
if (target is SimpleIdentifier &&
    target.name == arg.parameters?.parameters.first.identifier?.name) {
  rule.reportLint(node.methodName);
}
```

This way calculating the identifier can be short circuited and avoided if necessary, plus `name` isn't nullable if the right hand side evaluates to `null` the expression won't evaluate to true anyway.

pq: Nice.  Thanks!

PR #2457, opened 2021-02-16
enable some nullsafety lints

a14n:

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -42,6 +42,7 @@ linter:
     - camel_case_types
     - cancel_subscriptions
 #    - cascade_invocations
+    - cast_nullable_to_non_nullable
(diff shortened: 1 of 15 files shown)


pq:
Cool!  I've got a mega PR about to go out that includes these changes (though not the lint enablements but I can add them 👍 ).  Sorry for the duplicated work...  Maybe it's time we setup an announce list or a discourse channel so we can all better coordinate?

pq:
See also: #2453

a14n:
No worries :)

bwilkerson:
> ... though not the lint enablements but I can add them ...

Bonus points if you separate out the lint enablement and, when we enable the lints we ensure that we have bulk fix support for fixing all the violations.

PR #2458, opened 2021-02-16
nnbd tidying

pq:
More tidying up.

On balance I think this makes the codebase a lot better and surely more safe.  There were a number of places where I was tempted to do deeper structural re-organization but I opted not to in the interest of keeping this more manageable.  At some point I do think a full audit would be in order.

There are a few test utilities left after this but this should get `lib` under control.

Thanks in advance for muddling through!

/cc @bwilkerson

### DIFF
--- a/lib/src/rules/pub/sort_pub_dependencies.dart
+++ b/lib/src/rules/pub/sort_pub_dependencies.dart
@@ -3,6 +3,7 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'package:analyzer/src/lint/pub.dart'; // ignore: implementation_imports
+import 'package:source_span/source_span.dart';
 
 import '../../analyzer.dart';
 
@@ -47,16 +48,25 @@ class Visitor extends PubspecVisitor<void> {
   }
 
   void _visitDeps(PSDependencyList dependencies) {
+    int compare(SourceLocation? lc1, SourceLocation? lc2) {
+      if (lc1 == null || lc2 == null) {
+        return 0;
+      }
+      return lc1.compareTo(lc2);
+    }
+
     final depsByLocation = dependencies.toList()
-      ..sort((d1, d2) => d1.name!.span.start.compareTo(d2.name!.span.start));
+      ..sort((d1, d2) => compare(d1.name?.span.start, d2.name?.span.start));
     var previousName = '';
     for (final dep in depsByLocation) {
-      final name = dep.name!.text!;
-      if (name.compareTo(previousName) < 0) {
-        rule.reportPubLint(dep.name!);
-        return;
+      final name = dep.name?.text;
+      if (name != null) {
+        if (name.compareTo(previousName) < 0) {
+          rule.reportPubLint(dep.name!);
+          return;
+        }
+        previousName = name;
       }
-      previousName = name;
     }
   }
 }
--- a/lib/src/rules/pub/sort_pub_dependencies.dart
+++ b/lib/src/rules/pub/sort_pub_dependencies.dart
@@ -59,13 +59,16 @@ class Visitor extends PubspecVisitor<void> {
       ..sort((d1, d2) => compare(d1.name?.span.start, d2.name?.span.start));
     var previousName = '';
     for (final dep in depsByLocation) {
-      final name = dep.name?.text;
+      final name = dep.name;
       if (name != null) {
-        if (name.compareTo(previousName) < 0) {
-          rule.reportPubLint(dep.name!);
-          return;
+        final text = name.text;
+        if (text != null) {
+          if (text.compareTo(previousName) < 0) {
+            rule.reportPubLint(name);
+            return;
+          }
+          previousName = text;
         }
-        previousName = name;
       }
     }
   }
--- a/lib/src/rules/tighten_type_of_initializing_formals.dart
+++ b/lib/src/rules/tighten_type_of_initializing_formals.dart
@@ -77,13 +77,17 @@ class _Visitor extends SimpleAstVisitor<void> {
             : e.leftOperand is NullLiteral
                 ? e.rightOperand
                 : null)
-        .where((e) => e != null && context.typeSystem.isNullable(e.staticType!))
         .whereType<Identifier>()
+        .where((e) {
+          var staticType = e.staticType;
+          return staticType != null &&
+              context.typeSystem.isNullable(staticType);
+        })
         .map((e) => e.staticElement)
         .whereType<FieldFormalParameterElement>()
         .forEach((e) {
-      rule.reportLint(
-          node.parameters.parameters.firstWhere((p) => p.declaredElement == e));
-    });
+          rule.reportLint(node.parameters.parameters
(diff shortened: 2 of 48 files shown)


lib/src/rules/pub/sort_pub_dependencies.dart
```diff
@@ -47,16 +48,25 @@ class Visitor extends PubspecVisitor<void> {
...
     final depsByLocation = dependencies.toList()
-      ..sort((d1, d2) => d1.name!.span.start.compareTo(d2.name!.span.start));
+      ..sort((d1, d2) => compare(d1.name?.span.start, d2.name?.span.start));
     var previousName = '';
     for (final dep in depsByLocation) {
-      final name = dep.name!.text!;
-      if (name.compareTo(previousName) < 0) {
-        rule.reportPubLint(dep.name!);
-        return;
+      final name = dep.name?.text;
```
bwilkerson, line 62: Consider rewriting this to
```
final name = dep.name;
if (name != null) {
  final text = name.text;
```

lib/src/rules/tighten_type_of_initializing_formals.dart
```diff
@@ -77,8 +77,7 @@ class _Visitor extends SimpleAstVisitor<void> {
             : e.leftOperand is NullLiteral
                 ? e.rightOperand
                 : null)
-        .where((e) => e != null)
-        .where((e) => context.typeSystem.isNullable(e!.staticType!))
+        .where((e) => e != null && context.typeSystem.isNullable(e.staticType!))
```
bwilkerson, line 80: I'll note that the `.whereType<Identifier>` would also guarantee that the remaining elements are not null, so you could eliminate the null check entirely by reversing (or combining) this line and the next.

lib/src/rules/use_late_for_private_fields_and_variables.dart
```diff
@@ -163,10 +167,12 @@ class _Visitor extends UnifyingAstVisitor<void> {
...
       return;
     }
-    if (context.typeSystem.isNonNullable(variable.declaredElement!.type)) {
+    var declaredElement = variable.declaredElement;
+    if (declaredElement == null ||
+        context.typeSystem.isNonNullable(declaredElement.type)) {
       return;
     }
-    lateables[currentUnit]!.add(variable);
+    lateables[currentUnit]?.add(variable);
```
bwilkerson, line 175: No change required, just some musings: This one feels a little odd, like we're saying "if we're keeping data about this key, then add the new data". On the other hand the only alternative I've thought of (`lateables.putIfAbsent(currentUnit, () => ...).add(variable)`) doesn't seem great either because we don't expect to ever need to invoke the closure.

parlough:
Something that is not super consistent with these changes and elsewhere in null_safety migrations is the difference between code similar to the following:

```dart
var person = group.leader;
return person != null && person.isCool;
```

versus

```dart
var person = group.leader;
return person?.isCool == true;
```

I'm not sure which I prefer, but I wonder if there should be a suggestion as part of `effective dart` for this situation and we should add a potential lint. I'm assuming the prefer would be the first one, as while it is not as concise, it is a bit more clear in intentions. Not super related to this PR, but thought I should inquire about it. Perhaps @munificent has some opinions around it.

lib/src/util/condition_scope_visitor.dart
```diff
@@ -129,10 +129,10 @@ abstract class ConditionScopeVisitor extends RecursiveAstVisitor {
   ConditionScope? outerScope;
   final breakScope = BreakScope();
 
-  Iterable<Expression?> getFalseExpressions(Iterable<Element?> elements) =>
+  Iterable<Expression?>? getFalseExpressions(Iterable<Element?> elements) =>
```
bwilkerson, line 132: Would it be reasonable to return an empty `Iterable` rather than `null` here?

lib/src/util/condition_scope_visitor.dart
```diff
@@ -315,9 +317,9 @@ abstract class ConditionScopeVisitor extends RecursiveAstVisitor {
     });
   }
 
-  Iterable<Expression?> _getExpressions(Iterable<Element?> elements,
+  Iterable<Expression?>? _getExpressions(Iterable<Element?> elements,
           {bool value = true}) =>
-      outerScope!.getExpressions(elements, value: value);
+      outerScope?.getExpressions(elements, value: value);
```
bwilkerson, line 322: Maybe by adding `?? []` here?

bwilkerson (approved):
None of my comments are required, so take them or leave them.

pq:
@parlough: thanks for noticing this.  I almost wrote a note about it.  I admit I feel pretty uneasy about code like `foo?.isBar == true` but sometimes the alternative is even more cumbersome.  I'm not sure if there's a once-size-fits-all solution but would be happy if there were one if only so I didn't have to hem and haw each time.

lib/src/util/condition_scope_visitor.dart
parlough: I think this would be a bit cleaner but with the current code flow would end up with some minor unnecessary objects created and logic running. Both ways seem fine, but perhaps this could be addressed afterwards.

parlough (approved):

a14n:
> return person?.isCool == true;

The recommendation from [Effective dart](https://dart.dev/guides/language/effective-dart/usage#do-use--to-convert-null-to-a-boolean-value) is: 

> return person?.isCool ?? false;

Do we have a lint for that?

pq:
@a14n: I had missed that entirely.  I appreciate the rationale too.  We should absolutely lint for it though IMO.  

https://github.com/dart-lang/linter/issues/2460

lib/src/rules/tighten_type_of_initializing_formals.dart
pq: Good catch.  I also didn't see the dangling `!`.  Fixing.  Thank you!

lib/src/rules/use_late_for_private_fields_and_variables.dart
pq: 100% agreed.  @a14n: I think this is making a mess of your code a bit.  Maybe after the dust settles you could take a pass?  (Or I could.)  Sorry for that.

lib/src/util/condition_scope_visitor.dart
pq: Exactly.  When I mentioned in the description above that there were a few places where I stopped short of a a bigger redesign this is one of them.  It would be great if we did a sweep at some point.  The are a bunch of places where we're returning nullable values where we could do a bit better.  (A nullable collection feels like an anti-pattern if you ask me.)

pq: I'll add a TODO.

bwilkerson: There are times when there's a legitimate need to distinguish between an empty collection and `null`, but I agree that when that's not the case it's better to return an empty collection than `null`.

PR #2461, opened 2021-02-16
NNBD linting

pq:
Cheers for these awesome lints @a14n!

Follow-up from https://github.com/dart-lang/linter/pull/2457.

/cc @bwilkerson

### DIFF
--- a/test/integration_test.dart
+++ b/test/integration_test.dart
@@ -128,6 +128,7 @@ void coreTests() {
 
         final options = _getOptionsFromString(src);
         var configuredLints =
+            // ignore: cast_nullable_to_non_nullable
             (options['linter'] as YamlMap)['rules'] as YamlList;
 
         // rules are sorted
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -42,6 +42,7 @@ linter:
     - camel_case_types
     - cancel_subscriptions
 #    - cascade_invocations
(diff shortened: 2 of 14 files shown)


test/integration_test.dart
```diff
@@ -128,6 +128,7 @@ void coreTests() {
 
         final options = _getOptionsFromString(src);
         var configuredLints =
+            // ignore: cast_nullable_to_non_nullable
             (options['linter'] as YamlMap)['rules'] as YamlList;
```
bwilkerson, line 132: This seems like a false positive to me. IIRC, `options` is a `Map<String, dynamic>`. While it's true that `dynamic` is nullable, I think it's reasonable to cast it to either a nullable or non-nullable type.

bwilkerson (approved):

bwilkerson:
Only one of those lints has a quick fix associated with it. Out of curiosity, did you notice any patterns while you were fixing them that would be reasonable to offer as a quick fix?

pq:
> Only one of those lints has a quick fix associated with it. Out of curiosity, did you notice any patterns while you were fixing them that would be reasonable to offer as a quick fix?

Removing an unnecessary `!` seems like an easy one and would be handy as a bulk fix.  The others are trickier.

test/integration_test.dart
pq: fyi @a14n

PR #2464, opened 2021-02-17
Update required named parameters rule for new required keyword

parlough:
This updates the test to test the null safety feature with the `required` keyword and instead focuses on the `require` keyword in the rule documentation as that is what users should be using after the migration.


bwilkerson (approved):

pq (approved):
Nice.  Thanks!

PR #2465, opened 2021-02-17
upgrade to nullable http and github

pq:
Updates to new `http` APIs and nullable `github`.

Definitely room for cleaning up but this gets us a lot of the way there.

/cc @bwilkerson

### DIFF
--- a/tool/crawl.dart
+++ b/tool/crawl.dart
@@ -2,8 +2,6 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-// ignore_for_file: import_of_legacy_library_into_null_safe
-
 import 'dart:async';
 import 'dart:io';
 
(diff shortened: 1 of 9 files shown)


tool/crawl.dart
```diff
@@ -188,7 +191,11 @@ Future<List<String>> _fetchSdkTags() {
   return github.repositories
       .listTags(slug)
       .map((t) => t.name)
+      .where((t) => t != null)
       .where((t) {
+        if (t == null) {
```
bwilkerson, line 196: It seems like this test and the test on line 194 are doing the same thing, but maybe I'm missing something.

bwilkerson (approved):

tool/crawl.dart
pq: Indeed, thanks. The `.where((t) => t != null)` idea didn't pan out (since I still get a stream of nullable Strings) but I forgot to remove it.  👍

pq: (There's no `whereType<T>()` on streams?)

PR #2466, opened 2021-02-17
nullable list cleanup

pq:
/cc @bwilkerson

### DIFF
--- a/lib/src/util/score_utils.dart
+++ b/lib/src/util/score_utils.dart
@@ -10,9 +10,9 @@ final _pedanticOptionsRootUrl =
 final _pedanticOptionsUrl =
     _pedanticOptionsRootUrl.resolve('analysis_options.yaml');
 
-List<String?>? _pedanticRules;
+List<String>? _pedanticRules;
 
-Future<List<String?>> get pedanticRules async =>
+Future<List<String>> get pedanticRules async =>
     _pedanticRules ??= await _fetchPedanticRules();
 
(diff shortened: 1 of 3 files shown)


lib/src/util/score_utils.dart
```diff
@@ -10,9 +10,9 @@ final _pedanticOptionsRootUrl =
 final _pedanticOptionsUrl =
     _pedanticOptionsRootUrl.resolve('analysis_options.yaml');
 
-List<String?>? _pedanticRules;
+List<String>? _pedanticRules;
```
bwilkerson, line 13: One possible approach for collections like this is to initialize the collection to be empty, then use `isEmpty` rather than `== null` as the trigger for filling the collection. Just a possibility for a future PR.

bwilkerson (approved):

lib/src/util/score_utils.dart
pq: Interesting idea.  In this case that'd work since it should never be legitimately empty.

parlough: 👍 Make sure to use `const []` if possible.

bwilkerson: If we use `const []` then we have to create a new list for the actual values, so I don't see that as a big win, but I'm also not too concerned about it one way or the other.

parlough: Ah yeah I suppose it depends on if it's not a more complex path where you plan to add to it after built or over time.

PR #2467, opened 2021-02-17
soundness fixes

pq:
/cc @bwilkerson

### DIFF
--- a/lib/src/rules/prefer_single_quotes.dart
+++ b/lib/src/rules/prefer_single_quotes.dart
@@ -141,14 +141,14 @@ class _ImmediateChildrenVisitor extends UnifyingAstVisitor {
 /// its children.
 class _IsOrContainsStringVisitor extends UnifyingAstVisitor<bool> {
   /// Different way to express `accept` in a way that's clearer in this visitor.
-  bool? isOrContainsString(AstNode node) => node.accept(this);
+  bool isOrContainsString(AstNode node) => node.accept(this) ?? false;
 
   /// Scan as little of the tree as possible, by bailing out on first match. For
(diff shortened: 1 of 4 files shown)


lib/src/rules/prefer_single_quotes.dart
```diff
@@ -141,14 +141,14 @@ class _ImmediateChildrenVisitor extends UnifyingAstVisitor {
...
-  bool? isOrContainsString(AstNode node) => node.accept(this);
+  bool isOrContainsString(AstNode node) => node.accept(this) ?? false;
 
   /// Scan as little of the tree as possible, by bailing out on first match. For
   /// all leaf nodes, they will either have a method defined here and return
   /// true, or they will return false because leaves have no children.
   @override
-  bool visitNode(AstNode node) => _ImmediateChildrenVisitor.childrenOf(node)
-      .any(isOrContainsString as bool Function(AstNode));
+  bool visitNode(AstNode node) =>
```
pq, line 150: I'm pretty sure this unsafe cast was produced by the migration tool.

/fyi @srawlins

srawlins: Yes, there is a known bug for these unnecessary function casts

pq: Awesome.  Thanks!

pq:
The coverage dip is really curious and feels wrong.

/fyi @kevmoo @grouma in case they've seen any oddness w/ sound-mode and coverage reporting

bwilkerson (approved):

parlough (approved):

PR #2474, opened 2021-02-22
cast cleanup

pq:
/cc @bwilkerson

### DIFF
--- a/lib/src/util/tested_expressions.dart
+++ b/lib/src/util/tested_expressions.dart
@@ -88,9 +88,10 @@ class TestedExpressions {
       return _contradictions;
     }
 
-    final binaryExpression = testingExpression is BinaryExpression
-        ? testingExpression as BinaryExpression
-        : null;
+    var testingExpression = this.testingExpression;
+
+    final binaryExpression =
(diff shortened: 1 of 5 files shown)


lib/src/util/tested_expressions.dart
```diff
@@ -88,9 +88,10 @@ class TestedExpressions {
       return _contradictions;
     }
 
-    final binaryExpression = testingExpression is BinaryExpression
-        ? testingExpression as BinaryExpression
-        : null;
+    var testingExpression = this.testingExpression;
+
+    final binaryExpression =
+        testingExpression is BinaryExpression ? testingExpression : null;
```
bwilkerson, line 94: Consider writing this as
```
var facts;
if (testingExpression is BinaryExpression) {
  facts = [testingExpression.leftOperand, testingExpression.rightOperand];
} else {
  facts = [testingExpression];
}
```
I think it's a little easier to follow the logic.

bwilkerson (approved):

lib/src/util/tested_expressions.dart
pq: Nice one.  Agreed. 👍

parlough (approved):

PR #2480, opened 2021-02-25
add use_if_null_to_convert_null_to_bool

### LINKED ISSUE #2460: new lint: `if_null_boolean_conversions` (use ?? to convert null to a boolean value.)
I'm sure someone can come up with a better name but the rationale is in https://github.com/dart-lang/linter/pull/2458#issuecomment-780088506.  The advice from effective Dart is 

**DO use ?? to convert null to a boolean value.**

See: https://dart.dev/guides/language/effective-dart/usage#do-use--to-convert-null-to-a-boolean-value

Cheers @a14n for the suggestion 🍻 .

a14n:
# Description

Use if-null operator to convert null to bool.

**BAD:**
```
if (nullableBool == true) {
}
```

**GOOD:**
```
if (nullableBool ?? false) {
}
```

Fixes #2460

### DIFF
--- a/lib/src/rules/use_if_null_to_convert_null_to_bool.dart
+++ b/lib/src/rules/use_if_null_to_convert_null_to_bool.dart
@@ -6,6 +6,7 @@ import 'package:analyzer/dart/analysis/features.dart';
 import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/token.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/type.dart';
 
 import '../analyzer.dart';
 
@@ -19,11 +20,15 @@ Use if-null operator to convert null to bool.
 ```
 if (nullableBool == true) {
 }
+if (nullableBool != false) {
+}
 ```
 
 **GOOD:**
 ```
 if (nullableBool ?? false) {
+}
+if (nullableBool ?? true) {
 }
 ```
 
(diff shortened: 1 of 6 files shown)


bwilkerson (commented):
Do we want to catch cases where the operands are reversed (`true == e`)?
Do we want to catch uses of `!=` (`e != false`)?

a14n:
> Do we want to catch cases where the operands are reversed (`true == e`)?

I don't want to bother yoda conditions lovers :)  (this could be another lint BTW - `avoid_yoda_conditions`)

> Do we want to catch uses of `!=` (`e != false`)?

I will add some code to catch that.

bwilkerson (approved):

lib/src/rules/use_if_null_to_convert_null_to_bool.dart
```diff
@@ -0,0 +1,84 @@
...
+
+import 'package:analyzer/dart/analysis/features.dart';
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/token.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/type.dart';
+
+import '../analyzer.dart';
+
+const _desc = r'Use if-null operator to convert null to bool.';
```
pq, line 13: =>

'Use if-null operators to convert nulls to bools.';

(We standardized on plurals -- mostly 😬 )

lib/src/rules/use_if_null_to_convert_null_to_bool.dart
```diff
@@ -0,0 +1,84 @@
...
+
+const _details = r'''
+
+Use if-null operator to convert null to bool.
```
pq, line 17: =>

Use if-null operators to convert nulls to bools.

pq (commented):
Consider renaming to `use_if_null_to_convert_nulls_to_bools`?

I feel like we've mostly pluralized.  For example:

`use_key_in_widget_constructors`
`use_late_for_private_fields_and_variables`
`use_setters_to_change_properties`

etc.

pq:
PS: this is super awesome!

pq (approved):
Looks great!

pq:
Thanks!

PR #2481, opened 2021-02-25
update to analyzer 1.1.0 apis

pq:
/cc @bwilkerson @scheglov

### DIFF
--- a/lib/src/ast.dart
+++ b/lib/src/ast.dart
@@ -233,15 +233,15 @@ File? locatePubspecFile(CompilationUnit compilationUnit) {
   }
 
   final file = resourceProvider.getFile(fullName);
-  var folder = file.parent;
+  var folder = file.parent2;
 
   // Look for a pubspec.yaml file.
-  while (folder != null) {
+  while (!folder.isRoot) {
     final pubspecFile = folder.getChildAssumingFile('pubspec.yaml');
     if (pubspecFile.exists) {
       return pubspecFile;
     }
-    folder = folder.parent;
+    folder = folder.parent2;
   }
 
   return null;
--- a/lib/src/ast.dart
+++ b/lib/src/ast.dart
@@ -233,15 +233,13 @@ File? locatePubspecFile(CompilationUnit compilationUnit) {
   }
 
   final file = resourceProvider.getFile(fullName);
-  var folder = file.parent2;
 
   // Look for a pubspec.yaml file.
-  while (!folder.isRoot) {
+  for (var folder in file.parent2.withAncestors) {
     final pubspecFile = folder.getChildAssumingFile('pubspec.yaml');
     if (pubspecFile.exists) {
       return pubspecFile;
     }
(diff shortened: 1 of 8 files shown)


lib/src/ast.dart
```diff
@@ -233,15 +233,15 @@ File? locatePubspecFile(CompilationUnit compilationUnit) {
   }
 
   final file = resourceProvider.getFile(fullName);
-  var folder = file.parent;
+  var folder = file.parent2;
 
   // Look for a pubspec.yaml file.
-  while (folder != null) {
+  while (!folder.isRoot) {
```
scheglov, line 239: You could use `FolderExtension.withAncestors` here, e.g.
```dart
for (var current in file.parent2.withAncestors) {
  ...
}
```

lib/src/rules/avoid_returning_null_for_future.dart
```diff
@@ -63,17 +63,17 @@ class _Visitor extends SimpleAstVisitor<void> {
     if (parent == null) return;
 
     DartType? returnType;
```
scheglov, line 65: The `declaredElement` must be always set (to `ExecutableElement`) in resolved units, so the return type is never `null`.

lib/src/rules/avoid_returning_null_for_future.dart
```diff
@@ -63,17 +63,17 @@ class _Visitor extends SimpleAstVisitor<void> {
...
-    bool? isAsync;
+    bool isAsync;
     if (parent is FunctionExpression) {
       returnType = parent.declaredElement?.returnType;
-      isAsync = parent.body?.isAsynchronous;
+      isAsync = parent.body.isAsynchronous;
```
scheglov, line 69: With the change to `declaredElement` you could get `isAsynchronous` from `ExecutableElement`.

lib/src/rules/avoid_returning_null_for_void.dart
```diff
@@ -89,14 +89,14 @@ class _Visitor extends SimpleAstVisitor<void> {
     bool? isAsync;
     if (parent is FunctionExpression) {
       type = parent.declaredElement?.returnType;
-      isAsync = parent.body?.isAsynchronous;
+      isAsync = parent.body.isAsynchronous;
```
scheglov, line 92: Looks like a dup of the above.
Consider adding a helper to get the `ExecutableElement` and use in both places to access the return type and `isAsynchronous`.

scheglov (approved):

lib/src/ast.dart
pq: Nice!

lib/src/rules/avoid_returning_null_for_future.dart
pq: Are you suggesting `!`s below?:

    returnType = parent.declaredElement!.returnType;

pq:
Landing and can follow up with more tweaks.

PR #2483, opened 2021-02-25
null check cleanup

pq:
A little inelegant but trying to winnow down the `!`s.

/cc @a14n @bwilkerson

### DIFF
--- a/lib/src/rules/use_late_for_private_fields_and_variables.dart
+++ b/lib/src/rules/use_late_for_private_fields_and_variables.dart
@@ -7,6 +7,7 @@ import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/token.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
 import 'package:analyzer/dart/element/element.dart';
+import 'package:analyzer/dart/element/type.dart';
 import 'package:analyzer/error/error.dart';
 
 import '../analyzer.dart';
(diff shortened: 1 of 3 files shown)


lib/src/rules/use_late_for_private_fields_and_variables.dart
```diff
@@ -177,9 +181,9 @@ class _Visitor extends UnifyingAstVisitor<void> {
 
   void _checkAccess(Iterable<CompilationUnitElement> units) {
     final allNullableAccess =
-        units.expand((unit) => nullableAccess[unit]!).toSet();
+        units.expand((unit) => nullableAccess[unit] ?? {}).toSet();
```
bwilkerson, line 184: Consider making empty collections that are never stored (and hence never added to) be `const`.

lib/src/rules/use_late_for_private_fields_and_variables.dart
```diff
@@ -111,6 +112,9 @@ class _Visitor extends UnifyingAstVisitor<void> {
       element = DartTypeUtilities.getCanonicalElementFromIdentifier(node);
     }
 
+    bool isNonNullable(DartType? type) =>
```
bwilkerson, line 115: Seems like this would be useful to have in the utilities library.

bwilkerson (approved):

lib/src/rules/use_late_for_private_fields_and_variables.dart
pq: Good catch!

lib/src/rules/use_late_for_private_fields_and_variables.dart
pq: Good idea!

PR #2500, opened 2021-03-08
rename score.yaml to core.yaml

devoncarew:
- rename `score.yaml` to `core.yaml`

@pq


pq:
I like this!

@munificent @mit-mit : how do you feel about generally renaming the "score" set to "core"?

devoncarew:
Before this rename, it wasn't immediately obvious to me the relationship between the 'score' lints and the 'recommended' lints.

mit-mit:
SGTM

pq (approved):
👍

munificent:
I don't have a super strong opinion, but for what it's worth, I slightly prefer "score" because then the name clearly states what it means. "Core" is pretty vague.

PR #2510, opened 2021-03-16
stop AST manipulation

pq:
Some work to move away from deprecated APIs.

@scheglov: could you advise on the best approach for `unnecessary_const`?

/cc @scheglov

### DIFF
--- a/lib/src/rules/unnecessary_const.dart
+++ b/lib/src/rules/unnecessary_const.dart
@@ -60,11 +60,10 @@ class _Visitor extends SimpleAstVisitor {
 
   @override
   void visitInstanceCreationExpression(InstanceCreationExpression node) {
-    if (node.keyword == null) return;
+    if (node.keyword?.type != Keyword.CONST) return;
 
-    // todo (pq): remove cast when `inConstantContext` is public on Element
-    var inConstContext = (node as ExpressionImpl).inConstantContext;
-    if (inConstContext && node.keyword?.type == Keyword.CONST) {
+    // todo (pq): remove cast when `inConstantContext` is public on Expression
+    if ((node as ExpressionImpl).inConstantContext) {
       rule.reportLint(node);
     }
   }
@@ -78,11 +77,10 @@ class _Visitor extends SimpleAstVisitor {
   }
 
   void _visitTypedLiteral(TypedLiteral node) {
-    if (node.constKeyword == null) return;
+    if (node.constKeyword?.type != Keyword.CONST) return;
 
-    // todo (pq): remove cast when `inConstantContext` is public on Element
-    var inConstContext = (node as ExpressionImpl).inConstantContext;
-    if (inConstContext && node.constKeyword?.type == Keyword.CONST) {
+    // todo (pq): remove cast when `inConstantContext` is public on Expression
+    if ((node as ExpressionImpl).inConstantContext) {
       rule.reportLint(node);
     }
   }
--- a/lib/src/rules/unnecessary_const.dart
+++ b/lib/src/rules/unnecessary_const.dart
@@ -62,7 +62,7 @@ class _Visitor extends SimpleAstVisitor {
   void visitInstanceCreationExpression(InstanceCreationExpression node) {
     if (node.keyword == null) return;
 
-    // todo (pq): move to a new LinterContext API
+    // todo (pq): remove cast when `inConstantContext` is public on Element
     var inConstContext = (node as ExpressionImpl).inConstantContext;
     if (inConstContext && node.keyword?.type == Keyword.CONST) {
       rule.reportLint(node);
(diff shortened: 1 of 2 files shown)


lib/src/rules/unnecessary_const.dart
```diff
@@ -82,6 +82,8 @@ class _Visitor extends SimpleAstVisitor {
   void _visitTypedLiteral(TypedLiteral node) {
     if (node.constKeyword == null) return;
 
+    // todo (pq): update to use linter context API
```
bwilkerson, line 85: I think you should be able to ask `node.inConstantContext`.

scheglov: Yes, it looks that it will work.
One complication is that it declared in `ExpressionImpl`.

bwilkerson: Should we make it public? Do we need to wrap it in `LintContext`?

scheglov: We don't need manipulations with AST to implement it, so I think we can make it public.

Do you like its name? Should it be `isInConstantContext`? OTOH we also have `inGetterContext`, so it seems consistent. Just don't know if consistently good, or consistently bad :-)

pq: Thanks!  Seems like this could go in `LinterContext`?

pq:
@scheglov, @bwilkerson: not pretty, but better?

lib/src/rules/unnecessary_const.dart
pq: (whoops.  missed your earlier comment.)

> We don't need manipulations with AST to implement it, so I think we can make it public.

I can't speak to consistency.  Either work for me.  I guess I'd lean towards `inConstantContext`.  If nothing else, it's shorter!

lib/src/rules/unnecessary_const.dart
```diff
@@ -60,13 +62,9 @@ class _Visitor extends SimpleAstVisitor {
...
-    // remove keyword and check if there's const error
-    final oldKeyword = node.keyword;
-    node.keyword = null;
-    final isConstWithoutKeyword = node.isConst;
-    node.keyword = oldKeyword;
-
-    if (isConstWithoutKeyword && node.keyword?.type == Keyword.CONST) {
+    // todo (pq): remove cast when `inConstantContext` is public on Element
+    var inConstContext = (node as ExpressionImpl).inConstantContext;
+    if (inConstContext && node.keyword?.type == Keyword.CONST) {
```
scheglov, line 67: I would have switched the order, and checked `node.keyword?.type == Keyword.CONST` first (as you have in the first statement of this method). It is cheaper.

lib/src/rules/unnecessary_const.dart
```diff
@@ -82,13 +80,9 @@ class _Visitor extends SimpleAstVisitor {
...
-    // remove keyword and check if there's const error
-    final oldKeyword = node.constKeyword;
-    node.constKeyword = null;
-    final isConstWithoutKeyword = node.isConst;
-    node.constKeyword = oldKeyword;
-
-    if (isConstWithoutKeyword && node.constKeyword?.type == Keyword.CONST) {
+    // todo (pq): remove cast when `inConstantContext` is public on Element
+    var inConstContext = (node as ExpressionImpl).inConstantContext;
+    if (inConstContext && node.constKeyword?.type == Keyword.CONST) {
```
scheglov, line 85: Can `constKeyword` be anything else?

lib/src/rules/unnecessary_const.dart
```diff
@@ -60,13 +62,9 @@ class _Visitor extends SimpleAstVisitor {
...
     if (node.keyword == null) return;
 
-    // remove keyword and check if there's const error
-    final oldKeyword = node.keyword;
-    node.keyword = null;
-    final isConstWithoutKeyword = node.isConst;
-    node.keyword = oldKeyword;
-
-    if (isConstWithoutKeyword && node.keyword?.type == Keyword.CONST) {
+    // todo (pq): remove cast when `inConstantContext` is public on Element
```
bwilkerson, line 65: "Element" --> "Expression"

scheglov (approved):

lib/src/rules/unnecessary_const.dart
bwilkerson, line 67: Consider moving `node.keyword?.type == Keyword.CONST` to line 63. It's a less expensive check so it should be done first, and there's really no reason to check that the keyword is non-null in t addition to checking that it is the right keyword.

bwilkerson (approved):
> Should it be `isInConstantContext`?

No, I don't think the "is" adds any value in this case, and it's more consistent.

(Not sure why, but I can't seem to reply to that question in-line.)

lib/src/rules/unnecessary_const.dart
bwilkerson: There's already a check on line 81.

lib/src/rules/unnecessary_const.dart
pq: Good catch.  Konstantin had a similar suggestion. Updated.

PR #2511, opened 2021-03-16
extract a flutter utility method

pq:
/cc @bwilkerson

### DIFF
--- a/lib/src/util/flutter_utils.dart
+++ b/lib/src/util/flutter_utils.dart
@@ -69,10 +69,9 @@ class _Flutter {
   bool hasWidgetAsAscendant(ClassElement? element,
       [Set<ClassElement>? alreadySeen]) {
     alreadySeen ??= {};
-    if (element == null || alreadySeen.contains(element)) {
+    if (element == null || !alreadySeen.add(element)) {
       return false;
     }
-    alreadySeen.add(element);
     if (_isExactWidget(element, _nameWidget, _uriFramework)) {
       return true;
     }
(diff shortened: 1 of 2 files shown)


lib/src/util/flutter_utils.dart
```diff
@@ -61,13 +66,38 @@ class _Flutter {
         _uriContainer = Uri.parse('$uriPrefix/src/widgets/container.dart'),
         _uriFramework = Uri.parse('$uriPrefix/src/widgets/framework.dart');
 
-  bool isExactWidgetTypeContainer(DartType? type) =>
-      type is InterfaceType &&
-      _isExactWidget(type.element, _nameContainer, _uriContainer);
+  bool hasWidgetAsAscendant(ClassElement? element,
+      [Set<ClassElement>? alreadySeen]) {
+    alreadySeen ??= {};
+    if (element == null || alreadySeen.contains(element)) {
```
bwilkerson, line 72: `Set.add` returns `true` if the set was modified, so consider using `!add` in place of `contains`. It's more performant that way.

lib/src/util/flutter_utils.dart
```diff
@@ -61,13 +66,38 @@ class _Flutter {
...
+  }
+
+  bool isBuildContext(DartType? type, {bool skipNullable = false}) {
+    if (type is! InterfaceType) {
+      return false;
+    }
+    if (skipNullable && type.nullabilitySuffix == NullabilitySuffix.question) {
+      return false;
+    }
+    var element = type.element;
```
bwilkerson, line 89: Consider using `_isExactWidget` here.

bwilkerson (approved):

lib/src/util/flutter_utils.dart
pq: Oh, thanks!  I forgot I'd inlined that.

lib/src/util/flutter_utils.dart
pq: Nice one.  Thanks!

PR #2517, opened 2021-03-16
Remove avoid_private_typedef_functions from recommend.yaml

goderbauer:
As discussed in our meeting today.

In the future: could be replaced with https://github.com/dart-lang/linter/issues/2519.

/cc @pq @Hixie @mit-mit @munificent @devoncarew


pq (approved):
👍

pq:
The failing bot has been disabled (w/ #2518) but is not getting removed from this build so the error can safely be ignored.

Issue #2509, opened 2021-03-15
[linter] Support for non-function type aliases


pq:
Follow-up from https://github.com/dart-lang/sdk/issues/44078 and https://github.com/dart-lang/sdk/issues/44951, we need to shore up linter support for non-function type aliases.

- [x] consider existing lints 
  - [x] ~`avoid_private_typedef_functions`~ https://github.com/dart-lang/linter/issues/2535
  - [x] `camel_case_types`
  - [x] `prefer_mixin`
  - [x] `public_member_api_docs`
  - [x] `slash_for_doc_comments`
- [x] add test cases
- [x] consider new lints

pq:
I did a quick look through existing lints and besides the fix to `prefer_mixin` didn't come up with anything else to address.  I didn't feel a strong pull to add test cases either though I'm super open to suggestions.

@srawlins, @bwilkerson, @scheglov : curious if anything jumps out at you?

As for new lints, I'm curious if there's any thinking about naming conventions.  CamelCase?  If yes, then thoughts on extending [camel_case_types](https://dart-lang.github.io/linter/lints/camel_case_types.html)?  Alternatively I guess we could add `camel_case_nonfunction_typedefs` 😬 ...

bwilkerson:
Doesn't `camel_case_types` already cover the names defined via `typedef`? I would expect it to just continue to work, but a test for that case would be in order.

As for other places that might need fixed up, you could search for places where `staticElement` is being invoked on a `TypeAnnotation`, though that's probably going to return lots of places where it's used on something other than a type annotation.

pq:
> Doesn't camel_case_types already cover the names defined via typedef? 

Yes it does cover function type aliases.

The question is whether it should flag the following:

```dart
class A { }

typedef a = A;
```

I think yes?

(Currently it does not.)

Thanks for the tip on `staticElement` greps!

bwilkerson:
Ah, that's because it's visiting `FunctionTypeAlias`s when I think it ought to be visiting `GenericTypeAlias`s (it might need to visit both until the experiment is enabled). So you can also look for other implementations of `visitFunctionTypeAlias`.

scheglov:
FWIW, this means that the lint did not work for "modern" function type aliases as well, like `typedef b = void Function();`.
`FunctionTypeAlias` is not going anywhere, I think, because it is a different syntax than `GenericTypeAlias`.

pq:
[...]

PR #2523, opened 2021-03-17
`prefer_mixin` support for nonfunction-type-aliases

pq:
See: #2509.

/cc @bwilkerson @srawlins


srawlins (commented):
Haha wow, there you go.

test/rules/experiments/nonfunction-type-aliases/rules/prefer_mixin.dart
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) 2018, the Dart project authors. Please see the AUTHORS file
```
bwilkerson, line 1: nit: year

pq: The approach that we've taken in the past is to keep the copyright from the original since the new contents will get copied back to it when the experiment is finished.

bwilkerson (approved):

PR #2530, opened 2021-03-18
add lint function_expression_with_incorrect_types

a14n:
# Description

Use the parameter types expected by the target in function expression.

**BAD:**
```
f(void Function(int a) p) {}

f((int? a) {});
f((num a) {});
```
**GOOD:**
```
f(void Function(int a) p) {}

f((a) {});
f((int a) {});
```

### DIFF
--- a/lib/src/rules/function_expression_with_incorrect_types.dart
+++ b/lib/src/rules/function_expression_with_incorrect_types.dart
@@ -1,0 +1,125 @@
+// Copyright (c) 2021, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/analysis/features.dart';
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/type.dart';
+import 'package:collection/collection.dart';
+
+import '../analyzer.dart';
+
+const _desc =
+    r'Use the parameter types expected by the target in function expression.';
+
+const _details = r'''
+
+Use the parameter types expected by the target in function expression.
+
+**BAD:**
+```
+f(void Function(int a) p) {}
+
+f((int? a) {});
+f((num a) {});
+```
+
+**GOOD:**
+```
+f(void Function(int a) p) {}
+
+f((a) {});
+f((int a) {});
+```
+
+''';
+
+class FunctionExpressionWithIncorrectTypes extends LintRule
+    implements NodeLintRule {
+  FunctionExpressionWithIncorrectTypes()
+      : super(
+            name: 'function_expression_with_incorrect_types',
+            description: _desc,
+            details: _details,
+            group: Group.style);
+
+  @override
+  void registerNodeProcessors(
+      NodeLintRegistry registry, LinterContext context) {
+    var visitor = _Visitor(this, context);
+    registry.addFunctionExpression(this, visitor);
+  }
+}
+
+class _Visitor extends SimpleAstVisitor<void> {
+  _Visitor(this.rule, this.context);
+
+  final LintRule rule;
+  final LinterContext context;
+
+  @override
+  void visitFunctionExpression(FunctionExpression node) {
+    var currentType = node.staticType;
+    if (currentType is! FunctionType) return;
+
+    var parameterElement = node
+        .thisOrAncestorMatching<Expression>(
+            (e) => e.parent is! ParenthesizedExpression)
+        ?.staticParameterElement;
+    if (parameterElement == null) return;
+
+    var expectedType = parameterElement.type;
+    if (expectedType is! FunctionType) return;
+
+    // too hard to check cross nullability
+    var currentNullability = context.isEnabled(Feature.non_nullable);
+    var expectedNullability =
+        (parameterElement.library ?? parameterElement.declaration.library)
+            ?.featureSet
+            .isEnabled(Feature.non_nullable);
+    if (expectedNullability == null ||
+        currentNullability ^ expectedNullability) {
+      return;
+    }
+
+    // check type of each parameters
+    var currentParams = currentType.parameters;
+    var expectedParams = expectedType.parameters;
+    for (var i = 0; i < currentParams.length; i++) {
+      var currentParam = currentParams[i];
+      // if param is mutated, no lint
+      if (node.body.isPotentiallyMutatedInScope(currentParam) ||
+          node.body.isPotentiallyMutatedInClosure(currentParam)) {
+        continue;
+      }
+
+      // if param is not found on target, no lint
+      var expectedParam = currentParam.isPositional
+          ? expectedParams[i]
+          : expectedParams
+              .where((e) => e.name == currentParam.name)
+              .firstOrNull;
+      if (expectedParam == null) {
+        continue;
+      }
+
+      var typeSystem = context.typeSystem;
+      var currentType = currentParam.type;
+      var expectedType = expectedParam.type;
+      // if type are not consistent, lint!
+      if (currentParam.isRequiredPositional && currentType != expectedType ||
+          currentParam.isOptional &&
+              (currentNullability &&
+                      (typeSystem.isNonNullable(expectedType) &&
+                              expectedType !=
+                                  typeSystem.promoteToNonNull(currentType) ||
+                          typeSystem.isNullable(expectedType) &&
+                              expectedType != currentType) ||
+                  !currentNullability && expectedType != currentType)) {
+        rule.reportLint(node.parameters!.parameters
+            .firstWhere((e) => e.declaredElement == currentParam));
+      }
+    }
+  }
+}
--- a/example/all.yaml
+++ b/example/all.yaml
@@ -69,6 +69,7 @@ linter:
     - exhaustive_cases
     - file_names
     - flutter_style_todos
+    - function_expression_with_incorrect_types
     - hash_and_equals
     - implementation_imports
     - invariant_booleans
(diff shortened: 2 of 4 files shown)


lib/src/rules/function_expression_with_incorrect_types.dart
```diff
@@ -0,0 +1,94 @@
...
+class _Visitor extends SimpleAstVisitor<void> {
+  final LintRule rule;
+
+  _Visitor(this.rule);
+
+  @override
+  void visitFunctionExpression(FunctionExpression node) {
+    DartType? expectedType;
+    var parent = node.unParenthesized.parent;
+    if (parent is ArgumentList) {
```
bwilkerson, line 65: Use `node.unParenthesized.staticParameterElement` to get the element corresponding to the argument. This will correctly handle named arguments as well as arguments not inside an argument list.

lib/src/rules/function_expression_with_incorrect_types.dart
```diff
@@ -0,0 +1,94 @@
...
+      }
+    } else if (parent is AssignmentExpression) {
+      expectedType = parent.writeType;
+    }
+
+    // exit if the case is not covered
+    if (expectedType == null || expectedType is! FunctionType) return;
+
+    // check type of each parameters
+    var currentType = node.staticType! as FunctionType;
```
bwilkerson, line 82: Not sure what value there is in null checking the `staticType` before casting to a non-nullable type. It just changes the kind of exception that will be thrown.

lib/src/rules/function_expression_with_incorrect_types.dart
```diff
@@ -0,0 +1,94 @@
...
+    var currentParams = currentType.parameters;
+    var expectedParams = expectedType.parameters;
+    for (var i = 0; i < min(expectedParams.length, currentParams.length); i++) {
```
bwilkerson, line 85: I'm guessing that this won't work for named parameters if the parameters are defined in different orders or if the closure allows named parameters that aren't in the expected type. For example
```
void g(void Function({String? s, int? i}) f) {}
void h() {
  g(({int? i, String? s}) {});
}
```

bwilkerson (commented):
I have to wonder how much value this adds, but I assume you've seen this pattern in a few places or you wouldn't have thought of the rule. Can you give me an idea of how common a violation of this lint would be?

lib/src/rules/function_expression_with_incorrect_types.dart
a14n: Without null check [cast_nullable_to_non_nullable](https://dart-lang.github.io/linter/lints/cast_nullable_to_non_nullable.html) triggers.

lib/src/rules/function_expression_with_incorrect_types.dart
a14n: Fixed.

lib/src/rules/function_expression_with_incorrect_types.dart
a14n: Done

a14n:
>I have to wonder how much value this adds, but I assume you've seen this pattern in a few places or you wouldn't have thought of the rule. Can you give me an idea of how common a violation of this lint would be?

This lint was done after https://github.com/flutter/flutter/pull/78143#pullrequestreview-614805784. I expected several occurance of this lint in Flutter where types are required everywhere but surprisingly it didn't come up a lot.

lib/src/rules/function_expression_with_incorrect_types.dart
a14n: Not fixed at all as names for positionnal parameters can change... Sorry

lib/src/rules/function_expression_with_incorrect_types.dart
bwilkerson: That just points out that I didn't think this through carefully enough yesterday. The style we're adopting on the analyzer team is to avoid runtime exceptions whenever possible now that we have sound typing. So I'm going to request that we have neither the null check nor the cast by doing something like
```
var currentType = node.staticType;
if (currentType is! FunctionType) {
  return;
}
```

lib/src/rules/function_expression_with_incorrect_types.dart
```diff
@@ -0,0 +1,110 @@
...
+
+class _Visitor extends SimpleAstVisitor<void> {
+  final LintRule rule;
+
+  _Visitor(this.rule);
+
+  @override
+  void visitFunctionExpression(FunctionExpression node) {
+    DartType? expectedType;
+    var parent = node.unParenthesized.parent;
```
bwilkerson, line 63: I just double checked and I don't think `unParenthesized` is what you want either here or below; it will return either the node or a child of the node. I think you actually want the inverse, which doesn't exist (maybe something like `thisOrAncestorNotInParentheses`).

I wonder how often we've made the same mistake in other places.

lib/src/rules/function_expression_with_incorrect_types.dart
```diff
@@ -0,0 +1,110 @@
...
+    for (var currentParam in currentParams) {
+      // if param is mutated, no lint
+      if (node.body.isPotentiallyMutatedInScope(currentParam) ||
+          node.body.isPotentiallyMutatedInClosure(currentParam)) {
+        continue;
+      }
+
+      // if param is not found on target, no lint
+      var expectedParam =
+          expectedParams.where((e) => e.name == currentParam.name).firstOrNull;
```
bwilkerson, line 86: This will work for named parameters because they are required to have the same name, but the names of positional parameters are not important. For example:
```
void f(void Function(int a) p) {}
void m() {
  f((int b) {}); // OK
```

lib/src/rules/function_expression_with_incorrect_types.dart
```diff
@@ -0,0 +1,110 @@
...
+      if (expectedType.isDynamic &&
+          currentType.isDartCoreObject &&
+          currentType.nullabilitySuffix != NullabilitySuffix.none) {
+        continue;
+      }
+
+      // if type are not consistent, lint!
+      if (currentParam.isRequiredPositional && currentType != expectedType ||
+          currentParam.isOptional &&
+              currentType.element != expectedType.element) {
```
bwilkerson, line 104: This condition doesn't cover the case of a required named parameter. It seems like it ought to.

bwilkerson (commented):
I have to question how much value this brings. Most users will follow the guideline of not adding type annotations in closures, and if it isn't doing much for Flutter then it's adding cognitive burden for users searching for a lint for not much benefit. Perhaps we shouldn't be adding it. @goderbauer Thoughts?

goderbauer:
I didn't know how common this pattern was. If we think it's not that common (which the data seems to indicate) I don't feel too strongly about having this lint.

a14n:
@bwilkerson I face a strange behaviour with `.staticParameterElement`. The result seems to have a `.library` at `null` for method with generics (eg. Map.forEach). Is it expected?

bwilkerson:
> The result seems to have a `.library` at `null` for method with generics (eg. `Map.forEach`). Is it expected?

I didn't think about it earlier, but I think I know what's happening. I suspect that the element being returned is a "view on an element" (a subclass of `Member`). This "view" holds the actual types after type parameters have been replaced by their corresponding arguments. They used to know what library they came from, but some language changes made that sometimes be a lie, so we changed the API. You can safely ask any element (including a member) for the `declaration` and then ask the declaration for the library and that should give you the result you want. Note however that the declaration can return `null` in which case there is no reliable way to get a library.

a14n:
> You can safely ask any element (including a member) for the declaration and then ask the declaration for the library and that should give you the result you want.

I still get `null` for `parameterElement.library` and `parameterElement.declaration.library` if the target method is parameterized.

a14n:
`.enclosingElement` is also `null` weither on `parameterElement` or `parameterElement.declaration`. That seems to be logic regarding `ParameterElementImpl.declaration` that returns `this` (`parameterElement` or `parameterElement.declaration` are both `ParameterElementImpl` for generic method or not).

I made my tests on:

```dart
class MyMap<K, V> {
  void f(void Function(int a) p) {}
  void ff(void Function(K a) p) {}
}
m() {
  MyMap().f((int? a) {}); // parameterElement.library is correctly set
  MyMap().ff((Object? a) {}); // parameterElement.library is null
}
```

bwilkerson:
@scheglov I'm guessing more has changed than I realized.

bwilkerson:
@scheglov ?

scheglov:
I think what happens here is that when a generic method is invoked, we instantiate the corresponding `FunctionType` and use its parameters for the context. And `ParameterElement`(s) of `FunctionType` are in nowhere - not attached to any library.

a14n:
So IIUC it is a bug that need to be fixed on the analyzer side?

scheglov:
Correct, this is an issue that should be fixed in the analyzer.

a14n:
@scheglov is there a specific issue already created to track this problem or do you want I create one?

scheglov:
@a14n please create a new one.

a14n:
> @a14n please create a new one.

Done with https://github.com/dart-lang/sdk/issues/45964

srawlins:
Apologies for the delay in reviewing (though perhaps this is mainly blocked on https://github.com/dart-lang/sdk/issues/45964).

In order to have a more focused discussion on whether this rule has enough value, would you mind, @a14n , opening a lint proposal issue, as per our new [lint lifecycle guidelines](https://github.com/dart-lang/linter/blob/main/doc/lint-lifecycle.md)?

a14n:
> In order to have a more focused discussion on whether this rule has enough value, would you mind, @a14n , opening a lint proposal issue, as per our new [lint lifecycle guidelines](https://github.com/dart-lang/linter/blob/main/doc/lint-lifecycle.md)?

Done with #4281

devoncarew:

PR #2531, opened 2021-03-18
`camel_case_types` support for nonfunction type aliases

pq:
See: #2509.

/cc @bwilkerson

### DIFF
--- a/test/rules/experiments/nonfunction-type-aliases/rules/camel_case_types.dart
+++ b/test/rules/experiments/nonfunction-type-aliases/rules/camel_case_types.dart
@@ -1,0 +1,35 @@
+// Copyright (c) 2015, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// test w/ `pub run test -N camel_case_types`
+
+class A {}
+
+class DB {}
+
+class FooBar {}
+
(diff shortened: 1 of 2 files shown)


test/rules/experiments/nonfunction-type-aliases/rules/camel_case_types.dart
```diff
@@ -0,0 +1,35 @@
+// Copyright (c) 2015, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// test w/ `pub run test -N camel_case_types`
```
bwilkerson, line 5: Consider adding tests for
- generic function type aliases (`typedef f = void Function();`)
- class type aliases (`class c = Object with M;`)
- enums (`enum e { a, b }`)
- mixins (`mixin m {}`)
- extensions (`extension ext on Object {}`)

bwilkerson (approved):

test/rules/experiments/nonfunction-type-aliases/rules/camel_case_types.dart
pq: Ok, wow.  enums, mixins and extensions are not supported! 😬   Since that could have some internal impact, I'm going to break that change into a separate issue.  Thanks!

PR #2534, opened 2021-03-18
`avoid_private_typedef_functions` support for nonfunction type aliases

pq:
See: #2509.

/cc @bwilkerson


bwilkerson (commented):
This rule seems to be very specifically about function types. I can certainly see an argument for generalizing it, but we might want to change the name and we should certainly update the documentation if we decide to go that way. I think it might be worth opening an issue to discuss the question and involve Morgan and Bob, and maybe someone from the Flutter team if it's a lint they use.

pq:
That sounds reasonable to me.  Closing in favor of https://github.com/dart-lang/linter/issues/2535.

PR #2536, opened 2021-03-19
Fix iterable_contains_unrelated_type

### LINKED ISSUE #1003: iterable_contains_unrelated_type not working for List
In IntelliJ, I see this triggers the lint:

```
  Iterable<int> foo = [1];
  print(foo.contains('3'));
```

and this:

```
  Set<int> foo = Set.from([1]);
  print(foo.contains('3'));
```

but not this:

```
  List<int> foo = [1];
  print(foo.contains('3'));
```

I'm at a loss as to why not; it looks like there are test cases that should catch this in the linter code, and they pass. Maybe there is something in the test setup that is now different from the actual `List`?

xvrh:
Fixes #1003

This is an attempt to fix `iterable_contains_unrelated_type`. This rule works in the test suite but not with a real SDK.

This is because the [MockSdk] defines the hierarchy as `List<T> implements Iterable<T>` ([analyzer/mock_sdk.dart](https://github.com/dart-lang/sdk/blob/caa31f55be518e2de0bf723c81c853c2362dafb5/pkg/analyzer/lib/src/test_utilities/mock_sdk.dart#L419)) while the real SDK defines `List<T> extends EfficientLengthIterable<T>` ([sdk/core/list.dart](https://github.com/dart-lang/sdk/blob/c5d01ec4f7078a75b038a2ed04312afb50154323/sdk/lib/core/list.dart#L52))

### DIFF
--- a/test/rules/iterable_contains_unrelated_type.dart
+++ b/test/rules/iterable_contains_unrelated_type.dart
@@ -151,12 +151,6 @@ abstract class B3 extends A3 {}
 
 bool takesB3(B3 b) => b.contains('a'); // OK
 
-abstract class A2 implements Iterable<String> {}
-
-abstract class B2 extends A2 {}
-
-bool takesB2(B2 b) => b.contains('a'); // OK
-
 abstract class AddlInterface {}
 
 abstract class SomeIterable<E> implements Iterable<E>, AddlInterface {}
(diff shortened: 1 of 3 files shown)


test/rules/iterable_contains_unrelated_type.dart
```diff
@@ -151,12 +151,6 @@ abstract class B3 extends A3 {}
 
 bool takesB3(B3 b) => b.contains('a'); // OK
 
-abstract class A2 implements Iterable<String> {}
```
xvrh, line 154: I remove it because this is the same case as A1 above

bwilkerson (approved):

pq:
> This is because the [MockSdk] defines the hierarchy as List<T> implements Iterable<T> (analyzer/mock_sdk.dart) while the real SDK defines List<T> extends EfficientLengthIterable<T> (sdk/core/list.dart)

Ah.  That's a problem.

It'd be good to get that updated.  (fyi @scheglov)

pq (approved):
💎 

Thanks!

scheglov:
https://dart-review.googlesource.com/c/sdk/+/192048

PR #2549, opened 2021-03-23
Enable `unnecessary_final`

pq:
As per the updated advice in "Effective Dart": https://dart.dev/guides/language/effective-dart/usage#do-follow-a-consistent-rule-for-var-and-final-on-local-variables, this enforces an opinion on `var` over `final` for locals.

/cc @bwilkerson

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -98,6 +98,7 @@ linter:
     - throw_in_finally
 #    - type_annotate_public_apis
     - unnecessary_brace_in_string_interps
+    - unnecessary_final
     - unnecessary_getters_setters
     - unnecessary_lambdas
     - unnecessary_null_aware_assignments
--- a/lib/src/analyzer.dart
+++ b/lib/src/analyzer.dart
@@ -53,10 +53,10 @@ Future<Iterable<AnalysisErrorInfo>> lintFiles(
     DartLinter linter, List<File> filesToLint) async {
   // Setup an error watcher to track whether an error was logged to stderr so
   // we can set the exit code accordingly.
-  final errorWatcher = ErrorWatchingSink(errorSink);
+  var errorWatcher = ErrorWatchingSink(errorSink);
   errorSink = errorWatcher;
 
-  final errors = await linter.lintFiles(filesToLint);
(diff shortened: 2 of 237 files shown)


parlough:
I'm curious what the rationale for going this direction is. I personally find using `final` helps makes code easier to understand as readers know it won't be reassigned, and also prevents users from doing so accidentally. 

I'm happy to standardize on whatever is decided, just curious.

pq:
Thanks for chiming in!  I have some personal feelings about this one (and actually argued for `final` for a long time) but feel that its merits are sort of negligible/tenuous in practice and would prefer standardization (who's benefits are more easily measured).  If you look at the SDK, while consistency is not enforced, there is a tendency towards `var` and so I'd like to just bite the bullet and get on board.  (FWIW: the current mixed standard *is* distracting to me personally so if we didn't turn this on, I'd want to turn on `prefer_final_locals` but that feels like we're swimming away from a kind of growing standard in the SDK and I just don't see the benefit.)

pq:
> I personally find using final helps makes code easier to understand as readers know it won't be reassigned, and also prevents users from doing so accidentally.

Anecdotally, I thought this was true for me too until I spent a bunch of time in a codebase that didn't use `final` for locals and frankly never missed them.   Obviously that's just me though!

parlough:
@pq Thanks for the extra context. I agree the mix is distracting, so I'm glad the linter is standardizing on something. 

I think the reasoning seems fair and like you said it seems common for people to gravitate towards `var` already. 

I'm happy to adapt towards a standard but it'll take changing some muscle memory ;) 

https://dart-review.googlesource.com/c/sdk/+/192720 will help - thanks!

pq:
I hear you about muscle memory.  That's big for me too.  I realize this is a bit of a change and may be jarring.  The good news is that if we decide we don't like it, we can go back.  If there isn't a comparable set of fixes for `prefer_final_locals`, we can add em!

> https://dart-review.googlesource.com/c/sdk/+/192720 will help - thanks!

Entirely the motivation.  Much easier when you can just run `dart fix`!

parlough (approved):

bwilkerson (approved):

PR #2557, opened 2021-03-24
Add require_trailing_comma lint rule.

### LINKED ISSUE #1338: [flutter] (prefer) trailing_commas
Following-up from https://github.com/flutter/flutter-intellij/issues/2980#issuecomment-451105228, we might consider a rule that enforces the [flutter style guide suggestion](https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#use-a-trailing-comma-for-arguments-parameters-and-list-items-but-only-if-they-each-have-their-own-line):

> Use a trailing comma for arguments, parameters, and list items, but only if they each have their own line.

Since a major motivation is to give analyzer something to hang quick-fixes from, we might consider enforcing just the first part and allowing `dartfmt` to handle line-breaks but maybe this would make the rule less useful for flutter repo committers.

@Hixie: any thoughts?

/cc @jaumard

lauweijie:
Add `require_trailing_comma` lint rule.

**DO** use trailing commas for all function calls and definitions unless the function call or definition, from the start of the function name up to the closing parenthesis, fits in a single line.

**GOOD:**
```dart
void run() {
  method(
    'does not fit on one line',
    'test test test test test test test test test test test',
  );
}
```

**BAD:**
```dart
void run() {
  method('does not fit on one line',
      'test test test test test test test test test test test');
}
```

**Exception:** If the final parameter is positional (vs named) and either a function literal implemented using curly braces, a literal map, a literal set or a literal array. This exception only applies if the final parameter does not fit entirely on one line.

Fixes #1338

### DIFF
--- a/example/all.yaml
+++ b/example/all.yaml
@@ -139,6 +139,7 @@ linter:
     - provide_deprecation_message
     - public_member_api_docs
     - recursive_getters
+    - require_trailing_comma
     - sized_box_for_whitespace
     - slash_for_doc_comments
     - sort_child_properties_last
--- a/example/all.yaml
+++ b/example/all.yaml
@@ -139,7 +139,7 @@ linter:
     - provide_deprecation_message
     - public_member_api_docs
     - recursive_getters
-    - require_trailing_comma
+    - require_trailing_commas
     - sized_box_for_whitespace
     - slash_for_doc_comments
     - sort_child_properties_last
--- a/lib/src/rules/require_trailing_comma.dart
+++ b/lib/src/rules/require_trailing_comma.dart
@@ -8,11 +8,11 @@ import 'package:analyzer/dart/ast/visitor.dart';
 
 import '../analyzer.dart';
 
-const _desc = r'Use trailing commas for all function calls and definitions.';
+const _desc = r'Use trailing commas for all function calls and declarations.';
 
 const _details = r'''
 
-**DO** use trailing commas for all function calls and definitions unless the
+**DO** use trailing commas for all function calls and declarations unless the
 function call or definition, from the start of the function name up to the
 closing parenthesis, fits in a single line.
 
@@ -34,10 +34,10 @@ void run() {
 }
 ```
 
-**Exception:** If the final parameter is positional (vs named) and either a
-function literal implemented using curly braces, a literal map, a literal set or
-a literal array. This exception only applies if the final parameter does not fit
-entirely on one line.
+**Exception:** If the final parameter/argument is positional (vs named) and is
+either a function literal implemented using curly braces, a literal map, a
+literal set or a literal array. This exception only applies if the final
+parameter does not fit entirely on one line.
 
 ''';
 
@@ -60,7 +60,8 @@ class RequireTrailingComma extends LintRule implements NodeLintRule {
       ..addCompilationUnit(this, visitor)
       ..addArgumentList(this, visitor)
       ..addFormalParameterList(this, visitor)
-      ..addAssertStatement(this, visitor);
+      ..addAssertStatement(this, visitor)
+      ..addAssertInitializer(this, visitor);
   }
 }
 
@@ -111,6 +112,16 @@ class _Visitor extends SimpleAstVisitor<void> {
     );
   }
 
+  @override
+  void visitAssertInitializer(AssertInitializer node) {
+    super.visitAssertInitializer(node);
+    _checkTrailingComma(
+      node.leftParenthesis,
+      node.rightParenthesis,
+      node.message ?? node.condition,
+    );
+  }
+
   void _checkTrailingComma(
     Token leftParenthesis,
     Token rightParenthesis,
@@ -119,10 +130,10 @@ class _Visitor extends SimpleAstVisitor<void> {
     // Early exit if trailing comma is present.
     if (lastNode.endToken.next?.type == TokenType.COMMA) return;
 
-    // No trailing comma is needed if the function call or definition, up to the
-    // closing parenthesis, fits on a single line. Ensuring the left and right
-    // parenthesis are on the same line is sufficient since dartfmt places the
-    // left parenthesis right after the identifier (on the same line).
+    // No trailing comma is needed if the function call or declaration, up to
+    // the closing parenthesis, fits on a single line. Ensuring the left and
+    // right parenthesis are on the same line is sufficient since dartfmt places
+    // the left parenthesis right after the identifier (on the same line).
     if (_isSameLine(leftParenthesis, rightParenthesis)) return;
 
     // Check the last parameter to determine if there are any exceptions.
@@ -132,8 +143,8 @@ class _Visitor extends SimpleAstVisitor<void> {
   }
 
   bool _isSameLine(Token token1, Token token2) =>
-      _lineInfo!.getLocation(token1.charOffset).lineNumber ==
-      _lineInfo!.getLocation(token2.charOffset).lineNumber;
+      _lineInfo!.getLocation(token1.offset).lineNumber ==
+      _lineInfo!.getLocation(token2.offset).lineNumber;
 
   bool _shouldAllowTrailingCommaException(AstNode lastNode) {
     // No exceptions are allowed if the last parameter is named.
--- a/lib/src/rules/require_trailing_comma.dart
+++ b/lib/src/rules/require_trailing_comma.dart
@@ -44,17 +44,17 @@ entirely on one line.
 class RequireTrailingComma extends LintRule implements NodeLintRule {
   RequireTrailingComma()
       : super(
-    name: 'require_trailing_comma',
-    description: _desc,
-    details: _details,
-    group: Group.style,
-  );
+          name: 'require_trailing_comma',
+          description: _desc,
+          details: _details,
+          group: Group.style,
+        );
 
   @override
   void registerNodeProcessors(
-      NodeLintRegistry registry,
-      LinterContext context,
-      ) {
+    NodeLintRegistry registry,
+    LinterContext context,
+  ) {
     var visitor = _Visitor(this);
     registry
       ..addCompilationUnit(this, visitor)
@@ -112,10 +112,10 @@ class _Visitor extends SimpleAstVisitor<void> {
   }
 
   void _checkTrailingComma(
-      Token leftParenthesis,
-      Token rightParenthesis,
-      AstNode lastNode,
-      ) {
+    Token leftParenthesis,
+    Token rightParenthesis,
+    AstNode lastNode,
+  ) {
     // Early exit if trailing comma is present.
     if (lastNode.endToken.next?.type == TokenType.COMMA) return;
 
@@ -133,7 +133,7 @@ class _Visitor extends SimpleAstVisitor<void> {
 
   bool _isSameLine(Token token1, Token token2) =>
       _lineInfo!.getLocation(token1.charOffset).lineNumber ==
-          _lineInfo!.getLocation(token2.charOffset).lineNumber;
+      _lineInfo!.getLocation(token2.charOffset).lineNumber;
 
   bool _shouldAllowTrailingCommaException(AstNode lastNode) {
     // No exceptions are allowed if the last parameter is named.
--- a/lib/src/rules/require_trailing_comma.dart
+++ b/lib/src/rules/require_trailing_comma.dart
@@ -1,0 +1,155 @@
+// Copyright (c) 2021, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/token.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+
+import '../analyzer.dart';
+
+const _desc = r'Use trailing commas for all function calls and definitions.';
+
+const _details = r'''
+
+**DO** use trailing commas for all function calls and definitions unless the
+function call or definition, from the start of the function name up to the
+closing parenthesis, fits in a single line.
+
+**GOOD:**
+```dart
+void run() {
+  method(
+    'does not fit on one line',
+    'test test test test test test test test test test test',
+  );
+}
+```
+
+**BAD:**
+```dart
+void run() {
+  method('does not fit on one line',
+      'test test test test test test test test test test test');
+}
+```
+
+**Exception:** If the final parameter is positional (vs named) and either a
+function literal implemented using curly braces, a literal map, a literal set or
+a literal array. This exception only applies if the final parameter does not fit
+entirely on one line.
+
+''';
+
+class RequireTrailingComma extends LintRule implements NodeLintRule {
+  RequireTrailingComma()
+      : super(
+    name: 'require_trailing_comma',
+    description: _desc,
+    details: _details,
+    group: Group.style,
+  );
+
+  @override
+  void registerNodeProcessors(
+      NodeLintRegistry registry,
+      LinterContext context,
+      ) {
+    var visitor = _Visitor(this);
+    registry
+      ..addCompilationUnit(this, visitor)
+      ..addArgumentList(this, visitor)
+      ..addFormalParameterList(this, visitor)
+      ..addAssertStatement(this, visitor);
+  }
+}
+
+class _Visitor extends SimpleAstVisitor<void> {
+  static const _trailingCommaCode = LintCode(
+    'require_trailing_comma',
+    'Missing a required trailing comma.',
+  );
+
+  final LintRule rule;
+
+  LineInfo? _lineInfo;
+
+  _Visitor(this.rule);
+
+  @override
+  void visitCompilationUnit(CompilationUnit node) => _lineInfo = node.lineInfo;
+
+  @override
+  void visitArgumentList(ArgumentList node) {
+    super.visitArgumentList(node);
+    if (node.arguments.isEmpty) return;
+    _checkTrailingComma(
+      node.leftParenthesis,
+      node.rightParenthesis,
+      node.arguments.last,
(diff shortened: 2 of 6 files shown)


test/rules/require_trailing_comma.dart
```diff
@@ -0,0 +1,223 @@
...
+          () {},
+    );
+
+    test('no exception for function literal as it fits entirely on 1 line',
+            () {}); // LINT
+
+    assert(true);
+
+    assert('a very very very very very very very very very long string'
+        .isNotEmpty); // LINT
```
srawlins, line 205: Heads up: I suspect this part will be controversial. I don't have a strong opinion yet, but I suspect for many people, requiring a trailing comma for a single parameter/argument (on a long line) may be a bridge too far.

lib/src/rules/require_trailing_comma.dart
```diff
@@ -0,0 +1,155 @@
...
+
+**BAD:**
+```dart
+void run() {
+  method('does not fit on one line',
+      'test test test test test test test test test test test');
+}
+```
+
+**Exception:** If the final parameter is positional (vs named) and either a
```
bwilkerson, line 37: Based on the test cases I think this exception covers both parameters and arguments, but the text sounds more like you're talking about arguments even though you use the word 'parameter'. Consider clarifying that this applies to both.

lib/src/rules/require_trailing_comma.dart
```diff
@@ -0,0 +1,155 @@
...
+    if (node.parameters.isEmpty) return;
+    _checkTrailingComma(
+      node.leftParenthesis,
+      node.rightParenthesis,
+      node.parameters.last,
+    );
+  }
+
+  @override
+  void visitAssertStatement(AssertStatement node) {
```
bwilkerson, line 105: Are we sure we want to treat assert statements as invocations? If so, should we also treat asserts in initializer lists as invocations? If so, you'll also need to visit `AssertInitializer`s.

lib/src/rules/require_trailing_comma.dart
```diff
@@ -0,0 +1,155 @@
...
+
+    // Check the last parameter to determine if there are any exceptions.
+    if (_shouldAllowTrailingCommaException(lastNode)) return;
+
+    rule.reportLintForToken(rightParenthesis, errorCode: _trailingCommaCode);
+  }
+
+  bool _isSameLine(Token token1, Token token2) =>
+      _lineInfo!.getLocation(token1.charOffset).lineNumber ==
+      _lineInfo!.getLocation(token2.charOffset).lineNumber;
```
bwilkerson, line 136: Consider using `offset` rather than `charOffset`. It's slightly more efficient and more readable (imo).

test/rules/require_trailing_comma.dart
```diff
@@ -0,0 +1,223 @@
...
+
+  void method2(
+      Object param1,
+      Object param2,
+      Object param3,
+      Object param4,
+      Object param5,
+      ) {}
+
+  void method3(
```
bwilkerson, line 40: I might have missed it, but I don't see an example of a named parameter with a default value.

lib/src/rules/require_trailing_comma.dart
```diff
@@ -0,0 +1,155 @@
...
+    Token rightParenthesis,
+    AstNode lastNode,
+  ) {
+    // Early exit if trailing comma is present.
+    if (lastNode.endToken.next?.type == TokenType.COMMA) return;
+
+    // No trailing comma is needed if the function call or definition, up to the
+    // closing parenthesis, fits on a single line. Ensuring the left and right
+    // parenthesis are on the same line is sufficient since dartfmt places the
+    // left parenthesis right after the identifier (on the same line).
```
srawlins, line 125: To strengthen this assertion, I think it would be good to have a few extra tests:

* self-executing closure: `(a, b) {\n multiple lines \n}(1, 2)`
* executing the result of an expression which is not a simple identifier: `(one ??\n  two)(a, b)`

test/rules/require_trailing_comma.dart
```diff
@@ -0,0 +1,223 @@
+// Copyright (c) 2021, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+class RequireTrailingCommaExample {
+  RequireTrailingCommaExample.constructor1(Object param1, Object param2);
+
+  RequireTrailingCommaExample.constructor2(
```
srawlins, line 8: The only function type I see missing is operators. Since trailing commas don't make sense on an operator call (a BinaryExpression, UnaryExpression, or IndexExpression), I would only suggest adding a test for an operator declaration, like `void operator [](LongType longParameterName) {`

bwilkerson (commented):
Overall the implementation looks good.

I didn't address any questions about whether this implements the semantics we want, only whether the semantics this implements were implemented correctly.

I do have a larger concern. This implementation depends on having first run the formatter over the code. This has some negative consequences.
1. Sometimes users won't be told about missing commas until after they format the code. This could lead to a workflow in which users need to format their code, add commas, and then format again, which is less than ideal.
2. Sometimes users will be told that a comma is missing even if the formatter could put the whole invocation on a single line, which could lead to unnecessary commas.
3. Users will never be told when they could remove a trailing comma (such as when one or more of the arguments are extracted into local variables).

Removing that requirement would be very difficult, so I'm not sure the best way to proceed from here.

lib/src/rules/require_trailing_comma.dart
```diff
@@ -0,0 +1,155 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/token.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+
+import '../analyzer.dart';
+
+const _desc = r'Use trailing commas for all function calls and definitions.';
```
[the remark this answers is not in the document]
lauweijie: Done, replaced 'definitions' with 'declarations'.

lib/src/rules/require_trailing_comma.dart
lauweijie: Added test cases for self-executing closure.

test/rules/require_trailing_comma.dart
lauweijie: Added test case for operator declaration.

lib/src/rules/require_trailing_comma.dart
lauweijie: Updated to specify both parameter/argument.

lib/src/rules/require_trailing_comma.dart
lauweijie: Changed `charOffset` to `offset`.

test/rules/require_trailing_comma.dart
lauweijie: Added test case for named param with default value.

lib/src/rules/require_trailing_comma.dart
lauweijie: Yes, it should help readability there too. Thanks for pointing out the missing `AssertInitializer` visitor. I've added that and accompanying tests.

lauweijie:
Thanks for the review, @bwilkerson and @srawlins. I've addressed most comments.

> Sometimes users won't be told about missing commas until after they format the code. This could lead to a workflow in which users need to format their code, add commas, and then format again, which is less than ideal.

You're right, this assumes `dartfmt` has been run over the code, which could indeed be less than ideal. As you mentioned, figuring out when to require a trailing comma without `dartfmt` is tricky. Suggestions welcome!

> Users will never be told when they could remove a trailing comma (such as when one or more of the arguments are extracted into local variables).

True, this lint does not enforce the reverse when everything fits on a single line.
Perhaps that could be a separate lint: `avoid_trailing_comma`?

example/all.yaml
```diff
@@ -139,6 +139,7 @@ linter:
     - provide_deprecation_message
     - public_member_api_docs
     - recursive_getters
+    - require_trailing_comma
```
pq, line 142: We've (almost) standardized on pluralized names so I think I might suggest => `require_trailing_commas`

pq:
> You're right, this assumes dartfmt has been run over the code

I agree that this is awkward (and complicates quick fixes) but I'm not sure it's a show-stopper.  It would definitely be good to capture this in the rule documentation somehow.

pq:
> Perhaps that could be a separate lint: avoid_trailing_comma?

This is an interesting idea.  Maybe `unnecessary_trailing_commas` would be a good partner lint?

example/all.yaml
lauweijie: Done, renamed to `require_trailing_commas`.

lib/src/rules/require_trailing_commas.dart
```diff
@@ -0,0 +1,168 @@
...
+      ..addAssertStatement(this, visitor)
+      ..addAssertInitializer(this, visitor);
+  }
+}
+
+class _Visitor extends SimpleAstVisitor<void> {
+  static const _trailingCommaCode = LintCode(
+    'require_trailing_commas',
+    'Missing a required trailing comma.',
+  );
```
pq, line 74: I don't think you need to create your own lint code instance and can just call `rule.reportLintForToken(token)` (which will pickup the default lint code implementation returned by the `lintCode` getter which should be the same as what you're creating here).

pq: (Places where you *do* want your own lint codes are when you want to define more than one to capture additional context. See for example: https://github.com/dart-lang/linter/blob/983cbf8b2f8a5fd27658e20f0871ac8e8c0276b9/lib/src/rules/always_declare_return_types.dart#L67-L77)

bwilkerson:
> Perhaps that could be a separate lint: `avoid_trailing_comma`?

My perspective is that this is a rule designed to enforce a particular style concerning when (and when not) to use a trailing comma in parameter and argument lists. I think it would be much better for it to enforce the whole style than for it to require a companion rule to enforce some portion of the style.

And from that perspective, perhaps we should rename the lint to indicate that it is (or could be) more encompassing (though I don't have a good suggestion off the top of my head).

Also, I'm concerned that `require_trailing_commas` and `avoid_trailing_commas` sound like opposites, and hence appear to be mutually exclusive.

> As you mentioned, figuring out when to require a trailing comma without `dartfmt` is tricky.

It would be prohibitively expensive, but `dart_style` does provide a programatic interface to invoke the formatter, so theoretically we could format the content first, then parse it, figure out where commas should have been, and then map those back to the current source code.

We might also be able to do some close-enough computation of the length of various code snippets to guess at whether multiple lines are required. For example, given an invocation like
```
method(
  a,
  b,
  c,
);
```
we could add the lengths of the individual argument, plus commas and spaces, and determine that this would all fit on one line if the trailing comma were removed. I'm not sure what kind of nasty corner cases we might run into, but it might be worth exploring at some point.

lib/src/rules/require_trailing_commas.dart
bwilkerson: FWIW, I think that this is a much better message than the default (which would be "Use trailing commas for all function calls and declarations.").

lauweijie: Agree with @bwilkerson here.

lauweijie:
> We might also be able to do some close-enough computation of the length of various code snippets to guess at whether multiple lines are required. For example, given an invocation like
> ```
> method(
>   a,
>   b,
>   c,
> );
> ```
> we could add the lengths of the individual argument, plus commas and spaces, and determine that this would all fit on one line if the trailing comma were removed. I'm not sure what kind of nasty corner cases we might run into, but it might be worth exploring at some point.

Removing trailing commas by simply counting the length of each individual argument could be tricky due to corner cases like this:

```dart
// Comment prevents everything from fitting on 1 line.
method(
  a, // comment
  b,
);
```

For this case, I should be able to easily detect that the token spans multiple lines and skip:

```dart
// Function literal that has few characters but prevents everything from being on 1 line.
method(
  a,
  () {
    // doSomething
  },
);
```

If we do count chars, we also have to remember to count in the named params:

```dart
method(
  aVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryExtremelyLongNamedParameter: a,
  b,
);
```

My thought: if we can get this to a good state that's agreeable by most, shall we move forward with what we have currently and propose further improvements later?

lib/src/rules/require_trailing_commas.dart
pq: Whoops!  I didn't see that. 😬   I totally agree. Far better.  Sorry for the noise!

bwilkerson:
> if we can get this to a good state that's agreeable by most, shall we move forward with what we have currently and propose further improvements later?

Yes, I'm fine with that. If the changes we're contemplating would be breaking, then we might want to mark it as experimental in the meantime, but other than that I'm always in favor of making incremental progress on a feature.

lib/src/rules/require_trailing_commas.dart
```diff
@@ -0,0 +1,168 @@
...
+      'test test test test test test test test test test test');
+}
+```
+
+**Exception:** If the final parameter/argument is positional (vs named) and is
+either a function literal implemented using curly braces, a literal map, a
+literal set or a literal array. This exception only applies if the final
+parameter does not fit entirely on one line.
+
+**Note:** This lint rule assumes `dartfmt` has been run over the code.
```
bwilkerson, line 42: Consider adding "and might produce false positives until that's happened" to make the implications more clear.

lauweijie: Updated, although I feel that this could be a given.

The assumption that `dartfmt` has been run is not unique to this lint rule.

[`curly_braces_in_flow_control_structures`](https://github.com/dart-lang/linter/blob/master/lib/src/rules/curly_braces_in_flow_control_structures.dart) also has similar assumptions but is undocumented.

lauweijie:
> Yes, I'm fine with that. If the changes we're contemplating would be breaking, then we might want to mark it as experimental in the meantime, but other than that I'm always in favor of making incremental progress on a feature.

Sounds good! What are the implications of a rule being marked experimental? Does that mean it's unenforced by default but can be manually turned on?

bwilkerson:
> What are the implications of a rule being marked experimental?

It's basically just a signal to the community that we reserve the right to change its behavior. With lints that aren't marked as being experimental we're very careful to consider the possible impact on clients (such as new lints being generated) before making changes to them.

> Does that mean it's unenforced by default but can be manually turned on?

All lints are disabled by default and need to be explicitly enabled in the analysis options file. So yes, but not because it's marked experimental.

lauweijie:
@bwilkerson Marked the lint as experimental, PTAL

pq (approved):

bwilkerson (approved):

a14n:
Only 7211 lints found on flutter package :laughing: :hot_face:

Some random samples that might be discutable / need an exception:

- single argument (that [I excluded in my implementation](https://github.com/a14n/linter/blob/prefer_trailing_commas/lib/src/rules/prefer_trailing_commas.dart#L90) of this lint 18 month ago :))

```dart
  print('''
    --help      Print a usage message.
    --temp-dir  A location where temporary files may be written. Defaults to a
                directory in the system temp folder. If a temp_dir is not
                specified, then the default temp_dir will be created, used, and
                removed automatically.
    ''');
```

```dart
    await tester.pumpWidget(Wrap(
      alignment: WrapAlignment.spaceAround,
      spacing: 5.0,
      textDirection: TextDirection.rtl,
      children: const <Widget>[
        SizedBox(width: 100.0, height: 10.0),
        SizedBox(width: 200.0, height: 20.0),
        SizedBox(width: 310.0, height: 30.0),
      ],
    ));
```

```dart
    assert(() {
      _debugIsRenderViewInitialized = true;
      return true;
    }());
```

- test/map/closure pattern

```dart
    testWidgets('debugIsLocalCreationLocation test', (WidgetTester tester) async {
      // several statements
    }, skip: !WidgetInspectorService.instance.isWidgetCreationTracked()); // Test requires --track-widget-creation flag.
```

```dart
        Timeline.instantSync('ImageCache.evict', arguments: <String, dynamic>{
          'type': 'keepAlive',
          'sizeInBytes': image.sizeBytes,
        });
```

lauweijie:
> Only 7211 lints found on flutter package 😆 🥵

I've written a tool to automate most of the fixes and will be happy to help with the effort to apply this to the flutter package.

> single argument

I'm of the thought that single argument calls shouldn't be given this exception if it spans more than one line. Of course, open to discussion and would like to hear from the community.

> test/map/closure pattern

The lint rule already excepts the trailing comma if the last arg is a function literals or literal set/map/array. Exemption does not apply if the last param is named rather than positional.

Again, open to discussion, but I think keeping the rule simple while covering most cases should be the goal.

bwilkerson:
> I've written a tool to automate most of the fixes ...

Ideally we should add a fix in the analysis server for this lint and make it possible to apply the fix through `dart fix`.

a14n:
Regarding the lints reported on flutter I would recommand that the lint triggers:

- always when there are 2+ arguments where the argument list is on several lines.
- never when there is only 1 argument except if this argument is on a new line.

lauweijie:
> Ideally we should add a fix in the analysis server for this lint and make it possible to apply the fix through `dart fix`.

Agree. Shall we start by opening an issue on dart-lang/sdk for this request?

> Regarding the lints reported on flutter I would recommand that the lint triggers:
>
>- always when there are 2+ arguments where the argument list is on several lines.
>- never when there is only 1 argument except if this argument is on a new line.

@jefflim-google any thoughts on the single-arg exception proposed by @a14n?

bwilkerson:
> Shall we start by opening an issue on dart-lang/sdk for this request?

If you'd like to create an issue, then that's the best place for it. Not sure whether we need an issue, though.

PR #2560, opened 2021-03-25
add prefer_null_aware_method_calls

a14n:
# Description

Instead of checking nullability of a function/method `f` before calling it you
can use `f?.call()`.

**BAD:**

```
if (f != null) f!();
```

**GOOD:**

```
f?.call();
```

Requested by https://github.com/flutter/flutter/pull/79003#issuecomment-806156141

### DIFF
--- a/lib/src/rules/prefer_null_aware_method_calls.dart
+++ b/lib/src/rules/prefer_null_aware_method_calls.dart
@@ -1,0 +1,94 @@
+// Copyright (c) 2021, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/token.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+
+import '../analyzer.dart';
+
+const _desc = r'Prefer null aware method call.';
+
+const _details = r'''
+
+Instead of checking nullability of a function/method `f` before calling it you
+can use `f?.call()`.
+
+**BAD:**
+```
+if (f != null) f!();
+```
+
+**GOOD:**
+```
+f?.call();
+```
+
+''';
+
+class PreferNullAwareMethodCalls extends LintRule implements NodeLintRule {
+  PreferNullAwareMethodCalls()
+      : super(
+          name: 'prefer_null_aware_method_calls',
+          description: _desc,
+          details: _details,
+          group: Group.style,
+        );
+
+  @override
+  void registerNodeProcessors(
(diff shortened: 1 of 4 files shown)


lib/src/rules/prefer_null_aware_method_calls.dart
```diff
@@ -0,0 +1,94 @@
...
+
+  _Visitor(this.rule);
+
+  @override
+  void visitIfStatement(IfStatement node) {
+    var condition = node.condition;
+    if (condition is BinaryExpression &&
+        condition.operator.type == TokenType.BANG_EQ &&
+        condition.rightOperand is NullLiteral &&
+        node.elseKeyword == null) {
```
bwilkerson, line 59: Neither of these conditions check to see that the left hand side is a reference to a function, nor do they check that the function being invoked is the same as the one being tested. I think it's safe to ignore the first check as long as we check the second condition. For example, let's add a test of the form
```
var f, g;
void m() {
  if (f != null) g!();
  f != null ? g!() : null;
}
```

a14n: The equality between condition and target of call is done at line 89. The test you mentioned is similar to lines 19 and 25 of the test file.

bwilkerson: Sorry, I missed that.

lib/src/rules/prefer_null_aware_method_calls.dart
```diff
@@ -0,0 +1,94 @@
...
+      _check(node.thenExpression, condition);
+    }
+  }
+
+  void _check(Expression expression, BinaryExpression condition) {
+    if (expression is FunctionExpressionInvocation) {
+      var target = expression.function;
+      if (target is PostfixExpression &&
+          target.operator.type == TokenType.BANG &&
+          target.operand.toString() == condition.leftOperand.toString()) {
```
bwilkerson, line 89: The `analyzer` APIs reserve `toString` for debugging purposes. Technically these should be converted to use `toSource`.

bwilkerson (approved):

lib/src/rules/prefer_null_aware_method_calls.dart
```diff
@@ -0,0 +1,94 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/token.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+
+import '../analyzer.dart';
+
+const _desc = r'Prefer null aware method call.';
```
pq, line 11: =>

    r'Prefer null aware method calls.'

?

pq (approved):
👍

PR #2566, opened 2021-04-01
relax require_trailing_commas

a14n:
This PR relaxes a lint `require_trailing_commas` to ignore some line splits and allow patterns reported in https://github.com/dart-lang/linter/pull/2557#issuecomment-808406867 and https://github.com/dart-lang/linter/pull/2557#issuecomment-810554136

### DIFF
--- a/lib/src/rules/require_trailing_commas.dart
+++ b/lib/src/rules/require_trailing_commas.dart
@@ -208,6 +208,7 @@ class _ArgVisitor extends GeneralizingAstVisitor<void> {
   @override
   void visitListLiteral(ListLiteral node) {
     if (currentLine == lineOf(node.leftBracket.offset) &&
+        node.elements.isNotEmpty &&
         node.elements.last.endToken.next?.type == TokenType.COMMA) {
       currentLine = lineOf(node.end);
       return;
@@ -219,6 +220,7 @@ class _ArgVisitor extends GeneralizingAstVisitor<void> {
   @override
   void visitSetOrMapLiteral(SetOrMapLiteral node) {
     if (currentLine == lineOf(node.leftBracket.offset) &&
+        node.elements.isNotEmpty &&
         node.elements.last.endToken.next?.type == TokenType.COMMA) {
       currentLine = lineOf(node.end);
       return;
--- a/lib/src/rules/require_trailing_commas.dart
+++ b/lib/src/rules/require_trailing_commas.dart
@@ -34,10 +34,10 @@ void run() {
 }
 ```
 
-**Exception:** If the final parameter/argument is positional (vs named) and is
-either a function literal implemented using curly braces, a literal map, a
-literal set or a literal array. This exception only applies if the final
-parameter does not fit entirely on one line.
+**Exception:** If a parameter/argument is either a function literal
+implemented using curly braces, a literal map, a literal set or a literal array
+then any line break will be ignore and the expression will be treated as a
+single line expression.
 
 **Note:** This lint rule assumes `dartfmt` has been run over the code and may
 produce false positives until that has happened.
@@ -208,8 +208,7 @@ class _ArgVisitor extends GeneralizingAstVisitor<void> {
   @override
   void visitListLiteral(ListLiteral node) {
     if (currentLine == lineOf(node.leftBracket.offset) &&
-        node.elements.isNotEmpty &&
-        node.elements.last.endToken.next?.type == TokenType.COMMA) {
+        node.elements.isNotEmpty) {
       currentLine = lineOf(node.end);
       return;
     }
@@ -220,8 +219,7 @@ class _ArgVisitor extends GeneralizingAstVisitor<void> {
   @override
   void visitSetOrMapLiteral(SetOrMapLiteral node) {
     if (currentLine == lineOf(node.leftBracket.offset) &&
-        node.elements.isNotEmpty &&
-        node.elements.last.endToken.next?.type == TokenType.COMMA) {
+        node.elements.isNotEmpty) {
       currentLine = lineOf(node.end);
       return;
     }
--- a/lib/src/rules/require_trailing_commas.dart
+++ b/lib/src/rules/require_trailing_commas.dart
@@ -61,11 +61,11 @@ class RequireTrailingCommas extends LintRule implements NodeLintRule {
   ) {
     var visitor = _Visitor(this);
     registry
-      ..addCompilationUnit(this, visitor)
       ..addArgumentList(this, visitor)
-      ..addFormalParameterList(this, visitor)
+      ..addAssertInitializer(this, visitor)
       ..addAssertStatement(this, visitor)
-      ..addAssertInitializer(this, visitor);
+      ..addCompilationUnit(this, visitor)
+      ..addFormalParameterList(this, visitor);
   }
 }
 
@@ -91,7 +91,7 @@ class _Visitor extends SimpleAstVisitor<void> {
     _checkTrailingComma(
       node.leftParenthesis,
       node.rightParenthesis,
-      node.arguments.last,
+      node.arguments,
     );
   }
 
@@ -102,7 +102,7 @@ class _Visitor extends SimpleAstVisitor<void> {
     _checkTrailingComma(
       node.leftParenthesis,
       node.rightParenthesis,
-      node.parameters.last,
+      node.parameters,
     );
   }
 
@@ -112,7 +112,10 @@ class _Visitor extends SimpleAstVisitor<void> {
     _checkTrailingComma(
       node.leftParenthesis,
       node.rightParenthesis,
-      node.message ?? node.condition,
+      [
+        node.condition,
+        if (node.message != null) node.message!,
+      ],
     );
   }
 
@@ -122,17 +125,20 @@ class _Visitor extends SimpleAstVisitor<void> {
     _checkTrailingComma(
       node.leftParenthesis,
       node.rightParenthesis,
-      node.message ?? node.condition,
+      [
+        node.condition,
+        if (node.message != null) node.message!,
+      ],
     );
   }
 
   void _checkTrailingComma(
     Token leftParenthesis,
     Token rightParenthesis,
-    AstNode lastNode,
+    List<AstNode> nodes,
   ) {
     // Early exit if trailing comma is present.
-    if (lastNode.endToken.next?.type == TokenType.COMMA) return;
+    if (nodes.last.endToken.next?.type == TokenType.COMMA) return;
 
     // No trailing comma is needed if the function call or declaration, up to
     // the closing parenthesis, fits on a single line. Ensuring the left and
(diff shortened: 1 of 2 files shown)


lauweijie:
I'm not sure I agree entirely with the relaxed rule.

Anonymous functions in asserts LGTM:
```
assert(() {
  // code
  return true;
} ());
```

I feel these should still have trailing commas:
```
 await tester.pumpWidget(Wrap(
  alignment: WrapAlignment.spaceAround,
  spacing: 5.0,
  textDirection: TextDirection.rtl,
  children: const <Widget>[
    SizedBox(width: 100.0, height: 10.0),
    SizedBox(width: 200.0, height: 20.0),
    SizedBox(width: 310.0, height: 30.0),
  ],
));
```

```
Timeline.instantSync('ImageCache.evict', arguments: <String, dynamic>{
  'type': 'keepAlive',
  'sizeInBytes': image.sizeBytes,
});
```

cc: @jefflim-google

a14n:
I don't really understand why the follonwing code is allowed:

```dart
    test('exception for map literal as it spans multiple lines', const {
      'one': '1',
      'two': '2',
      'three': '3',
    });
```

and the following code is forbidden (named argument instead of positionnal):

```dart
    test('exception for map literal as it spans multiple lines', args: const {
      'one': '1',
      'two': '2',
      'three': '3',
    });
```

About test pattern adding trailing comma will only add additional 2-space indent and will not improve readability:

```dart
// currently not allowed
test('my awesome test', () {
  // ....
}, skip: true);
// adding a skip on a test will require to have a big diff because a trailing comma will be required
test(
  'my awesome test', 
  () {
    // ....
  }, 
  skip: true,
);
```

a14n:
What can be done to unblock this PR? It's a little frustrating to have no answer.

As [mentioned in this comment](https://github.com/dart-lang/linter/issues/1338#issuecomment-459618941) Flutter would totally use this lint... but only in this relaxed version imho.

jefflim-google:
Re named vs positional:

    test('exception for map literal as it spans multiple lines', args: const {
      'one': '1',
      'two': '2',
      'three': '3',
    });

This was intentionally chosen for our team because having "args:"  consistently positioned at the start of the line was the reason we chose not to allow named parameters have the exception.

Timeline.instantSync('ImageCache.evict', arguments: <String, dynamic>{
  'type': 'keepAlive',
  'sizeInBytes': image.sizeBytes,
});

vs

Timeline.instantSync(
  'ImageCache.evict', 
  arguments: <String, dynamic>{
    'type': 'keepAlive',
    'sizeInBytes': image.sizeBytes,
  },
);

The arguments: parameter here is much clearer.

Agree that the assert one can be cleaned up.

a14n:
> This was intentionally chosen for our team

It could be worth to have a softer lint accepted by several teams as we don't want to end up with 2 versions of the same lint :)

I don't speak for the Flutter team but regarding the current Flutter code base we have, the current version of the lint will enforce several changes like:

```dart
Timeline.instantSync(
  'ImageCache.evict', 
  arguments: <String, dynamic>{
    'type': 'keepAlive',
    'sizeInBytes': image.sizeBytes,
  },
);
// instead of 
Timeline.instantSync('ImageCache.evict', arguments: <String, dynamic>{
  'type': 'keepAlive',
  'sizeInBytes': image.sizeBytes,
});

// -------------------

FlutterError.reportError(
  FlutterErrorDetails(
    exception: exception,
    stack: stack,
    library: 'animation library',
    context: ErrorDescription('while notifying listeners for $runtimeType'),
    informationCollector: collector,
  ),
);
// instead of
FlutterError.reportError(FlutterErrorDetails(
  exception: exception,
  stack: stack,
  library: 'animation library',
  context: ErrorDescription('while notifying listeners for $runtimeType'),
  informationCollector: collector,
));

// -------------------

_sheetOpacity = _opacityTween.animate(
  CurvedAnimation(
    parent: animation!,
    curve: const Interval(0.9, 1.0),
  ),
);
// instead of
_sheetOpacity = _opacityTween.animate(CurvedAnimation(
  parent: animation!,
  curve: const Interval(0.9, 1.0),
));

// -------------------

child: Builder(
  builder: (BuildContext context) {
    return shouldBuildTab[index] ? widget.tabBuilder(context, index) : Container();
  },
),
// instead of
child: Builder(builder: (BuildContext context) {
  return shouldBuildTab[index] ? widget.tabBuilder(context, index) : Container();
}),

// -------------------

invokeCallback<void>(
  'onStart', 
  () => onStart!(
    ForcePressDetails(
      pressure: pressure,
      globalPosition: _lastPosition.global,
      localPosition: _lastPosition.local,
    ),
  ),
);
// instead of
invokeCallback<void>('onStart', () => onStart!(ForcePressDetails(
  pressure: pressure,
  globalPosition: _lastPosition.global,
  localPosition: _lastPosition.local,
)));
```

Another important issue is well illustrated by tests. If you want to skip a test you will have a 10 lines diff:

```dart
// before skiping
testWidgets('correctly wraps a row', (WidgetTester tester) async {
  // a lot of lines
});
// all the callback need to be re-indent and will end up with a big diff of a single skip added
testWidgets(
  'correctly wraps a row',
  (WidgetTester tester) async {
    // a lot of lines
  },
  skip: isBrowser,
);
```

@goderbauer @Hixie does those changes looks good to you or do you prefer to change the lint to accept the current formatting?

Hixie:
@a14n This is all entirely aesthetics, but I agree with you on many of these.

This one seems pretty bad in particular:
> ```dart
> FlutterError.reportError(
>   FlutterErrorDetails(
>     exception: exception,
>     stack: stack,
>     library: 'animation library',
>     context: ErrorDescription('while notifying listeners for $runtimeType'),
>     informationCollector: collector,
>   ),
> );
> // instead of
> FlutterError.reportError(FlutterErrorDetails(
>   exception: exception,
>   stack: stack,
>   library: 'animation library',
>   context: ErrorDescription('while notifying listeners for $runtimeType'),
>   informationCollector: collector,
> ));
> ```

This one too:
> ```
> child: Builder(
>   builder: (BuildContext context) {
>     return shouldBuildTab[index] ? widget.tabBuilder(context, index) : Container();
>   },
> ),
> // instead of
> child: Builder(builder: (BuildContext context) {
>   return shouldBuildTab[index] ? widget.tabBuilder(context, index) : Container();
> }),
> ```

And I agree with you that this one is egregiously bad, bad enough to want to avoid the lint in its current state:
> ```dart
> // before skiping
> testWidgets('correctly wraps a row', (WidgetTester tester) async {
>   // a lot of lines
> });
> // all the callback need to be re-indent and will end up with a big diff of a single skip added
> testWidgets(
>   'correctly wraps a row',
>   (WidgetTester tester) async {
>     // a lot of lines
>   },
>   skip: isBrowser,
> );
> ```

a14n:
@pq @bwilkerson what do you think about this PR and what can be done?

pq:
Thanks for the nudge @a14n.  I guess I was hoping some easy consensus would shake out of the conversation.  That said, if this change is incompatible w/ @jefflim-google and @lauweijie are currently enforcing with it, the proposed change is kind of breaking and not something I feel comfortable landing w/o agreement from them.

I wonder if any better options recommend themselves if we (re)consider some of `dart format`s strong opinions here?  (fyi @munificent)

Hixie:
FWIW I could live with the first two in my comment above (the builder one in particular is probably fine at the end of the day), but I think the testWidgets one is really bad. It would, for example, mess up the blame on test files when a test gets skipped or unskipped (as well as making those diffs hard to read).

pq:
Awesome.  Thanks @Hixie!

a14n:
>the proposed change is kind of breaking and not something I feel comfortable landing w/o agreement from them.

I made this PR before the lint was released (in a _experimental_ version) and this PR will make the lint trigger less often so I would not consider this change as breaking.

TimWhiting:
I just tried this lint out today, and disabled it within a few minutes because of changes like the ones @a14n mentioned. I have very few lints disabled and this just did not seem helpful in the current state, though I really wanted it and enabled it as soon I saw it was available. I would not consider this change breaking. If the old behaviour is desired at all, at least create a prefer_trailing_commas lint that is more relaxed.

lauweijie:
Thanks for the inputs, some comments:

1. Rerun `dartfmt` on the tests - I think they currently fail after `dartfmt` does it's magic
2. Take a stab at updating the lint `_details` to capture the relaxed rule you're proposing?

Currently:
```
**DO** use trailing commas for all function calls and declarations unless the
function call or definition, from the start of the function name up to the
closing parenthesis, fits in a single line.

**Exception:** If the final parameter/argument is positional (vs named) and is
either a function literal implemented using curly braces, a literal map, a
literal set or a literal array. This exception only applies if the final
parameter does not fit entirely on one line.
```

a14n:
> 1. Rerun `dartfmt` on the tests - I think they currently fail after `dartfmt` does it's magic

There's a section that is specific to the flutter codebase style and I'd like to keep it. I added a comment around that code.

In the linter tests we usually never format the code to be sure to be able to reproduce some special cases. Moreover dartfmt can change its output when a new version lands.

Althoughts running `dartfmt` before applying this lint is recommanded I think it is not required (Flutter codebase is a good example that the lint can be used without dartfmt)

> 2\. Take a stab at updating the lint `_details` to capture the relaxed rule you're proposing?

Done

munificent:
> I wonder if any better options recommend themselves if we (re)consider some of `dart format`s strong opinions here? (fyi @munificent)

I would love to evolve dartfmt towards a single consistent/coherent/consensus style around trailing commas and argument list formatting. But I don't know what that style is, would have to find the time to implement it (whatever it turns out to be), and have absolutely no idea how we'd handle reformatting the millions of lines of existing Dart code in google3 that don't follow it.

I'm not opposed to it, I just don't know how we get there from here, nor do I really know where "there" is.

a14n:
What needs to be done or what is blocking to merge this PR? I'm waiting this relaxing to apply/enable the lint on Flutter.

pq:
Thanks for your patience as we work through this.  This lint was introduced initially internally by a team to meet their specific needs.  We moved it externally in the hopes it would be generally useful.  Ideally this would have been as-is but as you've pointed out, there are possible refinements. What's unclear to me at this point is whether these refinements can be introduced without compromising the initial author's intent and making the lint unusable (or less valuable) internally.  If that causes a fork of the lint to get introduced and maintained internally I think we're back where we started and I'd really like to avoid that (since there are other advantages to having rules externally -- in particular for testing, maintenance and general evolution).  That said, having an external lint that's only useful internally is not ideal either.  The best solution would be to reach some kind of agreement on how this lint could serve everyone.  It feels like we're close but I'll have to defer to @lauweijie and @jefflim-google to chime in.  Another alternative would be to relax the lint and add another targeted one that captures the cases the proposed change would lose.  That would work for me too and I think is more desirable than two slightly different but incompatible rules.

Thanks again for the engagement and continued patience!

a14n:
Now that I've clean up all places in flutter package where the relaxed version of the lint triggered I'd like to enable it. Without feedback here I'm going to open a new PR for a new lint. Is it ok for you?

bwilkerson:
Generally speaking, it is not good for the ecosystem to have multiple lints that differ in minor ways, so I think we need to have a very compelling reason for having more than one lint related to where trailing commas are required.

If we really can't reach any kind of consensus as to how this lint should work, then I think we should take a step back and think about other possible solutions, including (a) not having any such lint until there's more consensus in the community or (b) breaking the lint into multiple smaller lints and requiring users to enable multiple lints in order to express their style.

a14n:
>Generally speaking, it is not good for the ecosystem to have multiple lints that differ in minor ways

+1 

The issues mentioned in several comments make the current version of the lint impossible to use.

I'd love to have feedback about using the relaxed lint vs. using the current lint. Correct me if I'm wrong but I don't thing the difference is huge if you use it after a first dartfmt pass... Unfortunatelly I can run this experiment myself on the internal google project from where the first version of the lint comes.

a14n:
@lauweijie @jefflim-google any chance you could try the relaxed version of the lint on your code and see if the result would be acceptable for you and how is the delta with the current version?

a14n:
Has anyone an idea how to make progress here?

a14n:
>Thanks for your patience as we work through this. 

This PR was opened 2 months ago. I just earned the silver-badge of patience :)

> This lint was introduced initially internally by a team to meet their specific needs. We moved it externally in the hopes it would be generally useful. 

It would be useful if the refinements pointed out by this PR was considerated. But I see the workflow of making an internal lint public as a real problem to make things evolve as this long standing PR shows.
 
> Ideally this would have been as-is but as you've pointed out, there are possible refinements. 

Not only refinements but also blocking issues and bugs.

> What's unclear to me at this point is whether these refinements can be introduced without compromising the initial author's intent and making the lint unusable (or less valuable) internally. If that causes a fork of the lint to get introduced and maintained internally I think we're back where we started and I'd really like to avoid that (since there are other advantages to having rules externally -- in particular for testing, maintenance and general evolution). 

I expected the changes I introduced was soft (it relaxes only a few number of cases) and would not corrupt the initial intent. But everytime I look at this PR to make progress I disagree that "_there are other advantages to having rules externally -- in particular for testing, maintenance and general evolution_" (especially _general evolution_). It's like this code has been written into stone that no one is able to change.

> That said, having an external lint that's only useful internally is not ideal either. 

+1

> The best solution would be to reach some kind of agreement on how this lint could serve everyone. It feels like we're close but I'll have to defer to @lauweijie and @jefflim-google to chime in. 

No comment since you requested feedback.

>Another alternative would be to relax the lint and add another targeted one that captures the cases the proposed change would lose. That would work for me too and I think is more desirable than two slightly different but incompatible rules.

The relaxed version of the lint is a big subset of the initial version. For instance running the initial lint version on linter reports 806 diagnostics and the relaxed version reports 772 diagnostics (~4% of difference - this is really low)
If the loosing cases are really a problem for the internal usage I'd suggest to make an additional lint internally to catch them. But it's hard to think about that without feedback.

>Thanks again for the engagement and continued patience!

I really want this (relaxed) lint :D

pq:
> I really want this (relaxed) lint :D

I hear you!

@Hixie just asked me about scope of application and I realize I don't know...  Are you imaging this would be applied to `flutter/flutter` code, `flutter/plugins` code, default `flutter_lints`, or something else?

/fyi @goderbauer

goderbauer:
> Are you imaging this would be applied to flutter/flutter code, flutter/plugins code, default flutter_lints, or something else?

I don't think this is a great candidate for `flutter_lints` as the lint is more about a specific aesthetics style rather than enforcing good coding practices to avoid bugs. I can see how some projects (especially those using dartfmt) would want to buy into this style, though, and enable a lint like this for their specific project.

I also don't see this as a great candidate for `flutter/flutter` code. My understanding is that the main advantage of adding the trailing comma is to get the formatter to auto-format widget code into something nicer. However, `flutter/flutter` doesn't use the formatter. So, what do we get out of enforcing this?

That aside (and ignoring that this is all about aesthetics), I can live with most of the differences pointed out in https://github.com/dart-lang/linter/pull/2566#issuecomment-816592564 except for the one about `testWidgets`. Producing a giant diff just to skip a test does not seem helpful in code reviews and would make this lint more harmful than useful.

a14n:
> Are you imaging this would be applied to `flutter/flutter` code, `flutter/plugins` code, default `flutter_lints`, or something else?

I already made [several PRs over time](https://github.com/flutter/flutter/pulls?q=is%3Apr+trailing+commas+author%3Aa14n+is%3Aclosed) to add missing trailing commas in Flutter codebase. With this relaxed lint we could enforce this rule.

> I also don't see this as a great candidate for `flutter/flutter` code. My understanding is that the main advantage of adding the trailing comma is to get the formatter to auto-format widget code into something nicer. However, `flutter/flutter` doesn't use the formatter. So, what do we get out of enforcing this?

Enabling this rule on `flutter/flutter` has several upcomes: 

* better diff in PR when you add a new argument/parameter at the end of a list (the diff will not contain the previous line with only a trailing comma added)
* better behaviour when someone copy-paste code to use it somewhere `dart format` is used (see my snippet on https://github.com/dart-lang/linter/issues/1338#issuecomment-459801311)
* better indication for flutter dev in the way the code should be formatted (if the lint appears when I code it reminds me to split every parameter/arguments to have one per line). This leads to more homogeneity in the code style.

Outside of `flutter/flutter` I personnaly will also enable this lint because I prefer the output of dartfmt with trailing commas when the arguments/parameters are splitted on several lines (especially when a widget tree is involved).

Side note: I have a future PR for this lint (once/if this PR lands) to also check trailing commas on set/list/map. This was not initially covered but I think it makes sense to also handle those cases.

a14n:
Has there been private discussions about this PR?

bwilkerson:
Not that I'm aware of. Phil is on vacation this week, so that probably slowed down the discussions.

pq:
> Has there been private discussions about this PR?

As Brian mentioned, I was out of town and just now spinning back up.

@goderbauer: do you have any thoughts wrt @a14n's responses above?

My sense is that landing these changes don't make sense unless flutter wants to adopt this rule broadly in one of their rulesets.  Even then we need to cycle back with @lauweijie and @jefflim-google.

I know this is frustrating...  Sorry for that.  And thanks for your continued perseverance!

a14n:
> My sense is that landing these changes don't make sense unless flutter wants to adopt this rule broadly in one of their rulesets.

Fron @Hixie in https://github.com/dart-lang/linter/issues/1338#issuecomment-459618941

> Flutter would totally use this lint.

Hixie:
It really depends on where we're talking about enabling it. If it's for the default set of lints our developers use when creating a new app with Flutter, then I don't think we should enable any styling-specific lints (that's what the formatter is for if people want something like that, and if they don't, then they also don't want lints — we don't need both enabled). If it's about formatting our own code where we don't use the formatter, then I think it makes sense to enable it if it's sufficiently relaxed to not make things worse than they are. I believe that's what @a14n's changes are about.

jefflim-google:
@lauweijie can you confirm the dependence on the auto formatter for this lint to work properly?

I think it's used for both detecting the violations reliably, and then relies on it for final formatting to be correct?

My feeling on this lint is I'm happy to compromise (relax) the rule *if* it will be taken up by some part of the flutter team, but if there isn't any consensus on this, then I'd prefer to keep it as is.

lauweijie:
Yes, it assumes `dartfmt` has been run on the code.

I'm aligned with @jefflim-google on this.

a14n:
> Yes, it assumes dartfmt has been run on the code.

Actually this lint also works on code that hasn't been formatted with `dartfmt`. I use this relaxed version regularly on Flutter codebase (that don't use dartfmt) without problems.

> My feeling on this lint is I'm happy to compromise (relax) the rule [...]

\o/

> [...] if it will be taken up by some part of the flutter team

That was my goal from the beginning (enable this lint on Flutter codebase) when I tried the original version of the lint and made this PR. 

Another goal is to enable the lint on all my packages but in it's current version there are issues (described in several comments of this PR) that make it impossible.

Moreover this lint has not yet been released publically and I think there will be more +1 on this PR as soon as people will try it.

a14n:
:birthday: :disappointed:

Issue #2573, opened 2021-04-08
[use_build_context_synchronously] missing a prefixed mounted check


pq:
![image](https://user-images.githubusercontent.com/67586/114077446-ab1ab700-985c-11eb-8bde-1c8ad08abda2.png)

pq:
Fixed w/ https://github.com/dart-lang/linter/commit/db022e9f85ce6a466b86be20fc8b9670b27ba1a8.

PR #2575, opened 2021-04-08
attempted repro for 2573

pq:
See #2573.

/cc @bwilkerson 

/fyi @jefflim-google -- I must be missing some context since this doesn't repro 🤔


pq:
stale

PR #2578, opened 2021-04-11
Reduce lines_longer_than_80_chars range to only the offending part of the line

### LINKED ISSUE #2570: Line length errors should start on the last non-offending character on the relevant line
**The problem**

When using the `lines_longer_than_80_chars` linter rule, when a problem is detected, the linter reports the correct line, the character 1:

```bash
info • example\echo_example.dart:1:1 • Avoid lines longer than 80 characters. • lines_longer_than_80_chars
```

In my opinion, it would be better to report:

```bash
info • example\echo_example.dart:1:80 • Avoid lines longer than 80 characters. • lines_longer_than_80_chars
```

I (and I imagine many others) will be using the analyzer inside an editor like VS Code, and the first thing you're going to do upon receiving this error is to go to character 80.

It is worth noting that I am totally blind, and as such am using a screen reader, so this is particularly true, since I can't see whatever code ruler tells you visually where character 80 is. This change would significantly speed up my work flow.

**To Reproduce**

To reproduce, create a dart file, and make a long line of code:

```dart
/// This is going to be such a long comment that the dart linter is bound to complain. I have achieved this by writing rubbish.
```

**Expected behaviour**

In VS Code, I should be taken to the 80th character (the m of complain), and the analyser should show the output I printed above.

Thank you for your consideration.

DanTup:
This came up at https://github.com/Dart-Code/Dart-Code/issues/3263 / https://github.com/dart-lang/linter/issues/2570. Previously the entire line would be included in the range which means clicking the lint would place the caret at the start of the line and not in the area where you need to make the modification. Jumping to 80 saves time time - especially if using a screenreader as the original reporter.

Screenshot showing long lines of code with only the characters past 80 squiggled

Fixes #2570.

### DIFF
--- a/lib/src/rules/lines_longer_than_80_chars.dart
+++ b/lib/src/rules/lines_longer_than_80_chars.dart
@@ -198,7 +198,10 @@ class _Visitor extends SimpleAstVisitor {
       }
       var length = end - start;
       if (length > 80) {
-        var line = _LineInfo(index: i, offset: start, end: end);
+        // Use 80 as the start of the range so that navigating to the lint
+        // will place the caret at exactly the location where the line needs
+        // to wrap.
+        var line = _LineInfo(index: i, offset: start + 80, end: end);
         longLines.add(line);
       }
     }
--- a/test/integration/lines_longer_than_80_chars.dart
+++ b/test/integration/lines_longer_than_80_chars.dart
@@ -32,10 +32,10 @@ void main() {
       expect(
           collectingOut.trim(),
           stringContainsInOrder([
-            'a.dart 7:1 [lint] Avoid lines longer than 80 characters.',
-            'a.dart 11:1 [lint] Avoid lines longer than 80 characters.',
-            'a.dart 20:1 [lint] Avoid lines longer than 80 characters.',
-            'a.dart 25:1 [lint] Avoid lines longer than 80 characters.',
+            'a.dart 7:81 [lint] Avoid lines longer than 80 characters.',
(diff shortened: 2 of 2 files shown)


bwilkerson:
@chrisnorman7 In #2570 you suggested highlighting from the start of the line through the 80th character, placing the end of the highlight range just before where the line needs to wrap. Does it work as well for you to highlight from the 81st character to the end of the line?

chrisnorman7:
Works fine for me mate, absolutely. I only suggested the 80th because
that's where I always go when the line length is too long.

Thank you for your help.

Take care,

Chris Norman

DanTup:
@bwilkerson I think the request was to start (not end) on the last non-offending character (the example output included `line:col` not `offset:length` which perhaps gave the impression of the opposite).

I did slightly misunderstand and started the range on the first offending character though. Although I think that may work better as it doesn't unnecessarily include an character that isn't offending in the range and keeps the squiggle to just the offending part (it may look like a bug if the squiggle includes a single character to the left of the ruler visible in the editor).

It's simple enough to update if there's a preference for something else though. Thanks!

bwilkerson:
I prefer having the range include only characters past the line limit. I think it's a better UX to only highlight the regions that are wrong, and I think that pressing left-arrow ought to place the cursor at the point where the line needs to be broken. I'm happy with the current implementation, I just wanted to make sure it met all the requirements.

bwilkerson (approved):

pq (approved):
👍

pq:
Thanks for the thoughtful back and forth.  LGTM!

PR #2579, opened 2021-04-12
lowerCamelCase check local variables

### LINKED ISSUE #2576: update `non_constant_identifier_names` to include local variables
Currently we don't check local variables in `non_constant_identifiers` but should.   I checked w/ @munificent who confirmed this is consistent w/ the intent of the style guide entry:

https://dart.dev/guides/language/effective-dart/style#do-name-other-identifiers-using-lowercamelcase

I'm guessing this was simply overlooked in implementation.

We'll want to get a sense for how breaking this will be to current adopters and look into internal compatibility but at least @jefflim-google's team has asked for it.

/fyi @davidmorgan 

Also, see https://github.com/dart-lang/linter/issues/330 for other issues related to internal adoption.

pq:
Fixes #2576 

/cc @bwilkerson

### DIFF
--- a/lib/src/rules/non_constant_identifier_names.dart
+++ b/lib/src/rules/non_constant_identifier_names.dart
@@ -48,6 +48,7 @@ class NonConstantIdentifierNames extends LintRule implements NodeLintRule {
     registry.addFunctionDeclaration(this, visitor);
     registry.addMethodDeclaration(this, visitor);
     registry.addVariableDeclaration(this, visitor);
+    registry.addVariableDeclarationStatement(this, visitor);
   }
 }
 
@@ -94,6 +95,13 @@ class _Visitor extends SimpleAstVisitor<void> {
   void visitMethodDeclaration(MethodDeclaration node) {
     if (!node.isOperator) {
       checkIdentifier(node.name);
+    }
+  }
+
+  @override
+  void visitVariableDeclarationStatement(VariableDeclarationStatement node) {
+    for (var variable in node.variables.variables) {
+      checkIdentifier(variable.name);
     }
   }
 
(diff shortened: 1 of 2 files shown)


pq:
FWIW, I'll fix up the `dart format` churn in another PR.

bwilkerson (commented):
I think you also need some tests for variables declared in for loops (both varieties) and in a catch clause. I suspect that you'll need a couple more visit methods to cover those cases.

lib/src/rules/non_constant_identifier_names.dart
```diff
@@ -43,11 +43,14 @@ class NonConstantIdentifierNames extends LintRule implements NodeLintRule {
   void registerNodeProcessors(
       NodeLintRegistry registry, LinterContext context) {
     var visitor = _Visitor(this);
+    registry.addCatchClause(this, visitor);
     registry.addConstructorDeclaration(this, visitor);
+    registry.addForEachPartsWithDeclaration(this, visitor);
     registry.addFormalParameterList(this, visitor);
     registry.addFunctionDeclaration(this, visitor);
     registry.addMethodDeclaration(this, visitor);
     registry.addVariableDeclaration(this, visitor);
```
bwilkerson, line 52: I think this will flag non-local variables (such as top-level variables and fields), but it's hard to imagine that nobody has ever run into this bug before.

On the other hand, I think you need to visit `ForPartsWithDeclarations`. `visitVariableDeclaration` is probably what's causing that test to pass.

pq: > I think this will flag non-local variables (such as top-level variables and fields)

Right.  And we *do* want to flag those.  IOW: feature not  bug?

bwilkerson: Oh, right. I'd gotten it in my head that this lint was for local variables. But it's only this PR that's about local variables.

bwilkerson (approved):

lib/src/rules/non_constant_identifier_names.dart
pq: Thanks for the extra look!

PR #2583, opened 2021-04-13
short circuit await discovery for terminating blocks

pq:
Fixes: #2574

I'm not sure about how (and if) we should be handling `break`s, `continue`, `whiles`s  and `for`-loops.  This doesn't feel complete as is.  (Filling out some test cases would help.)

/cc @bwilkerson


bwilkerson (approved):
I think the PR is an improvement, so I'm OK with you landing it an iterating, but I think it's missing some cases. What I don't know is whether they're cases we care about.

pq:
Thanks!  I'll land and iterate since I'm simultaneously working on a dependent change.

PR #2596, opened 2021-04-16
Fix `directives_ordering` for non-pub packages.

davidmorgan:
`directives_ordering` had a gap for google3 because those don't count as pub packages; this seems to do the job.

### DIFF
--- a/lib/src/rules/directives_ordering.dart
+++ b/lib/src/rules/directives_ordering.dart
@@ -271,7 +271,14 @@ class _Visitor extends SimpleAstVisitor<void> {
     _checkSectionInOrder(lintedNodes, relativeExports);
 
     var projectName = project?.name;
-    if (projectName != null) {
+    if (projectName == null) {
+      // Not a pub package. Package directives should be sorted in one block.
+      var packageImports = importDirectives.where(_isPackageDirective);
+      var packageExports = exportDirectives.where(_isPackageDirective);
+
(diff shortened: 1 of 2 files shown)


lib/src/rules/directives_ordering.dart
```diff
@@ -271,7 +271,14 @@ class _Visitor extends SimpleAstVisitor<void> {
     _checkSectionInOrder(lintedNodes, relativeExports);
 
     var projectName = project?.name;
-    if (projectName != null) {
+    if (projectName == null) {
+      // Not a pub package. Package directives should be sorted in one block.
```
bwilkerson, line 275: It would be good to have some tests for this case, but I don't know whether our current test framework supports test cases in non-pub packages.

bwilkerson (approved):

pq (approved):

lib/src/rules/directives_ordering.dart
davidmorgan: Thanks.

I hadn't noticed the test for this rule under `integration`; now I did notice it and tried to add coverage--but didn't figure out a way to trigger this codepath.

davidmorgan:
Please let me know re: test coverage.

Also, please merge if you want to, I don't have merge privileges here :)

pq:
We can loop back about improved test coverage.  I'll land for now.  Thanks!

PR #2601, opened 2021-04-20
Remove lints without full alignment

goderbauer:
As discussed in our meeting today this PR removes the three lints from the `core.yaml` / `recommend.yaml` set for which we don't have full agreement between Flutter and Dart yet.

For details see also https://github.com/flutter/flutter/issues/78432.


devoncarew:
Thanks for the heads up! Adding @pq and @munificent for review.

munificent (approved):
LGTM, thanks!

pq (approved):
💎

PR #2613, opened 2021-04-28
Add prefer_final_parameters lint

### LINKED ISSUE #1914: Prefer final function parameters
Some people, including myself, prefer `final` function parameters. A linter rule `prefer_final_function_parameters` should detect when a parameter is not `final`.

I don't know if such a rule makes sense and if there are cases where you should not make function parameters `final`.

**Example**
```dart
	public void foo(String bar) {} 		  // invalid
	public void foo(final String bar) {}  // valid
```

parlough:
I know this one doesn't fit everyone's style, but it's how I instinctively write my code and goes well along with `prefer_final_locals`. It's strange to have your locals be `final` but not your parameters which are similar to locally declared variables.

It also goes well with `parameter_assignments` as once they removal those assignments, then this lint can trigger. Then they won't trigger the `parameter_assignments` lint again.

Fixes #1914

### DIFF
--- a/lib/src/rules/prefer_final_parameters.dart
+++ b/lib/src/rules/prefer_final_parameters.dart
@@ -84,14 +84,14 @@ class _Visitor extends SimpleAstVisitor<void> {
   void _reportApplicableParameters(
       FormalParameterList? parameters, FunctionBody body) {
     if (parameters != null) {
-      parameters.parameters.forEach((e) {
-        if (e.isFinal || e.isConst || e is FieldFormalParameter) {
+      parameters.parameters.forEach((param) {
+        if (param.isFinal || param.isConst || param is FieldFormalParameter) {
           return;
         }
-        var declaredElement = e.declaredElement;
+        var declaredElement = param.declaredElement;
         if (declaredElement != null &&
             !body.isPotentiallyMutatedInScope(declaredElement)) {
-          rule.reportLint(e);
+          rule.reportLint(param);
         }
       });
     }
--- a/lib/src/rules/prefer_final_parameters.dart
+++ b/lib/src/rules/prefer_final_parameters.dart
@@ -68,9 +68,9 @@ class PreferFinalParameters extends LintRule implements NodeLintRule {
   void registerNodeProcessors(
       NodeLintRegistry registry, LinterContext context) {
     var visitor = _Visitor(this);
-    registry.addFunctionDeclaration(this, visitor);
-    registry.addMethodDeclaration(this, visitor);
+    registry.addConstructorDeclaration(this, visitor);
     registry.addFunctionExpression(this, visitor);
+    registry.addMethodDeclaration(this, visitor);
   }
 }
 
@@ -85,7 +85,7 @@ class _Visitor extends SimpleAstVisitor<void> {
       FormalParameterList? parameters, FunctionBody body) {
     if (parameters != null) {
       parameters.parameters.forEach((e) {
-        if (e.isFinal || e.isConst) {
+        if (e.isFinal || e.isConst || e is FieldFormalParameter) {
           return;
         }
         var declaredElement = e.declaredElement;
@@ -98,21 +98,14 @@ class _Visitor extends SimpleAstVisitor<void> {
   }
 
   @override
-  void visitFunctionExpression(FunctionExpression node) {
-    var parameters = node.parameters;
-    _reportApplicableParameters(parameters, node.body);
-  }
+  void visitConstructorDeclaration(ConstructorDeclaration node) =>
+      _reportApplicableParameters(node.parameters, node.body);
 
   @override
-  void visitFunctionDeclaration(FunctionDeclaration node) {
-    var functionExpression = node.functionExpression;
-    var parameters = functionExpression.parameters;
-    _reportApplicableParameters(parameters, functionExpression.body);
-  }
+  void visitFunctionExpression(FunctionExpression node) =>
+      _reportApplicableParameters(node.parameters, node.body);
 
   @override
-  void visitMethodDeclaration(MethodDeclaration node) {
-    var parameters = node.parameters;
-    _reportApplicableParameters(parameters, node.body);
-  }
+  void visitMethodDeclaration(MethodDeclaration node) =>
+      _reportApplicableParameters(node.parameters, node.body);
 }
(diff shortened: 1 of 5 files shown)


lib/src/rules/prefer_final_parameters.dart
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  @override
+  void visitFunctionExpression(FunctionExpression node) {
+    var parameters = node.parameters;
+    _reportApplicableParameters(parameters, node.body);
+  }
+
+  @override
+  void visitFunctionDeclaration(FunctionDeclaration node) {
```
bwilkerson, line 107: This doesn't need to visit `FunctionDeclaration`s because they always contain a `FunctionExpression` and the latter will be visited.

lib/src/rules/prefer_final_parameters.dart
```diff
@@ -0,0 +1,118 @@
...
+  @override
+  List<String> get incompatibleRules => const ['unnecessary_final'];
+
+  @override
+  void registerNodeProcessors(
+      NodeLintRegistry registry, LinterContext context) {
+    var visitor = _Visitor(this);
+    registry.addFunctionDeclaration(this, visitor);
+    registry.addMethodDeclaration(this, visitor);
+    registry.addFunctionExpression(this, visitor);
```
bwilkerson, line 73: Consider alphabetizing these registry calls.

Do we want to enforce `final` for constructor parameters? If so, this will need to add `ConstructorDeclaration` to the list and `_reportApplicableParameters` will need to allow `FieldFormalParameter`s to not have a `final`.

test_data/rules/prefer_final_parameters.dart
```diff
@@ -0,0 +1,49 @@
+// Copyright (c) 2021, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// test w/ `dart test -N prefer_final_parameters`
+
+void badMethod(String label) { // LINT
```
bwilkerson, line 7: nit: this is a top-level function, not a method

test_data/rules/prefer_final_parameters.dart
```diff
@@ -0,0 +1,49 @@
...
+void badMixed(String bad, final String good) { // LINT
+  print(bad);
+  print(good);
+}
+
+bool _testingVariable;
+
+void set badSet(bool setting) => _testingVariable = setting; // LINT
+
+var badCallback = (Object random) { // LINT
```
bwilkerson, line 22: FYI, the spec (and most of the documentation) refers to these as closures, not callbacks.

test_data/rules/prefer_final_parameters.dart
```diff
@@ -0,0 +1,49 @@
+// Copyright (c) 2021, the Dart project authors. Please see the AUTHORS file
```
bwilkerson, line 1: Consider adding at least one method and one constructor inside a class. It wouldn't hurt to also have a getter, setter, and operator.

Consider having at least one example each of an optional positional, required named, and optional named parameter.

test_data/rules/prefer_final_parameters.dart
parlough: Good catch, thanks!

lib/src/rules/prefer_final_parameters.dart
parlough: I was planning on support constructor parameters, but completely forgot as I was diving in to the implementation, thanks!

test_data/rules/prefer_final_parameters.dart
parlough: Great ideas, thanks. Let me know what you think of the latest additions.

test_data/rules/prefer_final_parameters.dart
```diff
@@ -0,0 +1,113 @@
...
+
+void set badSet(bool setting) => _testingVariable = setting; // LINT
+
+void set goodSet(final bool setting) => _testingVariable = setting; // OK
+
+var badClosure = (Object random) { // LINT
+  print(random);
+};
+
+var goodClosure = (final Object random) { // OK
```
bwilkerson, line 53: Sorry I didn't think of this on the first pass, but what about a case like
```
var uncertainClosure = (parameterWithNoType) => 0;
```
Such closures are common when passed as arguments to another function. Should those also be `final`? I believe that they will be linted, and I'm guessing that's the style you want to enforce. (A test isn't necessary, I mostly want to know that this is intentional.)

bwilkerson (approved):

lib/src/rules/unnecessary_final.dart
```diff
@@ -44,7 +44,8 @@ class UnnecessaryFinal extends LintRule implements NodeLintRule {
             group: Group.style);
 
   @override
-  List<String> get incompatibleRules => const ['prefer_final_locals'];
+  List<String> get incompatibleRules =>
+      const ['prefer_final_locals', 'prefer_final_parameters'];
```
pq, line 48: 💎

test_data/rules/prefer_final_parameters.dart
pq: Great question.  If the answer is no (and maybe either way) it might be good to update the docs to weigh in.

pq (approved):
Modulo the closure case, LGTM!

test_data/rules/prefer_final_parameters.dart
parlough: You are correct in that they will currently be linted, but do you think that this will discourage people from using the lint as a whole?

I am worried that this structure so common, especially in functional like methods like the ones present in `Stream` and `Iterable` where you end up with multiple calls in a row that accept closures, usually not modifying anything anyway.

Perhaps in this case(where the parameter type can be assumed?) we shouldn't trigger the lint or even remove all the closure cases from this lint introduce a separate one specific to closures?

Let me know what you both think, thanks!

bwilkerson: I don't know for sure, but I do know that when you wrote this lint you didn't make any of the parameters final anywhere. Are you sure this is your preferred style? :-)

As for closures, would you have wanted to have written
```
parameters.parameters.forEach((final e) {
```

parlough: Since we went with the `var` locals in linter, I tried my best to stick with that for parameters as well.

Truly, I would prefer to not write `final e` and instead have that be the default for Dart.

Thinking about it though, while I haven't really used `final` in my closures much, I think it makes sense to do so if I am making the rest of my parameters `final` as well. Especially since the `prefer_final_locals` will trigger in the closure too. The lint is for those who want that extra static guarantee, so I think it fits for those who will use it.

pq:
It looks like this is ready to land?

🦅

PR #2618, opened 2021-05-04
add noop_string_calls

a14n:
Some operation on String are idempotent and return the same string.

**BAD:**

```dart
f() {
  var s = 'hello';
  s + '';
  s * 1;
  s.toString();
  s.substring(0);
  s.padLeft(0);
  s.padRight(0);
}
```

### DIFF
--- a/lib/src/rules/noop_string_calls.dart
+++ b/lib/src/rules/noop_string_calls.dart
@@ -44,6 +44,7 @@ class NoopStringCalls extends LintRule implements NodeLintRule {
     LinterContext context,
   ) {
     var visitor = _Visitor(this, context);
+    registry.addAdjacentStrings(this, visitor);
     registry.addBinaryExpression(this, visitor);
     registry.addMethodInvocation(this, visitor);
   }
@@ -54,6 +55,15 @@ class _Visitor extends SimpleAstVisitor<void> {
 
   final LintRule rule;
   final LinterContext context;
+
+  @override
+  void visitAdjacentStrings(AdjacentStrings node) {
+    for (var literal in node.strings) {
+      if (literal.stringValue?.isEmpty ?? false) {
+        rule.reportLint(literal);
+      }
+    }
+  }
 
   @override
   void visitBinaryExpression(BinaryExpression node) {
--- a/lib/src/rules/noop_string_calls.dart
+++ b/lib/src/rules/noop_string_calls.dart
@@ -61,31 +61,31 @@ class _Visitor extends SimpleAstVisitor<void> {
 
     var right = node.rightOperand;
 
-    // string.toString()
+    // string + ''
     if (node.operator.type == TokenType.PLUS &&
         right is StringLiteral &&
-        right.stringValue == '') {
-      rule.reportLint(node);
+        (right.stringValue?.isEmpty ?? false)) {
+      rule.reportLintForToken(node.operator);
       return;
     }
 
-    // string.toString()
+    // string * 1
     if (node.operator.type == TokenType.STAR &&
         right is IntegerLiteral &&
         right.value == 1) {
-      rule.reportLint(node);
+      rule.reportLintForToken(node.operator);
       return;
     }
   }
 
   @override
   void visitMethodInvocation(MethodInvocation node) {
-    if (!(node.target?.staticType?.isDartCoreString ?? false)) return;
+    if (!(node.realTarget?.staticType?.isDartCoreString ?? false)) return;
 
     // string.toString()
     if (node.methodName.name == 'toString' &&
         context.typeSystem.isNonNullable(node.target!.staticType!)) {
-      rule.reportLint(node);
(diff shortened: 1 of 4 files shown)


lib/src/rules/noop_string_calls.dart
```diff
@@ -0,0 +1,111 @@
...
+  final LintRule rule;
+  final LinterContext context;
+
+  @override
+  void visitBinaryExpression(BinaryExpression node) {
+    if (!(node.leftOperand.staticType?.isDartCoreString ?? false)) return;
+
+    var right = node.rightOperand;
+
+    // string.toString()
```
bwilkerson, line 64: Wrong comment, here and below.

lib/src/rules/noop_string_calls.dart
```diff
@@ -0,0 +1,111 @@
...
+    if (node.operator.type == TokenType.PLUS &&
+        right is StringLiteral &&
+        right.stringValue == '') {
```
bwilkerson, line 67: Consider using `isEmpty`.

lib/src/rules/noop_string_calls.dart
```diff
@@ -0,0 +1,111 @@
...
+        right is IntegerLiteral &&
+        right.value == 1) {
+      rule.reportLint(node);
+      return;
+    }
+  }
+
+  @override
+  void visitMethodInvocation(MethodInvocation node) {
+    if (!(node.target?.staticType?.isDartCoreString ?? false)) return;
```
bwilkerson, line 83: Consider using `realTarget` rather than `target` to catch invocations in cascades.

lib/src/rules/noop_string_calls.dart
```diff
@@ -0,0 +1,111 @@
...
+
+    // string.toString()
+    if (node.methodName.name == 'toString' &&
+        context.typeSystem.isNonNullable(node.target!.staticType!)) {
+      rule.reportLint(node);
```
bwilkerson, line 88: In all of these invocation-based cases, consider highlighting the name of the method being invoked rather than the whole invocation.

lib/src/rules/noop_string_calls.dart
```diff
@@ -0,0 +1,111 @@
...
+  void visitBinaryExpression(BinaryExpression node) {
+    if (!(node.leftOperand.staticType?.isDartCoreString ?? false)) return;
+
+    var right = node.rightOperand;
+
+    // string.toString()
+    if (node.operator.type == TokenType.PLUS &&
+        right is StringLiteral &&
+        right.stringValue == '') {
+      rule.reportLint(node);
```
bwilkerson, line 68: In both of these operator-based cases, consider highlighting the operator (or the operator and the right operand) rather than the whole expression.

bwilkerson (commented):
Do we have any evidence that these cases occur frequently enough that a rule to catch them would add value?

a14n:
>Do we have any evidence that these cases occur frequently enough that a rule to catch them would add value?

This kind of lint was requested in https://github.com/flutter/flutter/pull/81634#issuecomment-830742979. You can see what it catches on flutter at https://github.com/flutter/flutter/pull/81859/files (mostly `string.toString()`)

bwilkerson:
Are you thinking of having separate lints for `toDouble` and `round`, or not checking those at all?

If we're not adding support for those other cases now, should we generalize this rule to allow it to cover other cases like those above in the future (such as `noop_invocations`?

a14n:
I was planning to have separate lints `noop_string_calls`, `noop_double_calls`, `noop_int_calls` to keep each lint focus and simple.

a14n:
BTW we could also flag empty string in adjacent strings.

a14n:
After taking a closer look at lints on String, int, and double we should perhaps merge all the cases in a `noop_core_expressions` (or `noop_invocations` if we discard linting empty String in adjacent strings) that will cover only the most frequent issues :

```dart
  doubleValue.toDouble(); 

  intValue.toInt(); 
  intValue.round(); 
  intValue.ceil(); 
  intValue.floor(); 
  intValue.truncate(); 
  
  s.toString();
  s = 'hello\n'
      'world\n'
      ''; // LINT
```

WDYT?

bwilkerson:
I like the idea of flagging an empty string in adjacent strings.

I'm fine with making this a single lint. Perhaps rather than 'core' it should be 'primitive types' as the distinguishing characteristic. If we ever decide to have lints for things like `list.sublist(0)` those would probably want to be in a separate lint even though they're defined in `core`. Perhaps `noop_primitive_operations`? Happy to have a better name if you can think of one.

I don't have any problem with catching a few less common cases. It wasn't the value of the individual cases that I wanted to ensure, it was the value of the lint overall, and I'm now convinced that it's reasonable to add it.

a14n:
I will open another PR for `noop_primitive_operations`. Thanks for your feedback!

PR #2638, opened 2021-05-11
refactor `prefer_contains` analysis to use visitMethodInvocation

pq:
Another significant gain moving away from visiting simple identifiers. (5ms vs 26)

/cc @bwilkerson

### DIFF
--- a/lib/src/rules/prefer_contains.dart
+++ b/lib/src/rules/prefer_contains.dart
@@ -85,8 +85,7 @@ class _Visitor extends SimpleAstVisitor<void> {
 
     var parentType = node.target?.staticType;
     if (parentType == null ||
-        !DartTypeUtilities.implementsAnyInterface(
-            parentType, <InterfaceTypeDefinition>[
+        !DartTypeUtilities.implementsAnyInterface(parentType, [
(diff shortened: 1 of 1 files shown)


lib/src/rules/prefer_contains.dart
```diff
@@ -79,37 +78,22 @@ class _Visitor extends SimpleAstVisitor<void> {
...
-
-    if (!DartTypeUtilities.implementsAnyInterface(
-        type, <InterfaceTypeDefinition>[
-      InterfaceTypeDefinition('Iterable', 'dart.core'),
-      InterfaceTypeDefinition('String', 'dart.core'),
-    ])) {
+    var parentType = node.target?.staticType;
+    if (parentType == null ||
+        !DartTypeUtilities.implementsAnyInterface(
+            parentType, <InterfaceTypeDefinition>[
```
bwilkerson, line 89: nit: Note that you probably don't need the type argument list in front of the list literal. It ought to be inferred.

bwilkerson (approved):

lib/src/rules/prefer_contains.dart
pq: Good catch.

PR #2648, opened 2021-05-12
relax sort_child_properties_last to accept other widget afterward

a14n:
Some widget may take several widgets as parameter and the `child` one may not be the good one to put at the last position. That's why this PR skip this case where a widget is used after `child`.

For instance in Flutter:

```
ExpansionTile(
  initiallyExpanded: true,
  title: TestText('Expanded', key: expandedTitleKey),
  backgroundColor: Colors.red,
  children: const <Widget>[ListTile(title: Text('0'))],
  trailing: TestIcon(key: expandedIconKey),
)
```


test_data/rules/sort_child_properties_last.dart
```diff
@@ -114,3 +114,19 @@ class WithClosure extends Widget {
...
+
+  @override
+  Widget build(BuildContext context) {
+    return WithSeveralWidgetCreation(
+      header: Text('a'),
+      child: Text('b'), // OK
+      footer: Text('c'),
+    );
+  }
+}
```
jacob314, line 132: nit: missing trailing line break.

lib/src/rules/sort_child_properties_last.dart
```diff
@@ -75,8 +75,11 @@ return Scaffold(
 );
 ```
 
-Exception: It's allowed to have parameter with a function expression after the
```
jacob314, line 78: I'd rather treat the two exceptions consistently by allowing widget parameters after child like you add function parameters after child.
For example given:

```dart
class WithSeveralWidgetCreation extends Widget {
  WithSeveralWidgetCreation({this.boldHeader, this.header, this.child, this.footer});
  final bool boldHeader;
  final Widget header;
  final Widget child;
  final Widget footer;

  @override
    return WithSeveralWidgetCreation(
      header: Text('a'),
      child: Text('b'), // OK
      footer: Text('c'),
    );
    return WithSeveralWidgetCreation(
      header: Text('a'),
      child: Text('b'), // OK
      footer: Text('c'),
      boldHeader: true, // LINT_ERROR
    );
  }
}
```

a14n: I initially wanted something like that but reverted my changes to have diagnostics consistent with the name of the lint.

bwilkerson: Perhaps the name of the lint needs to be changed. It doesn't seem to reflect the style guideline that it's moving toward supporting. Let's start by agreeing what style we want to support (an issue might be better than a PR for that purpose) and then name the lint to match the style.

It sounds to me like the real style guideline is to put multi-line named arguments at the end of the argument list. As a proxy we're looking at arguments of type `Widget`, `List<Widget>` and `Function`, but one question is whether it's necessary to use a proxy, or whether we can just allow any named argument whose value is computed by an expression that's too long to fit on one line. Doing so would certainly help make the lint more general. Perhaps it should be named `sort_multiline_arguments_last`.

If the lint must be specific to Flutter, then perhaps it should be named something like `sort_widget_and_function_arguments_last`.

a14n: I may be wrong but I don't think we want generalize this lint too much. IIRC this lint was introduced to help the UI guide to play nicely on the code. I'm not sure UI guide treat other field name than `child` or `children`. This change is just a pragmatic change regarding the issues I faced applying it on Flutter codebase to avoid some `// ignore` that are not greatly welcome.

a14n: @jacob314 WDYT?

jacob314: sorry I missed this. UI guides don't key off child and children. They key off the Flutter outline with is based on the types of the parameters of the widget.

a14n: Re-reading the initial issue/PR about `sort_child_properties_last` I think [this comment](https://github.com/dart-lang/linter/issues/1542#issuecomment-490584598) is what should be done (in an ideal world) but unfortunately it would required reviewing/changing a lot of declarations to sort parameters correctly (not only in Flutter codebase). For instance [ExpansionTile](https://github.com/flutter/flutter/blob/58ab556f4161548ae9cc65e0d231f22682c0f35d/packages/flutter/lib/src/material/expansion_tile.dart#L39-L59) has almost all its properties after `children`/`trailing`/...

That's the reason I've simply disabled the lint if several widgets appear as argument. For now we can not enable this lint in Flutter because the logical order would be broken.

a14n:
PTAL

jacob314 (approved):
lgtm

lib/src/rules/sort_child_properties_last.dart
```diff
@@ -108,31 +108,27 @@ class _Visitor extends SimpleAstVisitor {
...
-        .where((element) =>
-            element is NamedExpression &&
-            element.expression is! FunctionExpression)
-        .isEmpty;
-    if (!onlyClosuresAfterChild) {
-      rule.reportLint(arguments.firstWhere(isChildArg));
+    var reversedArguments = arguments.reversed.toList();
+    var lastWidget = reversedArguments
+        .where((e) => isWidgetProperty(e.staticType))
+        .firstOrNull;
```
bwilkerson, line 118: Seem like this would be more efficient if it used `lastIndexWhere`. We wouldn't need to build `reversedArguments` and we could just iterate from `index` to `length` below when looking for a non-closure rather than build iterators. But maybe iterators are more performant than I'm guessing they are. No need to change this if you're reasonably confident that it wouldn't provide a performance boost.

bwilkerson (approved):

lib/src/rules/sort_child_properties_last.dart
a14n: Thanks for the suggestion. Done.

a14n:
@pq I think the PR is now ready to be merged

PR #2653, opened 2021-05-14
remove redundant/deprecated rules

pq:
/cc @bwilkerson

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -13,14 +13,12 @@ linter:
     - always_put_required_named_parameters_first
 #    - always_specify_types
     - avoid_annotating_with_dynamic
-#    - avoid_as
 #    - avoid_bool_literals_in_conditional_expressions # under review (see #1068)
     - avoid_catches_without_on_clauses
     - avoid_catching_errors
 #    - avoid_classes_with_only_static_members
 #    - avoid_double_and_int_checks # under review (see #1068)
 #    - avoid_field_initializers_in_const_classes # under review (see #1068)
-#    - avoid_function_literals_in_foreach_calls
 #    - avoid_implementing_value_types
 #    - avoid_js_rounded_ints # under review (see #1068)
     - avoid_positional_boolean_parameters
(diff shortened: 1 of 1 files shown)


bwilkerson (approved):
Instead of deleting these lines do we want to just annotate them with the reason why they're commented out? I don't know whether this list is normally complete or whether there's a different reason for all the commented out lint names.

pq:
I'm not sure to be honest but my inclination is to remove all the commented out lints unless they're under active consideration.  I think it's just noise.  If folks think there's value I'd consider inlining commented out core and recommended lints for reference.

/cc @a14n @parlough for thoughts.

a14n:
On flutter we have the full list in the root [analysis_options.yaml](https://github.com/flutter/flutter/blob/master/analysis_options.yaml#L46) and a trailing comment on disabled lints explaining why there are commented out.

a14n:
On flutter we have the full list in the root [analysis_options.yaml](https://github.com/flutter/flutter/blob/master/analysis_options.yaml#L46) and a trailing comment on disabled lints explaining why there are commented out.

parlough:
I think for the case of these, it's fine to just include the ones under active consideration with either inline information or a link. 

If you want to annotate each one with why it's not enabled, I suppose that would work as well, but not sure it's necessary for the linter. It makes more sense for Flutter where it's a high traffic repository and users will refer to it as a reference.

bwilkerson:
I'm not really trying to argue for either approach, but if you annotate them then you avoid the possibility re-evaluating a lint multiple times because you can't remember whether you've already evaluated it. Of course, if all of the evaluation is in issues, then an issue search would also serve to avoid that situation.

Mostly I just noticed that we were keeping some commented out and deleting others, and the inconsistency made me raise the question.

pq:
I've written and deleted a few responses that almost contradict each other.  Can you tell I'm entirely on the fence? 😄 

I do agree w/ Brian that the inconsistency is not great but am open to ideas.  (FWIW: if we do add annotated references to all rules, I'd want a validation step to make sure it's up to date w/ the latest published lints.)

PR #2659, opened 2021-05-19
add lint always_depend_on_packages_you_use

### LINKED ISSUE #29: Lint for used package not defined in pubspec

jakemac53:
# Description

Adds a new lint, `always_depend_on_packages_you_use`. This applies to all imports and exports.

For files under the `lib` or `bin` dir, only packages in the `dependencies` section may be referenced. For all other files any package listed in `dev_dependencies` is also allowed.

Fixes #29

### DIFF
--- a/lib/src/rules/always_depend_on_packages_you_use.dart
+++ b/lib/src/rules/always_depend_on_packages_you_use.dart
@@ -1,0 +1,112 @@
+// Copyright (c) 2021, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+
+import '../analyzer.dart';
+import '../ast.dart';
+
+const _desc = r'Depend on packages you use.';
+
+const _details = r'''
+
+**DO** depend on packages you use.
+
+When importing a package, *always* add a dependency on it to your pubspec.
+
+Depending explicitly on packages that you use ensures they will always exist
+and allows you to put a dependency constraint on them to guard you against
+breaking changes.
+
+Whether this should be a regular dependency or dev_dependency depends on if it
+is imported from a public file (one under either `lib` or `bin`), or some other
+file.
+
+**BAD:**
+```dart
+import 'package:a/a.dart';
+import 'package:b/b.dart';
+```
+
+```yaml
+dependencies:
+  a: ^1.0.0
+```
+
+**GOOD:**
+```dart
+import 'package:a/a.dart';
+import 'package:b/b.dart';
+```
+
+```yaml
+dependencies:
+  a: ^1.0.0
+  b: ^1.0.0
+```
+
+''';
+
+class AlwaysDependOnPackagesYouUse extends LintRule implements NodeLintRule {
+  AlwaysDependOnPackagesYouUse()
+      : super(
+            name: 'always_depend_on_packages_you_use',
+            description: _desc,
+            details: _details,
+            group: Group.pub);
+
+  @override
+  void registerNodeProcessors(
+      NodeLintRegistry registry, LinterContext context) {
+    // Only lint if we have a pubspec.
+    var package = context.package;
+    if (package is! PubWorkspacePackage) return;
+    var pubspec = package.pubspec;
+    if (pubspec == null) return;
+
+    var visitor = _Visitor(this, context, pubspec,
+        isPublicFile: isInPublicDir(context.currentUnit.unit, context.package));
+    registry.addImportDirective(this, visitor);
+    registry.addExportDirective(this, visitor);
+  }
+}
+
+class _Visitor extends SimpleAstVisitor {
+  final AlwaysDependOnPackagesYouUse rule;
+  final LinterContext context;
+  final Pubspec pubspec;
+  final bool isPublicFile;
+  late final availableDeps = [
+    if (pubspec.dependencies != null)
+      for (var dep in pubspec.dependencies!)
+        if (dep.name?.text != null) dep.name!.text!,
+    if (!isPublicFile && pubspec.devDependencies != null)
+      for (var dep in pubspec.devDependencies!)
+        if (dep.name?.text != null) dep.name!.text!
+  ];
+
+  _Visitor(this.rule, this.context, this.pubspec, {required this.isPublicFile});
+
+  void _checkDirective(UriBasedDirective node) {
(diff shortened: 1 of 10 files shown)


lib/src/rules/always_depend_on_packages_you_use.dart
```diff
@@ -0,0 +1,112 @@
...
+and allows you to put a dependency constraint on them to guard you against
+breaking changes.
+
+Whether this should be a regular dependency or dev_dependency depends on if it
+is imported from a public file (one under either `lib` or `bin`), or some other
+file.
+
+**BAD:**
+```dart
+import 'package:a/a.dart';
```
bwilkerson, line 29: Consider removing the import on line 29. It isn't a bad import so it might be confusing to readers.

lib/src/rules/always_depend_on_packages_you_use.dart
```diff
@@ -0,0 +1,112 @@
...
+    registry.addExportDirective(this, visitor);
+  }
+}
+
+class _Visitor extends SimpleAstVisitor {
+  final AlwaysDependOnPackagesYouUse rule;
+  final LinterContext context;
+  final Pubspec pubspec;
+  final bool isPublicFile;
+  late final availableDeps = [
```
bwilkerson, line 81: Consider initializing this in the constructor in order to avoid the uses of the null check operator.

lib/src/rules/always_depend_on_packages_you_use.dart
```diff
@@ -0,0 +1,112 @@
...
+        if (dep.name?.text != null) dep.name!.text!,
+    if (!isPublicFile && pubspec.devDependencies != null)
+      for (var dep in pubspec.devDependencies!)
+        if (dep.name?.text != null) dep.name!.text!
+  ];
+
+  _Visitor(this.rule, this.context, this.pubspec, {required this.isPublicFile});
+
+  void _checkDirective(UriBasedDirective node) {
+    // Is it a package: import?
```
bwilkerson, line 93: nit: "import" --> "URI". Also rename local variables.

test/integration/always_depend_on_packages_you_use.dart
```diff
@@ -0,0 +1,93 @@
+// Copyright (c) 2020, the Dart project authors. Please see the AUTHORS file
```
bwilkerson, line 1: nit: "2020" --> "2021"

bwilkerson (approved):

example/all.yaml
```diff
@@ -3,6 +3,7 @@
 linter:
   rules:
     - always_declare_return_types
+    - always_depend_on_packages_you_use
```
pq, line 6: in general I think we're trending away from "always" and similar prefixes.  Bike shedding names...

* `direct_package_dependencies`

?

pq: `explicit_package_dependencies`

?

pq:
So awesome.  Modulo name bike-shedding, this looks great!

lib/src/rules/always_depend_on_packages_you_use.dart
```diff
@@ -0,0 +1,112 @@
...
+      var importUri = Uri.parse(importUriContent);
+      if (importUri.pathSegments.isEmpty) return;
+      var packageName = importUri.pathSegments.first;
+      if (availableDeps.contains(packageName)) return;
+      rule.reportLint(node.uri);
+    } on FormatException catch (_) {}
+  }
+
+  @override
+  void visitImportDirective(ImportDirective node) => _checkDirective(node);
```
pq, line 108: @bwilkerson: do you think this would be better done at the compilation unit level?  That way we would only build the list of available deps once per compilation (not repeatedly for each directive).

Notice the relative perf: https://github.com/dart-lang/linter/pull/2659/checks?check_run_id=2622882764#step:5:43

Seems like we can do better...

lib/src/rules/always_depend_on_packages_you_use.dart
jakemac53: It feels a bit weird having the dependencies totally empty in the pubspec

example/all.yaml
bwilkerson: Or `depend_on_referenced_packages`

lib/src/rules/always_depend_on_packages_you_use.dart
jakemac53: I can try using an iterable instead of a list also

bwilkerson: > That way we would only build the list of available deps once per compilation (not repeatedly for each directive).

The available deps is cached in a field and is only built once, so there shouldn't be any duplicated effort there.

jakemac53: Ok I tried a few things here. The most impactful one by far seemed to be not using `Uri.parse` and doing a bit of hacky munging instead.

I did also move up the creation of this list to the rule itself, and passed that into the visitor. That has the effect I believe of doing it only once per file. I could cache it per package as well, but I don't know how invalidation would work?

I also tried using an iterable and a set to see if they would help but they seemed the same.

I will push up the change to avoid Uri.parse, PTAL.

example/all.yaml
jakemac53: I am fine with any of these, I like `depend_on_referenced_packages` probably the most?

lib/src/rules/always_depend_on_packages_you_use.dart
jakemac53: https://github.com/dart-lang/linter/runs/2623499572#step:5:115 just a wee bit better ;)

jakemac53: > The available deps is cached in a field and is only built once, so there shouldn't be any duplicated effort there.

Oh true, I didn't need to hoist it. Happy to undo that change if you would like, but this seems a bit better?

bwilkerson: Either way is fine with me. We can look at caching the list with the `WorkspacePackage` when we move the API.

example/all.yaml
jakemac53: I went with that name for now, LMK if you want to change it

pq:
`depend_on_referenced_packages` works for me.  @bwilkerson: WDYT?

pq:
And hey, nice perf improvement!

https://github.com/dart-lang/linter/pull/2659/checks?check_run_id=2624270876#step:5:111

Looks like this is now down from 15ms in benchmarks to 2!

pq (approved):

PR #2660, opened 2021-05-20
fix depend_on_referenced_packages lint to allow imports of the current package

jakemac53:
I realized last night this would trigger for the current package as well, which we definitely don't want 🤣 .

This fixes that, and also makes the tests a bit more robust by ensuring we don't lint for any lines with `// OK` in them.

Also fixes some new unnecessary import lints (this lint now triggers for imports that don't expose anything not exposed already by another import).

### DIFF
--- a/lib/src/analyzer.dart
+++ b/lib/src/analyzer.dart
@@ -9,13 +9,13 @@ import 'dart:io';
 import 'dart:math' as math;
 
 import 'package:analyzer/error/error.dart';
+import 'package:analyzer/src/generated/engine.dart' show AnalysisErrorInfo;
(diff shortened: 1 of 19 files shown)


pq (approved):
💎

pq:
modulo some unnecessary imports, LGTM!

jakemac53:
Fixed the new import warnings @pq so this should be good to go :).

There was actually a weird pattern here that I think was triggering a bug in the new unnecessary import stuff - but I can't create a clean repro :/.

pq:
Interesting!  If you figure it out let me know.  Thanks!

pq (approved):

Issue #2637, opened 2021-05-11
False positive on only_throw_errors lint


jonahwilliams:
The following code leads to an `only_throw_errors` diagnostic, but `foo` is known to be an Exception at this point. Casting to exception leads to an unnecessary cast lint. This does not repro if the static type of `foo` is changed to Object or dynamic.

```dart
abstract class Foo {}

void bar(Foo foo) {
  if (foo is Exception) {
    throw foo;
  }
}
```

@pq

bwilkerson:
I believe that the reason for the observed behavior is that the `foo is Exception` doesn't promote `foo` because there is no relationship between `Foo` and `Exception`. The reason is works in the second case is because there _is_ a relationship between `Foo` and `FooException`.

I'm not sure that we can "fix" the lint, given that the type system can't tell us that `foo` is an `Exception`.

jonahwilliams:
Then maybe the problem is the cast diagnostic instead? This doesn't repro on my example but in the issue I linked I also got a diagnostic for an unnecessary cast error in flutter_reactive_ble - so that example might be different.

jonahwilliams:
i.e. might be a bad reduced example

bwilkerson:
Ah, then yes, this is a bug in the lint. When we have a `TypeParameterType` we need to check the `promotedBound` as well as the original type.

srawlins:
Given that `dynamic` is allowed by the lint, I think an unbound type parameter should also be allowed:

```dart
f(dynamic e) => throw e; // OK
g<T>(T e) => throw e; // OK
```

PR #2661, opened 2021-05-20
Fix only_throw_errors and prefer_contains for type variables

srawlins:
# Description

This updates prefer_contains and only_throw_errors to use a DartType's effective type when doing interface checks. For example, given the type variable `T extends Foo`, the check will use `Foo`. Given type parameters `T` and `U extends Foo`, and a type promotion like `f(T e) { if (e is U) { ... } }`, the check will use Foo.

I've added tests for both only_throw_errors and prefer_contains.

Fixes https://github.com/dart-lang/linter/issues/2637

### DIFF
--- a/lib/src/util/dart_type_utilities.dart
+++ b/lib/src/util/dart_type_utilities.dart
@@ -2,9 +2,6 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-// See https://github.com/dart-lang/linter/issues/2079.
-// ignore_for_file: comment_references
-
 import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/element/element.dart';
 import 'package:analyzer/dart/element/type.dart';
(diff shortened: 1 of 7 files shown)


lib/src/util/dart_type_utilities.dart
```diff
@@ -560,3 +565,24 @@ class InterfaceTypeDefinition {
         library == other.library;
   }
 }
+
+extension DartTypeExtensions on DartType {
+  /// Returns the type which should be used when conducting "interface checks"
+  /// on [this].
```
bwilkerson, line 571: Does Dartdoc handle `[this]`? The analyzer doesn't because there isn't anywhere interesting to link it to, so we've aways written it as in-line code (I can't figure out how to escape a backquote to illustrate).

bwilkerson (approved):

pq:
Nice!

Am I right in thinking (from #2637) that the user-facing effect of this change is generally fewer false positives?

(Thinking about ecosystem impact.)

lib/src/util/dart_type_utilities.dart
pq: /fyi @jcollins-g

pq (approved):
👍

lib/src/util/dart_type_utilities.dart
srawlins: Reverted.

srawlins:
Actually, its more complicated 😁 😁 😁 

* For only_throw_errors, fewer false positives will be reported.
* For prefer_contains, more real positives will be reported (type parameters promoted to or bound to Iterable).

srawlins:
Given that this CL may _increase_ diagnostics, should I scrap the fix for prefer_contains and file a bug to be resolved with a breaking change and migration guide?

pq:
We should talk about whether it's "breaking" exactly but I would definitely favor separating that change so we can do some assessment of its downstream impact and figure out whether we need a guide and do a bunch of internal (and Flutter) cleanup.  So yeah, maybe a bug would be good.  Thanks!

srawlins:
I cannot merge. :(

> The base branch restricts merging to authorized users. Learn more about protected branches.

PR #2667, opened 2021-05-21
prefer_interpolation_to_compose_strings should apply only for String.+

### LINKED ISSUE #792: prefer_interpolation_to_compose_strings is applied too broadly
Hi, it's me again. :)
I'm doing something like this frequently in my code:
```dart
Route get route => client.api + 'guilds/$id';
```
This triggers the `prefer_interpolation_to_compose_strings` rule.  However, `client.api` is a `Route` instance, and not a string instance.  The rule interprets this as a string being composed; thus it lints this case.

a14n:
# Description

`prefer_interpolation_to_compose_strings` should apply only if the left operand has a `String` type.

Fixes #792


bwilkerson (approved):

pq (approved):
💎

pq:
This is fantastic.  Thank you!

PR #2676, opened 2021-05-26
Report unnecessary_parenthesis when the parent precedence is equal to the expression precedence.

scheglov:
The internal presubmit looks green.
https://test.corp.google.com/ui#id=OCL:375864901:BASE:375864817:1622008515263:e0e91d35&include_skipped=true


pq:
I do wonder if this will identify any new issues in Flutter (or the ecosystem) but this seems like a change in the right direction, so let's go for it.

Thanks!

pq (approved):

bwilkerson (approved):

a14n:
https://github.com/flutter/flutter/pull/83433 fixes the new diagnotics.

PR #2684, opened 2021-05-30
Update use_key_in_widget_constructors.dart

Semvrij:
"The parameter 'key' can't have a value of 'null' because of its type, but the implicit default value is 'null'.
Try adding either an explicit non-'null' default value or the 'required' modifier."


Semvrij:
@googlebot I signed it!

a14n (approved):
This lint landed before the nullsafety introduction. This example is valid for dart<2.12 but I think it makes sense to migrate this example to nullsafety.

bwilkerson (approved):

pq:
Thanks for the contribution!

pq (approved):

PR #2688, opened 2021-06-01
Make `directives_ordering` match "organize imports" in the analyzer.

davidmorgan:
I added the tests I could find in the analyzer about sort ordering--not enough coverage by itself but it's a start.

Unfortunately it looks like we do need a breaking change to the lint. At least, it's not a _new_ breaking change with the last change to the lint, it was there before :)

### DIFF
--- a/lib/src/rules/directives_ordering.dart
+++ b/lib/src/rules/directives_ordering.dart
@@ -359,7 +359,7 @@ class _Visitor extends SimpleAstVisitor<void> {
         var directiveUri = directive.uriContent;
         if (previousUri != null &&
             directiveUri != null &&
-            previousUri.compareTo(directiveUri) > 0) {
+            compareDirectives(previousUri, directiveUri) > 0) {
           reportDirective(directive);
         }
       }
@@ -372,4 +372,18 @@ class _Visitor extends SimpleAstVisitor<void> {
 
   Iterable<ImportDirective> _getImportDirectives(CompilationUnit node) =>
       node.directives.whereType<ImportDirective>();
+}
+
+/// Compares directives by package then file in package.
+///
+/// Package is everything until the first `/`.
+int compareDirectives(String a, String b) {
+  if (!a.contains('/') || !b.contains('/')) {
+    return a.compareTo(b);
+  }
+  var as = a.split('/');
+  var bs = b.split('/');
+  var result = as[0].compareTo(bs[0]);
+  if (result != 0) return result;
+  return as.skip(1).join('/').compareTo(bs.skip(1).join('/'));
 }
--- a/lib/src/rules/directives_ordering.dart
+++ b/lib/src/rules/directives_ordering.dart
@@ -378,12 +378,10 @@ class _Visitor extends SimpleAstVisitor<void> {
 ///
 /// Package is everything until the first `/`.
 int compareDirectives(String a, String b) {
-  if (!a.contains('/') || !b.contains('/')) {
-    return a.compareTo(b);
-  }
-  var as = a.split('/');
-  var bs = b.split('/');
-  var result = as[0].compareTo(bs[0]);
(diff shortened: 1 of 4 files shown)


pq:
Thanks for this fix!  My biggest concern vis a vis breaking is how impacting this will be for Flutter and the ecosystem.

> At least, it's not a new breaking change with the last change to the lint, it was there before :)

Can you elaborate on this?  Not sure I understand...  How breaking is this?  Or maybe, better still, could you provide some prose to describe the change?

Many thanks.

davidmorgan:
The change is that previously the lint expected URIs to be sorted as strings; now they must be sorted as pairs of strings, where the first element is the package and the second the path to the file in the package.

But now that I think about it, I think this only matters when there is a `.` in the package name. Otherwise, they happen to be the same because `/` sorts before letters.

This test case highlights the change; before:

```
import 'package:product.ui/entity.dart';
import 'package:product.ui.api/entity1.dart';
import 'package:product.ui.api.aaa/manager2.dart';
import 'package:product.ui.api.bbb/manager1.dart';
import 'package:product.ui.api/entity2.dart';
import 'package:product2.client/entity.dart';
```

now:

```
import 'package:product.ui/entity.dart';
import 'package:product.ui.api/entity1.dart';
import 'package:product.ui.api/entity2.dart';
import 'package:product.ui.api.aaa/manager2.dart';
import 'package:product.ui.api.bbb/manager1.dart';
import 'package:product2.client/entity.dart';
```

pq:
Ah, got it.  So maybe more impacting than I'd hoped.  We'll want to work with @goderbauer to vet impact and then be careful about roll-out.

lib/src/rules/directives_ordering.dart
```diff
@@ -373,3 +373,17 @@ class _Visitor extends SimpleAstVisitor<void> {
...
 }
+
+/// Compares directives by package then file in package.
+///
+/// Package is everything until the first `/`.
+int compareDirectives(String a, String b) {
+  if (!a.contains('/') || !b.contains('/')) {
+    return a.compareTo(b);
+  }
+  var as = a.split('/');
```
bwilkerson, line 384: Instead of using `String.split`, consider something like
```
var indexA = a.indexOf('/');
var indexB = B.indexOf('/');
var result = a.substring(indexA).compareTo(b.substring(indexB));
if (result != 0) return result;
return a.substring(indexA + 1).compareTo(b.substring(indexB + 1));
```
That will involve creating fewer temporary strings and avoid the need to join them all back together at the end.

I don't know whether there would be a significant performance improvement, but it seems like a reasonable optimization.

srawlins:
Small note: we may wish to consider https://github.com/dart-lang/sdk/issues/26020 closed with this PR; it would make "organize imports" be aligned with directives_ordering, and I don't think we desire to support multiple orderings.

bwilkerson (approved):

lib/src/rules/directives_ordering.dart
pq: Yes, please.  I think this is great idea.

> I don't know whether there would be a significant performance improvement, but it seems like a reasonable optimization.

Every little bit counts! 👍

goderbauer:
We should double-check, but I don't think flutter depends on any packages with dots in the name.

lib/src/rules/directives_ordering.dart
bwilkerson: Yeah, I was fairly surprised by how much of a performance improvement similar changes were able to get us in code completion.

davidmorgan: Done.

davidmorgan:
Updated implementation--PTAL.

bwilkerson:
LGTM, thanks!

pq (approved):
Awesome.  Thanks!

PR #2695, opened 2021-06-03
enable `depend_on_referenced_packages` and `noop_primitive_operations`

pq:
/cc @bwilkerson @a14n

### DIFF
--- a/tool/doc.dart
+++ b/tool/doc.dart
@@ -247,7 +247,7 @@ ${parser.usage}
 ''');
 }
 
-String qualify(LintRule r) => r.name.toString() + describeMaturity(r);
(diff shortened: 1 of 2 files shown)


tool/doc.dart
```diff
@@ -247,7 +247,7 @@ ${parser.usage}
 ''');
 }
 
-String qualify(LintRule r) => r.name.toString() + describeMaturity(r);
+String qualify(LintRule r) => r.name + describeMaturity(r);
```
pq, line 250: @a14n : check it out. 🎉

bwilkerson (approved):

tool/doc.dart
a14n: :+1:

PR #2701, opened 2021-06-07
Don't stop identifying non-final parameters once a final one is found

### LINKED ISSUE #2700: `prefer_final_parameters` false negative if the first parameter is final
**Describe the issue**
`prefer_final_parameters` gives false negative if the first parameter is `final`, but the rest of the parameters are not `final` and not reassigned.

**To Reproduce**
All of these should LINT

```dart
void badMixed2(final String label1, String label2) { // LINT
  print(label1);
  print(label2);
}

void badMixed3(final String label1, String label2, String label3) { // LINT
  print(label1);
  print(label2);
  print(label3);
}
```

**Expected behavior**
`prefer_final_parameters` should report `badMixed2` and `badMixed3`. However, `dart test -N prefer_final_parameters` fails if both of them are added into `linter/test_data/rules/prefer_final_parameters.dart`.

**Additional context**
Pinging the author of this rule @parlough

parlough:
This rule stopped looking for non-final parameters if it found a `final` one causing lints to not be triggered when they should have been. This modification only skips that parameter instead of stopping processing completely.

Fixes #2700

### DIFF
--- a/test_data/rules/prefer_final_parameters.dart
+++ b/test_data/rules/prefer_final_parameters.dart
@@ -68,9 +68,21 @@ void useGoodTypedClosureArgument() {
   _testingList.forEach((final int element) => print(element + 4)); // OK
 }
 
-void badMixed(String bad, final String good) { // LINT
+void badMixedLast(String bad, final String good) { // LINT
   print(bad);
   print(good);
+}
+
+void badMixedFirst(final String goodFirst, String badSecond) { // LINT
+  print(goodFirst);
+  print(badSecond);
+}
+
+// LINT [+1]
+void badMixedMiddle(String goodFirst, final String badSecond, String badThird) { // LINT
+  print(goodFirst);
+  print(badSecond);
+  print(badThird);
 }
 
 void goodMultiple(final String bad, final String good) { // OK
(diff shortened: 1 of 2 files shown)


test_data/rules/prefer_final_parameters.dart
```diff
@@ -68,11 +68,23 @@ void useGoodTypedClosureArgument() {
...
   print(good);
 }
 
+void badMixedFirst(final String goodFirst, String badSecond) { // LINT
+  print(goodFirst);
+  print(badSecond);
+}
+
+// LINT [+1]
+void badMixedMiddle(String goodFirst, final String badSecond, String badThird) { // LINT
```
parlough, line 82: @pq Is there a better way to identify multiple lints triggered on one line?

bwilkerson (approved):

test_data/rules/prefer_final_parameters.dart
pq: Gross right? Not really, but at least it's possible...  Cheers to @a14n for that!

pq (commented):
🎉

test_data/rules/prefer_final_parameters.dart
a14n: You can be even more precise with something like:
```dart
// LINT [+2,column:length]
// LINT [+1,column:length]
void badMixedMiddle(String goodFirst, final String badSecond, String badThird) {
```

PR #2705, opened 2021-06-09
Remove dependency on package:charcode.

lrhn:
Removes the dependency on the `charcode` package. That package is not owned by the Dart team, and tools.dart.dev packages should not depend unnecessarily on third-party code.
(The `charcode` package can generate stand-alone constant declarations that you can use instead).


pq:
First of all: this is awesome, thank you!

With regard to your rationale:

> That package is not owned by the Dart team, and tools.dart.dev packages should not depend unnecessarily on third-party code.

This is really reasonable and something to be mindful of moving forward.  Should we do (or have you done) an audit of team-owned package dependencies generally?  Is tooling being considered to keep new undesired dependencies from creeping in?

pq (approved):

lrhn:
> Is tooling being considered to keep new undesired dependencies from creeping in?

Any part of the "general rules for Dart team owned code" which can be automated, probably should. I don't know if it's part of our first attempts at getting an overview, but I hope it eventually will be.
(It's not trivial, because there can be exceptions for particularly well-managed third-party packages that we don't want to duplicate, but third-party dependencies should never be added blindly. I personally thing `charcode` is well (hands-off) managed, but it's also quite unnecessary in the long run, and the migration to null safety was more painful than necessary because so many different packages depended on it.)

pq:
Great.  Thanks!  I'll think on tooling a little (some things jump to mind immediately but I haven't thought too hard about it yet).  

Are blessed packages enumerated somewhere?  Maybe a column could be added to https://dart-lang.github.io/repo_manager/#sort_column=repo?

/cc @kevmoo

lrhn:
I don't think we have blessed any packages except those owned by the dart.dev or dart.tools.dev organizations (and we're not completely done moving things to the right organizations yet - I think, unless @kevmoo got them all). The goal is to not have packages (or the SDK) depend on something with lower reliability than what we want for the package/SDK itself. That goes hand in hand with introducing higher reliability requirements/promises for our own packages.

pq:
Thanks!

/fyi also @natebosch

Issue #2713, opened 2021-06-15
Investigate new `sort_child_properties_last` diagnostics when relaxed


pq:
#2648 introduces new diagnostics to libraries in (at least) 20 internal packages (the SDK roller stopped trying to add exceptions at 20).  We'll need to investigate if these are false positives or if we need to coordinate fixes before the next roll.

pq:
Example breakage:

```dart
  @override
  Widget build(BuildContext context) => Scaffold(
      appBar: AppBar(
        brightness: widget.brightness,
        leading: IconButton(
          icon: const BackButtonIcon(), // <= LINT
          tooltip: MaterialLocalizations.of(context).backButtonTooltip,
          onPressed: () {
             ...
          },
          key: backButton,
        ),
  ...
```

pq:
Good to know.  Given the impact, we'll have to come up with a plan for migration (or reconsider the additional diagnostics).

Thanks!

/fyi @jacob314

jacob314:
Happy to help migrate. Perhaps it would be worth it to add a quick fix to help migrate? Moving all violating properties to after the last non-child & non closure property would probably be a fine default.
In the example given, `backgroundColor` should be at the top of the method next to the `brightness` property. As is it is, it takes a while to spot what widget the backgroundColor is associated with.

pq:
> Perhaps it would be worth it to add a quick fix to help migrate?

FWIW there is a fix but I expect it will need to be updated w/ the new and improved semantics.

PR #2714, opened 2021-06-15
Revert "relax sort_child_properties_last to accept other widget afterward"

pq:
Reverts dart-lang/linter#2648

This change is flagging a number of new violations internally and needs to be investigated (https://github.com/dart-lang/linter/issues/2713).  This revert will unblock the other changes that we want to roll.

/cc @bwilkerson @a14n @jacob314

### DIFF
--- a/lib/src/rules/sort_child_properties_last.dart
+++ b/lib/src/rules/sort_child_properties_last.dart
@@ -11,7 +11,7 @@ import '../util/flutter_utils.dart';
 const _desc = r'Sort child properties last in widget instance creations.';
 
 const _details = r'''
-Sort arguments to end with a Widget in widget instance creations.  This improves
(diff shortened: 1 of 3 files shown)


bwilkerson (approved):
Can we produce some examples of the kinds of code being flagged? My understanding was that the original PR was suppose to flag fewer cases, not more, so understanding what's being flagged by the change that wasn't before seems critical to re-landing it.

pq:
> Can we produce some examples of the kinds of code being flagged?

Yes.  See #2713.  It looks like the scope of the change widened over the course of the review of #2648.

bwilkerson:
Thanks, yes I saw that email after I reviewed this PR. If only Gmail could sort my email more optimally. :-)

PR #2719, opened 2021-06-18
Report use_setters_to_change_properties on the method name

### LINKED ISSUE #2430: "ignore: use_setters_to_change_properties" doesn't work with a preceding dartdoc comment
With `use_setters_to_change_properties` enabled in my `analysis_options.yaml`, running `dart analyze` against the following code:

```dart
class Foo {
  int member = 0;

  /// Documentation goes here.
  // ignore: use_setters_to_change_properties
  void setMember(int value) {
    member = value;
  }
}
```

gives me:

> info • Use a setter for operations that conceptually change a property at foo.dart:4:3 • (use_setters_to_change_properties)

I find this surprising since the `ignore:` line does work if I either move it above the dartdoc comment or if I remove the dartdoc comment completely.  It seems strange to me that this somehow seems sensitive to the placement of the dartdoc comment (and if anything, I'm even more surprised that it works with the intervening dartdoc comment).

I'm using Dart SDK version: 2.12.353624752-edge+google3-v2 (google3) (google3) on "linux_x64".

srawlins:
# Description

Fixes #2430 by reporting lint on the method name, not the entire method node including its doc comment and body.

Also tidy up by renaming a _local_ function to not start with an underscore, as per Effective Dart.

### DIFF
--- a/lib/src/rules/use_setters_to_change_properties.dart
+++ b/lib/src/rules/use_setters_to_change_properties.dart
@@ -61,7 +61,8 @@ class _Visitor extends SimpleAstVisitor<void> {
         node.returnType?.type?.isVoid != true) {
       return;
     }
-    void _visitExpression(Expression expression) {
+
+    void visitExpression(Expression expression) {
       if (expression is AssignmentExpression &&
           expression.operator.type == TokenType.EQ) {
         var leftOperand =
@@ -70,7 +71,7 @@ class _Visitor extends SimpleAstVisitor<void> {
             expression.rightHandSide);
         var parameterElement = node.declaredElement?.parameters.first;
         if (rightOperand == parameterElement && leftOperand is FieldElement) {
-          rule.reportLint(node);
+          rule.reportLint(node.name);
         }
       }
     }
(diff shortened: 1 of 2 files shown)


lib/src/rules/use_setters_to_change_properties.dart
```diff
@@ -61,7 +61,8 @@ class _Visitor extends SimpleAstVisitor<void> {
         node.returnType?.type?.isVoid != true) {
       return;
     }
-    void _visitExpression(Expression expression) {
+
+    void visitExpression(Expression expression) {
```
pq, line 65: Does this now override `visitExpression` in `SimpleAstVisitor`?  Maybe another name to avoid confusion?

bwilkerson: A different name would be good because it obviously did create confusion and probably will again, but `SimpleAstVisitor` doesn't implement `visitExpression`. The method `visitExpression` is in `GeneralizingAstVisitor`.

bwilkerson (approved):
You should also add an entry to the change log.

pq (approved):

PR #2721, opened 2021-06-18
Allow interpolations on adjacent strings

### LINKED ISSUE #2489: missing_whitespace_between_adjacent_strings false positive
In the following snippet, foo3 reports a false missing_whitespace_between_adjacent_strings. The other examples work as expected.

```dart
          String foo = 'Hello'
            ', world';

          String foo2 = 'Hello'
            '${icon != null ? ', world' : ''}';

          String foo3 = 'Hello'
            ', world' // missing_whitespace_between_adjacent_strings incorrectly reported
            '${icon != null ? ', world' : ''}';
```

srawlins:
# Description

This lint rule was too strict on adjacent strings that start or end with interpolations. Since it cannot know anything about string interpolations, it should treat them as "valid" (containing whitespace).

Fixes #2489 

Also:

* change helper functions to extension methods; let me know if this is undesirable
* update comments to format like sentences, as per [Effective Dart](https://dart.dev/guides/language/effective-dart/documentation#do-format-comments-like-sentences).

### DIFF
--- a/lib/src/rules/missing_whitespace_between_adjacent_strings.dart
+++ b/lib/src/rules/missing_whitespace_between_adjacent_strings.dart
@@ -57,7 +57,7 @@ class _Visitor extends RecursiveAstVisitor<void> {
 
   @override
   void visitAdjacentStrings(AdjacentStrings node) {
-    // skip regexp
+    // Skip strings passed to `RegExp()` or any method named `matches`.
     var parent = node.parent;
     if (parent is ArgumentList) {
       var parentParent = parent.parent;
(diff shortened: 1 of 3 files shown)


lib/src/rules/missing_whitespace_between_adjacent_strings.dart
```diff
@@ -90,27 +90,39 @@ class _Visitor extends RecursiveAstVisitor<void> {
...
   }
 }
 
+extension on String {
+  bool get hasWhitespace => _whitespaces.any(contains);
+  bool get endsWithWhitespace => _whitespaces.any(endsWith);
+  bool get startsWithWhitespace => _whitespaces.any(startsWith);
+}
+
 const _whitespaces = [' ', '\n', '\r', '\t'];
```
bwilkerson, line 128: If this is now only used in the extension, consider making it a static member of the extension (and dropping the leading underscore).

bwilkerson (approved):

lib/src/rules/missing_whitespace_between_adjacent_strings.dart
srawlins: done

PR #2722, opened 2021-06-18
Allow unnecessary getters and setters with annotations

### LINKED ISSUE #2706: unnecessary_getters_setters should allow setter or getter with annotations
**Describe the issue**

If I create a getter and a setter, that are simple, and one of these methods is annotated with `@visibleForTesting`, the lint rule triggers. I believe it should not.

I think this should be the case for _any_ annotations. But if we must enumerate them, I think that annotations on either getter or setter of `@Deprecated`, `@visibleFor`_X_, `@protected`, `@internal` should prevent the lint rule from triggering.

**To Reproduce**

```dart
class C {
  int _i;

  int get i => _i;

  @visibleForTesting
  set i(int value) => _i = value;
}
```

**Expected behavior**

No lint.

srawlins:
# Description

A getter or setter which is otherwise unnecessary, but has an annotation should be allowed; the annotation may carry meaning to humans, static analyzers, or code generators

Fixes #2706

### DIFF
--- a/lib/src/rules/unnecessary_getters_setters.dart
+++ b/lib/src/rules/unnecessary_getters_setters.dart
@@ -91,13 +91,11 @@ class _Visitor extends SimpleAstVisitor<void> {
 
     // Only select getters with setter pairs
     for (var id in getters.keys.where((id) => setters.keys.contains(id))) {
-      _visitGetterSetter(getters[id], setters[id]);
+      _visitGetterSetter(getters[id]!, setters[id]!);
     }
   }
 
-  void _visitGetterSetter(
-      MethodDeclaration? getter, MethodDeclaration? setter) {
-    if (getter == null || setter == null) return;
+  void _visitGetterSetter(MethodDeclaration getter, MethodDeclaration setter) {
     var getterElement = getter.declaredElement;
     var setterElement = setter.declaredElement;
     if (getterElement == null || setterElement == null) return;
--- a/lib/src/rules/unnecessary_getters_setters.dart
+++ b/lib/src/rules/unnecessary_getters_setters.dart
@@ -4,6 +4,7 @@
 
 import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/element.dart';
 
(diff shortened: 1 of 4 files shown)


lib/src/rules/unnecessary_getters_setters.dart
```diff
@@ -89,21 +90,21 @@ class _Visitor extends SimpleAstVisitor<void> {
...
+      _visitGetterSetter(getters[id], setters[id]);
+    }
   }
 
   void _visitGetterSetter(
       MethodDeclaration? getter, MethodDeclaration? setter) {
-    if (getter != null &&
-        setter != null &&
-        isSimpleSetter(setter) &&
+    if (getter == null || setter == null) return;
```
bwilkerson, line 100: Do we need this test if we've already guaranteed that both maps contain the id?

Better yet, can we remove the `where` on line 93 in favor of this condition (given that `[]` is nullable and we need the test to get around the type system)?

Even better, can we remove both the `where` _and_ this line by writing
```
var getterElement = getter?.declaredElement;
var setterElement = setter?.declaredElement;
```
and testing for `null` on line 103?

bwilkerson (approved):

lib/src/rules/unnecessary_getters_setters.dart
srawlins: Oh ho, good trick. Refactored.

lib/src/rules/unnecessary_getters_setters.dart
```diff
@@ -89,21 +90,20 @@ class _Visitor extends SimpleAstVisitor<void> {
...
+    }
   }
 
-  void _visitGetterSetter(
-      MethodDeclaration? getter, MethodDeclaration? setter) {
-    if (getter != null &&
-        setter != null &&
-        isSimpleSetter(setter) &&
+  void _visitGetterSetter(MethodDeclaration getter, MethodDeclaration? setter) {
+    if (setter == null) return;
```
srawlins, line 99: Unfortunately, `if (setter?.declaredElement == null) return;`  does not promote `setter` to non-null. There  is a language issue for this. So this reads nicer as is, rather than null-checking `setter` below.

PR #2723, opened 2021-06-18
allow overrides with a doc comment only if they are abstract

### LINKED ISSUE #2277: The `unnecessary_overrides` lint allows override for comments.
The [unnecessary_overrides](https://dart-lang.github.io/linter/lints/unnecessary_overrides.html) lint lists one of the exceptions as
> * if documentation comments are present on the member,

In that case, the *body* of the override is still unnecessary. You can declare an abstract member with the new documentation instead. (You can possibly also do that for annotations, depending on what the annotation means).
You should not "do a super method invocation with same parameters.", and I think it's fine to warn about that, so I'd remove that exception.

**To Reproduce**
Example:
```dart
class C {
  /// Do the foo.
  int foo(int x) => x
  /// Do the bar.
  int bar(int x) => x;
}
class D extends C {
  /// Let's do the foo again.
  int foo(int x) => super.foo(x);  // No lint.
  /// Let's do the bar again.
  int bar(int x);  // What you should have written.
}
```

**Expected behavior**
Still warn if the only reason to have a super-forwarding call is new documentation.

srawlins:
# Description

Update `unnecessary_overrides` to allow overrides with a doc comment only if they are abstract.

Also tidy comments to format like sentences.

Fixes #2277

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
   and setters with annotations.
 - update `missing_whitespace_between_adjacent_strings` to allow String
   interpolations at the beginning and end of String literals.
+- update `unnecessary_overrides` to allow overrides with a doc comment only if
+  they are abstract.
 
 # 1.6.1
 
--- a/lib/src/rules/unnecessary_overrides.dart
+++ b/lib/src/rules/unnecessary_overrides.dart
@@ -39,10 +39,11 @@ class A extends B {
 
 It's valid to override a member in the following cases:
 
-* if a type (return type or a parameter type) is not the exactly the same as the
-super method,
+* if a return type or a parameter type is not the exact same type as the
+  super method,
 * if the `covariant` keyword is added to one of the parameters,
-* if documentation comments are present on the member,
+* if a documentation comment is present on the member, and the member is
+  abstract,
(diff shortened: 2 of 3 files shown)


bwilkerson (approved):
Note that this is a breaking change because it can flag cases that it didn't use to flag. Not sure the extra checking is worth the potential cost.

srawlins:
@pq brought up this issue recently; I think the (temporary?) plan is to take such changes and patch them into the SDK and create a gerrit CL and test on the google bot. WDYT, @pq?

pq:
Yeah.  In this case it may be better to get a pulse check before we draft a release since it could be pretty widely impacting.  

Probably the easiest way to do this would be to update the lint in Google3 and analyze the world.

Maybe we can sync up on this on Monday?

Also: if we do land this, the quickfix will need updating too probably?

srawlins:
Google3 global presubmit shows 0 breakages. I'm not surprised that no one overrides a method just to update the documentation.

pq:
Ok, cool!  In that case, let's land it.  We've got enough fixes that this might even prompt a release to get these improvements baking internally.

Thanks!

pq (approved):
👍

PR #2724, opened 2021-06-18
Update unnecessary_getters_setters to allow for non-basic assignments

### LINKED ISSUE #2402: unnecessary_getters_setters false positive/inconsistent with null-aware assignment operator
This is giving `unnecessary_getters_setters`, but I believe it shouldn't be:

```dart
bool get foo => _foo; // Avoid wrapping fields in getters and setters just to be "safe".
set foo(bool b) =>  _foo ??= b; // Avoid wrapping fields in getters and setters just to be "safe".
bool _foo;
```

However, this doesn't raise the warning, even though they are effectively the same (i.e., there's an inconsistency):

```dart
bool get foo => _foo;
set foo(bool b) =>  _foo = _foo ?? b;
bool _foo;
```

srawlins:
# Description

Assignments other than `=` are allowable in setters, like `??=` or `+=`.

Fixes #2402 

Also updates related doc comments to use Markdown code blocks with language tags; dartdoc started warning about code blocks which don't use language tags recently.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
   and setters with annotations.
 - update `missing_whitespace_between_adjacent_strings` to allow String
   interpolations at the beginning and end of String literals.
+- update `unnecessary_getters_setters` to allow for setters with non-basic
+  assignments (for example, `??=` or `+=`).
 
 # 1.6.1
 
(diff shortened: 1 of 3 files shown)


bwilkerson (approved):

pq (commented):
Nice!

PR #2726, opened 2021-06-18
Report adjacent strings in if-elements and for-elements

### LINKED ISSUE #2343: no_adjacent_strings_in_list should trigger when the collection if is missing a trailing comma
**Describe the issue**

**To Reproduce**
```
List<String> list = <String>[
  'a',
  'b'
  'c'
];
```
triggers the lint but
```
List<String> list = <String>[
  'a',
  if (true) 'b'
  'c'
];
```
does not.

**Expected behavior**

**Additional context**
Real world bug:

srawlins:
# Description

This enhances no_adjacent_string_in_list to include if-elements and for-elements (see tests for examples).

Note that this will trigger on if-elements and for-elements in sets as well 😲 . The lint broadly does not cover sets. Should it? New lint rule? Deprecate this one and make a new lint rule for collection literals and collection elements?

Fixes #2343

Also fixes some grammar in the docs, and makes the test code more idiomatic.

### DIFF
--- a/lib/src/rules/no_adjacent_strings_in_list.dart
+++ b/lib/src/rules/no_adjacent_strings_in_list.dart
@@ -11,9 +11,9 @@ const _desc = r"Don't use adjacent strings in list.";
 
 const _details = r'''
 
-**DON'T** use adjacent strings in list.
+**DON'T** use adjacent strings in a list.
 
-This can be sign of forgotten comma.
+This can indicate a forgotten comma.
 
 **GOOD:**
 ```dart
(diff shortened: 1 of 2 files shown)


bwilkerson (commented):
The purpose for the lint is to catch missing commas. We can't have the same problem in maps that we do in lists and sets, but yes I think it makes sense to catch these in sets as well.

Either way this is a breaking change, but it's a bigger breaking change if we catch this in sets. I don't know what we want to do about sets (add them or add a new lint), but I do believe that we either need to not report a lint in `if` and `for` elements in sets or we need to handle all elements in sets. In its current form this rule is inconsistent.

It probably makes the most sense to limit the behavior to lists for now and open an issue for community feedback on the idea of adding support for sets.

srawlins:
Filed an issue: https://github.com/dart-lang/linter/issues/2727

srawlins:
Replaced by https://github.com/dart-lang/linter/pull/3741, as I don't use the source repo found in this PR, any longer.

PR #2725, opened 2021-06-18
Update badge colors.

bernaferrari:
image

Question: should effective_dart become gray with pedantic? Or we keep it as is for now?

https://github.com/dart-lang/linter/pull/2717

PR #2729, opened 2021-06-19
Deprecate effective dart badge.

bernaferrari:
Sequel to https://github.com/dart-lang/linter/pull/2725.

Before:
image
After: 
image

Reason: with the official Dart and Flutter lints, effective_dart is in a weird place since it is "community made" but not "official" and I believe it can be confusing for new people to differentiate between it and the official rules. The new rules are also being kept in sync with the last advances in dart_sdk and this is nice. `prefer_mixins` is going to be 'fixed' before being introduced in the official lint, while in effective_dart it is on by default, causing issues with ChangeListener.


pq (approved):
Thanks!

PR #2734, opened 2021-06-29
use_build_context_synchronously: support switches; improve flow analysis

pq:
/cc @bwilkerson

### DIFF
--- a/lib/src/rules/use_build_context_synchronously.dart
+++ b/lib/src/rules/use_build_context_synchronously.dart
@@ -119,24 +119,33 @@ class _Visitor extends SimpleAstVisitor {
   }
 
   void check(AstNode node) {
+    bool checkStatements(AstNode child, NodeList<Statement> statements) {
+      var index = statements.indexOf(child as Statement);
+      for (var i = index - 1; i >= 0; i--) {
+        var s = statements.elementAt(i);
+        if (isMountedCheck(s)) {
+          return false;
+        } else if (isAsync(s)) {
+          rule.reportLint(node);
+          return true;
+        }
+      }
+      return true;
+    }
+
     // Walk back and look for an async gap that is not guarded by a mounted
     // property check.
     AstNode? child = node;
     while (child != null && child is! FunctionBody) {
       var parent = child.parent;
-      // todo (pq): refactor to handle SwitchCase's
       if (parent is Block) {
-        var statements = parent.statements;
-        var index = statements.indexOf(child as Statement);
-        for (var i = index - 1; i >= 0; i--) {
-          var s = statements.elementAt(i);
-          if (isMountedCheck(s)) {
-            return;
-          } else if (isAsync(s)) {
-            rule.reportLint(node);
-            return;
-          }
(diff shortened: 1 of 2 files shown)


lib/src/rules/use_build_context_synchronously.dart
```diff
@@ -119,24 +119,33 @@ class _Visitor extends SimpleAstVisitor {
   }
 
   void check(AstNode node) {
+    bool checkStatements(AstNode child, NodeList<Statement> statements) {
+      var index = statements.indexOf(child as Statement);
+      for (var i = index - 1; i >= 0; i--) {
+        var s = statements.elementAt(i);
```
bwilkerson, line 125: Consider using `statements[i]`.

lib/src/rules/use_build_context_synchronously.dart
```diff
@@ -119,24 +119,33 @@ class _Visitor extends SimpleAstVisitor {
...
+        if (isMountedCheck(s)) {
+          return false;
+        } else if (isAsync(s)) {
+          rule.reportLint(node);
+          return true;
```
bwilkerson, line 130: If we've already reported a lint, is there any reason to keep looking up the flow path taken to get here? Could this produce multiple lints on the same node?

lib/src/rules/use_build_context_synchronously.dart
```diff
@@ -119,24 +119,33 @@ class _Visitor extends SimpleAstVisitor {
...
-            rule.reportLint(node);
-            return;
-          }
+        var keepChecking = checkStatements(child, parent.statements);
+        if (!keepChecking) {
+          return;
         }
+      } else if (parent is SwitchCase) {
+        checkStatements(child, parent.statements);
+        return;
```
bwilkerson, line 148: It seems wrong to not look above the switch statement. Is the rule behaving as expected for the following?
```
void awaitInSwitchCase_mountedCheckBeforeSwitch(BuildContext context) async {
    await Future<void>.delayed(Duration());
    if (!mounted) return;
    switch (1) {
      case 1:
        await Navigator.of(context).pushNamed('routeName'); // OK
        break;
    }
  }
```

bwilkerson (approved):

lib/src/rules/use_build_context_synchronously.dart
pq: done. thanks!

lib/src/rules/use_build_context_synchronously.dart
pq: yep.  That works (and test case added) but I'm not surprised you're confused since this is getting unwieldy.  I'll take a refactoring conversation offline...

pq: (actually it worked but only by accident.)

I think it's actually fixed now but would still like to explore a general refactoring so will follow up on that.

lib/src/rules/use_build_context_synchronously.dart
pq: Hmmm.  Interesting.  Will have to think on this.  Good question.

PR #2735, opened 2021-06-29
Fix missing_whitespace_between_adjacent_strings handling of no-whitespace strings

srawlins:
# Description

This involves a sizeable refactor; I pushed most of the logic of whether a
StringLiteral ends with, starts with, or contains whitespace into extensions on
StringLiteral. I think this makes a lot of sense; they're all getters that
return information about the instance.

Fixes a bug that @pq found.

This is slightly more performant; string literals that do start with whitespace or end with whitespace are not iterated.

### DIFF
--- a/lib/src/rules/missing_whitespace_between_adjacent_strings.dart
+++ b/lib/src/rules/missing_whitespace_between_adjacent_strings.dart
@@ -95,44 +95,34 @@ class _Visitor extends RecursiveAstVisitor<void> {
 }
 
 extension on StringLiteral {
-  /// Returns whether this ends with whitespace, where any
+  /// Returns whether this ends with whitespace, where an initial
   /// [InterpolationExpression] counts as whitespace.
   bool get endsWithWhitespace {
-    if (this is SimpleStringLiteral) {
-      return (this as SimpleStringLiteral).value.endsWithWhitespace;
-    } else if (this is StringInterpolation) {
-      var last = (this as StringInterpolation).elements.last;
-      if (last is InterpolationExpression) {
-        // Treat an interpolation expression as containing whitespace so as to
-        // avoid over-reporting strings that end with an interpolation
-        // expression.
-        return true;
-      } else if (last is InterpolationString) {
-        return last.value.endsWithWhitespace;
-      }
+    var self = this;
+    if (self is SimpleStringLiteral) {
+      return self.value.endsWithWhitespace;
+    } else if (self is StringInterpolation) {
+      var last = self.elements.last as InterpolationString;
+      return last.value.isEmpty || last.value.endsWithWhitespace;
     }
     throw ArgumentError(
-        'Expected SimpleStringLiteral or StringInterpolation, got $runtimeType');
+        'Expected SimpleStringLiteral or StringInterpolation, but got '
+        '$runtimeType');
   }
 
-  /// Returns whether this starts with whitespace, where any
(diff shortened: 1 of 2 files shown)


lib/src/rules/missing_whitespace_between_adjacent_strings.dart
```diff
@@ -119,6 +94,61 @@ class _Visitor extends RecursiveAstVisitor<void> {
...
 
+extension on StringLiteral {
+  /// Returns whether this ends with whitespace, where any
+  /// [InterpolationExpression] counts as whitespace.
+  bool get endsWithWhitespace {
+    if (this is SimpleStringLiteral) {
+      return (this as SimpleStringLiteral).value.endsWithWhitespace;
+    } else if (this is StringInterpolation) {
+      var last = (this as StringInterpolation).elements.last;
+      if (last is InterpolationExpression) {
```
bwilkerson, line 105: This condition won't ever be true. The first and last elements in a `StringInterpolation` are always an `InterpolationString` (see `StringInterpolation.firstString` and `StringInterpolation.lastString`). In the case of a string interpolation such as `'$a'`, both the first and last elements will be empty strings. As a result, I think the code needs to consider an empty `InterpolationString` to be the same as whitespace.

Note that there are no tests for the case where an interpolation expression is either first or last but the other end of the string contains whitespace. For example: `'a $b' 'c'` or `'a' '$b c'`.

lib/src/rules/missing_whitespace_between_adjacent_strings.dart
```diff
@@ -119,6 +94,61 @@ class _Visitor extends RecursiveAstVisitor<void> {
   }
 }
 
+extension on StringLiteral {
+  /// Returns whether this ends with whitespace, where any
+  /// [InterpolationExpression] counts as whitespace.
+  bool get endsWithWhitespace {
+    if (this is SimpleStringLiteral) {
```
bwilkerson, line 101: Consider assigning `this` to a local variable. Konstantin's been using `self` as the name of the local in such cases.

bwilkerson (approved):
The code is much cleaner now.

lib/src/rules/missing_whitespace_between_adjacent_strings.dart
srawlins: Done

lib/src/rules/missing_whitespace_between_adjacent_strings.dart
srawlins: Ah ha good catch. Much simpler now.

pq (approved):
Nice!

lib/src/rules/missing_whitespace_between_adjacent_strings.dart
```diff
@@ -95,44 +95,34 @@ class _Visitor extends RecursiveAstVisitor<void> {
...
-        // expression.
-        return true;
-      } else if (last is InterpolationString) {
-        return last.value.endsWithWhitespace;
-      }
+    var self = this;
+    if (self is SimpleStringLiteral) {
+      return self.value.endsWithWhitespace;
+    } else if (self is StringInterpolation) {
+      var last = self.elements.last as InterpolationString;
```
bwilkerson, line 105: Not necessary, but you could simplify this further by using `self.lastString` which casts the element for you.

bwilkerson (approved):

PR #2736, opened 2021-06-30
Fixed false positive use_full_hex_values_for_flutter_colors

Semvrij:
If you use an uppercase 'X' in the hexadecimal prefix, it gives a false positive.


bwilkerson (approved):
Thanks!

pq (approved):
💎

PR #2744, opened 2021-07-02
fix false positive in `use_late_for_private_fields_and_variables` in the presence of const constructors

### LINKED ISSUE #2718: `use_late_for_private_fields_and_variables` recommends something strange
**Describe the issue**
`use_late_for_private_fields_and_variables` wants me to make my field late instead of nullable in case where this causes me to make my class non-const

**To Reproduce**
```dart
class ContentChannelException {
  const ContentChannelException._(this.value, [this._exception]);

  final String value;

  @override
  bool operator ==(Object other) {
  if (other.runtimeType != runtimeType)
    return false;
  return other is ContentChannelException 
      && other.value == value;
  }

  @override
  int get hashCode => value.hashCode;

  PlatformException get exception => _exception!;
  /// Exceptions thrown with [_throw] all have [exception] non-null.
  /// Exceptions used as constants do not.
  /// There's no way for me to use late here without making this class non-const.
  final PlatformException? _exception;

  static const unexpected = ContentChannelException._('UNEXPECTED_ERROR');
  static const someError = ContentChannelException._('SOME_ERROR');

  static ContentChannelException _throw(PlatformException exception, List<ContentChannelException> expectedExceptions) {
    final toThrow = ContentChannelException._(exception.code, exception);
    assert(toThrow == unexpected || expectedExceptions.contains(toThrow));
    return toThrow;
  }
}

void main() {
  // Usage example
  try {
    throw ContentChannelException.someError;
  } on PlatformException catch (ex) {
    throw ContentChannelException._throw(ex, const [ContentChannelException.someError]);
  }
}
```

**Expected behavior**
I'm not really sure, but ideally in this case I wouldn't want to change it to late

**Additional context**
No

pq:
Fixes #2718

/cc @bwilkerson @a14n


lib/src/rules/use_late_for_private_fields_and_variables.dart
```diff
@@ -77,6 +77,17 @@ class _Visitor extends UnifyingAstVisitor<void> {
   CompilationUnitElement? currentUnit;
   _Visitor(this.rule, this.context);
 
+  @override
+  void visitClassDeclaration(ClassDeclaration node) {
+    // See: https://dart.dev/tools/diagnostic-messages#late_final_field_with_const_constructor
+    if (node.members.any((member) =>
+        member is ConstructorDeclaration && member.constKeyword != null)) {
```
bwilkerson, line 84: Consider re-writing as a for loop for performance reasons. Not sure whether the difference will be meaningful.

bwilkerson (approved):

lib/src/rules/use_late_for_private_fields_and_variables.dart
pq: Sure.  If there's any possible difference, I'll take it!

PR #2749, opened 2021-07-06
add github authentication

pq:
Speeds up doc creation considerably.

/cc @bwilkerson


pq:
PS: sorry for the sorting churn.

Really, maybe we should consider a `verify_sorted_test` a la server?

bwilkerson (approved):
I have no idea what's actually changed here, but given that it won't impact users because it's a tool, and assuming that you ran the tool to verify that it runs, I'm fine with whatever changes you made.

pq:
This just makes it possible to use explicit token authentication for github API calls.  This is much faster than the default anonymous access.

Issue #2751, opened 2021-07-06
Should `avoid_literal_in_foreach` lint reconsider when the collection is nullable.


lrhn:
See: https://stackoverflow.com/questions/68271469/linter-rule-and-null-safety
In short, if the iterable is nullable, you can write `foo?.forEach((e) { ... })`, but need much more verbosity to turn it into a `for` loop, like:
```dart
var v = foo; // If it's an expression, not already a variable.
if (v != null) {
  for (var e in v) { 
    ... 
  }
}
```
over
```
foo?.forEach((e) {
  ...
});
```

According to the stack overflow report, the latter triggers the lint. Maybe it shouldn't.

pq:
This is interesting.  I actually just recently hit this and begrudgingly made the required changes when `foo?.forEach` was so much cleaner.

I'd be really curious to get other people's thoughts...

/fyi @bwilkerson @munificent @natebosch @jakemac53 @eernstg

lrhn:
One workaround suggested was `for (var e in [...?foo]) { ... }`. That sadly has a semantic which requires building the intermediate list.
Another was `for var e in foo ?? const <Foo>[]) { ... }` which can be optimized better, but you need to write the `<Foo>` type to preserve the type of `e`, which is also annoying.

Maybe we just need `for (var e in? foo) { ... }` like we have `[...? foo]`. 😁

munificent:
Yeah, there's no great pattern for this. I'm OK with relaxing the lint to allow `foo?.forEach(...)`.

pq:
OK!  I think we can consider this closed w/ #2752.

natebosch:
> One workaround suggested was `for (var e in [...?foo]) { ... }`. That sadly has a semantic which requires building the intermediate list.


> Yeah, there's no great pattern for this. I'm OK with relaxing the lint to allow `foo?.forEach(...)`.

Another time it would be nice to have a syntax to generate iterables.

PR #2752, opened 2021-07-06
skip nullable iterables in `avoid_function_literals_in_foreach_calls`

pq:
Fixes #https://github.com/dart-lang/linter/issues/2751

/cc @bwilkerson

### DIFF
--- a/lib/src/rules/avoid_function_literals_in_foreach_calls.dart
+++ b/lib/src/rules/avoid_function_literals_in_foreach_calls.dart
@@ -4,6 +4,8 @@
 
 import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/nullability_suffix.dart';
+import 'package:analyzer/dart/element/type.dart';
 
 import '../analyzer.dart';
 import '../util/dart_type_utilities.dart';
@@ -30,6 +32,27 @@ for (var person in people) {
 people.forEach(print);
 ```
 ''';
+
+bool _hasMethodChaining(MethodInvocation node) {
+  var exp = node.target;
+  while (exp is PrefixedIdentifier ||
+      exp is MethodInvocation ||
+      exp is PropertyAccess) {
+    if (exp is PrefixedIdentifier) {
+      exp = exp.prefix;
+    } else if (exp is MethodInvocation) {
+      return true;
+    } else if (exp is PropertyAccess) {
+      exp = exp.target;
+    }
+  }
+  return false;
+}
+
+bool _isNonNullableIterable(DartType? type) =>
+    type != null &&
+    type.nullabilitySuffix != NullabilitySuffix.question &&
+    DartTypeUtilities.implementsInterface(type, 'Iterable', 'dart.core');
 
 class AvoidFunctionLiteralInForeachMethod extends LintRule
     implements NodeLintRule {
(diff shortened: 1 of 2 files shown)


test_data/rules/avoid_function_literals_in_foreach_calls.dart
```diff
@@ -9,12 +9,14 @@ class Person {
 }
 
 void main() {
-  Iterable<String> people;
+  Iterable<String> people = [];
```
bwilkerson, line 12: Should we add something to the test framework to fail a test if there are any diagnostics other than the lints being tested for? Seems like this is something that should have been caught when switching to analyze test code in null safety by default.

test_data/rules/avoid_function_literals_in_foreach_calls.dart
```diff
@@ -9,12 +9,14 @@ class Person {
...
 
-  for (var person in people) { // OK
+  for (var person in people) {
+    // OK
     print('$person!');
   }
-  people.forEach((person) { // LINT
+  people.forEach((person) // LINT
```
bwilkerson, line 18: I'm not sure of the value of moving the opening bracket, but ok.

lib/src/rules/avoid_function_literals_in_foreach_calls.dart
```diff
@@ -31,6 +33,27 @@ people.forEach(print);
...
+    } else if (exp is PropertyAccess) {
+      exp = exp.target;
+    }
+  }
+  return false;
+}
+
+bool _isNonNullableIterable(DartType? type) =>
+    type != null &&
+    type.nullabilitySuffix != NullabilitySuffix.question &&
```
bwilkerson, line 54: I'm not sure how this works for `dynamic`. Consider
```
void f(dynamic iter) => iter?.forEach(...);
```
Is the suffix always a question even when it isn't explicitly given?

I'll just note that an equivalent test might be to look for the `?.` operator on the method invocation.

bwilkerson (approved):
LGTM assuming it works with `dynamic`.

test_data/rules/avoid_function_literals_in_foreach_calls.dart
pq: It's a formatting thing.  The formatter moves the comment if it's after the brace, which breaks the expectation.

lib/src/rules/avoid_function_literals_in_foreach_calls.dart
pq: > Is the suffix always a question even when it isn't explicitly given?

I **_believe_** so?  (@scheglov may knot for sure.)

In any event, I added a `dynamic` test case and it seems to work as written.

Thanks for the catch!

test_data/rules/avoid_function_literals_in_foreach_calls.dart
pq: Good question. We're currently ignoring diagnostics and I suspect there are a ton.

https://github.com/dart-lang/linter/issues/2760

lib/src/rules/avoid_function_literals_in_foreach_calls.dart
scheglov: The type `dynamic` is always nullable, and it does not have any `nullabilitySuffix`, it is always `NullabilitySuffix.none`. Note that the suffix is not the same as nullability, e.g. `FutureOr<int?>` is nullable.

PR #2753, opened 2021-07-06
fix false positives for type aliases in `avoid_private_typedef_functions`

### LINKED ISSUE #2732: false positive for `avoid_private_typedef_functions` with generalized type aliases
[`avoid_private_typedef_functions`](https://dart-lang.github.io/linter/lints/avoid_private_typedef_functions.html) still assumes every `typedef` is a function alias.  

> AVOID private typedef functions used only once. Prefer inline function syntax.

**To Reproduce**

```dart
/// Likes, comments, shares
typedef _FacebookPostMetadata = Tuple3<int?, int?, int?>;
// Warning: Avoid private typedef functions
```

**Expected behavior**

No warning when the right-hand side of a `typedef` is not a `Function`.

**Additional context**

In a nested collection of primitives (like a viewmodel), a one-off `typedef` can add clarity to the internal structure where writing a class feels like overkill.

```sh
 $ dart --version
Dart SDK version: 2.13.3 (stable) (Wed Jun 9 12:44:44 2021 +0200) on "linux_x64"
```

pq:
Fixes #2732

/cc @bwilkerson


test_data/rules/experiments/nonfunction-type-aliases/rules/avoid_private_typedef_functions.dart
```diff
@@ -0,0 +1,36 @@
+// Copyright (c) 2018, the Dart project authors. Please see the AUTHORS file
```
bwilkerson, line 1: nit: year

bwilkerson (commented):
I really wish we had recorded the justification for this rule. I'm not sure why a user would want to avoid having a private typedef for a function type (unless it was being exposed through a public API which could be caught by a different lint).

Without understanding the justification for function typed typedefs, I can't evaluate whether it makes sense to restrict if to function types or whether all typedefs should be treated the same way. As a result, I can't decide whether this change is a positive change.

What I do know is that the argument in the issue, that sometimes it's useful to define a typedef in order to document a structure, makes as much sense for functions as it does for non-function types in my mind.

test_data/rules/experiments/nonfunction-type-aliases/rules/avoid_private_typedef_functions.dart
pq: Since this is a copy of the existing test (w/ one line added for the experiment), I kept the original copyright.  (Once the experiment is enabled by default, this will get folded back into the main test.)

Does that sound right?

pq:
@bwilkerson: as for a justification for this change, @mockturtl provides some motivation in https://github.com/dart-lang/linter/issues/2732.  I agree though, this could all be much better documented.

bwilkerson:
> as for a justification for this change, @mockturtl provides some motivation in #2732.

Yes, I understand the motivation from that issue, and I agree with it. I just happen to believe that the motivation given there applies to function type aliases as well as non-function type aliases. Based on that, my advice would be "Don't enable this lint. It can't tell the useful use cases from the non-useful use cases.".

What I'm missing is the motivation for enabling this rule at all. Without that, I can't decide whether this change is consistent with the intent of the rule.

srawlins:
Should we close this one out as stale?

srawlins:
I'll close as stale for now. Please re-open to land.

pq:
Re-opening as there's internal interest in landing this fix.

/fyi @oprypin

pq:
@bwilkerson: I think this is ready for a re-review.

If you'd like doc changes, I'm up for suggestions.

As for motivation, @oprypin might be able to provide more if needed. (There's internal interest in this lint.)

bwilkerson (approved):
I still don't know whether this lint makes any sense at all, but the changes look right for what we're trying to accomplish, and that's enough for now.

PR #2755, opened 2021-07-06
fix false positive in `unnecessary_string_interpolations` w/ nullable interpolated strings

### LINKED ISSUE #2658: false positive in `unnecessary_string_interpolations` with nullable strings
From the analyzer code base:

```dart
class _PSNode implements PSNode {
  @override
  final String? text;
  @override
  String toString() => '$text'; // <= LINT (but not safe to remove)
}

```

pq:
Fixes #2658

/cc @bwilkerson

### DIFF
--- a/lib/src/rules/unnecessary_string_interpolations.dart
+++ b/lib/src/rules/unnecessary_string_interpolations.dart
@@ -4,6 +4,7 @@
 
 import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/nullability_suffix.dart';
 
 import '../analyzer.dart';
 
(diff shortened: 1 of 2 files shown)


lib/src/rules/unnecessary_string_interpolations.dart
```diff
@@ -57,7 +58,9 @@ class _Visitor extends SimpleAstVisitor<void> {
       var end = node.elements[2] as InterpolationString;
       if (start.value.isEmpty && end.value.isEmpty) {
         var staticType = interpolation.expression.staticType;
-        if (staticType != null && staticType.isDartCoreString) {
+        if (staticType != null &&
+            staticType.isDartCoreString &&
+            staticType.nullabilitySuffix != NullabilitySuffix.question) {
```
bwilkerson, line 63: What do we want the behavior to be if the type is `*`, meaning that it comes from a legacy library? Should we continue to lint it even though it might be nullable?

bwilkerson (approved):

lib/src/rules/unnecessary_string_interpolations.dart
pq: My gut is that we do want to lint it but I'm really not sure.  It'll be interesting to see if this comes up in practice?

PR #2757, opened 2021-07-07
Remove a word from documentation.

asashour:
Correct a sentence in `WritingLints`.

`either` is usually followed by `or`, and it seems the sentence is better without it.


pq:
Good catch.  Thanks!

pq (approved):

PR #2761, opened 2021-07-07
support type aliases in `always_specify_types`

### LINKED ISSUE #2692: [analyzer] Type aliases are not working well with the `always_specify_types` rule
```dart
typedef JsonMap = Map<String, dynamic>;
typedef JsonList = List<JsonMap>;
```
image

This tracker is for issues related to: Analyzer

* Dart SDK Version: 2.13.1
* macOS BigSur

pq:
Maybe too naive? 🤔 

Fixes #2692

/cc @bwilkerson

### DIFF
--- a/test_data/rules/experiments/nonfunction-type-aliases/rules/always_specify_types.dart
+++ b/test_data/rules/experiments/nonfunction-type-aliases/rules/always_specify_types.dart
@@ -11,6 +11,9 @@ typedef StringList = List<JsonMap>; //OK
 typedef RawList = List; //LINT
 typedef StringMap<V> = Map<String, V>; //OK
 
+StringMap<String> sm = StringMap<String>(); //OK
+StringMap rm = StringMap(); //LINT
+
 Map<String, String> map = {}; //LINT
 List<String> strings = []; //LINT
 Set<String> set = {}; //LINT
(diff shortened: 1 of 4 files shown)


test_data/rules/experiments/nonfunction-type-aliases/rules/always_specify_types.dart
```diff
@@ -0,0 +1,97 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+// test w/ `dart test -N always_specify_types`
+
+import 'package:meta/meta.dart';
+
+typedef JsonMap = Map<String, dynamic>; //OK
+typedef StringList = List<JsonMap>; //OK
+typedef RawList = List; //LINT
+typedef StringMap<V> = Map<String, V>; //OK
```
pq, line 12: ^^^ are the new test cases.

bwilkerson: I'd like to see a slightly more complicated test added:
```
typedef MapList = List<StringMap>; // LINT
```
This should create a lint because `StringMap` is generic but doesn't specify type arguments, but I suspect that the condition you added will cause it to not get caught.

pq: Should now be fixed. 🤞

bwilkerson (approved):

Issue #1717, opened 2019-09-11
Analysis lint for trailing whitespace


dnfield:
Flutter has a custom script to check for whitespace at the end of files. It'd be nice to have this as an analyzer lint for dart files so the IDE could warn you ahead of time, at least for Dart, and assuming you're either not using dartfmt or haven't formatted yet.

srawlins:
The linter will not replicate whitespace rules already implemented in dartfmt.

An example of a formatting rule that _is_ implemented in the linter, because the formatter does not split strings or comments, includes [lines_longer_than_80_chars](http://dart-lang.github.io/linter/lints/lines_longer_than_80_chars.html).

The formatter does already collapse all whitespace at the end of a file, and ensures the file ends in a newline. I think these are exactly the two requests you are making, @dnfield. Is there a reason end-of-file whitespace is unique, or separate from other whitespace lint rule possibilities?

dnfield:
We don't use dartfmt in Flutter - for various reasons.  We do expect users to not have whitespace at the end of lines and to have a new line at EOL.  It seems like it would be nice if the analyzer could help with that, although I understand the reluctance given that the formatter can already do this.  We can just continue with our home grown script, but it would be one way to remove a surprise from a relatively small group of users.

srawlins:
I see, while `class C{   }` has glaring whitespace issues, a file missing a trailing newline, or having two newlines, etc. is easily missed until the presubmit script hollars at you.

pq:
Hey @dnfield.  I was looking into this a bit ago and got distracted; sorry!  Confirming this would still be valuable?

dnfield:
I think so  - although I say that assuming this wouldn't be an especially difficult one to implement and that it shouldn't be very highly prioritized.  But it could give helpful feedback to Flutter contributors who are encouraged not to use dartfmt.

pq:
@dnfield: I'm hoping that @asashour's new `eol_at_end_of_file` should be sufficient for you all?

(It's not landed in the SDK yet but will be on our next roll.)

pq:
This is implemented with `eol_at_end_of_file`. 👍

PR #2762, opened 2021-07-08
Implement eol_at_end_of_file rule

### LINKED ISSUE #2756: eol_last
**Description**

This lint ensure a newline at the end of non-empty text files.

It should be applied to all text files by default, and ignored for binary ones.

A similar ESLint is [here](https://eslint.org/docs/rules/eol-last).

This rule follows the definition of [a line in POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206).

asashour:
This PR introduces a new `eol_at_end_of_file` rule.

Fixes #2756

### DIFF
--- a/lib/src/rules/eol_at_end_of_file.dart
+++ b/lib/src/rules/eol_at_end_of_file.dart
@@ -1,0 +1,66 @@
+// Copyright (c) 2021, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+
+import '../analyzer.dart';
+
+const _desc = r'Put a single newline at end of file.';
+
+const _details = r'''
+
+**DO** put a single newline at the end of non-empty files.
+
+**BAD:**
+```dart
+a {
+}
+```
+
+**GOOD:**
+```dart
+b {
+}
+    <-- newline
+''';
+
+class EolAtEndOfFile extends LintRule implements NodeLintRule {
+  EolAtEndOfFile()
+      : super(
+            name: 'eol_at_end_of_file',
+            description: _desc,
+            details: _details,
+            group: Group.style);
+
+  @override
+  void registerNodeProcessors(
+      NodeLintRegistry registry, LinterContext context) {
+    var visitor = _Visitor(this, context);
+    registry.addCompilationUnit(this, visitor);
+  }
+}
+
+class _Visitor extends SimpleAstVisitor<void> {
+  final LintRule rule;
+  final LinterContext context;
+
+  _Visitor(this.rule, this.context);
+
+  @override
+  void visitCompilationUnit(CompilationUnit node) {
+    var content = context.currentUnit.content;
+    if (content.isNotEmpty &&
+        (!content.endsWithNewline || content.endsWithMultipleNewlines)) {
+      rule.reportLintForOffset(content.trimRight().length, 1);
+    }
+  }
+}
+
+extension on String {
+  bool get endsWithNewline => newline.any(endsWith);
+  static const newline = ['\n', '\r'];
+  bool get endsWithMultipleNewlines => multipleNewlines.any(endsWith);
+  static const multipleNewlines = ['\n\n', '\r\r', '\r\n\r\n'];
+}
--- a/example/all.yaml
+++ b/example/all.yaml
@@ -67,6 +67,7 @@ linter:
     - empty_catches
     - empty_constructor_bodies
     - empty_statements
+    - eol_at_end_of_file
     - exhaustive_cases
     - file_names
     - flutter_style_todos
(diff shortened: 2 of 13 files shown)


lib/src/rules/eol_last.dart
```diff
@@ -0,0 +1,63 @@
+// Copyright (c) 2016, the Dart project authors. Please see the AUTHORS file
```
bwilkerson, line 1: nit: "2016" --> "2021"

Also in the test data file.

lib/src/rules/eol_last.dart
```diff
@@ -0,0 +1,63 @@
...
+  @override
+  void visitCompilationUnit(CompilationUnit node) {
+    var endToken = node.endToken;
+    var previous = endToken.previous;
+    if (previous == null) {
+      return;
+    }
+    var comments = endToken.precedingComments;
+    if (previous.end == endToken.offset ||
+        (comments != null && comments.end == endToken.offset)) {
```
bwilkerson, line 59: The `preceedingComments` is a (doubly) linked list of tokens with one token per comment, ordered such that the token with the smallest offset is first in the list. If there's more than one comment at the end of the file then this will look at the end of the first comment when it should be looking at the end of the last comment. For example:
```
class C {}
// This is just a placeholder.
// This file should be replaced with useful code.
```

An easier solution might be to pass the `LinterContext` to the `_Visitor` constructor and store it in an instance field, then use `context.currentUnit.content` to get the content of the file and check to see whether it ends with an end-of-line marker (assuming that it isn't an empty string).

asashour: Thanks a lot for the hint.

lib/src/rules/eol_last.dart
```diff
@@ -0,0 +1,68 @@
...
+```dart
+b {
+}
+
+''';
+
+class EolLast extends LintRule implements NodeLintRule {
+  EolLast()
+      : super(
+            name: 'eol_last',
```
bwilkerson, line 32: Consider renaming this to `eol_at_end_of_file` or `single_eol_at_end_of_file`. I think it might be easier for users to understand, and if we want to also enforce that there's a single eol marker (as was requested in https://github.com/dart-lang/linter/issues/1717) then I think it's general enough to cover that case as well.

lib/src/rules/eol_last.dart
```diff
@@ -0,0 +1,68 @@
...
+  _Visitor(this.rule, this.context);
+
+  @override
+  void visitCompilationUnit(CompilationUnit node) {
+    var endToken = node.endToken;
+    var previous = endToken.previous;
+    if (previous == null) {
+      return;
+    }
+    if (previous.end == endToken.offset ||
```
bwilkerson, line 58: I'm guessing that this is to ensure that the lint doesn't trigger on empty files. I think this will work (I think it will always be true for an empty file and false for a file ending with either whitespace or a comment), but I was thinking of something like
```
var content = context.currentUnit.content;
if (content.isNotEmpty && !content.endsWithNewLine) ...
```
That seems more explicit to me. It took me a while to convince myself that I fully understood the semantics of the comparison.

bwilkerson (commented):
What do you think of extending this to also check that there aren't multiple eol markers in a file?

I can see the benefit in terms of cleanliness. The only downside I can see is that it might trigger when the user adds a couple of blank lines at the end of the file just before typing a new declaration. I suspect those false positives would be acceptable and rare because most people will type the "\n\nclass" fast enough that the linter wouldn't even run when the file's in a bad state.

lib/src/rules/eol_last.dart
pq: +1.  I prefer `eol_at_end_of_file`.

lib/src/rules/eol_last.dart
```diff
@@ -0,0 +1,68 @@
...
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+
+import '../analyzer.dart';
+
+const _desc = r'Put an end of line for non-empty files.';
+
+const _details = r'''
+
+**DO** put an end of line for non-empty files.
```
pq, line 14: I think you want to be clear that this is at the _end of a file_.  Maybe:

    **DO** put at least one newline at the end of non-empty files.

bwilkerson: Or "exactly one" if we extend it to include the full request in #1717.

lib/src/rules/eol_at_end_of_file.dart
```diff
@@ -0,0 +1,66 @@
...
+```
+
+**GOOD:**
+```dart
+b {
+}
+    <-- newline
+''';
+
+class EolLast extends LintRule implements NodeLintRule {
```
bwilkerson, line 29: We usually name the class the same as the rule, but I'm fine for that to get cleaned up in a later PR, either from you or someone else.

bwilkerson (approved):
Thanks for the contribution, and thanks for your patience with the review comments. We appreciate both.

lib/src/rules/eol_at_end_of_file.dart
asashour: Thanks, I noted it, and it was committed :)

lib/src/rules/eol_last.dart
asashour: Thanks for guiding new contributors.

pq (approved):
Fantastic.  Thank you!

pq:
See also: https://github.com/dart-lang/linter/issues/2765.

PR #2764, opened 2021-07-08
Unify kind in description.

asashour:
This PR tries to make the same pattern of the description in all rules.

- `**DO**` instead of `*DO*`
- `**DON'T**` instead of `**DO NOT**`
- Add `**` around `Avoid` and `Prefer`
- Precede the first word with an empty line


srawlins:
Hi @asashour , sorry this has not gotten attention. I certainly loooove consistency! If you can fix the merge conflicts, I can review!

asashour:
That's fine, kindly take a second look.

lib/src/rules/always_use_package_imports.dart
```diff
@@ -10,7 +10,9 @@ import '../ast.dart';
 
 const _desc = r'Avoid relative imports for files in `lib/`.';
 
-const _details = r'''*DO* avoid relative imports for files in `lib/`.
+const _details = r'''
+
```
srawlins, line 14: I actually don't understand why we need a _blank_ line after `r'''`. Can you explain?

srawlins (approved):

srawlins:
Any comments, @pq or @bwilkerson ? I love this.

lib/src/rules/avoid_redundant_argument_values.dart
```diff
@@ -9,7 +9,9 @@ import '../analyzer.dart';
 
 const _desc = r'Avoid redundant argument values.';
 
-const _details = r'''Avoid redundant argument values.
+const _details = r'''
+
+Avoid redundant argument values.
```
bwilkerson, line 14: This looks like an accidental copying of the `_desc` into the `_details` (at some previous time). Perhaps we should remove it.

bwilkerson (commented):
I agree that we should remove the extra blank line between the `r'''` and the first non-whitespace character, but even with that I'm fine to land it. I suspect that the extra whitespace, while unnecessary, will be ignored by the markdown processor.

lib/src/rules/always_use_package_imports.dart
asashour: This was done, since most of the lints are doing so. I would suggest that we either remove all, or ensure all.

bwilkerson: I'm fine with whichever is the least amount of work (which is probably having the blank line, or you would have removed them to make it all consistent).

asashour: No worry, replace by regular expression is fun to do :)

lib/src/rules/avoid_redundant_argument_values.dart
asashour: Done

lib/src/rules/always_use_package_imports.dart
pq: Interesting...  I'll take a look at adding validation.

https://github.com/dart-lang/linter/issues/3709

pq (approved):
Thanks!

PR #2771, opened 2021-07-15
Fix curly_braces_in_flow_control_structures when "if" is "else" of previous "if"

### LINKED ISSUE #2416: `curly_braces_in_flow_control_structures` misses some else-ifs
See: https://github.com/flutter/devtools/pull/2597#pullrequestreview-566571624

asashour:
Fixes #2416

### DIFF
--- a/lib/src/rules/curly_braces_in_flow_control_structures.dart
+++ b/lib/src/rules/curly_braces_in_flow_control_structures.dart
@@ -100,7 +100,9 @@ class _Visitor extends SimpleAstVisitor {
       }
     } else {
       _check(node.thenStatement);
-      if (elseStatement is! IfStatement) {
+      if (elseStatement is IfStatement) {
+        _check(elseStatement.thenStatement);
+      } else {
         _check(elseStatement);
       }
     }
--- a/lib/src/rules/curly_braces_in_flow_control_structures.dart
+++ b/lib/src/rules/curly_braces_in_flow_control_structures.dart
@@ -89,8 +89,8 @@ class _Visitor extends SimpleAstVisitor {
   void visitIfStatement(IfStatement node) {
     var elseStatement = node.elseStatement;
     if (elseStatement == null) {
-      if (node.parent is IfStatement &&
-          node == (node.parent! as IfStatement).elseStatement) {
+      var parent = node.parent;
+      if (parent is IfStatement && node == parent.elseStatement) {
         _check(node.thenStatement);
         return;
       }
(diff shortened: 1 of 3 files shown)


lib/src/rules/curly_braces_in_flow_control_structures.dart
```diff
@@ -100,7 +100,9 @@ class _Visitor extends SimpleAstVisitor {
       }
     } else {
       _check(node.thenStatement);
-      if (elseStatement is! IfStatement) {
+      if (elseStatement is IfStatement) {
+        _check(elseStatement.thenStatement);
```
bwilkerson, line 104: While I agree that the test case you added should be flagged, I think that the impact of this change would be to cause two lints to be generated for the following code:
```dart
void f() {
  if (true) {
  } else if (true)
    return;
}
```
One when visiting the outer `if` and `_check` discovers that the argument isn't a block, and one when visiting the inner `if` and we get to line 99. (I haven't applied your delta to confirm, so I might be wrong.)

If I'm right, then I believe the correct fix would be to add code between lines 91 and 92 that will `_check` the `thenStatement` if the `node` is the `elseStatement` of an outer `if` (and return to skip other checks).

asashour: Oh, that's interesting.

I tested with both

```dart
testIfElse() {
  void f() {
    if (true) {
    } else if (true) return;
  }
}
```

and

```dart
void f() {
  if (true) {
  } else if (true)
    return;
}
```

And lint is triggered only once.


BTW, without the proposed change, the below doesn't trigger lint:

```dart
  void f() {
    if (true) {
    } else if (true) return;
  }
```

asashour: Apologies, you are totally right.

lib/src/rules/curly_braces_in_flow_control_structures.dart
```diff
@@ -89,6 +89,11 @@ class _Visitor extends SimpleAstVisitor {
   void visitIfStatement(IfStatement node) {
     var elseStatement = node.elseStatement;
     if (elseStatement == null) {
+      if (node.parent is IfStatement &&
+          node == (node.parent! as IfStatement).elseStatement) {
```
bwilkerson, line 93: Our preferred style is to assign fields like this to a local variable so that type inference can do more of the work. In this case that would look like
```dart
var parent = node.parent;
if (parent is IfStatement && node == parent.elseStatement) {
  // ...
}
```
That allows both the null check (`!`) and the cast to be omitted.

bwilkerson (approved):
Thank you, that does address my concern. I have a stylistic comment, but I'm willing to commit this as is and clean it up later if you'd prefer. Please let me know your preference.

lib/src/rules/curly_braces_in_flow_control_structures.dart
asashour: Thanks for your feedback. I will change it when I return back, and will add a line in`changes`.

bwilkerson (approved):
Thanks!

PR #2773, opened 2021-07-16
Implement null_safety rule

asashour:
Implements the first requirement of #2702

I am not sure how best to process `pubspec.yaml`

Should the name be `sound_null_safety`, `unsound_null_safety`, ...?


lib/src/rules/null_safety.dart
```diff
@@ -0,0 +1,74 @@
...
+**GOOD:**
+```dart
+b {
+}
+''';
+
+class NullSafety extends LintRule implements NodeLintRule {
+  NullSafety()
+      : super(
+            name: 'null_safety',
```
bwilkerson, line 34: Although it isn't strictly a requirement, consider naming this with a verb phrase, something like `enable_null_safety`. We generally try to name lints so that they describe either what _should_ be done or what _shouldn't_ be done.

lib/src/rules/null_safety.dart
```diff
@@ -0,0 +1,74 @@
...
+  final LinterContext context;
+
+  _Visitor(this.rule, this.context);
+
+  @override
+  void visitCompilationUnit(CompilationUnit node) {
+    var content = context.currentUnit.content;
+    var lines = LineSplitter().convert(content);
+
+    var regExp = RegExp(r"//@dart\=2\.(\d)");
```
bwilkerson, line 58: This regex doesn't work with all of the possibilities. Users are allowed to have whitespace almost anywhere in the line, and we probably need to catch comments that use a major version of `1` (though that might be disallowed in general; I cant remember). We should probably share the regex used by `LanguageVersionOverrideVerifier` (in the analyzer package). Unfortunately that class isn't public API (and the field is private), so you'll have to copy it with a note that it needs to be kept in sync with the original (though I can't imagine that the format will change).

The use of `firstMatch` below means that it can match even if there's non-whitespace before the comment:
```
void main() { //@dart=2.2
}
```
but that isn't considered a version comment because the version comment has to come before any non-comment in the file. The use of `firstMatch` also means that it could match text in a string literal. Both of those are unlikely possibilities, but there is a slightly different approach that would solve those cases and be more performant when dealing with larger files.

The AST structure provides access to the underlying tokens, including comments. In particular you can ask any node for the comment tokens that proceed it. The class `LanguageVersionOverrideVerifier` uses this to limit the search to only those comments that appear before the first non-comment token. Consider using a similar approach.

test_data/integration/null_safety/src/null_safety.dart
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) 2021, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// @dart=2.8
```
bwilkerson, line 5: I'd recommend testing boundary conditions, so consider using `2.11` instead. We should also have a test with `2.12` to ensure that we're not generating a lint in that case.

bwilkerson (commented):
Thanks!

You should be able to handle the pubspec case by overriding `PubspecVisitor getPubspecVisitor()` in your subclass of `LintRule`. I haven't verified that it will give you access to everything you need, but we can extend it if necessary.

lib/src/rules/null_safety.dart
asashour: I see, it is now `enable_null_safety`. Another suggestion is `ensure_sound_null_safety`.

lib/src/rules/null_safety.dart
asashour: I see, in the previous code there was `.replaceAll(' ', '')` to remove all whitespaces before using the `RegExp`.

Anyhow, it is now changed to the regular expression from `LanguageVersionOverrideVerifier` which contains the initial line `^`, I believe that would make things better, since it will always consider line beginning.

Also, the comments of the first token are iterated.

Regarding the `PubspecVisitor`, by naked eye, I couldn't see something related to the `environment` or `sdk`, and since this is in `analyzer`, how about restricting this PR scope to the dart files. This `PubspecVisitor` seems to have dependency elsewhere.

bwilkerson: I missed the line to remove spaces, thanks.

Yes, I'm totally fine with merging this as-is and adding the pubspec handling later. I don't see support for the sdk version either, so we'll need to add that.

bwilkerson (approved):

bwilkerson:
@pq Do we want to merge this now (well, after the conflict is resolved) and risk it being a breaking change to have additional checks, or would it be better to merge this after the next roll into the SDK so that it can't be used yet? A third option would be to mark it as experimental until the pubspec checking has been added. (FYI, Flutter is interested in this lint.)

lib/src/rules/enable_null_safety.dart
```diff
@@ -0,0 +1,70 @@
...
+const _desc = r'Do use sound null safety.';
+
+const _details = r'''
+
+**DO** use sound null safety, by not specifying a dart version lower than `2.12`.
+
+**BAD:**
+```dart
+// @dart=2.8
+a {
```
pq, line 19: Perhaps:

```
a() {

}
```

?

lib/src/rules/enable_null_safety.dart
```diff
@@ -0,0 +1,70 @@
...
+}
+```
+
+**GOOD:**
+```dart
+b {
```
pq, line 25: And,

```
b() {

}
```

?

lib/src/rules/enable_null_safety.dart
```diff
@@ -0,0 +1,70 @@
...
+      : super(
+            name: 'enable_null_safety',
+            description: _desc,
+            details: _details,
+            group: Group.style);
+
+  @override
+  void registerNodeProcessors(
+      NodeLintRegistry registry, LinterContext context) {
+    var visitor = _Visitor(this, context);
```
pq, line 40: I don't know how expensive the comment parsing below is but I wonder if it's worth bailing if non-nullability is enabled?

```dart
    if (context.isEnabled(Feature.non_nullable)) {
      return;
    }
```

bwilkerson: The comments have already been parsed and are in the AST so it's fairly inexpensive to iterate over them, but it wouldn't hurt to check.

pq: 👍 

More specifically I mean the cost of the **_regexps_**.

asashour: Thanks for hinting.

```dart
if (context.isEnabled(Feature.non_nullable)) {
  return;
}
```

doesn't work, giving all cases are ok (2 should fail).

On the other hand, there was also a suggestion to do something like:

```dart
if (line.replaceAll(' ', '').startsWith('//@dart=')) {
    // parse the version
}
```

if there is a concern about `regExp`.

srawlins:
I think it would be super good to land this. Would you mind fixing the merge conflict, @asashour ?

srawlins:
@bwilkerson @pq this is the rule that bans `// @dart=2.9`, etc. Likely not needed for Dart >= 3.0, but maybe extra useful before then.

lib/src/rules/enable_null_safety.dart
```diff
@@ -0,0 +1,69 @@
...
+  final LintRule rule;
+  final LinterContext context;
+  // to be kept in sync with LanguageVersionOverrideVerifier (added groups for the version)
+  final regExp = RegExp(r'^\s*//\s*@dart\s*=\s*(\d+)\.(\d+)');
+
+  _Visitor(this.rule, this.context);
+
+  @override
+  void visitCompilationUnit(CompilationUnit node) {
+    CommentToken? comment = node.beginToken.precedingComments;
```
bwilkerson, line 54: I didn't find any documentation that explicitly mentions it, but I think it's valid to have both a script tag and a language version override comment in the same file, and I'm guessing that the comment follows the script tag.

That's an unlikely corner case, but if I'm right then the `beginToken` might be a script tag token, in which case you would need to skip to the next token before getting the `precedingComments`.

lib/src/rules/enable_null_safety.dart
bwilkerson: It might not be too bad, given that we only need to look at the comments before the first non-comment token in the file (modulo the corner case below). The performance is then dependent on the length of the file header and library documentation comments.

test/integration/enable_null_safety.dart
```diff
@@ -0,0 +1,33 @@
...
+import 'package:analyzer/src/lint/linter.dart';
+import 'package:linter/src/analyzer.dart';
+import 'package:linter/src/cli.dart' as cli;
+import 'package:test/test.dart';
+
+import '../mocks.dart';
+import '../test_constants.dart';
+
+void main() {
+  group('enable_null_safety', () {
```
bwilkerson, line 15: I'd rather have these written using `reflective_test_loader`, but doing so shouldn't hold up merging the PR.

bwilkerson (approved):

bwilkerson (approved):

pq (approved):
👍

pq:
I'd like to hold off on landing this until after the currently in progress roll is finished. (See: https://dart-review.googlesource.com/c/sdk/+/264961)

Thanks for your patience!

PR #2774, opened 2021-07-16
Remove `pub get` from action

kevmoo:


kevmoo:
I have a tool that looks for "bad strings" – this is getting caught up in that

devoncarew (approved):

pq (commented):
👍

PR #2786, opened 2021-07-20
Ignore unnecessary cast warning.

stereotype441:
https://dart-review.googlesource.com/c/sdk/+/207360 improved the
analyzer by making CompoundAssignmentExpression a subtype of
Expression; this means that once a version of the analyzer is
published with this improvement, and the linter can depend on that
version, it will be possible to promote from Expression to
CompoundAssignmentExpression (and the linter will be able to remove a
cast).  Until then, the cast is still required to work with older
analyzers, but will trigger an "unnecessary cast" warning when
analyzing against the newest analyzer.

So for now we ignore the "unnecessary cast" warning.  Once the linter
depends on a later version of the analyzer, we'll be able to clean
this up.

### DIFF
--- a/lib/src/rules/avoid_dynamic_calls.dart
+++ b/lib/src/rules/avoid_dynamic_calls.dart
@@ -231,8 +231,13 @@ class _Visitor extends SimpleAstVisitor<void> {
       return;
     }
     if (root is CompoundAssignmentExpression) {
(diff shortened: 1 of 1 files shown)


srawlins (approved):

pq (approved):
Thanks!

pq:
Thanks!

PR #2791, opened 2021-07-22
allow keywords for dart package_names

### LINKED ISSUE #2790: False positive for package_names
This lint disallows the name `async` with the lint "Use `lowercase_with_underscores` for package names."

pq:
Fixes #2790

`isValidDartIdentifier` rejects keywords too which is too restrictive for package names.

/cc @natebosch @bwilkerson

### DIFF
--- a/lib/src/utils.dart
+++ b/lib/src/utils.dart
@@ -62,8 +62,7 @@ bool isValidLibraryPrefix(String libraryPrefix) =>
 
 /// Returns true if this [id] is a valid package name.
 bool isValidPackageName(String id) =>
-    _lowerCaseUnderScoreWithLeadingUnderscores.hasMatch(id) &&
-    isValidDartIdentifier(id);
+    _lowerCaseUnderScoreWithLeadingUnderscores.hasMatch(id) && isIdentifier(id);
 
 class CamelCaseString {
   static final _camelCaseMatcher = RegExp(r'[A-Z][a-z]*');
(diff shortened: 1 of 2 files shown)


natebosch (approved):
Thanks!

parlough:
So after this change, are all reserved keywords are allowed as package names according to the lint or am I missing something?

Taking a look at `pub`, it [checks](https://github.com/dart-lang/pub/blob/master/lib/src/pubspec.dart#L120-L123) to make sure a name is not [one of these reserved words](https://github.com/dart-lang/pub/blob/master/lib/src/utils.dart#L32-L64) or else it errors.

For example, `break` seems to not trigger the lint, but you get this error when running a pub command:
```
Error on line 1, column 7 of pubspec.yaml: "name" field may not be a Dart reserved word.
  ╷
1 │ name: break
  │       ^^^^^
  ╵
```

pq:
Oh!  I forgot to close the loop on this.  I think that's not ideal.  We used to disallow all **_keywords_**  which was too much; disallowing reserved words does seem right though since that will align us w/ pub.

pq:
Thanks for catching this @parlough !  I've got a fix here unless you already have one?

parlough:
Sounds good to me, thanks ! That will also align with the current description: https://github.com/dart-lang/linter/blob/master/lib/src/rules/pub/package_names.dart#L19

PR #2796, opened 2021-07-23
migrate `prefer_is_not_empty` from visitSimpleIdentifier to more specific visitPrefixEpression

pq:
Improves performance a bit and is a little tidier (note the use of `unParenthesized`).

Benchmark runs:

**before:**

    -----------------------------------------------------------------------------
    Timings                                                                    ms
    -----------------------------------------------------------------------------

    prefer_is_not_empty [core, pedantic]                                       23

**after:**

    prefer_is_not_empty [core, pedantic]                                        2

([Task link](https://github.com/dart-lang/linter/pull/2796/checks?check_run_id=3147906276).)

🔥 

/cc @srawlins @bwilkerson

### DIFF
--- a/lib/src/rules/prefer_is_not_empty.dart
+++ b/lib/src/rules/prefer_is_not_empty.dart
@@ -3,13 +3,7 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'package:analyzer/dart/ast/ast.dart'
-    show
-        AstNode,
-        ParenthesizedExpression,
-        PrefixExpression,
-        PrefixedIdentifier,
-        PropertyAccess,
-        SimpleIdentifier;
+    show PrefixExpression, PrefixedIdentifier, PropertyAccess, SimpleIdentifier;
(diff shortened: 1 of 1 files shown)


lib/src/rules/prefer_is_not_empty.dart
```diff
@@ -63,48 +57,39 @@ class _Visitor extends SimpleAstVisitor<void> {
...
+    // Should be prefixed w/ a "!".
+    var prefix = node.operator;
+    if (prefix.type != TokenType.BANG) {
+      return;
     }
 
+    var expression = node.operand.unParenthesized;
+
+    // Should be a property access or prefixed identifier.
+    SimpleIdentifier? isEmptyIdentifier;
```
bwilkerson, line 70: Not important, but you could probably make this non-nullable (and remove the check on line 76) by rewriting slightly as
```
if (expression is Prop...) {
  ...
} else if (...) {
  ...
} else {
  return;
}
```

bwilkerson (approved):

PR #2799, opened 2021-07-25
Fix no_logic_in_create_state for empty body

### LINKED ISSUE #2345: no_logic_in_create_state abstract methods are false positives
An abstract class that overrides `createState` to change its signature but has no body will trigger the no_logic_in_create_state lint.

asashour:
Fixes #2345


pq (approved):
Super.  Thanks!

PR #2800, opened 2021-07-25
Fix unnecessary_parenthesis for PropertyAccess and MethodInvocation

### LINKED ISSUE #2176: unnecessary_parenthesis is not reported for (foo.bar).baz
**Describe the issue**
A clear and concise description of what the issue is.

If you're seeing unexpected behavior in a lint rule, please include the name of the rule in the issue description.

**To Reproduce**
```dart
main() {
  print((0.sign).isEven);
}
```

**Expected behavior**
I'd expect that `unnecessary_parenthesis` is reported for `0.sign`.

**Additional context**

asashour:
Fixes #2176

I wonder if are cases which shouldn't have the lints.

Also, should we have `var expression = node.expression` at the beginning of the method, since it is used many times.


srawlins (approved):

lib/src/rules/unnecessary_parenthesis.dart
```diff
@@ -120,6 +120,11 @@ class _Visitor extends SimpleAstVisitor<void> {
         if (target == node &&
             node.expression is SetOrMapLiteral &&
             parent.parent is ExpressionStatement) return;
+
+        if (node.expression is PropertyAccess ||
+            node.expression is MethodInvocation) {
```
bwilkerson, line 125: Do we really need to special case these two kinds of nodes, or would it be sufficient to change line 130 from `<` to `<=`? That's effectively what we're doing here. The precedence or property access and method invocation  are the same (`Precedence.postfix`), but this is missing a lot of other node types that also have that precedence. If we can't change line 130, then we should at least change this condition to be in terms of precedence values in order to capture those cases (and any kinds of expressions that might be added with the same precedence in the future).

asashour: Thanks for hinting about `Precedence`.

However, there two expressions with `Precedence.postfix`, one is ok, the other is not:

```dart
  (a++).toString(); // OK
  (0.sign).isEven; // LINT
```

Should we check all other expressions which belong to `postfix` and report a lint for a subset of them? Or is there anything else?

bwilkerson: Interesting. We should eventually figure out why precedence isn't sufficient and add an API to the AST to let us make this determination in a less ad-hoc manner, but until then we should expand the test to cover the other cases where it's valid to have a `prefix` precedence expression as the target of an invocation. We should also add tests for both the valid and invalid cases to prevent regressions.

bwilkerson: A TODO comment to remind us of the desire for a better API would be great.

asashour: Below are all expressions with `postfix` precedence, with the current status.

| Expression  | Status |
| ------------- | ------------- |
| ConstructorReference  | Don't know how to test |
| ExtensionOverride  | Don't know how to test |
| FunctionExpressionInvocation  | Test existed before |
| FunctionReference  | Don't know how to test |
| IndexExpression  | Test added |
| LibraryIdentifier  | Don't know how to test |
| MethodInvocation  | Test added |
| PostfixExpression  | Test existed before  |
| PrefixedIdentifier  | Test added |
| PropertyAccess  | Test added |
| TypeLiteral  | Don't know how to test  |

bwilkerson: `ConstructorReference` is support for a future language feature, and I think it isn't implemented enough to be testable at this point.

I don't believe it's valid for a `LibraryIdentifier` to appear in an expression (the type hierarchy is a bit wonky at that point).

`ExtensionOverride`:
```
extension E on String {
  String m() => this;
}

void f(String s) {
  (E(s)).m(); // Shouldn't be flagged because it's already an error for other reasons.
}
```

`FunctionReference`:
```
void f(int i) {
  (i.toString).call(); // LINT
}
```

`TypeLiteral`:
```
void f() {
  (int).toString(); // OK, removing the parens changes the semantics.
}
```

asashour: Thanks for hinting.

Using the provided sample, there is no `ExtensionOverride`, `FunctionReference` or `TypeLiteral`. There are other things (e.g. `SimpleIdentifier` and `PrefixedIdentifier`), but not what is needed to be tested.

I guess I am missing something.

<details>
<summary>ExtensionOverride</summary>

```dart
extension E on String {
  String m() => this;
}
void f(String s) {
  (E(s)).m(); // OK
}
```

```
ExtensionDeclarationImpl [extension E on String {String m() => this;}] 
  DeclaredSimpleIdentifier [E] 
  TypeNameImpl [String] 
    SimpleIdentifierImpl [String] 
  MethodDeclarationImpl [String m() => this;] 
    TypeNameImpl [String] 
      SimpleIdentifierImpl [String] 
    DeclaredSimpleIdentifier [m] 
    FormalParameterListImpl [()] 
    ExpressionFunctionBodyImpl [=> this;] 
      ThisExpressionImpl [this] 
FunctionDeclarationImpl [void f(String s) {(E(s)).m();}] 
  TypeNameImpl [void] 
    SimpleIdentifierImpl [void] 
  DeclaredSimpleIdentifier [f] 
  FunctionExpressionImpl [(String s) {(E(s)).m();}] 
    FormalParameterListImpl [(String s)] 
      SimpleFormalParameterImpl [String s] 
        TypeNameImpl [String] 
          SimpleIdentifierImpl [String] 
        DeclaredSimpleIdentifier [s] 
    BlockFunctionBodyImpl [{(E(s)).m();}] 
      BlockImpl [{(E(s)).m();}] 
        ExpressionStatementImpl [(E(s)).m();] // OK
          MethodInvocationImpl [(E(s)).m()] 
            ParenthesizedExpressionImpl [(E(s))] 
              MethodInvocationImpl [E(s)] 
                SimpleIdentifierImpl [E] 
                ArgumentListImpl [(s)] 
                  SimpleIdentifierImpl [s] 
            SimpleIdentifierImpl [m] 
            ArgumentListImpl [()] 
```
</details>

<details>
<summary>FunctionReference</summary>

```dart
void f(int i) {
  (i.toString).call(); // LINT
}
```

```
FunctionDeclarationImpl [void f(int i) {(i.toString).call();}] 
  TypeNameImpl [void] 
    SimpleIdentifierImpl [void] 
  DeclaredSimpleIdentifier [f] 
  FunctionExpressionImpl [(int i) {(i.toString).call();}] 
    FormalParameterListImpl [(int i)] 
      SimpleFormalParameterImpl [int i] 
        TypeNameImpl [int] 
          SimpleIdentifierImpl [int] 
        DeclaredSimpleIdentifier [i] 
    BlockFunctionBodyImpl [{(i.toString).call();}] 
      BlockImpl [{(i.toString).call();}] 
        ExpressionStatementImpl [(i.toString).call();] // LINT
          MethodInvocationImpl [(i.toString).call()] 
            ParenthesizedExpressionImpl [(i.toString)] 
              PrefixedIdentifierImpl [i.toString] 
                SimpleIdentifierImpl [i] 
                SimpleIdentifierImpl [toString] 
            SimpleIdentifierImpl [call] 
            ArgumentListImpl [()] 
```
</details>

<details>
<summary>TypeLiteral</summary>

```dart
void f() {
  (int).toString(); // OK
}
```

```
DeclaredSimpleIdentifier [f] 
FunctionExpressionImpl [() {(int).toString();}] 
  FormalParameterListImpl [()] 
  BlockFunctionBodyImpl [{(int).toString();}] 
    BlockImpl [{(int).toString();}] 
      ExpressionStatementImpl [(int).toString();] // OK
        MethodInvocationImpl [(int).toString()] 
          ParenthesizedExpressionImpl [(int)] 
            SimpleIdentifierImpl [int] 
          SimpleIdentifierImpl [toString] 
          ArgumentListImpl [()] 
```
</details>

bwilkerson: Ok, then I guess we don't need to worry about those cases. `FunctionReference` is part of the same language feature as `ConstructorReference`, I'd just forgotten that earlier. The other two cases are valid syntax, but obviously aren't parsed as the node types I thought they would be. I can't think of any other way to generate either an `ExtensionOverride` or a `TypeLiteral`, so I think we're good. Sorry about the confusion.

bwilkerson (approved):

srawlins:
`unnecessary_parenthesis` is not used internally, so no testing needed. Only testing in the flutter repos (flutter/flutter, flutter/engine, flutter/plugins, and I think flutter/gallery)

oprypin:
It would be good to pick up this change.
It has significant merge conflicts right now, though.

oprypin:
f7ced37e4f07465c06f43bf55918d38494b19d01 is my attempt at a merge.

The only real conflict is that this piece is not possible to reuse:
`if (parent is PropertyAccess || parent is MethodInvocation) {`
because it was moved into `wouldBeParsedAsStatementBlock`

pq (approved):
Once merge conflicts are sorted, LGTM!

oprypin:
Sure, I'm keeping track of google3. The lint is not even enabled there, so even less concern.

pq:
Great.  Thanks @oprypin!

pq:
Thanks, as always, for your contribution @asashour and your patience!

PR #2802, opened 2021-07-26
stop visiting simple identifiers in `prefer_void_to_null`

pq:
Improves perf a bit.

Benchmark runs:

**before:**

    -----------------------------------------------------------------------------
    Timings                                                                    ms
    -----------------------------------------------------------------------------

    prefer_void_to_null [recommended]                                          14

**after:**

    prefer_void_to_null [recommended]                                           7

/cc @bwilkerson


bwilkerson (approved):
We should consider adding a test that parses each rule's implementation and verified that
- if a registry call is made the corresponding visit method exists and
- if a visit method is defined there is a corresponding call to register it.

PR #2804, opened 2021-07-26
Improve RegExps.

### LINKED ISSUE #2803: False positive for `package_names` lint for name without underscores.
**Describe the issue**

The package `async` gets a warning about its package name not being lowercase-with-underscores.

**To Reproduce**

Make a package with a single-identifier name. Enable recommended lints.

**Expected behavior**

No warning. Single-identifier package names are common (`characters`, `test`, `flutter`, etc.).

**Additional context**

The RegExp in `utils.dart` is 
```dart
final _lowerCaseUnderScoreWithLeadingUnderscores =
    RegExp(r'^(_)*([a-z]+([_]?[a-z0-9]+)*)+$');
```

The final `+` should possibly be `*` instead.

lrhn:
Some of the current RegExps in `utils.dart` risk [catastrophic backtracking](https://swtch.com/~rsc/regexp/regexp1.html), which allows [denial of service attacks](https://en.wikipedia.org/wiki/ReDoS).

This change should fix those RegExps to guarantee nominally linear execution,
and generally clean up unnecessary complexity.

See: https://github.com/dart-lang/linter/issues/2803

The change should not alter the behavior of the RegExps other than where it's deliberately chosen to fix the linked issue, or where the current behavior is clearly erroneous. But do check that there are not further errors.

The change also attempts to define, in words, what each RegExp is supposed to match.

Fixes #2803

### DIFF
--- a/lib/src/utils.dart
+++ b/lib/src/utils.dart
@@ -97,8 +97,9 @@ class CamelCaseString {
   // * followed by any number of letters, digits, `?` or `$`s.
   //
   // This ensures that the text contains a `$`, `?` or upper-case letter
-  // before any lower-case letter or digit.
-  static final _camelCaseTester = RegExp(r'^[_$]*[$?A-Z][$?a-zA-Z\d]*$');
+  // before any lower-case letter or digit, and no letters or `?`s before an
+  // `_`.
+  static final _camelCaseTester = RegExp(r'^_*(?:\$+_+)*[$?A-Z][$?a-zA-Z\d]*$');
 
   final String value;
   CamelCaseString(this.value) {
--- a/lib/src/utils.dart
+++ b/lib/src/utils.dart
@@ -92,13 +92,13 @@ class CamelCaseString {
   static final _camelCaseMatcher = RegExp(r'[A-Z][a-z]*');
 
   // A camel case string here is defined as:
-  // * An arbitrary number of optional leading `_`s,
+  // * An arbitrary number of optional leading `_`s or `$`s,
   // * followed by an upper-case letter, `$` or `?`,
   // * followed by any number of letters, digits, `?` or `$`s.
   //
   // This ensures that the text contains a `$`, `?` or upper-case letter
   // before any lower-case letter or digit.
-  static final _camelCaseTester = RegExp(r'^_*[$?A-Z][$?a-zA-Z\d]*$');
+  static final _camelCaseTester = RegExp(r'^[_$]*[$?A-Z][$?a-zA-Z\d]*$');
 
   final String value;
   CamelCaseString(this.value) {
--- a/lib/src/utils.dart
+++ b/lib/src/utils.dart
@@ -92,10 +92,13 @@ class CamelCaseString {
   static final _camelCaseMatcher = RegExp(r'[A-Z][a-z]*');
 
   // A camel case string here is defined as:
-  // * An arbitrary number of optional leading `_`s or `$`s,
-  // * followed by an upper-case letter or `?`,
+  // * An arbitrary number of optional leading `_`s,
+  // * followed by an upper-case letter, `$` or `?`,
   // * followed by any number of letters, digits, `?` or `$`s.
-  static final _camelCaseTester = RegExp(r'^[_$]*[A-Z?][$?a-zA-Z\d]*$');
+  //
+  // This ensures that the text contains a `$`, `?` or upper-case letter
+  // before any lower-case letter or digit.
+  static final _camelCaseTester = RegExp(r'^_*[$?A-Z][$?a-zA-Z\d]*$');
 
   final String value;
   CamelCaseString(this.value) {
--- a/lib/src/utils.dart
+++ b/lib/src/utils.dart
@@ -2,25 +2,51 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-final _identifier = RegExp(r'^([(_|$)a-zA-Z]+([_a-zA-Z0-9])*)$');
+// An identifier here is defined as:
+// * A sequence of `_`, `$`, letters or digits,
+// * where no `$` comes after a digit.
+final _identifier = RegExp(r'^[_$a-z]+(\d[_a-z\d]*)?$', caseSensitive: false);
 
-final _lowerCamelCase = RegExp(
-    r'^(_)*[?$a-z][a-z0-9?$]*(([A-Z][a-z0-9?$]*)|([_][0-9][a-z0-9?$]*))*_?$');
+// A lower camel-case here is defined as:
+// * Any number of optional leading underscores,
+// * a lower case letter, `$` or `?` followed by a "word-tail"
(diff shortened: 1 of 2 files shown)


lib/src/utils.dart
```diff
@@ -2,25 +2,51 @@
...
+// A lower camel-case here is defined as:
+// * Any number of optional leading underscores,
+// * a lower case letter, `$` or `?` followed by a "word-tail"
+//   (a sequence of lower-case letters, digits, `$` or `?`),
+// * followed by any number of either
+//     * an upper case letter followed by a word tail, or
+//     * an underscore and then a digit followed by a word tail.
+// * and potentially ended by a single optional underscore.
+final _lowerCamelCase =
+    RegExp(r'^_*[?$a-z][a-z\d?$]*(?:(?:[A-Z]|_\d)[a-z\d?$]*)*_?$');
```
lrhn, line 19: The `?`s here look odd. Should they be `_`'s?

The current (and original) RegExp matches `???A?_2?`.
If it should be `_`s instead, the RegExp should be
```dart
    RegExp(r'^_*?[_$a-z][a-z\d_$]*(?:(?:[A-Z]|_\d)[a-z\d_$]*)*_?$');
```

lib/src/utils.dart
```diff
@@ -2,25 +2,51 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-final _identifier = RegExp(r'^([(_|$)a-zA-Z]+([_a-zA-Z0-9])*)$');
```
lrhn, line 5: The `(_|$)` is very likely a mistake since it allows identifiers to start with `(`, `|` or `)`.

lib/src/utils.dart
```diff
@@ -64,12 +90,17 @@ bool isValidPackageName(String id) =>
 
 class CamelCaseString {
   static final _camelCaseMatcher = RegExp(r'[A-Z][a-z]*');
-  static final _camelCaseTester = RegExp(r'^([_$]*)([A-Z?$]+[a-z0-9]*)+$');
+
+  // A camel case string here is defined as:
+  // * An arbitrary number of optional leading `_`s or `$`s,
+  // * followed by an upper-case letter or `?`,
+  // * followed by any number of letters, digits, `?` or `$`s.
+  static final _camelCaseTester = RegExp(r'^[_$]*[A-Z?][$?a-zA-Z\d]*$');
```
lrhn, line 98: There are unexpected `?`s in this RegExp too. Should they be `_`?

If so, the RegExp should be:
```dart
RegExp(r'^[$_A-Z][$_a-zA-Z\d]*$');
```

This RegExp is not matching the original, the original is quite hard to read. I'll update it again to fix that.

lib/src/utils.dart
```diff
@@ -2,25 +2,51 @@
...
+    RegExp(r'^_?[_a-z\d]*(?:\.[a-z][_a-z\d]*)*$');
 
+// A lower-case underscored (snake-case) with leading underscores is defined as
+// * An optional leading sequence of any number of underscores,
+// * followed by a sequence of lower-case letters, digits and underscores,
+// * with no two adjacent underscores,
+// * and not ending in an underscore.
 final _lowerCaseUnderScoreWithLeadingUnderscores =
-    RegExp(r'^(_)*([a-z]+([_]?[a-z0-9]+)*)+$');
+    RegExp(r'^_*[a-z](?:_?[a-z\d])*$');
```
lrhn, line 38: Here I deliberately changed the behavior to allow single identifiers (since it's used for package names, and package names can, e.g., `test`.

lib/src/utils.dart
pq: Oh wow.  Good catch!

pq:
Thanks for this!  I'll take a closer look this evening or tomorrow. 👍

pq:
Sorry for the slowness.  My biggest concern is any unexpected ecosystem breakages due to semantics changes.

I started a try run here:

https://dart-review.googlesource.com/c/sdk/+/208700

If it clears flutter and flutter-engine try-bots, that's a good thing.  Dart CI is the next indicator.  If that passes, we can run a global internal presubmit and see if anything shakes out across Google3.

lrhn:
There should be only two deliberate changes: The `($|_)` typo and changing `_lowerCaseUnderScoreWithLeadingUnderscores` to allow single identifiers (which was my original goal). If you already fixed the latter problem for #2790 , it can be change back.

I do think you should check whether the (now documented) behavior of the rest are what is *intended*. The two which accepts `?` as part of the name look somewhat suspicious to me, and `_lowerCamelCase` also seems a little too permissive if the `?` was supposed to be `_`.


(General rule that I try to live by: Never write a `RegExp` without writing a prose description of what it's intended to match. No-one can debug or test it if they don't know the intent.)

pq:
Internal smoke tests are failing on a use of `$` that looks like this:

```dart
   class  _$_FooBar<T>
```

(Which reminds me that I think some code generators use `$`  and `_`; maybe to avoid name collisions?)  Anyway, I do think we probably want to preserve this behavior.

As for specifying these (and all) regular expressions, you're totally right.  These have evolved over time and we haven't done a great job of capturing the rationale and outside the loose description in @munificent's effective Dart docs and some tests exercising special cases we don't have much formality -- not good since so much depends on them!

Another thing that's interesting is how likely it is that these patterns are being repeated by a number of clients (e.g., analyzer, pub, etc.) Odds are high there's some skew.  I've often thought it'd be nice to have a shared package to capture these and related patterns but haven't followed through.

lrhn:
That was an unintended change, so good testing! I've restored the older behavior. It's not great because `$$$$$$$$$-` can match in a polynomial number of ways. I'll try to find a safer appraoch which accepts the same strings.

The RE for CamelCase is somewhat questionable. It allows `_a`, which isn't particularly upper-camel-case (and then there are  the `?`s). Do consider what it's *intended* to match, it's very likely that it's easier match that efficiently than the current behavior.

pq:
Thanks for iterating!

I've update my "try" PR w/ this latest commit and will look forward to the test results:

https://dart-review.googlesource.com/c/sdk/+/208700

>  RegExps after all, they're write-only

💯 

I wish we'd been a bit better at keeping a running tally of changes w/ rationales.  Getting formal here, in retrospect, would have been worth the effort.

Thanks so much for helping shore this up!

pq:
The internal smoke tests run clean.  The next step is a global tap presubmit to see if anything else shakes out.  If you have a chance in off MTV hours to kick that off, feel free.  Otherwise I can try and do it early tomorrow morning (out tonight).

pq:
Running a fresh try: https://dart-review.googlesource.com/c/sdk/+/210500

pq:
Smoke tests look good enough that I'd like to go ahead and land and try and integrate this internally.

Thanks for your patience!

pq (approved):

lrhn:
Wohoo! Thanks!

pq:
Thank **you**! 👍

PR #2806, opened 2021-07-26
Fix invariant_booleans after reassignment

### LINKED ISSUE #2696: Yet another false positive with invariant_booleans
Hard to tell if this is duplicate of any of the many other open issues with `invariant_booleans`.

```dart
void f({required bool a, required bool b}) {
  var c = a;
  if (c) {
    return;
  }
  c = b;
  if (c) {  // <-- False positive here.
    return;
  }
  print('hi');
}
```

asashour:
Fixes #2696


srawlins:
We've deprecated this lint rule, so will not be landing any changes to it for correctness.

PR #2810, opened 2021-07-27
Fix false positive case in `invariant_booleans`

### LINKED ISSUE #2478: Analyzer false positive invariant_booleans
This tracker is for issues related to:

* Analyzer

* Dart SDK Version (`dart --version`)
2.12.0-223.0.dev

```dart
var i = 0;
var j = 0;

if (i < j) {
  i++;
} else if (j < i) { // here
  j++;
} else {
  i++;
  j++;
}
```

asashour:
Fixes #2478

Changes:

1. `_sameOperands` treated operands in the same order similar to the inverted one. This is now split into `_sameOperandsSameOrder` and `_sameOperandsInverted`
1.1. If `_sameOperandsSameOrder`, then same behavior as before
1.2. If `_sameOperandsInverted`, we need to `invert` the `operator`, e.g. `a > b` becomes `b < a`
1.3 If none of the above, then `return`. This is the same as the previous logic, with a little performance improvement since we now `return` earlier.
2. `cOperatorType` is now not-nullable (I think we have all cases in `BooleanExpressionUtilities.NEGATIONS` and `INVERSIONS`, but we can revert this change.

### DIFF
--- a/lib/src/util/boolean_expression_utilities.dart
+++ b/lib/src/util/boolean_expression_utilities.dart
@@ -31,6 +31,15 @@ class BooleanExpressionUtilities {
     TokenType.LT_EQ: TokenType.GT,
   });
 
+  static HashMap<TokenType, TokenType> INVERSIONS = HashMap.from(const {
+    TokenType.EQ_EQ: TokenType.EQ_EQ,
+    TokenType.BANG_EQ: TokenType.BANG_EQ,
+    TokenType.GT: TokenType.LT,
+    TokenType.GT_EQ: TokenType.LT_EQ,
+    TokenType.LT: TokenType.GT,
+    TokenType.LT_EQ: TokenType.GT_EQ,
+  });
+
   static HashSet<TokenType> TRICHOTOMY_OPERATORS =
       HashSet.from(const [TokenType.EQ_EQ, TokenType.LT, TokenType.GT]);
 
--- a/lib/src/util/tested_expressions.dart
+++ b/lib/src/util/tested_expressions.dart
@@ -56,14 +56,13 @@ bool _isNegationOrComparison(
   return isNegationOrComparison;
 }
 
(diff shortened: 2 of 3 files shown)


pq:
Apologies for slowness on these reviews!  I've been busy with the 1.8.0 release (which involves a lot of fixing up of our internal code base before landing in the SDK) and it's been a bit consuming.  I hope to have that done in the next day or so though.

As an aside, it's worth you being aware that we (us not you since you don't have access) have to vet any lint changes against our internal codebase to make sure we don't break anyone or block a roll of the Dart SDK internally.  Ideally this is not an issue with false positives but everything needs to be checked regardless.  Thanks for your patience... Some changes may look like they're languishing unmerged when in fact they've just not been checked for ecosystem impact.  (For visibility, I'd like to add some tags and will open an issue to track conversation around that.)

As always, thanks for your contributions!

asashour:
Thanks a lot for your detailed feedback. Giving the others more information would help to better set the expectations.

I understand that not everybody is ready to look into specific area all the time, and there may be other responsibilities which have more priority than PR reviewing. At the end, I trust all will be looked into.

It's worth mentioning, that tagging existing issues for being accepted as a bug, or makibg a decision in a particular case, would help others to participate without asking and taking more time from the team. I hope that long stale issues/PR are handled, so it is clear what to do there, but obviously it takes time and effort to clean them all.

From my side, I would try to dig for more issues to be fixed, try to learn more and hope not to bother the team with much noise :)

bwilkerson:
You're not bothering the team; we very much appreciate the contributions.

pq:
Emphatic +1 to what Brian said.  Thanks so much for your contributions!

As for our process, we'll work more on making it transparent.  Your feedback is really welcome and appreciated!

srawlins:
We've deprecated this lint rule, so will not be landing any changes to it for correctness.

PR #2811, opened 2021-07-27
Fix false positive in `use_rethrow_when_possible`

### LINKED ISSUE #2789: False positive for `use_rethrow_when_possible`
* Dart SDK Version: 2.13

At the moment, `rethrow` is prohibited in ternary expressions. However, the normal `throw` equivalent is allowed. To make matters interesting, `dart analyzer` emits a [`use_rethrow_when_possible`](https://dart-lang.github.io/linter/lints/use_rethrow_when_possible.html)  hint for the normal `throw` equivalent even though the suggested fix isn't valid in this context.

```dart
try {
  // do something
} on SomeError catch (e) {
  e.code == Code.something ? handle() : rethrow; // Fails to compile with "'rethrow' can't be used as an identifier because it's a keyword."
}

try {
// do something
} on SomeError catch (e) {
  e.code == Code.something ? handle() : throw e; // Compiles but dart_analyzer emits a "use_rethrow_when_possible" hint
}
```

In my opinion, this interaction is quite misleading and it would be nice if either `rethrow` is allowed in ternary expressions or `dart_analyzer` does not flag `throw e` in this context.

asashour:
Fixes #2789

I don't know how to have a compilable case with `return i ?? throw e;`

### DIFF
--- a/lib/src/rules/use_rethrow_when_possible.dart
+++ b/lib/src/rules/use_rethrow_when_possible.dart
@@ -62,7 +62,7 @@ class _Visitor extends SimpleAstVisitor<void> {
 
   @override
   void visitThrowExpression(ThrowExpression node) {
-    if (node.parent is ConditionalExpression) return;
+    if (node.parent is Expression || node.parent is ArgumentList) return;
 
     var element =
         DartTypeUtilities.getCanonicalElementFromIdentifier(node.expression);
--- a/lib/src/rules/use_rethrow_when_possible.dart
+++ b/lib/src/rules/use_rethrow_when_possible.dart
@@ -62,6 +62,8 @@ class _Visitor extends SimpleAstVisitor<void> {
 
   @override
   void visitThrowExpression(ThrowExpression node) {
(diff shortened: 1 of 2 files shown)


lib/src/rules/use_rethrow_when_possible.dart
```diff
@@ -62,6 +62,8 @@ class _Visitor extends SimpleAstVisitor<void> {
 
   @override
   void visitThrowExpression(ThrowExpression node) {
+    if (node.parent is ConditionalExpression) return;
```
bwilkerson, line 65: Should this just be `is Expression`? Consider
```
try {
} catch (e) {
  o.m(throw e);
}
```
It's poor code, but I believe that this would also be a false positive.

asashour: Oh! I never thought this could even be possible.

Thanks for hinting.

bwilkerson (approved):

pq (approved):
👍 Thanks!

PR #2812, opened 2021-07-27
Fix false positive in `omit_local_variable_types`

### LINKED ISSUE #2708: omit_local_variable_types - false positive on Null
**Describe the issue**

Enable `omit_local_variable_types`

Write this expression

```dart
const Null response = null;
```

The linter complains that I should omit this type

These workarounds don't work for me:

1. ***The '?' is unnecessary because 'Null' is nullable without it.***

```dart
const Null? response = null;
```

2. ***Missing variable type for 'response'.
Try adding an explicit type, or remove implicit-dynamic from your analysis options file***

```dart
const response = null;
```

I will use `Object?` for now, which works fine, but I would like to limit it to only `Null`

**To Reproduce**
See above

**Expected behavior**
Lint not triggered

**Additional context**
I'm writing tests and all of them should have similar structure - `request` and `response` variables should go at the beginning to be easily discoverable, so I intentionally keep this as a variable

asashour:
Fixes #2708

### DIFF
--- a/lib/src/rules/omit_local_variable_types.dart
+++ b/lib/src/rules/omit_local_variable_types.dart
@@ -101,7 +101,9 @@ class _Visitor extends SimpleAstVisitor<void> {
 
   void _visitVariableDeclarationList(VariableDeclarationList node) {
     var staticType = node.type?.type;
-    if (staticType == null || staticType.isDynamic) {
+    if (staticType == null ||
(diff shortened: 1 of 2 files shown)


bwilkerson (approved):

pq (approved):
👍

PR #2814, opened 2021-07-27
Handle block comments in `rule_test`

asashour:
During testing a new case in an existing file with `rule_test`, one is tempted to use the block comment `/* */` to comment cases which are not relevant, to focus and debug the new case only.

It seems `rule_test` reads the lines and lints regardless of whether they are in block comment or not.

This PR adds a primitive way to consider inside comment or outside comment.

Test scenario:

- Put a block comment around all code of a particular rule test
- Run `dart test -N <specific_rule_name>`

Expected: successful test and 0 lints
Actual: triggered lints are 0, however the expectations is always more than 0.

### DIFF
--- a/test/rule_test.dart
+++ b/test/rule_test.dart
@@ -223,10 +223,19 @@ void testRule(String ruleName, File file,
     var expected = <AnnotationMatcher>[];
 
     var lineNumber = 1;
+    var inComment = false;
     for (var line in file.readAsLinesSync()) {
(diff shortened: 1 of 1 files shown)


srawlins:
WDYT, @pq? Should this just be superseded by the new unit testing?

bwilkerson:
You didn't ask me, but yes.

pq:
I agree.  Thanks!

PR #2820, opened 2021-07-27
prune visits in `avoid_type_to_string`

pq:
More performance tuning.

Also, generally I think collecting data in "parent" visitors is a bit of an anti-pattern that we should probably document?

/cc @bwilkerson


pq:
Hrmmm, maybe we want to think on this.  It's possible that this is actually **_slower_**.

If that's the case, maybe this is a pattern to consider after all. 🤔

pq:
Ok.  Had a nice chat with @scheglov and the existing implementation is better.  For anyone watching, sorry for the noise! 😆

PR #2823, opened 2021-07-28
Fix false negative in `avoid_types_as_parameter_names`

asashour:
Covers the cases for #2770 

I am not sure about the documentation request, since this is a test case

    m3(f(int)) => null; // LINT


lib/src/rules/avoid_types_as_parameter_names.dart
```diff
@@ -60,13 +60,22 @@ class _Visitor extends SimpleAstVisitor<void> {
...
 
-    for (var parameter in node.parameters) {
+  void _visitParameters(
+      FormalParameterList node, NodeList<FormalParameter> parameters) {
+    for (var parameter in parameters) {
       var declaredElement = parameter.declaredElement;
       var identifier = parameter.identifier;
+      if (parameter is SimpleFormalParameter) {
+        var type = parameter.type;
+        if (type is GenericFunctionType) {
```
bwilkerson, line 73: We don't need to do this. If the type of the parameter is a `GenericFunctionType` then we'll visit the formal parameters of that type independently (as part of visiting the whole AST). I believe that you can remove lines 71-76 without breaking anything (and having them means that we'll get more than one lint reported, which is undesirable).

If we don't already have one it would be good to add a test to confirm that this case is covered, perhaps something like:
```
void f(int Function<T>(String T) g) {} // LINT
```

If you delete those lines then I don't think there's any value in splitting out this separate method, so consider reverting those changes as well.

lib/src/rules/avoid_types_as_parameter_names.dart
```diff
@@ -75,11 +84,30 @@ class _Visitor extends SimpleAstVisitor<void> {
   }
 
   bool _isTypeName(AstNode scope, SimpleIdentifier node) {
+    var parent = scope.parent;
+    if (parent is FunctionExpression) {
```
bwilkerson, line 88: I don't believe you need either lines 87-99 or lines 107-112 either. I believe that the easier fix is to extend the test on line 103 to include `element is TypeParameterElement`.

lib/src/rules/avoid_types_as_parameter_names.dart
asashour: You are right, I should have thought a little about this. Reverted.

lib/src/rules/avoid_types_as_parameter_names.dart
asashour: I am afraid `context.resolveNameInScope` always return `none` state with the new scenarios.

bwilkerson: Oh. That sounds like a bug in `resolveNameInScope`. The names of type parameters are definitely in scope in the formal parameter list. I've created an issue to track it (https://github.com/dart-lang/sdk/issues/46753).

Given that, please add a TODO to remove this code when the bug has been fixed (and referencing that issue).

lib/src/rules/avoid_types_as_parameter_names.dart
```diff
@@ -66,20 +67,47 @@ class _Visitor extends SimpleAstVisitor<void> {
...
-          declaredElement.hasImplicitType &&
           identifier != null &&
           _isTypeName(node, identifier)) {
         rule.reportLint(identifier);
       }
     }
   }
 
+  @override
+  void visitGenericFunctionType(GenericFunctionType node) {
```
bwilkerson, line 78: We don't need this either. An instance of `GenericFunctionType` contains an instance of `FormalParameterList`, and that instance will be visited by `visitFormalParameterList`. This also causes the parameters to be visited twice.

asashour: Thanks for your patience.

Clear. I probably misunderstood your other comment.

There was a check at the very beginning of `visitFormalParameterList` to check the parent, now removed in the last commit.

I hope it is now fine.

bwilkerson (approved):
Yes, that looks good, thanks.

lib/src/rules/avoid_types_as_parameter_names.dart
```diff
@@ -75,11 +72,33 @@ class _Visitor extends SimpleAstVisitor<void> {
...
+      return true;
+    }
+    if (parent is GenericFunctionType &&
+        _isTypeParameter(parent.typeParameters, node)) {
+      return true;
+    }
     var result = context.resolveNameInScope(node.name, false, scope);
     if (result.isRequestedName) {
       var element = result.element;
       return element is ClassElement || element is TypeAliasElement;
```
scheglov, line 94: To support type parameters we also need `|| element is TypeParameterElement` here.

scheglov (commented):
https://dart-review.googlesource.com/c/sdk/+/209220

srawlins:
I'd be happy to land this, but I'll need conflicts resolved in order to test inside Google first.

asashour:
Conflicts resolved, thanks.

bwilkerson:
If there are both a type and a name, then it seems reasonable that this lint not fire. As I recall, the motivation for this lint was to prevent cases where someone wrote a type name and forgot to write a parameter name, which clearly isn't the case if the parameter name is there.

srawlins:
In that case, we might be able to just revert the deleted `declaredElement.hasImplicitType &&` line, and continue.

(There were other breakages in google3 I'll clean up first)

lib/src/rules/avoid_types_as_parameter_names.dart
```diff
@@ -68,14 +68,11 @@ class _Visitor extends SimpleAstVisitor<void> {
...
   @override
   void visitFormalParameterList(FormalParameterList node) {
-    if (node.parent is GenericFunctionType) return;
-
     for (var parameter in node.parameters) {
       var declaredElement = parameter.declaredElement;
       var name = parameter.name;
       if (declaredElement != null &&
           declaredElement is! FieldFormalParameterElement &&
-          declaredElement.hasImplicitType &&
```
srawlins, line 78: Would you mind reverting this bit; it shouldn't be needed for the fix, and the motivation of the lint rule is about implicit typed parameters.

srawlins (changes requested):

lib/src/rules/avoid_types_as_parameter_names.dart
asashour: Done, please check the changed test cases in the same commit.

srawlins:
Ah, now we should change those test cases to have implicit parameter types, if possible. I think the function-typed ones _cannot_ have implicit types. So in `void f(int) {}`, I've created a parameter named `int` with type `dynamic`, but in a function type like `void f(void Function<T>(T) callback) {}`, there is no parameter named `T`, there is only an unnamed parameter with type `T`.

srawlins:
I'm still working on cleaning this one up internally.

srawlins:
We are now clean internally.

pq:
@srawlins: did you happen to look at Flutter too?  (No worries either way...)

srawlins:
Not yet. I hope have a moment this week. 🤞

srawlins:
Yes, the one change I requested is done.

srawlins (approved):

pq (approved):
👍

PR #2827, opened 2021-07-28
`annotate_overrides` perf: cache class elements when visiting class decls

pq:
Explores a pattern of caching class elements rather than (re)visiting the AST for each member.

The measurable performance change is smaller than I expected (but then the saved traversals were pretty short -- though frequent).

The open question is whether this is a pattern we want to promote.

Input welcome!

/cc @bwilkerson

### DIFF
--- a/lib/src/rules/annotate_overrides.dart
+++ b/lib/src/rules/annotate_overrides.dart
@@ -125,7 +125,6 @@ class _Visitor extends SimpleAstVisitor<void> {
     }
   }
 
-
   @override
   void visitMethodDeclaration(MethodDeclaration node) {
     if (node.isStatic) return;
--- a/lib/src/rules/annotate_overrides.dart
+++ b/lib/src/rules/annotate_overrides.dart
@@ -60,6 +60,8 @@ class AnnotateOverrides extends LintRule implements NodeLintRule {
 
     // Cache parent element.
     registry.addClassDeclaration(this, visitor);
+    registry.addExtensionDeclaration(this, visitor);
+    registry.addMixinDeclaration(this, visitor);
 
     // Do the work.
     registry.addFieldDeclaration(this, visitor);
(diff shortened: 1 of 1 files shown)


lib/src/rules/annotate_overrides.dart
```diff
@@ -78,7 +85,7 @@ class _Visitor extends SimpleAstVisitor<void> {
   }
 
   Element? getOverriddenMember(Element member) {
-    var classElement = member.thisOrAncestorOfType<ClassElement>();
+    var classElement = this.classElement;
```
bwilkerson, line 88: The one possible issue here is that we can't clear the field when we exit a class. So if we visit a class and then visit an extension we'll use the element from the previous class. We can address this by visiting `ExtensionDeclaration`s and `MixinDeclaration`s. I can't think of any other containers that contain fields or methods.

pq: Ah.  Good catch.

pq: Fixed.

lib/src/rules/annotate_overrides.dart
```diff
@@ -116,10 +125,17 @@ class _Visitor extends SimpleAstVisitor<void> {
...
   void visitMethodDeclaration(MethodDeclaration node) {
     if (node.isStatic) return;
 
     check(node.declaredElement, node.name);
   }
+
+  @override
+  void visitMixinDeclaration(MixinDeclaration node) {
+    // Clear the (possibly) cached element.
+    classElement = null;
```
bwilkerson, line 138: The declared element of a mixin is a `ClassElement`, so you could capture it here. And mixins can override members from the `on` class, so it's probably the right thing to do.

And given that no tests broke as a result of this change, we should add some tests of these cases.

srawlins:
Maybe this should be converted to a draft PR, so it is not in the queue to be reviewed?

pq:
Closing for now. May revisit down the road.

PR #2830, opened 2021-07-29
Fix false negative in `prefer_contains`

asashour:
Related to the `prefer_contains` part of #2662 

Please note the change is in `DartTypeUtilities`, which would potentially affect other rules as well.


lib/src/util/dart_type_utilities.dart
```diff
@@ -208,7 +208,10 @@ class DartTypeUtilities {
       return isAnyInterface(type) ||
           !element.isSynthetic && element.allSupertypes.any(isAnyInterface);
     } else {
-      return false;
+      var typeForInterfaceCheck = type.typeForInterfaceCheck;
+      return typeForInterfaceCheck.isDynamic ||
+          DartTypeUtilities.implementsAnyInterface(
+              typeForInterfaceCheck, definitions);
```
bwilkerson, line 214: The general approach is correct: we should indeed be checking the bound of a type parameter rather than just returning false.

Unfortunately the implementation is too broad. If this method is invoked with a `type` that's an instance of `FunctionType` then `typeForInterfaceCheck` will be the same instance, the check for `isDynamic` will return `false` and we'll recurse with the same parameter values as the last time, which would lead to a stack overflow.

I think what we want instead is code of the form
```
var realType = type;
if (type is TypeParameterType) {
  realType = type.typeForInterfaceCheck;
}
if (type is InterfaceType) {
  ...
} else {
  return false;
}
```

asashour: As always, thanks a lot.
Done.

bwilkerson (approved):

srawlins:
I'd love to test this internally, @asashour , if you want to rebase.

asashour:
Conflict was merged. Please note there is an issue with Dart 3.

lib/src/extensions.dart
```diff
@@ -186,6 +186,9 @@ extension DartTypeExtension on DartType? {
         definitions.any((d) => i.isSameAs(d.name, d.library));
 
     var self = this;
+    if (self is TypeParameterType) {
+      self = self.typeForInterfaceCheck;
```
bwilkerson, line 190: I'd prefer that we only use `self` as a synonym for `this` (and then only for the purposes of improving type promotion). If we're going to be setting the variable to something other than `this`, let's rename it. Consider `typeToCheck`.

asashour: Thanks, done.

srawlins:
All good internally.

PR #2834, opened 2021-07-29
Fix false positive in `test_types_in_equals`

### LINKED ISSUE #443: test_types_in_equals false positive
```dart
bool operator ==(Object other) {
    if (other is! Foo) return false;
   ...
}
```

asashour:
Fixes #443

### DIFF
--- a/lib/src/rules/test_types_in_equals.dart
+++ b/lib/src/rules/test_types_in_equals.dart
@@ -97,6 +97,10 @@ class _Visitor extends SimpleAstVisitor<void> {
       return;
     }
 
+    if (node.expression.staticType == node.staticType) {
+      return;
+    }
+
     var identifier = node.expression as SimpleIdentifier;
     var parameters = declaration?.parameters;
     var parameterName = parameters?.parameterElements.first?.name;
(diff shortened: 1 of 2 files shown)


bwilkerson:
I'm tempted to just close the issue without applying this PR. That issue was opened before null safety was enabled, and I don't think the same implementation pattern applies now. For example, I think the test case you added would be better written as
```
class Good2 {
  bool b = false;

  @override
  bool operator ==(Object other) {
    if (other is! Good2) return false;
    return b == other.b;
  }

  @override
  int get hashCode => b.hashCode;
}
```

Because type promotion is so much better than it used to be I don't think there are many (if any) reasons to use `as` in an `==` method. Hence, I don't think the rule needs to accept this case anymore.

asashour:
With null-safety, there is a `lint` unnecessary cast in 

    Good2 otherGood = other as Good2;

I am fine with closing the issue, with giving a reason.

pq:
> Because type promotion is so much better than it used to be I don't think there are many (if any) reasons to use as in an == method. Hence, I don't think the rule needs to accept this case anymore.

👍   I tend to agree.

asashour:
The lint is already not accepting `as` an `==` method, closing the PR.

PR #2837, opened 2021-07-30
Implement `FAILING` tests

### LINKED ISSUE #2815: Add a way to mark a failed test case
## Use case

Sometimes there is a need to commit a test case, before actually implementing it.

As part of [Test-driven development](https://en.wikipedia.org/wiki/Test-driven_development), this has the benefit of:
- Agreeing first if a scenario is a correct with expected behavior, before starting the implementation.
- One may lack the implementation knowledge, but a test case could be easier to contribute.
- Having a single place of all reported issues (in one file), instead of searching in github issues.
- When an implementation is (accidentally) fixed, a failing test would point to the tracked issue to be closed.

Some other projects use the same concept (`notYetImplemented`), e.g. [Groovy](https://docs.groovy-lang.org/latest/html/gapi/groovy/transform/NotYetImplemented.html) and [Selenium](https://github.com/SeleniumHQ/selenium/blob/trunk/java/client/test/org/openqa/selenium/testing/NotYetImplemented.java).

Needles to say, this doesn't mean implementation is done without testing. But it means tests can be committed, and implementation can later follow.

A similar request in `test` in https://github.com/dart-lang/test/issues/1178

## Proposal

An optional `FAILING` to be added to the lints comment (optionally having an issue number), and when set, the result of the test is reversed.

- If the test fails (as expected) then the build will be green
- If the implementation succeeds, then the build is red complaining `Test XYZ is marked as failing but already succeeds`.

```dart
void false_negative() {
  noLintTriggered(); // LINT FAILING
}
void false_posivite() {
  incorrectLintTriggered(); // OK FAILING #1234
}
```

As a side note, it seems `Annotation.message` is not used by any lint, except in `annotation_test`, and so it could be removed.

asashour:
Fixes #2815 

To test

Select any current rule test, e.g.

```dart
currentlyLinting(); //LINT
currentlyOk(); //OK
```

Change the expectations to the below, and the tests are still green.
```dart
// this is for false positive cases, means it should be ok, but `unfortunately` currently lints
currentlyLinting(); // OK FAILING
// this is for false negative ones, meaning it should lint, but `unfortunately` currently is ok
currentlyOk(); // LINT FAILING
```

If the `FAILING` ones actually pass, there is a message printed about the succession.

```dart
// this fails the build and gives a message that it actually works
currentlyLinting(); // LINT FAILING
// this fails the build and gives a message that it actually works
currentlyOk(); // OK FAILING
```

### DIFF
--- a/lib/src/test_utilities/annotation.dart
+++ b/lib/src/test_utilities/annotation.dart
@@ -5,22 +5,36 @@
 import 'package:analyzer/error/error.dart';
 import 'package:analyzer/source/line_info.dart';
 
+final lintRE = RegExp(
+    r'(//|#) ?LINT( FAILING)?( \[([\-+]\d+)?(,?(\d+):(\d+))?\])?( (.*))?$');
+final okFailingRE = RegExp(r'(//|#).*FAILING');
+
 Annotation? extractAnnotation(int lineNumber, String line) {
-  var regexp =
-      RegExp(r'(//|#) ?LINT( \[([\-+]\d+)?(,?(\d+):(\d+))?\])?( (.*))?$');
-  var match = regexp.firstMatch(line);
-  if (match == null) return null;
+  var match = lintRE.firstMatch(line);
+  if (match == null) {
+    match = okFailingRE.firstMatch(line);
+    if (match == null) return null;
+
+    // ignore commented out lines
+    var index = match.start;
+    var comment = match[1]!;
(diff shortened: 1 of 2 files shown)


srawlins:
CC @bwilkerson I think we will be moving to the unit-test-style tests, instead of the e2e tests, especially for cases where we need to mark a "failing test case".

bwilkerson:
Agreed. That's our plan.

PR #2838, opened 2021-07-30
Fix false negative in `prefer_is_empty`

### LINKED ISSUE #2818: `prefer_is_empty` does not evaluate constants

asashour:
Fixes #2818 

I hope it is fine to do this change, since `not evaluating constants deliberately` in the old TODO, I guess it was added to retain the old behavior in the other PR.

### DIFF
--- a/lib/src/rules/prefer_is_empty.dart
+++ b/lib/src/rules/prefer_is_empty.dart
@@ -81,15 +81,13 @@ class _Visitor extends SimpleAstVisitor<void> {
 
   @override
   void visitBinaryExpression(BinaryExpression node) {
-    // todo(pq): not evaluating constants deliberately but we *should*.
-    // see: https://github.com/dart-lang/linter/issues/2818
-    var value = getIntValue(node.rightOperand, null);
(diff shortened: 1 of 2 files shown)


bwilkerson:
I suspect that there was a reason why the TODO wasn't addressed before committing the lint, but I don't know / remember what it was. I think that Phil is out today, so this will need to wait until he's back.

pq:
Brian is right.  This change has the potential to introduce a bunch of new diagnostics internally and so we need to proceed with caution.  It's generally true of false negatives that they need to be landed with care since they can potentially be very impactful on the ecosystem.

oprypin:
I commented on the issue - I am against such a change :(
https://github.com/dart-lang/linter/issues/2818#issuecomment-1337225760

mosuem:
Closing this as it is stale and needs more background work, see the referenced issue.

Issue #197, opened 2016-03-11
"Specify type annotations." should say the type that the analyzer things it is (if not dynamic)


Hixie:
For example:
```
void main() {
  var x = 2;
}
```
...gives:
```
[lint] Specify type annotations. (test.dart, line 2, col 3)
```
...but if I say:
```
void main() {
  String x = 2;
}
```
...I get:
```
[warning] A value of type 'int' cannot be assigned to a variable of type 'String' (test.dart, line 2, col 14)
```
...so clearly the analyzer has the information to tell me what I really want to know for the first one, which is:
```
[lint] Specify type annotations. "var" could probably be replaced by "int". (test.dart, line 2, col 3)
```
...or some such.

zoechi:
There is a quick-fix to add type annotations but it seems incomplete.

```
class MyClass {
  isMouseOver = false;
}
```
suggests `bool` but it doesn't provide the quick fix for

```  
class MyClass {
  mouseEnterHandler() {
    isMouseOver = true;
  }
}
```

There is also #25945

bwilkerson:
@pq This would be a good use of the "correction" message.

@zoechi @scheglov I suspect that the reason for the current behavior is that the parser recognizes the first example as being a field declaration, despite the fact that there is no type or 'var' specified, while in the second case it gets parsed as an expression statement whose expression is an assignment expression. The quick fix is probably keyed on variable / field declarations with no explicit type. But perhaps it ought to apply when there is an assignment expression and the target of the assignment is not defined.

zoechi:
@bwilkerson Sorry for my unclear post. I expected a quick fix for the return type of `mouseEnterHandler()`. I didn't care about the content of the method and didn't see how confusing my example is.

bwilkerson:
@zoechi Ha! Yes, having a quick fix to add a return type to methods would also be cool. It's a little harder to do; we'd basically have to find all of the return statements and look at the types of their expressions, then pick a reasonable upper bound, but it's definitely possible.

alexeieleusis:
@bwilkerson @pq  where can I read more about the "correction" message?

bwilkerson:
https://github.com/dart-lang/sdk/blob/master/pkg/analyzer/lib/src/generated/error.dart#L3052

alexeieleusis:
I am delinting code at work today and I can confirm the output from this rule is not very useful, good thing I can do something about it :)

bwilkerson:
[...]

PR #2841, opened 2021-07-31
Better lint messages for `always_specify_types`

asashour:
Related to #197

I believe having a `fix` is currently outside the `linter`, this would be a second stage later.


lib/src/rules/always_specify_types.dart
```diff
@@ -99,10 +100,18 @@ class AlwaysSpecifyTypes extends LintRule implements NodeLintRule {
     registry.addTypeName(this, visitor);
     registry.addVariableDeclarationList(this, visitor);
   }
+
+  void _reportLintForTokenWithDescription(Token token, String description) {
+    reporter.reportErrorForToken(LintCode(name, description), token);
```
bwilkerson, line 105: I believe that the better implementation would be to follow the pattern used in `lib/src/rules/always_declare_return_types.dart`. Doing so
- removes the need for special reporting methods and
- allows the code to have a correction message, which is the better place to put this information.

lib/src/rules/always_specify_types.dart
```diff
@@ -114,8 +123,16 @@ class _Visitor extends SimpleAstVisitor<void> {
...
   void visitDeclaredIdentifier(DeclaredIdentifier node) {
-    if (node.type == null) {
-      rule.reportLintForToken(node.keyword);
+    var keyword = node.keyword;
+    if (node.type == null && keyword != null) {
+      var element = node.identifier.staticElement;
+      if (element is VariableElement) {
+        var description = keyword.keyword == Keyword.VAR
+            ? "'$keyword' could be '${element.type}'."
+            : "Specify '${element.type}' type.";
```
bwilkerson, line 132: By convention, the correction messages all follow the pattern "Try <solution>", so I'd propose a message like "Try specifying the type '{0}'.". That doesn't specify whether to replace the keyword so we wouldn't need the extra test.

Similarly below.

lib/src/rules/always_specify_types.dart
```diff
@@ -163,8 +187,51 @@ class _Visitor extends SimpleAstVisitor<void> {
...
+        types = _getTypes(parent.variables);
+      } else if (parent is FieldDeclaration) {
+        types = _getTypes(parent.fields);
+      } else if (parent is VariableDeclarationStatement) {
+        types = _getTypes(parent.variables);
+      }
+
+      if (types == null) return;
+
+      String? multipleTypesString;
```
bwilkerson, line 206: If there are multiple types then it's highly likely that none of them are correct for all the cases. What we really need to do in that case is to (a) capture the `DartType` object for each of the types rather than the names of the types and (b) use the `TypeSystem` to find the least upper bound of those types. We can then suggest the one correct type.

To do even better, if there's more than one type we might want a different correction message (which implies a different instance of `LintCode`) that read more like "Try specifying the type '{0}', or split the declaration into multiple declarations in order to use more specific types.".

asashour: Thanks for feedback, I didn't think about this.

However, I don't think we can simply use the upper bound of those types, because we must first ensure no sub-class methods are being invoked, and I am not sure if this is possible.

Suppose we have the following case

```dart
class A {
  void a() {}
}
class B extends A {
  void b() {}
}
void main() {
  var a = A(), b = B();
  a.a();
  b.b();
}
```

We can't simply say it can be changed to

```dart
A a = A(), b = B();
```

Because `b.b()` is not part of `A`.

Is there a way to know if all method invocations of `a` and `b` are relevant to only `class A` and not `B`?

bwilkerson: You're right, the least upper bound isn't sufficient in this case. The only guaranteed solution is to split the declarations. So, let's not provide any type in the message and shorten the correction message to only include the one valid option (something like "Try splitting the declaration into multiple declarations.").

lib/src/rules/always_specify_types.dart
```diff
@@ -163,8 +204,57 @@ class _Visitor extends SimpleAstVisitor<void> {
...
+          arguments = [];
+        }
+      }
+      rule.reportLintForToken(keyword,
+          arguments: arguments, errorCode: errorCode);
+    }
+  }
+
+  List<String> _getTypes(VariableDeclarationList list) {
+    var types = <String>[];
```
bwilkerson, line 251: The fact that `types` is a list makes me suspect that we won't catch cases where all of the types are the same, such as
```
var a = 0, b = 1;
```
This should produce a diagnostic that suggests the type `int`.

I suspect that you need to use a set here.

lib/src/rules/always_specify_types.dart
```diff
@@ -163,8 +204,57 @@ class _Visitor extends SimpleAstVisitor<void> {
...
+    for (var variable in list.variables) {
+      var type = variable.initializer?.staticType;
+      if (type != null) {
+        types.add(type.getDisplayString(withNullability: false));
```
bwilkerson, line 255: The fact that we're not including nullability information here means that we can't tell the difference between nullable and non-nullable types. Consider the following
```
String? s1 = '';
var s2 = '', s3 = s1;
```
I suspect that in this case we'll suggest using the type `String`, but that will lead to an error because `String?` (`s1`) isn't assignable to `String`.

lib/src/rules/always_specify_types.dart
```diff
@@ -102,6 +104,18 @@ class AlwaysSpecifyTypes extends LintRule implements NodeLintRule {
 }
 
 class _Visitor extends SimpleAstVisitor<void> {
+  static const LintCode keywordCouldBeTypeCode = LintCode(
+      "always_specify_types", "'{0}' could be '{1}'.",
```
bwilkerson, line 108: The problem message should indicate what's wrong. The problem isn't that the keyword could be a type, but that there is no type annotation when one is required. It would be better if the problem message were something like "Missing a type annotation.".

On the other hand, the correction message should indicate what to do to fix the problem, and that's a great place to include one possible type. Perhaps something like "Try replacing '{0}' with '{1}'."

Similarly with the other error codes.

lib/src/rules/always_specify_types.dart
```diff
@@ -102,6 +104,18 @@ class AlwaysSpecifyTypes extends LintRule implements NodeLintRule {
...
+      correction: "Try changing the type.");
+
+  static const LintCode keywordCouldBeSplitToTypesCode = LintCode(
+      "always_specify_types", "'{0}' could be split into types.",
+      correction: "Try splitting '{0}' to different types.");
```
bwilkerson, line 113: In addition to the comments above, this one is a bit confusing because I don't know what it means to split a keyword into types. I think what you're trying to say is that the declaration could be split into multiple declarations in order to be able to provide more specific type information. That's a valid way to correct the problem and definitely one that we should suggest, but let's talk about splitting the declaration rather than splitting a keyword. (And, of course, the problem isn't that the declarations could be split up.)

bwilkerson (commented):
For what it's worth, diagnostic message are very hard to write well and typically take a lot of iteration.

lib/src/rules/always_specify_types.dart
asashour: The List was converted to `Set` in the caller function, now `Set` is directly used.

lib/src/rules/always_specify_types.dart
asashour: I see, fixed

lib/src/rules/always_specify_types.dart
asashour: Forgive my inaccuracy. Hopefully fixed.

lib/src/rules/always_specify_types.dart
```diff
@@ -149,9 +179,25 @@ class _Visitor extends SimpleAstVisitor<void> {
...
+        var keyword = param.keyword!;
+        var type = param.declaredElement?.type;
+
+        if (keyword.type == Keyword.VAR &&
+            type != null &&
+            type is! DynamicType) {
+          rule.reportLintForToken(keyword,
+              arguments: [keyword, type], errorCode: keywordCouldBeTypeCode);
+        } else {
+          rule.reportLintForToken(keyword);
```
bwilkerson, line 191: Do we want to use a specific errorCode here?

lib/src/rules/always_specify_types.dart
```diff
@@ -149,9 +179,25 @@ class _Visitor extends SimpleAstVisitor<void> {
...
+        }
+      } else if (param.declaredElement != null) {
+        var type = param.declaredElement!.type;
+
+        if (type is DynamicType) {
+          rule.reportLint(param);
```
bwilkerson, line 197: Do we want to use a specific errorCode here?

lib/src/rules/always_specify_types.dart
```diff
@@ -163,8 +209,69 @@ class _Visitor extends SimpleAstVisitor<void> {
...
+        types = _getTypes(parent.variables);
+      } else if (parent is ForPartsWithDeclarations) {
+        types = _getTypes(parent.variables);
+      } else if (parent is FieldDeclaration) {
+        types = _getTypes(parent.fields);
+      } else if (parent is VariableDeclarationStatement) {
+        types = _getTypes(parent.variables);
+      }
+
+      if (types == null) return;
```
bwilkerson, line 226: It isn't a requirement that you do so, but consider making `types` be non-nullable and adding
```
} else {
  return;
```
before line 224. I think it makes the intent of the code a little more clear.

lib/src/rules/always_specify_types.dart
```diff
@@ -102,6 +104,23 @@ class AlwaysSpecifyTypes extends LintRule {
...
+      "always_specify_types", // ignore: prefer_single_quotes
+      "Specify type annotations.", // ignore: prefer_single_quotes
+      correction:
+          "Try splitting the declaration and specify the different type annotations."); // ignore: prefer_single_quotes
+
+  static const LintCode specifyTypeCode = LintCode(
+      "always_specify_types", // ignore: prefer_single_quotes
+      "Specify the type annotation.", // ignore: prefer_single_quotes
+      correction:
+          "Try specifying the type '{0}'."); // ignore: prefer_single_quotes
```
bwilkerson, line 122: If you're making other changes, then I'll note that this ignore isn't needed. If you're not making other changes then you can leave it.

Similarly, if you want to, it's perfectly fine to use single quotes for the name of the code (on lines 108, 113, and 119). Your choice.

lib/src/rules/always_specify_types.dart
asashour: This is related to the below cases, which specific `errorCode` should we use? Since we don't know which type is needed. I guess the default `description` is fine.

```dart
a(var x) {}
d(final x) {}
Foo(var bar);
```

lib/src/rules/always_specify_types.dart
asashour: Related to the below case, I am not sure which specific `errorCode` to be used.

```dart
b(s) {}
```

lib/src/rules/always_specify_types.dart
asashour: I understood from https://github.com/dart-lang/linter/pull/2863#discussion_r684798583, that all `LintCode` usage should have double quotes.

bwilkerson: Then I wasn't very clear. The convention we follow is that messages (the problem message and the correction message in this case) are always in double quotes so that they're consistent and allow for having single quotes in the messages themselves. The code name is usually in single quotes because it isn't a message and hence gets treated like any other string.

lib/src/rules/always_specify_types.dart
bwilkerson: The default description is the same as the message for `keywordCouldBeSplitToTypesCode`, but the message for `keywordCouldBeTypeCode` seems more appropriate in this case because there's only one identifier being declared here (though you can't use that code because we don't have a specific type to include in the correction message). Whether you want to define a new type code for this new case is up to you.

But mostly I was asking because it seemed inconsistent to be explicit about some of the codes and to use the default for others. It isn't wrong to do so, and it's OK if you intended to do so, I was just checking that it wasn't on oversight. I should have said as much earlier.

srawlins:
@asashour it looks like some of the open conversations need your attention, (and some need @bwilkerson's). Would you mind resolving conflicts and/or updating conversations, to make progress? Thanks!

lib/src/rules/always_specify_types.dart
asashour: > The default description is the same as the message for keywordCouldBeSplitToTypesCode

Not really, the default `correctionMessage` is `Try adding a type annotation.`, but for `keywordCouldBeSplitToTypesCode` it is `splitting the declaration ...`. Both don't need arguments, but the second one is meant for the `var` in the delcaration of different types.

In other words, for a case like `a(var x) {}`, we shouldn't use `keywordCouldBeSplitToTypesCode` as there are not declarations to split.

lib/src/rules/always_specify_types.dart
bwilkerson: nit: The ignore on line 110 shouldn't be needed because the string contains single quotes.

lib/src/rules/always_specify_types.dart
bwilkerson: Ok, we can leave it for now.

lib/src/rules/always_specify_types.dart
bwilkerson: Ok, we can leave it for now.

bwilkerson (approved):

lib/src/rules/always_specify_types.dart
```diff
@@ -103,8 +120,19 @@ class _Visitor extends SimpleAstVisitor<void> {
...
   void visitDeclaredIdentifier(DeclaredIdentifier node) {
-    if (node.type == null) {
-      rule.reportLintForToken(node.keyword);
+    var keyword = node.keyword;
+    if (node.type == null && keyword != null) {
+      var element = node.declaredElement;
+      if (element is VariableElement) {
+        if (keyword.keyword == Keyword.VAR) {
+          rule.reportLintForToken(keyword,
+              arguments: [keyword.lexeme, element!.type],
```
pq, line 129: I'm surprised by the `!`.  Is that needed?  (I'd expect the `is` check to ensure a non-null value?)

lib/src/rules/always_specify_types.dart
```diff
@@ -138,17 +166,95 @@ class _Visitor extends SimpleAstVisitor<void> {
     var name = param.name;
     if (name != null && param.type == null && !name.lexeme.isJustUnderscores) {
       if (param.keyword != null) {
-        rule.reportLintForToken(param.keyword);
-      } else {
-        rule.reportLint(param);
+        var keyword = param.keyword!;
```
pq, line 169: To avoid this `!`, consider moving the keyword local to before the check?

```dart
var keyword = param.keyword;
if (keyword != null) {
  var type = param.declaredElement?.type;

  ...

}
```




```

lib/src/rules/always_specify_types.dart
asashour: https://github.com/dart-lang/sdk/issues/50913

lib/src/rules/always_specify_types.dart
asashour: Done, thanks.

lib/src/rules/always_specify_types.dart
pq: Thanks!

pq (approved):
👍

PR #2843, opened 2021-07-31
Fix false positive in `prefer_void_to_null`

### LINKED ISSUE #2680: prefer_void_to_null should not fire for the `on` in an extension
Currently the lint fires for this `Null`:

```dart
extension _ on Null {
```

I think we should ignore this case. `extension _ on void` has a drastically different meaning, and is not something anyone should write, preferring instead `extension _ on Object?`. There are potentially legitimate reasons to write `extension _ on Null`.

asashour:
Fixes #2680


scheglov (approved):

pq (approved):
Thanks!

Issue #909, opened 2018-03-01
avoid_catches_without_on_clauses should not warn if there is a rethrow


zoechi:
```dart
    try {
      incrementPendingRequestCounter();
      result = await serverRequest();
      decrementPendingRequestsCounter();
      return result;
      // ignore: avoid_catches_without_on_clauses
    } catch (e) {
      decrementPendingRequestsCounter(error: e.message);
      rethrow;
    }
```

srawlins:
I'm all for this, and I think in particular it should not warn if there is a `rethrow` as one of the statements directly in the `catch` body. Meaning if a `rethrow` is found in an if-body, or whatever else, then `avoid_catches_without_on_clauses` is still reported. I don't want people is-checking the error inside an unguarded catch.

PR #2847, opened 2021-08-01
Add TODO to remove an `ignore` once an issue is fixed

asashour:
Since #909 is accepted as a bug, then the `ignore` should be removed once the issue is fixed.


bwilkerson (approved):

pq (approved):
👍

PR #2853, opened 2021-08-04
mark `avoid_dynamic_calls` stable

### LINKED ISSUE #2666: Graduate avoid_dynamic_calls from experimental
I have enabled avoid_dynamic_calls in a few repos, and really enjoy it's benefits. I'm not sure why it is marked experimental, but I think we should graduate it to not-experimental.

This is probably the first step to getting it into the lints package.

pq:
Fixes #2666

/cc @srawlins @bwilkerson 

/fyi @devoncarew @matanlurey


bwilkerson (approved):
As far as I know this should be fine. I don't remember why it was marked as experimental, though, so there might be a reason I don't know for not doing this.

pq:
It just needed some baking in.   I think we're well past that now.

Thanks!

PR #2856, opened 2021-08-06
fix `prefer_void_to_null` false positive on overriding returns

pq:
Fixes #https://github.com/dart-lang/linter/issues/2792

/cc @bwilkerson

### DIFF
--- a/lib/src/rules/prefer_void_to_null.dart
+++ b/lib/src/rules/prefer_void_to_null.dart
@@ -4,6 +4,8 @@
 
 import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/element.dart';
+import 'package:analyzer/dart/element/type.dart';
 
 import '../analyzer.dart';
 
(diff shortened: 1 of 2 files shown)


lib/src/rules/prefer_void_to_null.dart
```diff
@@ -56,15 +58,57 @@ class PreferVoidToNull extends LintRule implements NodeLintRule {
...
+    var libraryUri = classElement.library.source.uri;
+    return context.inheritanceManager.getInherited(
+      classElement.thisType,
+      Name(libraryUri, name),
+    );
+  }
+
+  bool isFutureOrVoid(DartType type) {
+    if (!type.isDartAsyncFutureOr) return false;
+    if (type is! InterfaceType) return false;
```
bwilkerson, line 94: I don't think that anything other than an `InterfaceType` could return `true` for `isDartAsyncFutureOr`, so I don't think we need this test.

bwilkerson (approved):

lib/src/rules/prefer_void_to_null.dart
pq: I think you're right but in the next line I need an `InterfaceType` and I kind of cringe at casting and (just barely) prefer an `is` check.

PR #2863, opened 2021-08-07
Enable `prefer_single_quotes`

asashour:
~~No need to ignore `prefer_single_quotes`. It is not part of the checked lints in the project.~~

Enable `prefer_single_quotes`, as per the below feedback.

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -63,6 +63,7 @@ linter:
 #    - prefer_mixin
     - prefer_null_aware_method_calls
     - prefer_relative_imports
+    - prefer_single_quotes
 #    - public_member_api_docs
 #    - sort_constructors_first
     - sort_pub_dependencies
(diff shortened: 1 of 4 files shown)


pq:
Ah!  Good catch.  I think I'd like that enabled actually.  (I'm guessing it was accidentally dropped when we migrated off pedantic.)  Maybe switch this to re-enable it?

lib/src/rules/always_declare_return_types.dart
```diff
@@ -64,17 +64,13 @@ class AlwaysDeclareReturnTypes extends LintRule {
 }
 
 class _Visitor extends SimpleAstVisitor<void> {
-  static const LintCode functionCode = LintCode(
-      "always_declare_return_types", // ignore: prefer_single_quotes
+  static const LintCode functionCode = LintCode('always_declare_return_types',
       "The function {0} should have a return type but doesn't.",
-      correction:
-          "Try adding a return type to the function."); // ignore: prefer_single_quotes
+      correction: 'Try adding a return type to the function.');
```
bwilkerson, line 69: This was mimicking the convention in the `analyzer` package that error codes always use double quotes to that we don't have to switch back and forth as we add or remove single quotes within the messages. I think we want to continue using that convention because it makes code reviews easier. (Changing the quotes on the lint name is fine though, because we don't use single quotes in lint names so it doesn't have the same issue.)

asashour: I see, thanks for your feedback.

bwilkerson (approved):

pq (approved):
Thanks!

PR #2867, opened 2021-08-10
make analyzer mocks less fragile

pq:
Follow-up from https://github.com/dart-lang/linter/pull/2866.

A few more tweaks to make our mocks less fragile.

/cc @bwilkerson


bwilkerson (approved):
IMO, it would be better to remove these classes, but this is still an improvement.

pq:
> IMO, it would be better to remove these classes

Agreed.  In fact, I'd like to do a full audit of analyzer deps and do some trimming.  (Maybe next fix-it Friday?)

PR #2872, opened 2021-08-13
fix false positive in `avoid_null_checks_in_equality_operators` w/ non-nullable param

### LINKED ISSUE #2864: remove `avoid_null_checks_in_equality_operators` lints redundant w/ NNBD diagnostics
For example:

```dart
class Person {
  final String name = '';

  @override
  operator ==(other) {
    return other != null &&
          other is Person &&
          name == other.name;
  }
}
```

produces a lint in addition to a `UNNECESSARY_NULL_COMPARISON_FALSE` (which alone should be sufficient).

See: https://dart-review.googlesource.com/c/sdk/+/209661 for more context.

pq:
Fixes #2864

/cc @bwilkerson

### DIFF
--- a/lib/src/rules/avoid_null_checks_in_equality_operators.dart
+++ b/lib/src/rules/avoid_null_checks_in_equality_operators.dart
@@ -2,10 +2,12 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'package:analyzer/dart/analysis/features.dart';
 import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/token.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
 import 'package:analyzer/dart/element/element.dart';
+import 'package:analyzer/dart/element/nullability_suffix.dart';
 
 import '../analyzer.dart';
 import '../util/dart_type_utilities.dart';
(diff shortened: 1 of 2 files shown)


lib/src/rules/avoid_null_checks_in_equality_operators.dart
```diff
@@ -82,34 +84,48 @@ class AvoidNullChecksInEqualityOperators extends LintRule {
...
-      DartTypeUtilities.traverseNodesInDFS(node.body)
-          .where(checkIfParameterIsNull)
-          .forEach(rule.reportLint);
+
+    if (node.name.token.type != TokenType.EQ_EQ || parameters.length != 1) {
+      return;
+    }
+
+    var parameter = DartTypeUtilities.getCanonicalElementFromIdentifier(
+        parameters.first.identifier);
```
bwilkerson, line 111: I don't understand why we're not just using `parameter.first.declaredElement`. Is there some case that I'm missing?

bwilkerson (approved):

lib/src/rules/avoid_null_checks_in_equality_operators.dart
pq: I'm not at all sure.  I notice a comment in `getCanonicalElement` with a bug reference that might be a hint?

```
        // A field element defined in a parameterized type where the values of
        // the type parameters are known.
        //
        // This concept should be invisible when comparing FieldElements, but a
        // bug in the analyzer causes FieldElements to not evaluate as
        // equivalent to equivalent FieldMembers. See
        // https://github.com/dart-lang/sdk/issues/35343.
```

It'd be worth tracking this down (but note that it's called in a bunch of places so we'd want to proceed with some extra caution).

bwilkerson: I added a question to the issue to double check, but I believe that `Element.declaration` should replace the need to check for `Member`s.

PR #2879, opened 2021-08-18
Fix false positive in `use_build_context_synchronously`

### LINKED ISSUE #2878: False positive in `use_build_context_synchronously`
**Describe the issue**
There are false positives when writing awaits inside anonymous functions defined above statements that use the `BuildContext` while using the `use_build_context_synchronously` lint.

**To Reproduce**
The linter warns about the `await` inside the `onTap` function.
```dart
void build(BuildContext context) {
  final childWidget = _Button(
    onTap: () async {
      await Future<void>.delayed(Duration());
    },
  );
  return Container(
    color: Theme.of(context).accentColor,
    child: childWidget,
  );
}
```

**Expected behavior**
The `await` inside the onTap shouldn't be taken into account with this lint rule, because the body inside the function is not executed.

davidmartos96:
# Description

There are false positives when writing awaits inside anonymous functions defined above.

Fixes #2878

PS: First time contributing to the linter so I may be taking wrong assumptions here, but this works for the false positive cases I've encountered.

### DIFF
--- a/lib/src/rules/use_build_context_synchronously.dart
+++ b/lib/src/rules/use_build_context_synchronously.dart
@@ -85,6 +85,12 @@ class _AwaitVisitor extends RecursiveAstVisitor {
   void visitAwaitExpression(AwaitExpression node) {
     hasAwait = true;
   }
+
+  @override
+  visitBlockFunctionBody(BlockFunctionBody node) {
+    // Stop visiting if it's a function body block.
+    // Awaits inside it shouldn't matter
+  }
 }
 
 class _Visitor extends SimpleAstVisitor {
--- a/lib/src/rules/use_build_context_synchronously.dart
+++ b/lib/src/rules/use_build_context_synchronously.dart
@@ -91,6 +91,11 @@ class _AwaitVisitor extends RecursiveAstVisitor {
     // Stop visiting if it's a function body block.
     // Awaits inside it shouldn't matter
   }
+
+  @override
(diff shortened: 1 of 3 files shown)


lib/src/rules/use_build_context_synchronously.dart
```diff
@@ -85,6 +85,12 @@ class _AwaitVisitor extends RecursiveAstVisitor {
   void visitAwaitExpression(AwaitExpression node) {
     hasAwait = true;
   }
+
+  @override
+  visitBlockFunctionBody(BlockFunctionBody node) {
```
bwilkerson, line 90: You probably also want to visit `ExpressionFunctionBody` to stop at the `=>` form.

bwilkerson (commented):
Welcome to the linter package!

lib/src/rules/use_build_context_synchronously.dart
davidmartos96: You are right, thanks!

bwilkerson (approved):

pq:
This is fantastic.  Thank you!

pq:
@davidmartos96: assuming you're not a Googler, how can we attribute your contribution in `AUTHORS`?

pq (approved):
💎

davidmartos96:
@pq Maybe something similar to this 😅:

Individuals:

David Martos <davidmartos96@gmail.com>

Issue #855, opened 2017-12-14
catch places where "map" is used instead of "forEach"


MichaelRFairhurst:
Much like unnecessary_statements.

I found myself wasting a bunch of time recently where I did things like:

```
list.map((k) => print(k));
// and
map = {};
list.map((k) => map[k.name] = k);
// and
list.map(recordItem);
```

and didn't think about how `map` is lazy and would do nothing in these cases. I thought I simply had empty lists in my tests! Linter could definitely catch this based on MethodElement and the fact that its return value is not used.

Seems like a standard enough function that either unnecessary_statements could catch it, or it could be its own rule. Note that unnecessary_statements doesn't catch it now because it assumes all method calls have effects.

jamesderlin:
People not reading and understanding that `Iterable.map` returns a lazy iterator and immediately discarding its return value has come up a number of times on StackOverflow:

* [List.map not executing despite List having elements inside(flutter)](https://stackoverflow.com/q/64584646/)
* [For some reason Flutter are ignoring my list.map](https://stackoverflow.com/q/65317038/)
* [My .map method doesnt work properly in function](https://stackoverflow.com/q/63901804/)
* [Flutter function return null value](https://stackoverflow.com/q/68744888/)

pq:
Thanks for the examples @jamesderlin!  Clearly this is confusing a bunch of folks.

I wonder if a notification that the result of `map` was unused (e.g., feedback from the newly added [`@useResult` annotation](https://pub.dev/documentation/meta/latest/meta/useResult-constant.html)) would be a sufficient nudge? 

(Technical aside: SDK libraries can't use `package:meta` so it's not as easy as that but I do wonder if the idea has any legs.)

/fyi @lrhn @eernstg @natebosch @jakemac53 @munificent (handful of language folks who might have thoughts)

natebosch:
I would be in favor of making a private SDK copy of the annotation which is also understood by the lint for that use case.

jakemac53:
Ya I think this would be a good idea, and if we need to hack it with something like Nate said that is worth while. There are probably other things in the SDK which could use this also. For instance `Iterable.take`, or really any method that returns an iterable.

jakemac53:
On second thought maybe it is worth actually adding a special lint around unused iterables. I can't think of a situation where that could ever be doing what the user intended? Does anybody have a counter example?

lrhn:
[...]

PR #2880, opened 2021-08-18
+ `use_iterables`

pq:
An experimental lint to detect unused (and so uninstantiated) lazy iterables.

**Open Question:**  should we extend this to:

* **_any_** method invocation/property access that leaves an iterable unused? or 
* just the ones defined on `Iterable` (`take`, `takeWhile`, etc?)

See: #855

/cc @bwilkerson


lib/src/rules/use_iterables.dart
```diff
@@ -0,0 +1,97 @@
...
+**BAD:**
+```dart
+[1, 2, 3].map(print);
+```
+
+**GOOD:**
+```dart
+[1, 2, 3].forEach(print);
+[1, 2, 3].map(print).last;
+[1, 2, 3].map(print).toList();
```
bwilkerson, line 28: I'm not sure I can agree that the last two are examples of "good" code. :-)

But consider also proposing a simple `for` loop, which is equivalent to `forEach` when a closure is passed in rather than a tear-off.

lib/src/rules/use_iterables.dart
```diff
@@ -0,0 +1,97 @@
...
+      return false;
+    }
+
+    if (parent is ParenthesizedExpression ||
+        parent is ConditionalExpression ||
+        parent is CascadeExpression) {
+      return _isUsed(parent);
+    }
+
+    return parent is ArgumentList ||
```
bwilkerson, line 86: I'm concerned about both false positives and false negatives. For example, I think the `BinaryExpression` needs to be in this list to prevent a false positive. There are probably others, like `PrefixExpression` and `IndexExpression`. And maybe `ExtensionOverride` needs to be skipped over like `ParenthesizedExpression` above. I'm not sure the list is complete even now.

What we really want is a getter on `Expression` that returns `true` if the value of the expression is used in some way.

bwilkerson (approved):

lib/src/rules/use_iterables.dart
pq: Yeah, I really dislike the last two but they are common work-arounds.  I'll defer to @natebosch, @jakemac53, @lrhn et al for whether we should exclude the work-arounds, common as they are.  (Alternatively, maybe some descriptive text?)

natebosch: I agree that we should not include those work-arounds in the "good" section.

I would use

```
[1, 2, 3].forEach(print);
for (var element in [1, 2, 3]) {
  print(element);
}
```

lib/src/rules/use_iterables.dart
pq: We definitely want to guard against false positives and negatives so I appreciate the attention to detail!

I added a few kind of odd test cases.  If you can think of some more, please share them and I'll make sure they're covered.

Thanks!

lib/src/rules/use_iterables.dart
pq: Done.  Thanks!

lrhn: Agree with Nate. 
Using `.toList()` is wasteful, but otherwise harmless.
Using `.last` is plain wrong. If the receiver of the `map` call was a list, then `.last` will only call the map function on the last element of the list: `[1, 2, 3].map(print).last` only prints `3`.

lib/src/rules/use_iterables.dart
```diff
@@ -0,0 +1,98 @@
...
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/type.dart';
+import 'package:analyzer/dart/element/type_provider.dart';
+
+import '../analyzer.dart';
+
+const _desc = r'Use lazy iterables.';
+
+const _details = r'''
+**DO** use lazy iterables created by `iterable.map`. Since they are lazily
```
lrhn, line 15: The word "use" is ambiguous/misleading.

If a style guide/lint tells me to "use" something, I'd think it tells me to *do* put lazy iterables into my code, and that they are not being evaluated is a good thing. Not that I should ensure that the value of said lazy iterable is not ignored.

Q.v. other lints saying things like "Use valid regular expression syntax.".

Maybe "don't ignore the value of" instead of "use"?

lib/src/rules/use_iterables.dart
lrhn: Do we have the concept of "being used" around `void` already?
The contexts where we'd allow `void`, which means it isn't being used, are:
* An initializer expression part, or any expression in the increment part, of a `for(;;)` statement.
* The expression of an expression statement.
* Either branch of a conditional expression if the conditional isn't being used.
* A body of parenthesized expression if the parenthesized expression isn't being used.
* The second part of a `??` operator where the `??` expression isn't being used.
* The operand of an `as` expression where the `as` expression isn't isn't being used.
* The operand of an `is` expression.
* The `=> e` expression of a `void`-typed function.
* Assignment to a `void` typed variable (questionable!)
* The one I forgot (there's always one!)

Currently an extension override is safe because you must call a member on it.

(Also the second operand of a `&&` or `||` operator which isn't being used, but types usually prevent that from being relevant).

Also consider `maybeIterable?.map(...)` and `iterable..map(...)` (followed by no selectors).

lib/src/rules/use_iterables.dart
pq: 💯 

Spot on.  Thanks!

eernstg:
This is cool! But shouldn't it target an arbitrary type declaration which is somehow marked as relevant, as in

```dart
@NotSilentlyDiscardable('Explain why a C should not be discarded, with link to doc')
class C {}
```

lib/src/rules/use_iterables.dart
bwilkerson: I agree with @lrhn, it would be much better to enumerate the places where the value isn't being used than to try to enumerate the places where it is. Even if it doesn't make the implementation easier it will help prevent false positives when new language constructs are added.

> Do we have the concept of "being used" around `void` already?

I don't believe that we have anything around this that the linter can currently re-use.

pq:
@eernstg: sorry for the slow reply!  I'd love to see this as generalized as possible.  We do already have something somewhat similar in [`@useResult`](https://pub.dev/documentation/meta/latest/meta/useResult-constant.html) but taking advantage of that (or doing something similar) has some challenges:

1. `package:meta` is not available to SDK core classes and though we could add another something for the SDK my sense is there's some resistance to hosting annotations in the SDK proper (@lrhn can maybe elaborate here)
2. annotations are typically enforced by server w/o lint configuration (that means they have a higher bar for no false positives)
3. we're not 100% sure what semantics we even want to enforce to get a good balance of value (over false positives)

The result is that for the short term iterating on a lint (ha) seems expedient and non-disruptive. That said, when we settle on semantics maybe ultimately this should be generalized and more fully baked into the APIs.  Thanks for what you've already added to #855 by the way.  Any other ideas are more than welcome!

eernstg:
Thx!

srawlins:
@pq did you want to keep working on this? I think there is a review request item to change to tracking _unuse_ rather than _use_. Maybe mark it as draft for now?

pq:
Good call.  Marked as a draft 👍

Issue #2914, opened 2021-08-31
consider flagging "raw" generic method invocations in `always_specify_types`


pq:
Specifically, we do not currently lint the following "raw" invocation but maybe we should?

```dart
List<T> makeList<T>() => <T>[];

void f() {
  makeList<int>(); // OK
  makeList(); // <== LINT?
}
```

By extension, this would likely carry over to similar advice for constructor tear-offs.

```dart
List<E> Function<E>(int, E) filledList = List.filled; // OK
filledList<int>(3, 3); // OK
filledList(3, 3); // LINT?
```

@davidmorgan: I believe this rule is banned internally but I wonder if you have an opinions?

@goderbauer: similarly, you all don't use this rule but wonder if you have thoughts?

/cc @bwilkerson

srawlins:
Just a semantic note, this is not a "raw" type in the sense that the `strict-raw-types` analysis mode considers; there is opportunity for inference.

`List<int> f() => makeList();` uses inference to fill in `T`.

That being said, `always_specify_types` does seem to have a "never use inference" subtitle to it, so maybe its valid to catch such code here.

It'd be hard to land, if Flutter uses it.

pq:
Thanks for clarifying! If not "raw", is there a better term for invocations like `makeList()`?

srawlins:
I think its just a value with an inferred type.

If inference falls back to `dynamic`, I'm not sure that it is proper to say "inference filled in `dynamic`" or "inference failed to fill anything in, which results in a fallback of `dynamic`." I think in the `strict-inference` mode, we consider it a "failure of inference" to fall back to `dynamic` (or the bound).

davidmorgan:
No opinion on `always_specify_types`, it's in contravention of the style guide so users are on their own ;)

eernstg:
Wouldn't it be up to some Flutter folks to make a decision about the main purpose of this lint? I would assume that [this](https://github.com/flutter/flutter/blob/36531fbc9f2503b3d464c942dff5ccbc4f78be29/analysis_options.yaml#L55) is by far the weightiest reference to `always_specify_types`. If that is true then this lint should basically mean whatever they want it to mean.

A different perspective on this could be "what makes sense, abstractly?", and for that I'd think that the following situation could be a relevant target: _An invocation of a generic method is subject to type inference, and one or more inferred type arguments had the empty set of constraints._

[...]

Issue #2693, opened 2021-06-03
Implement constructor tearoffs support in the linter


eernstg:
This issue tracks support of the [constructor tearoffs feature](https://github.com/dart-lang/language/blob/master/accepted/future-releases/constructor-tearoffs/feature-specification.md).  See the [enclosing meta-issue](https://github.com/dart-lang/sdk/issues/46228) for details.

---

## Rules that need investigation, tests, and possible updating

- [x] [`always_specify_types`](https://dart-lang.github.io/linter/lints/always_specify_types.html) (e.g. catch `List<List>;`, `List<List>.filled;`, ~`someFuture.then<List>`~) (https://github.com/dart-lang/linter/pull/2908)
- [x] `avoid_redundant_argument_values` (#2895)
- [x] ~`prefer_const_constructors`~
- [x] `unnecessary_const` (#2903)
- [x] #2911
- [x] #2909
- [x] #2912
- [x] #2910
- [x] `unnecessary_new` (#2902)
- [x] #2907
- [x] [`unnecessary_statements`](https://dart-lang.github.io/linter/lints/unnecessary_statements.html) (https://github.com/dart-lang/linter/pull/2901)

## Server updates

- [x] https://github.com/dart-lang/sdk/issues/47102
- [x] https://github.com/dart-lang/sdk/issues/47103

## New lints

- [x] #2904 (and https://github.com/dart-lang/site-www/issues/3525)

## Related issues / Future work

*  #2928 
*  #2914

devoncarew:
@pq - I see all the existing lints / new lints items checked off above. Is this issue complete (at least from the POV of any work we'd want to do before we ship)?

pq:
Yep.  I think we're good to go! 👍

PR #2915, opened 2021-08-31
test updates for `always_specify_types` w/ constructor tear-offs

pq:
Follow-up from #2693.

See also: #2914.

(FWIW: waiting to address TODOs until a new push of analyzer since ASTs are in flux.)

/cc @bwilkerson @srawlins

### DIFF
--- a/test_data/rules/experiments/constructor_tearoffs/rules/always_specify_types.dart
+++ b/test_data/rules/experiments/constructor_tearoffs/rules/always_specify_types.dart
@@ -10,8 +10,11 @@ import 'package:meta/meta.dart';
 
 /// Constructor tear-offs
 void constructorTearOffs() {
-  List.filled; // LINT
   List<List>.filled; // LINT
+  // todo(pq): should be OK.
+  // List<E> Function<E>(int, E) filledList = List.filled; // OK - generic function
+  filledList<int>(3, 3); // OK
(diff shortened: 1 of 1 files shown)


test_data/rules/experiments/constructor_tearoffs/rules/always_specify_types.dart
```diff
@@ -10,8 +10,11 @@ import 'package:meta/meta.dart';
 
 /// Constructor tear-offs
 void constructorTearOffs() {
-  List.filled; // LINT
   List<List>.filled; // LINT
+  // todo(pq): should be OK.
+  // List<E> Function<E>(int, E) filledList = List.filled; // OK - generic function
```
bwilkerson, line 15: I wonder, yet again, whether it would be worthwhile to have a way to mark a linter test as failing so that we didn't need to remember to occasionally check to see whether this is passing yet. Maybe something like `OK-FAILING` and `LINT-FAILING`.

bwilkerson (approved):

test_data/rules/experiments/constructor_tearoffs/rules/always_specify_types.dart
pq: I was thinking similarly.  @asashour filed #2815 and even has a PR!  I'll take a closer look later today.

PR #2925, opened 2021-09-02
element metadata access cleanup

pq:
/cc @bwilkerson

### DIFF
--- a/lib/src/ast.dart
+++ b/lib/src/ast.dart
@@ -81,26 +81,12 @@ bool hasConstantError(LinterContext context, Expression node) {
 }
 
 /// Returns `true` if this [element] has a `@literal` annotation.
-bool hasLiteralAnnotation(Element element) {
-  var metadata = element.metadata;
-  for (var i = 0; i < metadata.length; i++) {
-    if (metadata[i].isLiteral) {
-      return true;
-    }
-  }
-  return false;
-}
+@Deprecated('prefer: element.hasLiteral')
+bool hasLiteralAnnotation(Element element) => element.hasLiteral;
 
 /// Returns `true` if this [element] has an `@override` annotation.
-bool hasOverrideAnnotation(Element element) {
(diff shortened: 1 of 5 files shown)


lib/src/rules/always_specify_types.dart
```diff
@@ -59,22 +59,8 @@ main() {
...
-bool _isOptionallyParameterized(TypeParameterizedElement element) {
-  var metadata = element.metadata;
-  return metadata.any((ElementAnnotation a) => _isOptionalTypeArgs(a.element));
-}
-
-bool _isOptionalTypeArgs(Element? element) =>
-    element is PropertyAccessorElement &&
-    element.name == _optionalTypeArgsVarName &&
-    element.library.name == _metaLibName;
+bool _isOptionallyParameterized(TypeParameterizedElement element) =>
```
bwilkerson, line 62: Consider inlining this.

lib/src/ast.dart
```diff
@@ -81,26 +81,12 @@ bool hasConstantError(LinterContext context, Expression node) {
...
-bool hasLiteralAnnotation(Element element) {
-  var metadata = element.metadata;
-  for (var i = 0; i < metadata.length; i++) {
-    if (metadata[i].isLiteral) {
-      return true;
-    }
-  }
-  return false;
-}
+@Deprecated('prefer: element.hasLiteral')
```
bwilkerson, line 84: It looks like you already removed references from this package. Is there any reason not to remove these two functions?

bwilkerson (approved):

lib/src/rules/always_specify_types.dart
pq: Good idea.

lib/src/ast.dart
pq: It *might* be used internally?

PR #2931, opened 2021-09-05
New lint always_declare_return_types_async

lukas-h:
# Description

It's like the `always_declare_return_types` rule but for `async` functions *only*.

If you don't want to declare a return type for functions that have *normal* return types (for example you are too lazy to always write `void` for functions returning nothing), but you want to have the benefits of the static analysis for asynchronous functions and *especially* the `await` keyword (e.g. rule `unawaited_futures`).

For this reason I started to implement `always_declare_return_types_async`.

I would be thrilled to receive some feedback, do all the necessary changes and improvements and eventually see this merged.


lukas-h:
@googlebot I signed it!

grundid:
Here is a demo code that demonstrates the issue. A linter rule that forces you to declare Future<void> in the clear() method could prevent it.

```dart
import 'package:test/test.dart';

class Demo {
  List<String> data = ["one", "two", "three"];

  clear() async {
    // dome some other async cleanup stuff
    await Future.delayed(Duration(seconds: 5));
    data.clear();
  }

  run() {
    clear();
    data.add("new");
  }
}

void main() {
  test("unawaited", () {
    final Demo demo = Demo();
    demo.run();
    expect(demo.data, ["new"]);

    /* the test failes with:
    Expected: ['new']
    Actual: ['one', 'two', 'three', 'new']
    Which: at location [0] is 'one' instead of 'new'
    */
  });
}

```

The correct could would look like this:

```dart
import 'package:test/test.dart';

class Demo {
  List<String> data = ["one", "two", "three"];

  Future<void> clear() async {
    // dome some other async cleanup stuff
    await Future.delayed(Duration(seconds: 5));
    data.clear();
  }

  run() async {
    await clear();
    data.add("new");
  }
}

void main() {
  test("unawaited", () async {
    final Demo demo = Demo();
    await demo.run();
    expect(demo.data, ["new"]);
  });
}

```

mosuem:
Closing this as it is stale. Feel free to reopen in case you still want to land it!

PR #2945, opened 2021-09-13
Quote function name in always_declare_return_types

srawlins:
# Description

Quote function name in always_declare_return_types

Addresses the second point in https://github.com/dart-lang/sdk/issues/47142.


lib/src/rules/always_declare_return_types.dart
```diff
@@ -66,13 +66,13 @@ class AlwaysDeclareReturnTypes extends LintRule {
 class _Visitor extends SimpleAstVisitor<void> {
   static const LintCode functionCode = LintCode(
       "always_declare_return_types", // ignore: prefer_single_quotes
```
bwilkerson, line 68: We could also drop these ignore comments at some point. Not required for this PR, of course.

bwilkerson (approved):

pq (approved):

PR #2947, opened 2021-09-13
(temp) fix spelunking w/ constructor tearoffs

pq:
/cc @scheglov

### DIFF
--- a/test/rule_test.dart
+++ b/test/rule_test.dart
@@ -4,6 +4,7 @@
 
 import 'dart:io';
 
+import 'package:analyzer/dart/analysis/features.dart';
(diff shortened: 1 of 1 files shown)


scheglov (approved):
Note that I landed a few changes to the analyzer that make it use `currentVersion`, and remove `latestSdkLanguageVersion` altogether. I hope to stop using `FeatureSet.forTesting` eventually. I have removed all its uses in the analyzer. So, we might need to switch to an alternative solution soon.

pq:
SGTM.  I'm happy to change whenever you're ready.  This is just a stop-gap.  Thanks!

Issue #2944, opened 2021-09-10
False positive for `unnecessary_parenthesis`


Pante:
* Dart SDK Version: 2.14

I currently have the following snippet of code.
```dart
void foo() {
  final items = [];
  (() => [].add('something')).compose(() {});
}


extension A on void Function() {

  void Function() compose(void Function() other) => () {
    this();
    other();
  };

}
```

`dart analyze` falsely reports that the parenthesis around `(() => [].add('something'))` can be removed. However, it cannot be removed since the code's semantics will change and become `() => [].add('something').compose(() {})`. 

I think it's also worth noting that the parenthesis can be removed if the function used `{ ... }` instead of `=>`.

PR #2948, opened 2021-09-13
fix `unnecessary_parenthesis` false positive for function expressions

pq:
Fixes #https://github.com/dart-lang/linter/issues/2944

/cc @bwilkerson @srawlins

### DIFF
--- a/lib/src/rules/unnecessary_parenthesis.dart
+++ b/lib/src/rules/unnecessary_parenthesis.dart
@@ -70,7 +70,7 @@ class _Visitor extends SimpleAstVisitor<void> {
 
     // https://github.com/dart-lang/linter/issues/2944
     // (() => foo()).bar();
-    if (expression is FunctionExpression) {
+    if (expression is FunctionExpression && parent is MethodInvocation) {
       return;
     }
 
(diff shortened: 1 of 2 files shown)


lib/src/rules/unnecessary_parenthesis.dart
```diff
@@ -68,6 +68,12 @@ class _Visitor extends SimpleAstVisitor<void> {
       return;
     }
 
+    // https://github.com/dart-lang/linter/issues/2944
+    // (() => foo()).bar();
+    if (expression is FunctionExpression && parent is MethodInvocation) {
```
srawlins, line 73: In addition to MethodInvocation this same problem may happen for (deep breath):

* getter access: `(() => foo()).bar;`
* setter access: `(() => foo()).bar = 7;`
* operators: `(() => foo())[6];` or `(() => foo())++;`
* type instantiation: `(() => foo()).bar<int>;`

pq: Right!  Good call.  I'll update.

pq:
@srawlins: I think I got the cases covered.  PTAL?

test_data/rules/unnecessary_parenthesis.dart
```diff
@@ -11,13 +11,21 @@ import 'dart:async';
...
+  (() => '')[0]; // OK
 }
 
 extension A on void Function() {
   void Function() compose(void Function() other) => () {
     this();
     other();
   };
+
+  set f(int f) {}
```
srawlins, line 26: This stuff is a trip. extension on a function type... 🤯

srawlins (approved):

bwilkerson (approved):

PR #2951, opened 2021-09-14
Prepare const static function tearoff test for analyzer roll

srawlins:
# Description

The next analyzer release will allow for top-level and static function tearoffs to be const. This means this test data file has one new LINT inside, where indicated.

Instead of merging this PR by itself, the diff can just be used in a PR which bumps analyzer.


pq:
This is great.  Thanks for following up on this!

pq (approved):
Thanks!

PR #2952, opened 2021-09-14
Ignore initializing formals for `prefer_final_parameters`

### LINKED ISSUE #2950: False positive for `prefer_final_parameters`
False positioves for all final named parameters:

image

Dart 2.14.0

parlough:
# Description

If a parameter is an initializing formal, reference to it refer to the field, so adding an extra `final` isn't useful. 

Fixes #2950


pq (approved):
👍 

Thanks for the quick fix!

FaFre:
Thank you guys for taking care of it so quickly @pq @parlough :)

PR #2957, opened 2021-09-16
Allow print when it is behind the `kDebugMode` flag

bwilkerson:
Per our conversation.

### DIFF
--- a/lib/src/rules/avoid_print.dart
+++ b/lib/src/rules/avoid_print.dart
@@ -5,6 +5,7 @@
 import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
 import 'package:analyzer/dart/element/element.dart';
+import 'package:linter/src/util/flutter_utils.dart';
 
 import '../analyzer.dart';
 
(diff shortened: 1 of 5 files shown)


lib/src/rules/avoid_print.dart
```diff
@@ -5,6 +5,7 @@
 import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
 import 'package:analyzer/dart/element/element.dart';
+import 'package:linter/src/util/flutter_utils.dart';
```
pq, line 8: I think this needs to get relativized...

test_data/rules/avoid_print.dart
```diff
@@ -4,10 +4,13 @@
 
 // test w/ `dart test -N avoid_print`
 
+import 'package:flutter/foundation.dart';
+
 void main() {
   print('ha'); // LINT
   [1,2,3].forEach(print); // LINT
   Future.value('hello').then(print); // LINT
+  if (kDebugMode) print(''); // OK
```
pq, line 13: Maybe a case w/ brace too?

```dart
if (kDebugMode) {
  print(''); // OK
}
```

pq: I'm guessing we don't care about cases like:

```dart
if (!kDebugMode) {
  /// ... 
} else {
  print(''); // OK
}
```

(To be clear, I don't think we should...)

pq (approved):

lib/src/rules/avoid_print.dart
bwilkerson: Done

test_data/rules/avoid_print.dart
bwilkerson: Done. And no, I wasn't planning on supporting any kind of flow control logic unless users really need it.

PR #2960, opened 2021-09-19
fix `unnecessary_lambdas` false positive for  const constructor invocations

### LINKED ISSUE #2959: unnecessary_lambdas false positive on const constructor invocations
See this reported by @jefflim-google:

image

My understanding is there is no way to get the equivalent semantics w/ constructor tearoffs so this is a false positive.

Maybe @lrhn or @eernstg could confirm?

pq:
Fixes #2959

/cc @bwilkerson


pq:
(Coverage breakage tracked: https://github.com/dart-lang/coverage/issues/353)

bwilkerson (approved):

PR #2964, opened 2021-09-21
migrate to a (lazy) cached getter

pq:
See: https://github.com/dart-lang/sdk/issues/47221#issuecomment-922115669

💎 

/cc @bwilkerson 

/fyi @jcollins-g @eernstg

### DIFF
--- a/lib/src/ast.dart
+++ b/lib/src/ast.dart
@@ -14,9 +14,7 @@ import 'package:path/path.dart' as path;
 import 'analyzer.dart';
 import 'utils.dart';
 
-List<String>? _reservedWords;
-
-List<String> get reservedWords => _reservedWords ??= _collectReservedWords();
+late final List<String> reservedWords = _collectReservedWords();
 
 /// Returns direct children of [parent].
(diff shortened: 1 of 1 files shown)


bwilkerson (approved):

lib/src/ast.dart
```diff
@@ -14,9 +14,7 @@ import 'package:path/path.dart' as path;
 import 'analyzer.dart';
 import 'utils.dart';
 
-List<String>? _reservedWords;
-
-List<String> get reservedWords => _reservedWords ??= _collectReservedWords();
+late final List<String> reservedWords = _collectReservedWords();
```
lrhn, line 17: You can (and I would) drop the `late` here. Top-level/static variables are inherently lazy, and always have been.

pq: Thanks!  This was absolutely a misunderstanding of `late` semantics on my part.  In the meantime, we added `unnecessary_late` to catch just such cases: https://github.com/dart-lang/linter/pull/3124.

PR #2971, opened 2021-09-28
Cleanup unwanted test diagnostics

pq:
See: #2969.

/cc @bwilkerson

### DIFF
--- a/test_data/rules/avoid_type_to_string.dart
+++ b/test_data/rules/avoid_type_to_string.dart
@@ -7,14 +7,14 @@
 // SHARED
 
 class A {
-  String toString() {}
+  String toString() => '';
 }
 
-String takesFunction(Function f) {}
+String takesFunction(Function f)  => '';
 
 class TypeChildWithOverride extends Type {
   @override
-  String toString() {}
+  String toString()  => '';
 }
 
 class TypeGrandChildWithOverride extends TypeChildWithOverride {}
@@ -24,7 +24,7 @@ class TypeChildNoOverride extends Type {}
 class TypeGrandChildNoOverride extends TypeChildNoOverride {}
 
 mixin ToStringMixin {
-  String toString() {}
+  String toString()  => '';
 }
 
 // BAD
@@ -38,7 +38,7 @@ class Bad {
 }
 
 class BadWithType extends Type {
-  Function passedFunction;
+  Function passedFunction = (){};
 
   BadWithType(Function func) : this.withFunc(func);
   BadWithType.withoutFunc() {}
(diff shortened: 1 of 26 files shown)


srawlins (approved):
Very excited! Love this

test_data/rules/avoid_type_to_string.dart
```diff
@@ -145,6 +145,7 @@ extension ExtensionOnGoodWithTypeAndMixin on GoodWithTypeAndMixin {
 }
 
 extension on int Function(int) {
+  // ignore: unused_element
```
bwilkerson, line 148: You could also name the extension so that it isn't private (unless that's important to the test in some way that I'm not seeing.

test_data/rules/cascade_invocations.dart
```diff
@@ -57,13 +57,15 @@ class Foo {
 }
 
 void noCascadeWithGetter() {
-  final foo = new Foo();
+  final foo = Foo();
   foo.baz(); // LINT
   foo.bar; // LINT
   foo.foo(); // LINT
   foo.bar = 8; // LINT
+  // ignore: invalid_null_aware_operator
```
bwilkerson, line 65: It isn't clear to me that it's useful to test the behavior of the lint in code with compilation errors. When it was valid to use the null aware access operator (pre null-safety) it made sense to me, but now there's already a different diagnostic, and I don't think we generally attempt to suppress lints in those cases (though maybe we should).

test_data/rules/join_return_with_assignment.dart
```diff
@@ -24,15 +24,16 @@ int doubleFoo() {
 }
 
 class A {
-  int _a;
+  late int _a;
   int get myA {
+    // ignore: dead_null_aware_expression
```
bwilkerson, line 29: Could we get rid of this one by making `_a` nullable?

bwilkerson (approved):
The ignore comments are at least marking the places where we have (or might have) technical debt, so I'm ok committing as-is if you prefer.

test_data/rules/cascade_invocations.dart
pq: I agree 100%.  This was a mistake on my part.  I don't think this is a useful state to be testing and in any event in the case where it is, I'd like to break those out into separate unit tests rather than suppressing w/ ignores.

Thanks for the catch!

test_data/rules/join_return_with_assignment.dart
pq: Yep. Done.

test_data/rules/avoid_type_to_string.dart
pq: I wish I knew the intent here.   I'll look a little closer.

(Aside: these are exactly the kinds of questions I hope this work will help avoid in the future!)

pq:
> The ignore comments 

FWIW, I've been thinking it would be nice to do another round through these tests to audit all the ignores. 

Thanks for the review!

PR #2972, opened 2021-09-28
Remove unnecessary `begin of input` and raw string from `identifierPart`

asashour:
In `unnecessary_brace_in_string_interps.dart`, there is an extra `^`, which doesn't add a value, since the `identifierPart` is used only once in `startsWith()`.

This PR removes, as also hinted in https://dart-review.googlesource.com/c/sdk/+/212742/comments/e3e4560f_28eca233

Also, no need to have the starting as `raw`.


bwilkerson (approved):

pq (approved):
Thanks!

PR #2975, opened 2021-09-29
in-memory reflective test harness

### LINKED ISSUE #2974: create an in-memory rule test harness
See: #2969.

Like in server, this should support:

* multiple compilation units,
* package dependencies, and
* assertion of the presence (or absence) of diagnostics

I'm inclined to make something that borrows from `AbstractSingleUnitTest` (though probably not using reflective tests).  Input here is welcome.

/cc @bwilkerson

pq:
Fixes #2974

Extracts a set of tests with diagnostics (`avoid_init_to_null` -- see: #2969).

/cc @bwilkerson

### DIFF
--- a/test/rule_test_support.dart
+++ b/test/rule_test_support.dart
@@ -31,10 +31,12 @@ class AnalysisOptionsFileConfig {
   String toContent() {
     var buffer = StringBuffer();
 
-    buffer.writeln('analyzer:');
-    buffer.writeln('  enable-experiment:');
-    for (var experiment in experiments) {
-      buffer.writeln('    - $experiment');
+    if (experiments.isNotEmpty) {
+      buffer.writeln('analyzer:');
+      buffer.writeln('  enable-experiment:');
+      for (var experiment in experiments) {
+        buffer.writeln('    - $experiment');
+      }
     }
 
     buffer.writeln('linter:');
@@ -53,7 +55,7 @@ abstract class LintRuleTest extends PubPackageResolutionTest {
   @override
   List<String> get _lintRules => [if (lintRule != null) lintRule!];
 
-  Future<void> assertHasLint(String code) async {
+  Future<void> assertLint(String code) async {
     addTestFile(code);
     await resolveTestFile();
 
@@ -66,7 +68,7 @@ abstract class LintRuleTest extends PubPackageResolutionTest {
     fail('Expected: $lintRule, found none');
   }
 
-  Future<void> assertHasNoLint(String code) async {
+  Future<void> assertNoLint(String code) async {
     addTestFile(code);
     await resolveTestFile();
 
--- a/test/rule_test_support.dart
+++ b/test/rule_test_support.dart
@@ -1,0 +1,306 @@
+// Copyright (c) 2021, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/analysis/results.dart';
+import 'package:analyzer/error/error.dart';
+import 'package:analyzer/file_system/file_system.dart';
+import 'package:analyzer/src/dart/analysis/analysis_context_collection.dart';
+import 'package:analyzer/src/dart/analysis/byte_store.dart';
+import 'package:analyzer/src/dart/analysis/driver_based_analysis_context.dart';
+import 'package:analyzer/src/dart/analysis/experiments.dart';
+import 'package:analyzer/src/test_utilities/mock_packages.dart';
+import 'package:analyzer/src/test_utilities/mock_sdk.dart';
+import 'package:analyzer/src/test_utilities/package_config_file_builder.dart';
+import 'package:analyzer/src/test_utilities/resource_provider_mixin.dart';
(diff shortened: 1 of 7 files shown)


test/rule_test_support.dart
```diff
@@ -0,0 +1,306 @@
...
+
+  AnalysisOptionsFileConfig({
+    this.experiments = const [],
+    this.lints = const [],
+  });
+
+  String toContent() {
+    var buffer = StringBuffer();
+
+    buffer.writeln('analyzer:');
```
bwilkerson, line 34: Lines 34-38 might want to be conditional on there being some experiments to enable. It's probably safe the way it is, but closer to what users will actually do if we don't have unnecessary content in the file (and probably a tiny bit faster, which for large numbers of tests can add up to a significant savings).

test/rule_test_support.dart
```diff
@@ -0,0 +1,306 @@
...
+    }
+
+    fail('Expected: $lintRule, found none');
+  }
+
+  Future<void> assertHasNoLint(String code) async {
+    addTestFile(code);
+    await resolveTestFile();
+
+    for (var error in errors) {
```
bwilkerson, line 73: Why not just `if (errors.isNotEmpty)`?

test/rules/avoid_init_to_null.dart
```diff
@@ -0,0 +1,60 @@
...
+    defineReflectiveTests(AvoidInitToNullTest);
+  });
+}
+
+@reflectiveTest
+class AvoidInitToNullTest extends LintRuleTest {
+  @override
+  String get lintRule => 'avoid_init_to_null';
+
+  test_lintOnNullable() async {
```
bwilkerson, line 20: You can name tests however you like, but the style we've developed in the analyzer seems to work fairly well. For example, we try to keep the names as short as possible. To that end I'd drop the "lintOn" and "NoLintOn" prefixes (and always use lowercase after an underscore). More generally, the test name should describe the condition being tested, not the outcome of the test. The names of the "assert" methods tend to be enough for that in practice.

I'd also add "_topLevelVariable" in keeping with the pattern used in the rest of the methods.

test/rules/avoid_init_to_null.dart
```diff
@@ -0,0 +1,60 @@
...
+  test_NoLintOnInvalidAssignment_parameter() async {
+    // Produces an invalid_assignment compilation error.
+    await assertHasNoLint(r'''
+class X {
+  X({int a: null});
+}
+''');
+  }
+
+  test_NoLintOnInvalidAssignment_parameter2() async {
```
bwilkerson, line 44: "parameter2" isn't very descriptive, so you might consider something like "_namedParameter_fieldFormal" (and "_namedParameter" on line 35).

test/rule_test_support.dart
```diff
@@ -0,0 +1,306 @@
...
+  String? get lintRule;
+
+  @override
+  List<String> get _lintRules => [if (lintRule != null) lintRule!];
+
+  Future<void> assertHasLint(String code) async {
+    addTestFile(code);
+    await resolveTestFile();
+
+    for (var error in errors) {
```
bwilkerson, line 60: There's an interesting design question here. Currently this only validates that the lint rule produced at least one diagnostic. It doesn't verify that no other diagnostics (lint or otherwise) were produced. The tests would be stronger if we completely controlled the set of diagnostics.

Longer term, consider copying `GatheringErrorListener` from either `analyzer` or `analysis_server` (or moving the analyzer version into `src`). It does a really good job of comparing expected outcomes and producing a useful report when they don't match (including the code needed to update the test if the test expectations need to be changed.

It also allows you to specify the location of the diagnostic, which would be useful.

test/rule_test_support.dart
```diff
@@ -0,0 +1,306 @@
...
+      if (error.errorCode.name == lintRule) {
+        return;
+      }
+    }
+
+    fail('Expected: $lintRule, found none');
+  }
+
+  Future<void> assertHasNoLint(String code) async {
```
bwilkerson, line 69: Consider dropping the "Has" and "HasNot" from these names. I don't think they add to the readability.

bwilkerson (approved):
It's hard to tell which portions of this were copied, so I might not have looked at everything as carefully as I could have.

I think it's fine to land this as-is, but if you were wanting to start a conversation about the framework I've added a couple of comments related to the way the analyzer framework is used.

test/rule_test_support.dart
pq: Good idea.  Thanks!

test/rule_test_support.dart
pq: Because we intentionally want to support the presence of other diagnostics.  Ww might consider and error filter or something down the line if we want to get explicit.

test/rules/avoid_init_to_null.dart
pq: Done.  Thanks!

test/rule_test_support.dart
pq: Absolutely agreed. I think we'll want to do that for sure when we get to migrating integration tests.

pq:
Thanks for the feedback!

PR #2978, opened 2021-09-29
add support for asserting diagnostics && migrate `overridden_fields`

pq:
See: #2969

Diagnostics matching API can evolve and certainly the air-lifted test source should be revisited at some point...  (So many errors -- all the more value in this work!)

/cc @bwilkerson

### DIFF
--- a/test/rule_test_support.dart
+++ b/test/rule_test_support.dart
@@ -63,9 +63,6 @@ class AnalysisOptionsFileConfig {
 
 /// A description of a diagnostic that is expected to be reported.
 class ExpectedDiagnostic {
-  /// An empty array of diagnostic descriptors used when no errors are expected.
-  static List<ExpectedDiagnostic> noDiagnostics = <ExpectedDiagnostic>[];
-
   final DiagnosticMatcher diagnosticMatcher;
 
   /// The offset of the beginning of the diagnostic's region.
--- a/test/rule_test_support.dart
+++ b/test/rule_test_support.dart
@@ -13,11 +13,23 @@ import 'package:analyzer/src/test_utilities/mock_packages.dart';
 import 'package:analyzer/src/test_utilities/mock_sdk.dart';
 import 'package:analyzer/src/test_utilities/package_config_file_builder.dart';
 import 'package:analyzer/src/test_utilities/resource_provider_mixin.dart';
+import 'package:linter/src/analyzer.dart';
 import 'package:linter/src/rules.dart';
 import 'package:meta/meta.dart';
 import 'package:test/test.dart';
 
+export 'package:analyzer/src/error/codes.dart';
 export 'package:analyzer/src/test_utilities/package_config_file_builder.dart';
+
+ExpectedError error(ErrorCode code, int offset, int length,
+        {Pattern? messageContains}) =>
+    ExpectedError(code, offset, length, messageContains: messageContains);
+
+ExpectedLint lint(String lintName, int offset, int length,
+        {Pattern? messageContains}) =>
+    ExpectedLint(lintName, offset, length, messageContains: messageContains);
+
+typedef DiagnosticMatcher = bool Function(AnalysisError error);
 
 class AnalysisOptionsFileConfig {
   final List<String> experiments;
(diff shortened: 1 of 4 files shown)


test/rules/overridden_fields.dart
```diff
@@ -0,0 +1,134 @@
...
+
+class B extends A {
+  String field = 'otherValue';
+}
+''');
+  }
+
+  test_recursiveInterfaceInheritance() async {
+    // Produces a recursive_interface_inheritance diagnostic.
+    await assertNoLint(r'''
```
pq, line 35: FWIW this should probably become a new call to `assertDiagnostics` (TODO added).

test/rule_test_support.dart
```diff
@@ -49,12 +61,164 @@ class AnalysisOptionsFileConfig {
   }
 }
 
+/// A description of a diagnostic that is expected to be reported.
+class ExpectedDiagnostic {
+  /// An empty array of diagnostic descriptors used when no errors are expected.
+  static List<ExpectedDiagnostic> noDiagnostics = <ExpectedDiagnostic>[];
```
bwilkerson, line 67: I don't know whether this was in the original source, but there's no value in Dart to defining constants like this. It's better to add `const` before the right-hand side and then just inline the expression everywhere.

test/rule_test_support.dart
```diff
@@ -49,12 +61,164 @@ class AnalysisOptionsFileConfig {
...
+
+  final DiagnosticMatcher diagnosticMatcher;
```
bwilkerson, line 69: Personally I would prefer a style in which the subclasses override `matches` to invoke `super.matches` and then do the additional checks directly.

test/rule_test_support.dart
```diff
@@ -49,12 +61,164 @@ class AnalysisOptionsFileConfig {
   }
 }
 
+/// A description of a diagnostic that is expected to be reported.
+class ExpectedDiagnostic {
```
bwilkerson, line 65: Consider splitting this file into several smaller files, maybe in a subdirectory named `support` or `utilities` with each file following the normal file naming conventions.

test/rules/overridden_fields.dart
```diff
@@ -0,0 +1,130 @@
...
+    await assertDiagnostics(r'''
+class Base {
+  Object field = 'lorem';
+
+  Object something = 'change';
+}
+
+class Bad1 extends Base {
+  @override
+  final x = 1, field = 'ipsum'; // LINT
```
bwilkerson, line 56: The markers aren't needed anymore, and they might confuse people who don't know that they're not needed, so it's probably better to remove them.

test/rules/overridden_fields.dart
```diff
@@ -0,0 +1,130 @@
...
+    // Produces a recursive_interface_inheritance diagnostic.
+    await assertNoLint(r'''
+class A extends B {}
+class B extends A {
+  int field = 0;
+}
+''');
+  }
+
+  test_mixinInheritsFromNotObject() async {
```
bwilkerson, line 43: I would strongly encourage you to have a separate test method for each lint diagnostic. It's much easier to understand what's being tested.

test/rules/overridden_fields.dart
```diff
@@ -0,0 +1,130 @@
...
+    // See: https://github.com/dart-lang/linter/issues/2969
+    // Preserves existing testing logic but has so many misuses of mixins that
+    // that it's hard to know how much tested logic is intentional.
```
bwilkerson, line 46: In some cases we might need to toss out the old tests entirely and figure out all over again which cases need to be tested. It would be a lot of work, but better than having a bunch of tests whose value is questionable.

bwilkerson (approved):

test/rule_test_support.dart
pq: It was in the original source.  My guess is the intention was to make it clear that an empty list was meant to be unpopulated and not an accident but I don't feel strongly so I'll remove.

test/rule_test_support.dart
pq: 💯 

I'll do this in a follow-up.

test/rules/overridden_fields.dart
pq: Agreed.  This test source is not trivial to disentangle and I'm not confident that a lot of it has lasting value so I decided to leave it as is w/ the TODO below.

That said, in general, I'm with you 100%.

Thanks!

test/rules/overridden_fields.dart
pq: Agreed.

https://github.com/dart-lang/linter/issues/2981

PR #2984, opened 2021-09-30
more test diagnostic cleanup

pq:
See: #2969

/cc @bwilkerson

### DIFF
--- a/test/rule_test_support.dart
+++ b/test/rule_test_support.dart
@@ -18,6 +18,7 @@ import 'package:linter/src/rules.dart';
 import 'package:meta/meta.dart';
 import 'package:test/test.dart';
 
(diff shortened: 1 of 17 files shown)


test/rule_test_support.dart
```diff
@@ -214,32 +215,6 @@ abstract class LintRuleTest extends PubPackageResolutionTest {
       fail(buffer.toString());
     }
   }
```
pq, line 217: Hrmm.  Looks like I didn't rebase.  Luckily the change that removes `assertLint ` has landed.  Sorry for the noise!

bwilkerson: np

bwilkerson (approved):

test/rule_test_support.dart
pq: Thanks!

PR #2988, opened 2021-10-01
fix `prefer_collection_literals` named typed param false positive

### LINKED ISSUE #2985: prefer_collection_literals and empty named LinkedHashSet
As I want create a empty named LinkedHashSet, the lint https://dart-lang.github.io/linter/lints/prefer_collection_literals.html has a false positive.
Without a named parameter, the linter does not complain.  

```
class Foo {}

void a({required LinkedHashSet<Foo> some}) {}
void b(LinkedHashSet<Foo> some) {}

void c() {
  a(some: LinkedHashSet()); // lint false positive
  b( LinkedHashSet()); // no lint
}
```

pq:
Fixes #2985

/cc @bwilkerson

### DIFF
--- a/test/rules/prefer_collection_literals.dart
+++ b/test/rules/prefer_collection_literals.dart
@@ -1,0 +1,84 @@
+// Copyright (c) 2021, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:test_reflective_loader/test_reflective_loader.dart';
+
+import '../rule_test_support.dart';
+
+main() {
+  defineReflectiveSuite(() {
+    defineReflectiveTests(PreferCollectionLiteralsTest);
+  });
+}
+
+@reflectiveTest
+class PreferCollectionLiteralsTest extends LintRuleTest {
+  @override
+  String get lintRule => 'prefer_collection_literals';
+
+  /// https://github.com/dart-lang/linter/issues/2985
+  test_linkedHashSetParameter_named_type_required() async {
+    await assertDiagnostics(r'''
+import 'dart:collection';
+    
+class Foo {}
+
+void a({required LinkedHashSet<Foo> some}) {}
+
+void c() {
+  a(some: LinkedHashSet<Foo>());
+}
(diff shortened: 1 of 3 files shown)


test/rules/prefer_collection_literals.dart
```diff
@@ -0,0 +1,84 @@
...
+    
+class Foo {}
+
+void a({required LinkedHashSet<Foo> some}) {}
+
+void c() {
+  a(some: LinkedHashSet<Foo>());
+}
+''', [
+      // No lints
```
bwilkerson, line 33: Not related to this PR, and it might be more confusing, but consider making the list of diagnostics optional so that we don't have to have an explicit list in every "no diagnostics" call.

test/rules/prefer_collection_literals.dart
```diff
@@ -0,0 +1,84 @@
...
+    
+class Foo {}
+
+void a({required Set<Foo> some}) {}
+
+void c() {
+  a(some: LinkedHashSet<Foo>());
+}
+''', [
+      lint('prefer_collection_literals', 103, 20),
```
bwilkerson, line 49: Consider shortening this slightly, and removing duplication, by writing `lint(lintRule, 103, 20);`. Not sure it's better.

bwilkerson (approved):

test/rules/prefer_collection_literals.dart
pq: Interesting.  Alternatively how about a new `assertNoDiagnostics`?

test/rules/prefer_collection_literals.dart
pq: I'm not sure either.  I'll add this to a list of things to discuss offline.

pq:
Thanks!


(I'll address the analyzer API changes in another PR.)

test/rules/prefer_collection_literals.dart
bwilkerson: I like that better than making it optional. Less error prone.

pq:
Ha.  Of course Konstantin already got to it: https://github.com/dart-lang/linter/pull/2989. 😄

PR #2999, opened 2021-10-05
Stop using FeatureSet.forTesting()

scheglov:
I would like to remove it.

You always want to go through `AnalysisContextCollection`, which discovers the pertinent analysis options file.
Even in tests. It is a question of configuring the file system, the actual, or in-memory, or overlay.


pq (approved):
👍

PR #3051, opened 2021-11-02
+ unsafe_unicode

### LINKED ISSUE #3048: lint for: unsafe_unicode
Context: 

* https://krebsonsecurity.com/2021/11/trojan-source-bug-threatens-the-security-of-all-code/

pq:
Fixes #3048

@mit-mit: want to help w/ word-smithing?

/cc @rakudrama @bwilkerson @jacob314 @timsneath

### DIFF
--- a/lib/src/rules/unsafe_unicode.dart
+++ b/lib/src/rules/unsafe_unicode.dart
@@ -12,7 +12,7 @@ const _desc = r"Don't use bidirectional Unicode text.";
 
 const _details = r'''
 
-**AVOID** using bidirectional Unicode text.
+**DO NOT** use bidirectional Unicode text.
 
 [Bidirectional Unicode](https://unicode.org/reports/tr9/) text may be
 interpreted and compiled differently than how it appears in editors
@@ -21,6 +21,12 @@ leading to possible security vulnerabilities.
 See the Common Vulnerabilities and Exposures (CVE) publication:
 [CVE-2021-42574](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-42574)
 for more details.
+
+For cases where bidirectional unicode text is needed, consider replacing the
+codepoints with the related escape sequences or prefix the occurence with 
+`// ignore: unsafe_unicode` and a comment explaining the intended use. For more
+on configuring analysis see the analysis options file
+[docs](https://dart.dev/guides/language/analysis-options#the-analysis-options-file).
 ''';
 
 class UnsafeUnicode extends LintRule {
--- a/lib/src/rules/unsafe_unicode.dart
+++ b/lib/src/rules/unsafe_unicode.dart
@@ -1,0 +1,60 @@
+// Copyright (c) 2021, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+
+import '../analyzer.dart';
+import '../util/unicode_utils.dart';
+
+const _desc = r'Do not use bidirectional Unicode text.';
+
+const _details = r'''
+
+**AVOID** using bidirectional Unicode text.
+
+Bidirectional Unicode text may be interpreted and compiled differently than how
+it appears in editors leading to possible security vulnerabilities. 
+
+See the Common Vulnerabilities and Exposures (CVE) publication:
+[CVE-2021-42574](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-42574)
+for more details.
+''';
+
+class UnsafeUnicode extends LintRule {
+  UnsafeUnicode()
+      : super(
+            name: 'unsafe_unicode',
+            description: _desc,
+            details: _details,
+            // todo(pq): consider a new 'security' group
+            group: Group.errors);
+
+  @override
+  void registerNodeProcessors(
+      NodeLintRegistry registry, LinterContext context) {
+    var visitor = _Visitor(this, context);
+    registry.addCompilationUnit(this, visitor);
+  }
+}
+
+class _Visitor extends SimpleAstVisitor {
+  final LintRule rule;
+  final String source;
+
+  _Visitor(this.rule, LinterContext context)
+      : source = context.currentUnit.content;
+
+  @override
+  void visitCompilationUnit(CompilationUnit node) {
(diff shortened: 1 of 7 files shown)


test_data/rules/unsafe_unicode.dart
```diff
@@ -0,0 +1,7 @@
+// Copyright (c) 2021, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// test w/ `dart test -N unsafe_unicode`
+
+var u202e = '‮'; // LINT [-6]
```
pq, line 7: Cool to see how this (mis)renders.

Here's how it looks in the IDE:

image


A bit of a case in point eh?

bwilkerson (approved):

lib/src/rules/unsafe_unicode.dart
```diff
@@ -0,0 +1,60 @@
...
+import 'package:analyzer/dart/ast/visitor.dart';
+
+import '../analyzer.dart';
+import '../util/unicode_utils.dart';
+
+const _desc = r'Do not use bidirectional Unicode text.';
+
+const _details = r'''
+
+**AVOID** using bidirectional Unicode text.
```
parlough, line 15: I like that you link to the CVE publication but is there some other link to info about bidirectional text we can point to here? Or an example?

Hixie:
I think this is perfectly find on the short term but it would be nice on the long term if we allowed balanced bidi formatting characters in strings and comments, otherwise writing apps that combine English and Hebrew, or Arabic and French, or whatever, is going to be rather annoying and frustrating.

lib/src/rules/unsafe_unicode.dart
pq: I added a spec link but an example would be welcome.

pq:
@Hixie: it's a good point.  Would escapes be acceptable in that case or too cumbersome?

Hixie:
Escapes wouldn't affect the rendering, so would be just as frustrating. Imagine if you had to edit all your strings like this:

```dart
   String message = 'DLROW OLLEH';
```

lib/src/rules/unsafe_unicode.dart
```diff
@@ -0,0 +1,62 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+
+import '../analyzer.dart';
+import '../util/unicode_utils.dart';
+
+const _desc = r"Don't use bidirectional Unicode text.";
```
mit-mit, line 11: "Do not" here and "avoid" below seem inconsistent?

lib/src/rules/unsafe_unicode.dart
```diff
@@ -0,0 +1,62 @@
...
+
+const _details = r'''
+
+**AVOID** using bidirectional Unicode text.
+
+[Bidirectional Unicode](https://unicode.org/reports/tr9/) text may be
+interpreted and compiled differently than how it appears in editors
+leading to possible security vulnerabilities. 
+
```
mit-mit, line 20: Do we also want to say something like:

```
For rare cases where bidirectional unicode text is needed,
prefix the statement with `// ignore: unsafe_uncode` and a
comment explaining that the intended use is.
```

test/util/unicode_utils_test.dart
```diff
@@ -0,0 +1,42 @@
+// Copyright (c) 2021, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:linter/src/util/unicode_utils.dart';
+import 'package:test/test.dart';
+
+void main() {
```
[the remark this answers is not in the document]
pq: Once the lint lands and is added to core, absolutely yes we'll need an ignore.

lib/src/rules/unsafe_unicode.dart
pq: Updated.

test/util/unicode_utils_test.dart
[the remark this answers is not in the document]
bwilkerson: Or, given that you're testing a function that takes an int, skip the whole process of putting the code unit in a string and just write something like
```dart
for (var c  in [8235, ...]) {
```

But if you really want strings, why not write them the way they're written in `lib/src/util/unicode_utils.dart`, which also avoids triggering the lint?

lib/src/util/unicode_utils.dart
```diff
@@ -0,0 +1,21 @@
...
+  '\u202A',
+  '\u202B',
+  '\u202C',
+  '\u202D',
+  '\u202E',
+  '\u2066',
+  '\u2067',
+  '\u2068',
+  '\u2069',
+].map((u) => u.codeUnitAt(0));
```
rakudrama, line 18: This is a mapped list iterable which will evaluate the `(u) => ...` function for each element for each code unit that is passed to `unsafe`.
So add a `.toList()` or `.toSet()`.

You could use a set literal:
```dart
final _unsafe = { 0x202A, 0x202B, .... };
```
or
```
final _unsafe = {..."\u202A\u202B\u202C\u...".codeUnits};
```

The fastest would probably be to do a range check:
```dart
bool unsafe(int codeUnit) =>
    0x202A <= codeUnit && codeUnit <= 0x202E ||
    0x2066 <= codeUnit && codeUnit <= 0x2069;
```

It is probably also fast to find with a RegExp `[\u202A-\u202E\u2066-\u2069]`.
You could step through the matches with `.indexOf(pattern, start)` - that would give you the code-unit index which presumably can be converted into a line/column.

lib/src/rules/unsafe_unicode.dart
rakudrama: The implementation does not support this usage, so either the comment or the implementation should change.

lib/src/util/unicode_utils.dart
pq: > This is a mapped list iterable  

😬 

Wow.  This is hilarious especially given that I've been thinking about analyses to capture **exactly** this kind of misuse of a lazy iterable.  (See: https://github.com/dart-lang/linter/issues/855.)

> The fastest would probably be to do a range check.

💯 

> You could step through the matches with ...

Oh.  This is a great idea.

Thank you!

pq:
Closing as we're implementing this check in the SDK as a hint: dart-lang/sdk#47611

PR #3052, opened 2021-11-02
Add a new `unnecessary_late` lint

parlough:
Adds a new `unnecessary_late` lint which will detect usages of the late modifier `late` on static and top-level variables which are evaluated as if they are marked with `late` already.

I believe there are other situations where `late` may be unnecessary. The first that comes to mind is in a local variable declaration where the initializer will have no side effects. I have opted to not account for this and potential other  more complicated situations currently(unless I'm missing something obvious). I was thinking other situations could be incorporated into this lint in the future, **but if you think this** should have a more specific name such `unnecessary_static_late`, let me know :)

### DIFF
--- a/lib/src/rules/unnecessary_late.dart
+++ b/lib/src/rules/unnecessary_late.dart
@@ -1,0 +1,86 @@
+// Copyright (c) 2021, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+
+import '../analyzer.dart';
+
+const _desc = r"Don't specify the `late` modifier when it is not needed.";
+
+const _details = r'''
+
+**DO** not specify the `late` modifier for top-level and static variables
+when the declaration contains an initializer. 
+
+Top-level and static variables with initializers are already evaluated lazily
+as if they are marked `late`.
+
+**BAD:**
+```dart
+late String badTopLevelLate = "";
+```
+
+**GOOD:**
+```dart
+late String goodTopLevelLate;
+```
+
+**BAD:**
+```dart
+class BadExample {
+  static late String badStaticLate = "";
+}
+```
+
+**GOOD:**
+```dart
+class GoodExample {
+  static late String goodStaticLate;
+}
+```
+''';
+
+class UnnecessaryLate extends LintRule {
+  UnnecessaryLate()
+      : super(
+            name: 'unnecessary_late',
+            description: _desc,
+            details: _details,
+            group: Group.style);
+
+  @override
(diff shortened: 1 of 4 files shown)


lib/src/rules/unnecessary_late.dart
```diff
@@ -0,0 +1,86 @@
...
+    }
+  }
+
+  @override
+  void visitTopLevelVariableDeclaration(TopLevelVariableDeclaration node) {
+    _visitVariableDeclarations(node.variables);
+  }
+
+  @override
+  void visitFieldDeclaration(FieldDeclaration node) {
```
bwilkerson, line 81: Please sort this file.

test_data/rules/unnecessary_late.dart
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) 2021, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// test w/ `dart test -N unnecessary_late`
+
+late String unnecessaryTopLevel = ""; // LINT
+
+late String necessaryTopLevel; // OK
```
bwilkerson, line 9: Consider also adding a test of the form `String unnecessaryTopLevel = '';` just to test the other possible combination.

bwilkerson (approved):

pq (approved):
💎

pq:
> I believe there are other situations where late may be unnecessary.

Right.  Many seem to be harder to reliably check though making them not good candidates for a general lint.  My thinking there is that if there are such cases we may want to lint specifically for them.  Anyway, this is just a long way around to saying I think that `unnecessary_late` is a reasonable name and I say we go for it.

I'll add a few more folks to get their 2 cents, especially since I think this is a good candidate for inclusion in the core or recommended lint sets.

/fyi 📟  @jakemac53 @natebosch @munificent @lrhn @eernstg @leafpetersen 

Thanks as always @parlough for the contribution!

lrhn:
I think detecting a local `late` variable with an initializer not having a side effect, is going to be impossible. Also, I'd imagine using `late` to delay allocation, which is usually not considered a side-effect.

One other unnecessary `late` use is a local variable which is definitely assigned at all reads, and either not `final` or definitely unassigned at all writes. Not everybody is aware that you can declare an `final int x;` without an initializer, as long as you definitely initialize it once before any read.

parlough:
@bwilkerson Is this above currently feasible within the linter? If I'm understanding correctly, I believe this would require some sort of access to flow analysis, but I'm not sure what is exposed there.

pq:
Sorry for the lag on this.

@parlough: any objection to landing as-is?

parlough:
@pq Yeah I'm happy with this lint in the current state. I won't have time to investigate exposing and accessing the flow analysis for now.

pq:
Good deal.  Thanks!

PR #3055, opened 2021-11-03
Update justification for omit local variable types

### LINKED ISSUE #3054: Description for omit_local_variable_types is confusing
> Usually, the types of local variables can be easily inferred, so it isn't necessary to annotate them.

https://dart-lang.github.io/linter/lints/omit_local_variable_types.html

It isn't clear to me whether this means "easily inferred by the reader" (my guess) or "easily inferred by the compiler".

In Effective Dart we frame things a little differently.

> Local variables, especially in modern code where functions tend to be small, have very little scope. Omitting the type focuses the reader’s attention on the more important name of the variable and its initialized value.

https://dart.dev/guides/language/effective-dart/design#dont-redundantly-type-annotate-initialized-local-variables

Here we don't talk about inference at all - either from the compiler or the reader. When this has come up in past discussions, I don't think the framing of the type being easy to infer by a reader has been as convincing as the framing that very often the specific type isn't very important to the reader.

natebosch:
Fixes #3054

Copy the description from Effective Dart. Change from "CONSIDER" to
"DON'T" since Effective Dart changed to a stronger position in
https://github.com/dart-lang/site-www/pull/2966

### DIFF
--- a/lib/src/rules/omit_local_variable_types.dart
+++ b/lib/src/rules/omit_local_variable_types.dart
@@ -14,35 +14,54 @@ const _desc = r'Omit type annotations for local variables.';
 
 const _details = r'''
 
-**CONSIDER** omitting type annotations for local variables.
+**DON’T** redundantly type annotate initialized local variables.
 
-Usually, the types of local variables can be easily inferred, so it isn't
-necessary to annotate them.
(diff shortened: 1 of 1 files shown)


bwilkerson (approved):

pq (approved):
💎 

Thanks!

PR #3056, opened 2021-11-04
Lint: Don't use http urls in pubspec.yaml

sigurdm:
Fixes: https://github.com/dart-lang/pub/issues/3211


example/all.yaml
```diff
@@ -144,6 +144,7 @@ linter:
     - prefer_void_to_null
     - provide_deprecation_message
     - public_member_api_docs
+    - pubspec_no_http_urls
```
jonasfj, line 147: Should we consider calling it `avoid_http_urls` or `avoid_http_links` ?

I imagine that we should also extend it to cover HTTP links to git dependencies, and hosted-urls using HTTP (unless they are localhost)

lib/src/rules.dart
```diff
@@ -347,6 +349,7 @@ void registerLintRules({bool inTestMode = false}) {
     ..register(ProvideDeprecationMessage())
     ..register(PublicMemberApiDocs())
     ..register(PubPackageNames())
+    ..register(PubspecNoHttpUrls())
```
jonasfj, line 352: You probably forgot to include the file containing `PubspecNoHttpUrls` :rofl:

example/all.yaml
bwilkerson: We have generally been trying to move away from terms like "avoid" and "prefer" because the firmness of the advice tends to change over time.

I could perhaps suggest an alternative if I better understood what we're trying to avoid and why. Is there an issue discussing this lint? The 'details' suggests that this is to catch potentially unsecured access. If that's correct then consider a name that reflects the security concern, such as 'only_secure_dependencies'.

lib/src/rules/pub/no_http_urls.dart
```diff
@@ -0,0 +1,77 @@
+// Copyright (c) 2018, the Dart project authors. Please see the AUTHORS file
```
bwilkerson, line 1: nit: "2018" --> "2012"

lib/src/rules/pub/no_http_urls.dart
```diff
@@ -0,0 +1,77 @@
...
+import 'package:analyzer/src/lint/pub.dart'; // ignore: implementation_imports
+
+import '../../analyzer.dart';
+
+const _desc = r"Don't use http urls.";
+
+const _details = r'''
+**DON'T** use http urls in `pubspec.yaml`.
+
+Use https instead.
```
bwilkerson, line 14: Consider adding a 'good' and 'bad' example.

lib/src/rules/pub/no_http_urls.dart
```diff
@@ -0,0 +1,77 @@
+// Copyright (c) 2018, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/src/lint/pub.dart'; // ignore: implementation_imports
+
+import '../../analyzer.dart';
+
+const _desc = r"Don't use http urls.";
```
bwilkerson, line 9: The message isn't very helpful. Consider declaring a static field holding an instance of `LintCode` that has both a more complete problem message (something like "The 'http' scheme is not secure and shouldn't be used in a 'pubspec.yaml' file.") and a correction message (something like "Try using an 'https' scheme.".

example/all.yaml
sigurdm: I created an issue now: https://github.com/dart-lang/pub/issues/3211

lib/src/rules/pub/no_http_urls.dart
sigurdm: even 2021 :)

example/all.yaml
jonasfj: A name like `only_secure_dependencies` might be better too. Since we probably should discourage `git://` dependencies.

lib/src/rules/pub/no_http_urls.dart
bwilkerson: :-) Yes, that's what I was trying to type.

lib/src/rules/pub/no_http_urls.dart
```diff
@@ -0,0 +1,91 @@
...
+            name: 'pubspec_no_http_urls',
+            description: _desc,
+            details: _details,
+            group: Group.pub);
+
+  @override
+  PubspecVisitor getPubspecVisitor() => Visitor(this);
+
+  @override
+  LintCode get lintCode => const LintCode("Don't use http urls.",
```
bwilkerson, line 38: The first argument needs to be the name of the lint ('pubspec_no_http_urls').

example/all.yaml
pq: I like this suggestion.  Maybe we should be slightly more explicit though.  How about `only_secure_pub_dependencies`?

lib/src/rules/pub/secure_pubspec_urls.dart
```diff
@@ -0,0 +1,90 @@
...
+```yaml
+repository: 'https://github.com/dart-lang/example'
+```
+
+**BAD:**
+```dart
+repository: 'http://github.com/dart-lang/example'
+''';
+
+class PubspecNoHttpUrls extends LintRule {
```
bwilkerson, line 26: We generally name the class to be the same as the lint name. It's not critical, and it could be done in a later PR, but I'm mentioning it in case you intended to do so and just forgot.

lib/src/rules/pub/secure_pubspec_urls.dart
```diff
@@ -0,0 +1,90 @@
...
+            details: _details,
+            group: Group.pub);
+
+  @override
+  PubspecVisitor getPubspecVisitor() => Visitor(this);
+
+  @override
+  LintCode get lintCode => const LintCode(
+      'secure_pubspec_urls', 'The url should only only use secure protocols.',
+      correctionMessage: 'Try using https.');
```
bwilkerson, line 40: Should this read "Try using 'https' or 'git'."? (The protocol should be quoted in either case, but that's a minor nit.)

bwilkerson (approved):

lib/src/rules/pub/secure_pubspec_urls.dart
pq: I like this change too.

lib/src/rules/pub/secure_pubspec_urls.dart
pq: Yes, please.  This makes things much easier to keep track of.

pq (approved):
Modulo Brian's comments, looks great!

lib/src/rules/pub/no_http_urls.dart
sigurdm: Done

pq:
Awesome.  Thanks for the thoughtful back and forth!

PR #3061, opened 2021-11-13
`use_decorated_box`

minhqdao:
## Description

Use `DecoratedBox` over `Container` when only `Decoration` is specified.

Fixes https://github.com/dart-lang/linter/issues/3060

## Kind

Style advice.

## Bad Example

```dart
Widget buildArea() {
  return Container(
    decoration: const BoxDecoration(
      color: Colors.blue,
      borderRadius: BorderRadius.all(
        Radius.circular(5),
      ),
    ),
    child: const Text('...'),
  );
}
```

## Good Example

```dart
Widget buildArea() {
  return const DecoratedBox(
    decoration: BoxDecoration(
      color: Colors.blue,
      borderRadius: BorderRadius.all(
        Radius.circular(5),
      ),
    ),
    child: Text('...'),
  );
}
```

### DIFF
--- a/lib/src/rules/use_decorated_box.dart
+++ b/lib/src/rules/use_decorated_box.dart
@@ -1,0 +1,114 @@
+// Copyright (c) 2021, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+
+import '../analyzer.dart';
+import '../util/flutter_utils.dart';
+
+const _desc = r'Use `DecoratedBox`.';
+
+const _details = r'''
+
+**DO** use `DecoratedBox` when the `Container` only has `Decoration`.
+
+A `Container` is a heavier Widget than a `DecoratedBox`, and as bonus,
+`DecoratedBox` has a `const` constructor.
+
+**BAD:**
+```dart
+Widget buildArea() {
+  return Container(
+    decoration: const BoxDecoration(
+      color: Colors.blue,
+      borderRadius: BorderRadius.all(
+        Radius.circular(5),
+      ),
+    ),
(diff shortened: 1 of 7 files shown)


test_data/rules/use_decorated_box.dart
```diff
@@ -0,0 +1,60 @@
+// Copyright (c) 2021, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// test w/ `dart test -N use_decorated_box`
```
bwilkerson, line 5: We should also have a test case in which a `Container` is created with a `decoration` and at least one other argument other than `key` or `child`.

bwilkerson (commented):
Other than the one missing test case, this looks good.

bwilkerson (approved):

lib/src/rules/use_decorated_box.dart
```diff
@@ -0,0 +1,114 @@
...
+import 'package:analyzer/dart/ast/visitor.dart';
+
+import '../analyzer.dart';
+import '../util/flutter_utils.dart';
+
+const _desc = r'Use `DecoratedBox`.';
+
+const _details = r'''
+
+**DO** use `DecoratedBox` when `Container` has only `Decoration`.
```
pq, line 15: Consider:

    **DO** use `DecoratedBox` when `Container` has only a `Decoration`.

pq (changes requested):
Looks great. And thanks for kicking off the thoughtful corresponding conversation in #3060! 👍 

Could you add an attribution for yourself in `AUTHORS`?

Thanks!

lib/src/rules/use_decorated_box.dart
minhqdao: It's been changed, and I also added myself to `AUTHORS` 😏

pq: Thanks!

pq (approved):

PR #3062, opened 2021-11-15
+ no_leading_underscores_for_{library_prefixes, local_identifiers}

### LINKED ISSUE #1676: New rule for using a leading underscore for identifiers that aren't private

pq:
See: #1750.

Fixes #1676.

/cc @bwilkerson 

/fyi @jawahars16 @srawlins


test_data/rules/no_leading_underscores_for_local_identifiers.dart
```diff
@@ -0,0 +1,79 @@
...
+
+  void foo2(param) => null; // OK
+
+  void foo3(_) => null; // OK
+
+  void foo4(param_value) => null; // OK
+
+  void foo5(param, [_positional]) => null; // LINT
+
+  void foo6(param, [positional]) => null; // OK
```
bwilkerson, line 78: I don't see any tests involving
- named parameters (either required or optional),
- function typed formal parameters,
- function type defs, or
- `for` elements in collection literals.

lib/src/rules/no_leading_underscores_for_local_identifiers.dart
```diff
@@ -0,0 +1,115 @@
...
+
+  @override
+  void visitDeclaredIdentifier(DeclaredIdentifier node) {
+    checkIdentifier(node.identifier);
+  }
+
+  @override
+  void visitFormalParameterList(FormalParameterList node) {
+    for (var p in node.parameters) {
+      if (p is! FieldFormalParameter && !p.isOptionalNamed) {
```
bwilkerson, line 103: Consider adding a comment explaining why optional named parameters aren't being checked (I assume it's because it's already caught as a compile time error).

bwilkerson (approved):

lib/src/rules/no_leading_underscores_for_local_identifiers.dart
pq: Exactly right.  Comment added.

Issue #3063, opened 2021-11-16
proposal: `move_color_to_decoration`


minhqdao:
## `move_color_to_decoration`

##Description

Don't provide `Container` with both non-null `color` and `decoration`. Place `color` inside `decoration` instead.

## Details

When a `Container` is provided with both `color` and `decoration` (both non-null), an error is thrown at runtime, and there is currently nothing that warns you about that beforehand. This rule prevents you from running into the error by prompting to move `color` into `decoration`.

## Kind

Guard against errors.

## Good Examples 

```dart
Widget buildArea() {
  return Container(
    decoration: BoxDecoration(
      border: Border.all(color: Colors.white),
      color: Colors.black,
    ),
  );
}
```

## Bad Examples

```dart
Widget buildArea() {
  return Container(
    color: Colors.black,
    decoration: BoxDecoration(
      border: Border.all(color: Colors.white),
    ),
  );
}
```

## Discussion

Those who already ran into the error will know, but those who haven't will certainly do, and that's simply annoying. No one wants to dig through unexpected error messages, even if this one is rather easy to identify.

### Discussion checklist

- [x] List any existing rules this proposal complements, overlaps or conflict with.
Slightly complementing: https://github.com/dart-lang/linter/issues/3060 https://github.com/dart-lang/linter/pull/3061
- [x] List any relevant issues (reported here, the [SDK Tracker], or elsewhere).
Not found.
- [x] If there's any prior art (e.g., in other linters), please add references here.
None.
- [x] If this proposal corresponds to [Effective Dart] or [Flutter Style Guide] advice, please call it out. (If there isn’t any corresponding advice, should there be?)
- [x] If this proposal is motivated by real-world examples, please provide as many details as you can.  Demonstrating potential impact is _especially valuable_.

[Writing Lints]: https://github.com/dart-lang/linter/blob/master/doc/WritingLints.MD
[Effective Dart]: https://dart.dev/guides/language/effective-dart
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo]
[SDK Tracker]: https://github.com/dart-lang/sdk/issues

goderbauer:
This seems useful to inform people of an illegal configuration at code write time instead of compile time.

HansMuller:
I agree, if we can flag this problem early it's a win for developers.

pq:
[...]

PR #3064, opened 2021-11-16
Implement `move_color_to_decoration`

minhqdao:
## Description

When a `Container` is provided with both `color` and `decoration` (both non-null), an error is thrown at runtime, and there is currently nothing that warns you about that beforehand. This rule prevents you from running into the error by prompting to move `color` into `decoration`.

Fixes https://github.com/dart-lang/linter/issues/3063

### DIFF
--- a/example/all.yaml
+++ b/example/all.yaml
@@ -85,6 +85,7 @@ linter:
     - list_remove_unrelated_type
     - literal_only_boolean_expressions
     - missing_whitespace_between_adjacent_strings
+    - move_color_to_decoration
     - no_adjacent_strings_in_list
     - no_default_cases
     - no_duplicate_case_values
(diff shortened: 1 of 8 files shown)


bwilkerson (commented):
Ignoring the question of whether we should have a more general solution for flagging such situations, this PR lgtm.

srawlins:
CC @goderbauer what do you think?

srawlins:
An alternative implementation we might want is an annotation like `OnlyOneOf`. https://github.com/dart-lang/sdk/issues/47712

goderbauer:
Having a lint on this would be nice.

srawlins:
Sorry for the delay in review. I still think https://github.com/dart-lang/sdk/issues/47712 is a great alternative, but we haven't prioritized designing that annotation.

devoncarew:

Issue #1368, opened 2019-01-16
What to do with style lints that enforce rules that don't exist any more in Effective Dart


a14n:
For instance `prefer_constructors_over_static_methods` was created to enforce a lint that has disappeared:

> Prefer defining constructors instead of static methods to create instances.

- Should we deprecate them?
- Should we mark them with a special tag/category? Blocked on #1020
- Would it be easier to follow if Effective Dart is versioned? /cc @munificent

pq:
I'd be inclined to deprecate.

It's noteworthy that `prefer_constructors_over_static_methods` is not included in any of the rulesets we're trying to align (https://github.com/dart-lang/linter/issues/1365#issuecomment-454416014) so this one in particular would I think effect no-one.

@a14n, @kwalrath: are you aware of any others?

bwilkerson:
I disagree. The purpose of the linter is _not_ to enforce the style guide. The purpose of the linter is to enable users to customize the set of diagnostics that are produced by the analyzer.

If we have evidence that there are no users that are enabling this lint, then I think it's reasonable to remove it. But we should _not_ remove it just because it's no longer mentioned in the style guide. It's OK for users to choose to follow additional guidelines that are not in the style guide. At the risk of being heretical, it's even OK for users to choose to not follow guidelines that _are_ in the style guide.

kwalrath:
> I disagree. The purpose of the linter is not to enforce the style guide. The purpose of the linter is to enable users to customize the set of diagnostics that are produced by the analyzer.
>
> If we have evidence that there are no users that are enabling this lint, then I think it's reasonable to remove it. But we should not remove it just because it's no longer mentioned in the style guide. It's OK for users to choose to follow additional guidelines that are not in the style guide. At the risk of being heretical, it's even OK for users to choose to not follow guidelines that are in the style guide.

Agreed. But we should certainly remove lints that are irrelevant or problematic in Dart 2. Some used to be in the style guide but were removed (see dart-lang/site-www#863, dart-lang/site-www#889, and other [changes to Effective Dart](https://github.com/dart-lang/site-www/commits/master/src/_guides/language/effective-dart)). Here are some possibilities:

* `always_declare_return_types`
* `avoid_types_on_closure_parameters`
* `type_annotate_public_apis`
* `super_goes_last`: Irrelevant because Dart 2 enforces this. (I can't find this in the SDK changelog, but it's mentioned in dart-lang/site-www#863.)
[...]

PR #3077, opened 2021-11-29
migrate to analyzer 2.8.0

pq:
Fixes https://github.com/dart-lang/linter/issues/3069

/cc @scheglov @bwilkerson 

/fyi @srawlins

### DIFF
--- a/lib/src/rules/comment_references.dart
+++ b/lib/src/rules/comment_references.dart
@@ -95,9 +95,10 @@ class _Visitor extends SimpleAstVisitor<void> {
 
   @override
   void visitCommentReference(CommentReference node) {
-    var identifier = node.identifier;
-    if (!identifier.isSynthetic && identifier.staticElement == null) {
-      rule.reportLint(identifier);
+    var expression = node.expression;
+    if (expression.isSynthetic) return;
+    if (expression is Identifier && expression.staticElement == null) {
+      rule.reportLint(expression);
     }
   }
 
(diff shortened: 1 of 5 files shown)


lib/src/rules/comment_references.dart
```diff
@@ -95,9 +95,10 @@ class _Visitor extends SimpleAstVisitor<void> {
 
   @override
   void visitCommentReference(CommentReference node) {
-    var identifier = node.identifier;
-    if (!identifier.isSynthetic && identifier.staticElement == null) {
-      rule.reportLint(identifier);
+    var expression = node.expression;
+    if (expression.isSynthetic) return;
+    if (expression is Identifier && expression.staticElement == null) {
```
srawlins, line 100: +1 thanks for the pre-fix!

scheglov (approved):

test/rules/super_goes_last.dart
```diff
@@ -33,9 +32,7 @@ class C extends A {
 
 ''', [
       error(HintCode.UNUSED_FIELD, 61, 2),
-      // Update to `SUPER_INVOCATION_NOT_LAST` when analyzer 2.8.0+ is published
-      // https://github.com/dart-lang/linter/issues/3069
-      // error(CompileTimeErrorCode.INVALID_SUPER_INVOCATION, 84, 5),
+      error(CompileTimeErrorCode.SUPER_INVOCATION_NOT_LAST, 84, 5),
```
bwilkerson, line 35: Was this suppose to replace line 36? If both are currently being supported, should we stop reporting the lint when the error is being reported?

pq: I think the real answer is we remove the lint.  It's been deprecated for a while...

See also: https://github.com/dart-lang/linter/issues/1368#issuecomment-455253837

PR #3079, opened 2021-11-30
support lint test runs against an un-mocked SDK

pq:
/cc @bwilkerson 

/fyi @srujzs


bwilkerson (approved):

test_data/rules/unmocked/unsafe_html.dart
```diff
@@ -0,0 +1,32 @@
+// Copyright (c) 2021, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// A trimmed version of `unsafe_html` used to validate that tests can be run
```
srujzs, line 5: This is awesome to see that we can do this without significant pain, thanks Phil!

PR #3080, opened 2021-12-02
Add a lint for missing files in conditional imports

DanTup:
An issue came up in Dart-Code abut conditional imports not being validated:

```dart
import 'image_cropper/cropper_stub.dart'
    if (dart.library.io) 'image_cropper/cropper_mobile.dart' // no error/warning if file doesn't exist
    if (dart.library.html) 'image_cropper/cropper_web.dart';
```

Originally this was added to the analysis server (https://dart-review.googlesource.com/c/sdk/+/202924) but it triggered a lot of warnings internally for files that haven't been generated yet and it was agreed it would be better as a lint.

This is an attempt to move convert it to a lint, although I have a few outstanding questions:

- [x] what should it be called? I've used `missing_conditional_import` as a placeholder, but I suspect there are some conventions it should follow
- [x] is this tested correctly? I noticed a lot of lints have source in `test_data` but from reading #2969 I got the impression that might be the old way, and this may be the new way
- [ ] in the original analysis server version, the source was checked for existence differently (below), but as far as I can see I can't do that here, however calling `configuration.uriSource.exists()` seems to do what I want (the comment notes it won't include overlays, but I think that's probably fine - running the script through the VM also wouldn't find the overlay):
  ```
  bool _isExistingSource(Source source) {
    if (source is InSummarySource) {
    	return true;
    }
    return _library.isExistingFilePath(source.fullName);
  }
  ```

@pq @bwilkerson

### DIFF
--- a/lib/src/rules/missing_conditional_import.dart
+++ b/lib/src/rules/missing_conditional_import.dart
@@ -1,0 +1,70 @@
+// Copyright (c) 2021, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+
+import '../analyzer.dart';
+
+const _desc = r'Missing conditional import.';
+
+const _details = r'''
+
+**DON'T** reference files that do not exist in conditional imports.
+
+Code may fail at runtime if the condition evaluates such that the missing file
+needs to be imported.
+
+**BAD:**
+```dart
+import 'file_that_does_not_exist.dart';
+```
+
+**GOOD:**
+```dart
+import 'file_that_exists.dart';
+```
+
+''';
+
+class MissingConditionalImport extends LintRule {
+  MissingConditionalImport()
+      : super(
+            name: 'missing_conditional_import',
+            description: _desc,
+            details: _details,
+            group: Group.style);
+
+  @override
+  void registerNodeProcessors(
+      NodeLintRegistry registry, LinterContext context) {
+    var visitor = _Visitor(this);
+    registry.addCompilationUnit(this, visitor);
+  }
+}
+
+class _Visitor extends RecursiveAstVisitor<void> {
+  static const LintCode code = LintCode('missing_conditional_import',
+      "The target of the conditional URI '{0}' doesn't exist.",
+      correctionMessage: 'Try creating the file referenced by the URI, or '
+          'Try using a URI for a file that does exist.');
+
+  final LintRule rule;
+
+  _Visitor(this.rule);
+
+  @override
+  void visitConfiguration(Configuration configuration) {
+    String? uriContent = configuration.uri.stringValue;
+    if (uriContent != null) {
+      Source? source = configuration.uriSource;
+      if (!(source?.exists() ?? false)) {
+        rule.reportLint(configuration.uri,
+            arguments: [uriContent], errorCode: code);
+      }
+    }
+
+    super.visitConfiguration(configuration);
+  }
+}
--- a/test/rules/missing_conditional_import.dart
+++ b/test/rules/missing_conditional_import.dart
@@ -1,0 +1,78 @@
+// Copyright (c) 2021, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:test_reflective_loader/test_reflective_loader.dart';
+
+import '../rule_test_support.dart';
+
+main() {
+  defineReflectiveSuite(() {
+    defineReflectiveTests(MissingConditionalImportTest);
+  });
+}
+
+@reflectiveTest
+class MissingConditionalImportTest extends LintRuleTest {
+  @override
+  bool get addMetaPackageDep => true;
+
+  @override
+  String get lintRule => 'missing_conditional_import';
+
+  test_missingDartLibraries() async {
+    await assertDiagnostics(
+      r'''
+import ''
+    if (dart.library.io) 'dart:missing_1'
+    if (dart.library.html) 'dart:async'
+    if (dart.library.async) 'dart:missing_2';
+''',
+      [
+        error(HintCode.UNUSED_IMPORT, 7, 2),
+        lint('missing_conditional_import', 35, 16,
+            messageContains: 'dart:missing_1'),
+        lint('missing_conditional_import', 120, 16,
+            messageContains: 'dart:missing_2'),
+      ],
+    );
+  }
+
+  test_missingFiles() async {
+    newFile('$testPackageRootPath/lib/exists.dart');
+
+    await assertDiagnostics(
+      r'''
+import ''
+    if (dart.library.io) 'missing_1.dart'
+    if (dart.library.html) 'exists.dart'
+    if (dart.library.async) 'missing_2.dart';
+''',
+      [
+        error(HintCode.UNUSED_IMPORT, 7, 2),
+        lint('missing_conditional_import', 35, 16,
+            messageContains: 'missing_1.dart'),
+        lint('missing_conditional_import', 121, 16,
+            messageContains: 'missing_2.dart'),
+      ],
+    );
+  }
+
+  test_missingPackages() async {
+    await assertDiagnostics(
+      r'''
+import ''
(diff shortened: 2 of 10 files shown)


lib/src/rules/missing_conditional_import.dart
```diff
@@ -0,0 +1,70 @@
...
+```dart
+import 'file_that_exists.dart';
+```
+
+''';
+
+class MissingConditionalImport extends LintRule {
+  MissingConditionalImport()
+      : super(
+            name: 'missing_conditional_import',
```
bwilkerson, line 34: As for naming, I might suggest that we use a name similar to the error code (`URI_DOES_NOT_EXIST`) such as `conditional_uri_does_not_exist`. There's a companion error code in the analyzer (`URI_HAS_NOT_BEEN_GENERATED`) and we might want to have two separate messages for this lint to distinguish between those cases, but that's less clear to me.

(The only convention that I'm aware of for naming lints is that we're now actively discouraging using 'avoid', 'prefer' and other such designations from the style guide because those tend to change over time leaving us with misleading lint names. Other than that, lints should be named to either express the condition they're trying to ensure or the condition they're trying to prevent.)

lib/src/rules/missing_conditional_import.dart
```diff
@@ -0,0 +1,70 @@
...
+    var visitor = _Visitor(this);
+    registry.addCompilationUnit(this, visitor);
+  }
+}
+
+class _Visitor extends RecursiveAstVisitor<void> {
+  static const LintCode code = LintCode('missing_conditional_import',
+      "The target of the conditional URI '{0}' doesn't exist.",
+      correctionMessage: 'Try creating the file referenced by the URI, or '
+          'Try using a URI for a file that does exist.');
```
bwilkerson, line 51: nit: "Try" --> "try"

lib/src/rules/missing_conditional_import.dart
```diff
@@ -0,0 +1,70 @@
...
+  final LintRule rule;
+
+  _Visitor(this.rule);
+
+  @override
+  void visitConfiguration(Configuration configuration) {
+    String? uriContent = configuration.uri.stringValue;
+    if (uriContent != null) {
+      Source? source = configuration.uriSource;
+      if (!(source?.exists() ?? false)) {
```
bwilkerson, line 62: As you noted, this will probably work most of the time, but it would be better if we could ask the resource provider. Unfortunately there doesn't appear to be any way to access it currently.

My personal opinion is that we need to update `LinterContext` to provide access to the resource provider and then pass that in when constructing the visitor. Unfortunately, that's a multi-step process, and I'm not sure we want to delay merging this rule while that happens.

test/rules/missing_conditional_import.dart
```diff
@@ -0,0 +1,78 @@
...
+import '../rule_test_support.dart';
+
+main() {
+  defineReflectiveSuite(() {
+    defineReflectiveTests(MissingConditionalImportTest);
+  });
+}
+
+@reflectiveTest
+class MissingConditionalImportTest extends LintRuleTest {
```
bwilkerson, line 16: Yes, this is the preferred style for tests, thanks!

lib/src/rules/missing_conditional_import.dart
```diff
@@ -0,0 +1,70 @@
...
+
+  @override
+  void registerNodeProcessors(
+      NodeLintRegistry registry, LinterContext context) {
+    var visitor = _Visitor(this);
+    registry.addCompilationUnit(this, visitor);
+  }
+}
+
+class _Visitor extends RecursiveAstVisitor<void> {
```
bwilkerson, line 47: For performance reasons, the linter uses visitors as a dispatch mechanism rather than as a tree walking mechanism. Walking the tree is expensive, so we do it once and effectively run all of the lint rules simultaneously. As a result,
- `_Visitor` should extend `SimpleAstVisitor`,
- you should use `addConfiguration` rather than `addCompilationUnit` on line 43, and
- you should remove the invocation of `super`

lib/src/rules/missing_conditional_import.dart
```diff
@@ -0,0 +1,70 @@
...
+const _details = r'''
+
+**DON'T** reference files that do not exist in conditional imports.
+
+Code may fail at runtime if the condition evaluates such that the missing file
+needs to be imported.
+
+**BAD:**
+```dart
+import 'file_that_does_not_exist.dart';
```
pq, line 21: Consider updating to add a condition.

Maybe something like

```dart
import 'file_that_does_exist.dart' 
  if (condition) 'file_that_does_not_exist.dart';
```

(Or similar.)

lib/src/rules/missing_conditional_import.dart
DanTup: Thanks! I've changed it to `conditional_uri_does_not_exist`.

> There's a companion error code in the analyzer (URI_HAS_NOT_BEEN_GENERATED) and we might want to have two separate messages for this lint to distinguish between those cases, but that's less clear to me.

I don't know much about that so have no opinion, but if you think it would also be useful here I'm happy to add it - it doesn't seem like a complicated change (just changing the code/error depending on whether the uri ends with [these suffixes](https://github.com/dart-lang/sdk/blob/e995cb5f7cd67d39c1ee4bdbe95c8241db36725f/pkg/analyzer/lib/src/util/file_paths.dart#L75)?).

lib/src/rules/missing_conditional_import.dart
DanTup: Done, thanks!

lib/src/rules/missing_conditional_import.dart
DanTup: I'm happy to go with your decision here. If you think it's worth doing (and I know enough to do it), I'm happy to. Although I'm not sure if having to support `package:` URIs complicates using the resource provider. This `exists()` call seemed to handle that (and `dart:` URIs) for me which I thought might have complicated things :-)

lib/src/rules/missing_conditional_import.dart
DanTup: Ah, apparently the file I duplicated as a starting point was the only one using a RecursiveAstVisitor (`missing_whitespace_between_adjacent_strings.dart`) 😄 

This makes sense and I've made those changes, thanks!

lib/src/rules/missing_conditional_import.dart
DanTup: Oof, good call - those examples were definitely bad. Updated!

lib/src/rules/missing_conditional_import.dart
bwilkerson: Thinking about it a bit more, I'd propose that we leave it as is. If we get feedback from users that a different message would be helpful we can always update it later.

lib/src/rules/missing_conditional_import.dart
bwilkerson: I'm fine with leaving it as-is. You might consider adding a comment for future reference, because the fact that it isn't checking for overlays might not occur to a future reader.

lib/src/rules/missing_conditional_import.dart
bwilkerson: That rule should definitely be updated. (But obviously not an issue for this PR.)

bwilkerson (approved):

lib/src/rules/missing_conditional_import.dart
DanTup: Done!

Would this lint be invalidated/updated correctly if the user saved a new file with the name that was used here? Would the server know if a file has `import 'file_a.dart'` but that file does not exist, that creating a file at that location may require this file to be re-analyzed/linted?

lib/src/rules/missing_conditional_import.dart
DanTup: Ah, wasn't sure if there was a reason. I'll file a PR, seems like a trivial change. Thanks!

DanTup:
Should I update pubspec.yaml to 1.16.0 too, or is it done when ready to release?

pq:
> Should I update pubspec.yaml to 1.16.0 too, or is it done when ready to release?

No need.  I'll do that when I get the release cooking.

lib/src/rules/missing_conditional_import.dart
bwilkerson: I don't know whether it _is_ the case, but it ought to be re-analyzed when a new file is added, and we run lints as part of analyzing a file.

DanTup:
Ok, cool :)

I don't have permissions to merge here (or know the procedure) so I'll leave that to one of you two when appropriate. Thanks!

PR #3083, opened 2021-12-02
`avoid_redundant_argument_values` support for `named-arguments-anywhere`

### LINKED ISSUE #3082: Update avoid_redundant_argument_values for named arguments anywhere
No lint for:
```dart
void foo(int a, int b, {bool c = true}) {}

void f() {
  foo(0, c: true, 1);
}
```

pq:
Fixes #3082

/cc @scheglov @bwilkerson

### DIFF
--- a/lib/src/rules/avoid_redundant_argument_values.dart
+++ b/lib/src/rules/avoid_redundant_argument_values.dart
@@ -71,10 +71,11 @@ class _Visitor extends SimpleAstVisitor {
     for (var i = arguments.length - 1; i >= 0; --i) {
       var arg = arguments[i];
       var param = arg.staticParameterElement;
-      if (param == null || param.hasRequired || param.isRequiredNamed) {
+      if (param == null ||
+          param.hasRequired ||
+          param.isRequiredNamed ||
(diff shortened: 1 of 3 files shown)


scheglov (approved):

lib/src/rules/avoid_redundant_argument_values.dart
```diff
@@ -71,10 +71,11 @@ class _Visitor extends SimpleAstVisitor {
     for (var i = arguments.length - 1; i >= 0; --i) {
       var arg = arguments[i];
       var param = arg.staticParameterElement;
-      if (param == null || param.hasRequired || param.isRequiredNamed) {
+      if (param == null ||
+          param.hasRequired ||
+          param.isRequiredNamed ||
+          param.isRequiredPositional) {
```
bwilkerson, line 77: We should have an `isRequired` getter on `ParameterElement`, but in the meantime you could use `!param.isOptional`.

pq: https://github.com/dart-lang/linter/pull/3086

PR #3086, opened 2021-12-02
Tidy param check

pq:
Follow-up from https://github.com/dart-lang/linter/pull/3083#pullrequestreview-822046951.

/cc @bwilkerson

### DIFF
--- a/lib/src/rules/avoid_redundant_argument_values.dart
+++ b/lib/src/rules/avoid_redundant_argument_values.dart
@@ -71,7 +71,7 @@ class _Visitor extends SimpleAstVisitor {
     for (var i = arguments.length - 1; i >= 0; --i) {
       var arg = arguments[i];
       var param = arg.staticParameterElement;
-      if (param == null || !param.isOptional) {
+      if (param == null || param.hasRequired || !param.isOptional) {
         continue;
       }
       var value = param.computeConstantValue();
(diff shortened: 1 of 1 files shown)


lib/src/rules/avoid_redundant_argument_values.dart
```diff
@@ -71,10 +71,7 @@ class _Visitor extends SimpleAstVisitor {
     for (var i = arguments.length - 1; i >= 0; --i) {
       var arg = arguments[i];
       var param = arg.staticParameterElement;
-      if (param == null ||
-          param.hasRequired ||
-          param.isRequiredNamed ||
-          param.isRequiredPositional) {
+      if (param == null || param.hasRequired || !param.isOptional) {
```
bwilkerson, line 74: Unfortunately, `hasX` methods are used for annotations, so `hasRequired` means, "has the @required annotation", which isn't what you want here. We don't have an `isRequired`, which is what you really want, but it would be the inverse of `isOptional`.

pq: Actually that *is* what I intended.  I do think we still want to handle the (now legacy) case of params annotated to be required.

bwilkerson: Oop. I saw the `hasRequired` and missed the `!param.isOptional`. :-( Yes, this looks fine.

bwilkerson (approved):

PR #3113, opened 2021-12-16
enum-like: Account for equal-valued fields if one is deprecated

### LINKED ISSUE #3017: exhaustive_cases can't handle deprecated values that redirect to other values
`exhaustive_cases` complains about switch statements on FileSystemEntityType values that don't have cases for the deprecated uppercase variants.

For example this code:
```dart
    switch (type) {
      case FileSystemEntityType.file:
	// Do not overwrite files.                                                                                                                                                                       
	throwToolExit("Invalid project name: '$projectDirPath' - file exists.",
            exitCode: 2);
	break;
      case FileSystemEntityType.link:
	// Do not overwrite links.                                                                                                                                                                       
	throwToolExit("Invalid project name: '$projectDirPath' - refers to a link.",                                                                                                                     
            exitCode: 2);                                                                                                                                                                                
        break;                                                                                                                                                                                           
      case FileSystemEntityType.directory:                                                                                                                                                               
      case FileSystemEntityType.notFound:                                                                                                                                                                
        break;                                                                                                                                                                                           
    }
```

...gets these warnings:

```
[info] Missing case clause for 'DIRECTORY' (/home/ianh/dev/flutter/packages/flutter_tools/lib/src/commands/create_base.dart:290:5)
[info] Missing case clause for 'FILE' (/home/ianh/dev/flutter/packages/flutter_tools/lib/src/commands/create_base.dart:290:5)
[info] Missing case clause for 'LINK' (/home/ianh/dev/flutter/packages/flutter_tools/lib/src/commands/create_base.dart:290:5)
[info] Missing case clause for 'NOT_FOUND' (/home/ianh/dev/flutter/packages/flutter_tools/lib/src/commands/create_base.dart:290:5)
```

...even though DIRECTORY is defined as follows:

```dart
static const DIRECTORY = directory;
```

srawlins:
# Description

This adds some logic to DartTypeUtilities.asEnumLikeClass (IMHO this is begging to be an extension on ClassElement :D ) to account for equal-valued fields.

We store each field (enum-like value) in a mapping keyed to the field's constant value. Then only need to have a case for each constant value.

Fixes #3017

### DIFF
--- a/test_data/rules/exhaustive_cases.dart
+++ b/test_data/rules/exhaustive_cases.dart
@@ -146,4 +146,17 @@ void dep(DeprecatedFields e) {
       print('baz');
       break;
   }
+
+  switch (e) {
+    // OK
+    case DeprecatedFields.oldFoo:
+      print('oldFoo');
+      break;
+    case DeprecatedFields.bar:
+      print('bar');
+      break;
+    case DeprecatedFields.baz:
+      print('baz');
+      break;
+  }
 }
--- a/test_data/rules/exhaustive_cases.dart
+++ b/test_data/rules/exhaustive_cases.dart
@@ -111,3 +111,39 @@ void ae(ActualEnum e) {
       print('e');
   }
 }
+
+// Some fields are deprecated.
+class DeprecatedFields {
+  final int i;
+  const DeprecatedFields._(this.i);
+
+  @deprecated
+  static const oldFoo = newFoo;
+  static const newFoo = DeprecatedFields._(1);
+  static const bar = DeprecatedFields._(2);
+  static const baz = DeprecatedFields._(3);
+}
+
+void dep(DeprecatedFields e) {
+  switch (e) {
+    // OK
+    case DeprecatedFields.newFoo:
+      print('newFoo');
+      break;
+    case DeprecatedFields.bar:
+      print('bar');
+      break;
+    case DeprecatedFields.baz:
+      print('baz');
+      break;
+  }
+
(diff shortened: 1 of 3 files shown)


test_data/rules/exhaustive_cases.dart
```diff
@@ -111,3 +111,39 @@ void ae(ActualEnum e) {
...
+  const DeprecatedFields._(this.i);
+
+  @deprecated
+  static const oldFoo = newFoo;
+  static const newFoo = DeprecatedFields._(1);
+  static const bar = DeprecatedFields._(2);
+  static const baz = DeprecatedFields._(3);
+}
+
+void dep(DeprecatedFields e) {
```
bwilkerson, line 127: What happens with the following:
```dart
switch (e) {
  case DeprecatedFields.oldFoo:
  case DeprecatedFields.bar:
  case DeprecatedFields.baz:
    break;
}
```
I suspect that we'll get a lint (because we'll be expecting `newFoo` but won't find it), but I don't think we ought to.

If we do get a lint, then I think we'll need to implement the more general solution from the outset. Rather than computing a `Map<DartObject, FieldElement>`, just compute a `Map<DartObject, List<FieldElement>>` and remove the key if any element in the list is found. If the map isn't empty by the end of the `switch` then report a lint.

srawlins: It would report a lint. I've refactored to account for this.

bwilkerson (approved):

PR #3118, opened 2021-12-21
Do not report prefer_function_declarations_over_variables for non-final fields

srawlins:
# Description

This rule should not be reported for fields or for top-level variables which can be re-assigned (non-final).

Fixes https://github.com/dart-lang/linter/issues/1628


bwilkerson (approved):

pq (approved):
💎

PR #3120, opened 2021-12-21
prefer_contains: short circuit more

srawlins:
# Description

prefer_contains is the slowest "recommended" rule, and the 6th slowest rule overall (according to a recent CI benchmark). This PR should speed it up. Before this PR, the broad algorithm is:

For every binary expression:

1. If the right operand is an int:
   1. and if the left operand is an invocation of `indexOf`:
      1. and the value of the int combined with the operator should be linted, report lint.
2. Else, if the left operand is an int:
   1. and the right operand is an invocation of `indexOf`:
      1. and the value of the int combined with the operator should be linted, report lint.

I decided this was inefficient in two spots:

1. The algorithm doesn't check the operator until the innermost steps, after which other more expensive computation has been done.
2. The algorithm doesn't check whether an int operand is "interesting" (0, -1, or less than -1) until the innermost step, after deciding whether `indexOf` is being invoked.

The algorithm after this PR is more like:

1. If the operator is not interesting, stop.
2. If the right operand is an int:
   1. If that int is not interesting, stop.
   2. If the left operand is an invocation of `indexOf`:
      1. and the value of the int combined with the operator should be linted, report lint.
3. Else, if the left operand is an int:
   1. If that int is not interesting, stop.
   2. and the right operand is an invocation of `indexOf`:
      1. and the value of the int combined with the operator should be linted, report lint.

### DIFF
--- a/lib/src/rules/prefer_contains.dart
+++ b/lib/src/rules/prefer_contains.dart
@@ -90,14 +90,14 @@ class _Visitor extends SimpleAstVisitor<void> {
       return;
     }
     var value = getIntValue(node.rightOperand, context);
-    if (value is int && value <= 0) {
-      if (_isUnassignedIndexOf(node.leftOperand)) {
+    if (value is int) {
+      if (value <= 0 && _isUnassignedIndexOf(node.leftOperand)) {
         _checkConstant(node, value, node.operator.type);
       }
     } else {
       value = getIntValue(node.leftOperand, context);
-      if (value is int && value <= 0) {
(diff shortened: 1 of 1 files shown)


bwilkerson (approved):

srawlins:
The timings are quite mixed, but at least in the one CI run, prefer_contains drops from 51ms to 25ms. 😁 

If I should run a benchmark before and after on my hardware, I'd be happy to do so.

pq (commented):
Nice!

PR #3121, opened 2021-12-21
Reimpliment isOnlyUnderscores with a loop

srawlins:
# Description

isOnlyUnderscores is used in a few lint rules, including non_constant_identifier_names, which is the 2nd slowest core lint rule, and 8th slowest overall (according to a recent CI benchmark).

The idea here is simple, though unvetted by me in Dart: the regular expression engine may be heavy to fire up, so instead just do a simple character walk.

### DIFF
--- a/lib/src/util/ascii_utils.dart
+++ b/lib/src/util/ascii_utils.dart
@@ -58,3 +58,18 @@ bool isValidDartFileName(String name) {
   }
   return true;
 }
+
+extension StringExtensions on String {
+  /// Returns whether `this` is just underscores.
+  bool get isJustUnderscores {
+    if (isEmpty) {
+      return false;
+    }
+    for (var i = 0; i < length; i++) {
+      if (!isUnderScore(codeUnitAt(i))) {
+        return false;
+      }
+    }
+    return false;
+  }
+}
--- a/lib/src/util/ascii_utils.dart
+++ b/lib/src/util/ascii_utils.dart
@@ -70,6 +70,6 @@ extension StringExtensions on String {
         return false;
       }
     }
-    return false;
+    return true;
   }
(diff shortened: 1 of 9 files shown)


lib/src/util/ascii_utils.dart
```diff
@@ -58,3 +58,18 @@ bool isValidDartFileName(String name) {
...
 }
+
+extension StringExtensions on String {
+  /// Returns whether `this` is just underscores.
+  bool get isJustUnderscores {
+    if (isEmpty) {
+      return false;
+    }
+    for (var i = 0; i < length; i++) {
+      if (!isUnderScore(codeUnitAt(i))) {
```
bwilkerson, line 69: I wonder whether it would be faster to test some common short versions explicitly:
```dart
switch (length) {
  case 1: return this == '_';
  case 2: return this -- '__';
  // ...
  default:
    for ...
}
```

test/ascii_utils_test.dart
```diff
@@ -23,6 +23,28 @@ void main() {
       }
     });
   });
+
+  group('isOnlyUnderscores', () {
```
bwilkerson, line 27: Consider naming the group to match the getter being tested (`isJustUnderscores`).

Also, there should be a test for an identifier starting with an underscore but followed by non-underscores (`__a`).

srawlins: Oops, done!

lib/src/util/ascii_utils.dart
srawlins: That implementation sounds great. Done.

bwilkerson (approved):

pq (approved):

PR #3122, opened 2021-12-22
Remove subclasses of LintCode

bwilkerson:
This removes unnecessary subclasses of `LintCode` by using static fields instead. I think this is a better approach to having custom messages.


pq (approved):
Nice!

Issue #3059, opened 2021-11-11
proposal: `no_logic_in_state_constructor`


goderbauer:
## Description

The constructor of a `State` object should never contain any logic. That logic should always go into `State.initState`. When people put this kind of logic into their constructor, they often run into trouble.

## Kind

Guard against errors.

## Good Examples 
```dart
class FooState extends State<Foo> with SingleTickerProviderStateMixin {
  late AnimationController _controller;

  @override
  void initState() {
  super.initState();
  _controller = AnimationController(vsync: this);
  }
  // ...
}
```
## Bad Examples

```dart
class FooState extends State<Foo> with SingleTickerProviderStateMixin {
  FooState() {
   _controller = AnimationController(vsync: this);
  }

  late AnimationController _controller;

  // ...
}
```

## Discussion

*Add any other motivation or useful context here.*

### Discussion checklist

- [x] List any existing rules this proposal complements, overlaps or conflict with.
- [x] List any relevant issues (reported here, the [SDK Tracker], or elsewhere).
- [x] If there's any prior art (e.g., in other linters), please add references here.
- [x] If this proposal corresponds to [Effective Dart] or [Flutter Style Guide] advice, please call it out. (If there isn’t any corresponding advice, should there be?)
- [x] If this proposal is motivated by real-world examples, please provide as many details as you can.  Demonstrating potential impact is _especially valuable_.

[Writing Lints]: https://github.com/dart-lang/linter/blob/master/doc/WritingLints.MD
[Effective Dart]: https://dart.dev/guides/language/effective-dart
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo]
[SDK Tracker]: https://github.com/dart-lang/sdk/issues

/cc @dnfield

dnfield:
I wonder if we can/should also lint that you don't otherwise initialize variables in `State` objects, e.g.

```dart
class FooState extends State<Foo> {
  // Should be initializing this in initState rather than here
  final MySpecialControllerThatDoesNotNeedThis _controller = MySpecialControllerThatDoesNotNeedThis(bar: 3);

  ...
}
```

goderbauer:
That seems overly limiting. It's totally fine (and common) to initialize a `TextEditingController` like that, for example.

dnfield:
Maybe if it's `late final` but if it's `final` you might be initializing an object you'll never use right?

bwilkerson:
[...]

PR #3138, opened 2022-01-10
Add `no_logic_in_state_constructor`

minhqdao:
The constructor of a `State` object is not supposed to contain any logic. That logic should go into `State.initState()`.

Closes https://github.com/dart-lang/linter/issues/3059.

### DIFF
--- a/example/all.yaml
+++ b/example/all.yaml
@@ -92,6 +92,7 @@ linter:
     - no_leading_underscores_for_library_prefixes
     - no_leading_underscores_for_local_identifiers
     - no_logic_in_create_state
+    - no_logic_in_state_constructor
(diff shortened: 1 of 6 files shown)


srawlins:
Apologies on the delayed review, but we should resolve the discussion at https://github.com/dart-lang/linter/issues/3059 before landing an implementation.

srawlins:
I'll convert this to a draft PR until the discussion is resolved.

minhqdao:
> Apologies on the delayed review, but we should resolve the discussion at #3059 before landing an implementation.

No worries. I just thought in the beginning that this was going to be very straightforward and wouldn't require much discussion.

devoncarew:

PR #3140, opened 2022-01-10
1.18.0

pq:
# 1.18.0

- extend `camel_case_types` to cover enums
- fix `no_leading_underscores_for_local_identifiers` to not 
  mis-flag field formal parameters with default values
- fix `prefer_function_declarations_over_variables` to not
  mis-flag non-final fields
- performance improvements for `prefer_contains`
- update `exhaustive_cases` to skip deprecated values that
  redirect to other values

---

/cc @srawlins 

/fyi @goderbauer

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,14 @@
+# 1.18.0
+
+- extend `camel_case_types` to cover enums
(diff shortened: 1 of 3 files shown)


srawlins (approved):
❤️

PR #3153, opened 2022-01-16
Improve unnecessary_late to target variable name

parlough:
Previously the `unnecessary_late` lint triggered on the whole variable declaration range, which was a bit confusing, particularly in a list of declarations. Now the lint reports on just the name. I've also avoided looping if the list is not marked `late`.

### DIFF
--- a/test_data/rules/unnecessary_late.dart
+++ b/test_data/rules/unnecessary_late.dart
@@ -8,9 +8,11 @@ late String unnecessaryTopLevelLate = ''; // LINT
 
 late String necessaryTopLevelLate; // OK
 
-late String unnecessaryListLateOne = '', unnecessaryListLateTwo = ''; // LINT
+late String unnecessaryListLateOne = '', // LINT
+    unnecessaryListLateTwo = ''; // LINT
 
-late String necessaryListLate, forcedNecessaryListLate = ''; // OK
+late String necessaryListLate, // OK
+    unnecessaryListLate = ''; // LINT
 
 String unnecessaryTopLevel = ''; // OK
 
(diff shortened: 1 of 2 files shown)


pq (approved):

pq:
LGTM but I'd be curious to hear if @bwilkerson has any follow-on thoughts.

Thanks!

test_data/rules/unnecessary_late.dart
```diff
@@ -8,6 +8,10 @@ late String unnecessaryTopLevelLate = ''; // LINT
 
 late String necessaryTopLevelLate; // OK
 
+late String unnecessaryListLateOne = '', unnecessaryListLateTwo = ''; // LINT
```
bwilkerson, line 11: If we go with the second purpose, it would be good to put each declaration on a separate line to make it clear that we're getting two lints here.

bwilkerson (commented):
The intent of the lint is a bit unclear. If the only purpose of the lint is to signal when the `late` keyword can be removed without changing the semantics (that is, the lint is intended to reduce the number of characters in the code), then this change is fine.

But if the intent is to flag declarations that are marked as `late` when they don't need to be in order to avoid confusion about when `late` is or isn't necessary, then I think we should continue to flag individual variables in a variable declaration list even when some of the other variables in the list are required to be declared as `late`. Personally, I think the latter is a more appropriate purpose for the lint.

If we decide on the second purpose, then we should revert portions of these changes (testing for `late` before iterating over the declarations in the list is a good optimization), and report the lint against the name of the variable.

By highlighting the name the fix can still walk up the tree to find the variable declaration list and only propose a fix if all of the declarations in the list have initializers (and if the name is associated with the first declaration in the list). I think that will likely cover almost all of the cases we'll find in practice while never suggesting an invalid fix.

parlough:
Thanks for your thoughts. I agree it is nice to catch all variables that it may be unnecessary for and reporting the lint just for the name makes it much easier to follow in a list.

What do you think about the improvements now in their current state?

bwilkerson (approved):
lgtm, thanks.

pq:
Awesome.  Thank you both!

PR #3165, opened 2022-01-19
super_parameter support for `type_init_formals`

pq:
See: #3131

/cc @bwilkerson @scheglov

### DIFF
--- a/lib/src/rules/type_init_formals.dart
+++ b/lib/src/rules/type_init_formals.dart
@@ -48,26 +48,25 @@ class TypeInitFormals extends LintRule {
   @override
   void registerNodeProcessors(
       NodeLintRegistry registry, LinterContext context) {
-    var visitor = _Visitor(this, context);
+    var visitor = _Visitor(this);
     registry.addFieldFormalParameter(this, visitor);
     registry.addSuperFormalParameter(this, visitor);
   }
 }
 
 class _Visitor extends SimpleAstVisitor<void> {
-  final LinterContext context;
   final LintRule rule;
 
-  _Visitor(this.rule, this.context);
+  _Visitor(this.rule);
 
   @override
   void visitFieldFormalParameter(FieldFormalParameter node) {
     var nodeType = node.type;
     if (nodeType == null) return;
-    var cls = node.thisOrAncestorOfType<ClassDeclaration>()?.declaredElement;
-    if (cls == null) return;
+    var paramElement = node.declaredElement;
+    if (paramElement is! FieldFormalParameterElement) return;
 
-    var field = cls.getField(node.identifier.name);
+    var field = paramElement.field;
     // If no such field exists, the code is invalid; do not report lint.
     if (field != null) {
       if (nodeType.type == field.type) {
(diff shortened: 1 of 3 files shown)


lib/src/rules/type_init_formals.dart
```diff
@@ -47,29 +48,46 @@ class TypeInitFormals extends LintRule {
...
-        if (field != null) {
-          if (nodeType.type == field.type) {
-            rule.reportLint(nodeType);
-          }
-        }
+    if (nodeType == null) return;
+    var cls = node.thisOrAncestorOfType<ClassDeclaration>()?.declaredElement;
+    if (cls == null) return;
+
+    var field = cls.getField(node.identifier.name);
```
scheglov, line 70: Check `FieldFormalParameterElement.field` instead.

lib/src/rules/type_init_formals.dart
```diff
@@ -47,29 +48,46 @@ class TypeInitFormals extends LintRule {
...
+
+  @override
+  void visitSuperFormalParameter(SuperFormalParameter node) {
+    var nodeType = node.type;
+    if (nodeType == null) return;
+    var cls = node.thisOrAncestorOfType<ClassDeclaration>()?.declaredElement;
+    if (cls == null) return;
+
+    var inheritedElement = context.inheritanceManager
+        .getInherited2(cls, Name(null, node.identifier.name));
```
scheglov, line 87: For super-formal parameters we don't care about field, the super-constructor can do whatever it want with these parameters. But we inherit the type from the parameter of the super-constructor, see `SuperFormalParameterElement.superConstructorParameter`

lib/src/rules/type_init_formals.dart
pq: Nice.  Thanks!

lib/src/rules/type_init_formals.dart
pq: Ah.  That makes sense.  Updated.  PTAL. Thanks!

pq (commented):
Thanks!

lib/src/rules/type_init_formals.dart
```diff
@@ -60,17 +62,32 @@ class _Visitor extends SimpleAstVisitor<void> {
...
-        if (field != null) {
-          if (nodeType.type == field.type) {
-            rule.reportLint(nodeType);
-          }
-        }
+    if (nodeType == null) return;
+    var paramElement = node.declaredElement;
+    if (paramElement is! FieldFormalParameterElement) return;
+
+    var field = paramElement.field;
```
scheglov, line 69: Formatting of code in this method is inconsistent with the next method.

scheglov (approved):

lib/src/rules/type_init_formals.dart
pq: Good catch.  Fixing.  Thanks!

bwilkerson (commented):
The documentation needs to be updated. I haven't been remembering to check that in other PRs, so you should go through the ones you've already landed and double check them.

PR #3166, opened 2022-01-20
super_parameter support for `avoid_annotating_with_dynamic`

pq:
See: #3131.

/cc @bwilkerson

### DIFF
--- a/lib/src/rules/avoid_annotating_with_dynamic.dart
+++ b/lib/src/rules/avoid_annotating_with_dynamic.dart
@@ -49,6 +49,7 @@ class AvoidAnnotatingWithDynamic extends LintRule {
       NodeLintRegistry registry, LinterContext context) {
     var visitor = _Visitor(this);
     registry.addSimpleFormalParameter(this, visitor);
+    registry.addSuperFormalParameter(this, visitor);
   }
 }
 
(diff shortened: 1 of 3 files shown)


lib/src/rules/avoid_annotating_with_dynamic.dart
```diff
@@ -59,7 +60,15 @@ class _Visitor extends SimpleAstVisitor<void> {
 
   @override
   void visitSimpleFormalParameter(SimpleFormalParameter node) {
-    var type = node.type;
+    _checkNode(node, node.type);
+  }
+
+  @override
+  void visitSuperFormalParameter(SuperFormalParameter node) {
```
bwilkerson, line 67: It appears that we're not visiting `FieldFormalParameter`s. I think we should either include those in the lint or not include `SuperForrmalParameter`s (probably the former). Unfortunately, visiting `FieldFormalParameter`s is a breaking change.

bwilkerson (approved):

lib/src/rules/avoid_annotating_with_dynamic.dart
pq: I wondered about that too and after sitting on it agree that we should widen current support to include `FieldFormalParameter`s.  It's breaking but I think likely to impact few folks in practice (and needless to say is I think what they would *want* the lint to do).

pq: I'll land this change as is and follow-up w/ one to add `FieldFormalParameter`s  -- in case we ever want to revert just that.

PR #3167, opened 2022-01-20
super_parameters support for `tighten_type_of_initializing_formals`

pq:
See: #3131

/cc @bwilkerson


bwilkerson (approved):
This appears to have the changes from two PRs intermixed. I think I commented on the first set separately, but I think the changes for tighten_type_of_initializeing_formals looks right.

PR #3176, opened 2022-01-24
add docs for super initializer semantics

pq:
See #3131 

/cc @bwilkerson

### DIFF
--- a/lib/src/rules/deprecated_consistency.dart
+++ b/lib/src/rules/deprecated_consistency.dart
@@ -15,8 +15,12 @@ const _details = r'''
 Do apply `@Deprecated()` consistently:
 
 - if a class is deprecated, its constructors should also be deprecated.
-- if a field is deprecated, the constructor parameter pointing to it should also be deprecated.
-- if a constructor parameter pointing to a field is deprecated, the field should also be deprecated.
+- if a field is deprecated, the constructor parameter pointing to it should also
+  be deprecated.
+- if a constructor parameter pointing to a field is deprecated, the field should
+  also be deprecated.
+- if a super initializing formal parameter is marked deprecated, the parameter
+  it forwards to in the superclass should also be deprecated.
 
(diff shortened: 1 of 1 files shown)


lib/src/rules/deprecated_consistency.dart
```diff
@@ -15,8 +15,12 @@ const _details = r'''
...
 
 - if a class is deprecated, its constructors should also be deprecated.
-- if a field is deprecated, the constructor parameter pointing to it should also be deprecated.
-- if a constructor parameter pointing to a field is deprecated, the field should also be deprecated.
+- if a field is deprecated, the constructor parameter pointing to it should also
+  be deprecated.
+- if a constructor parameter pointing to a field is deprecated, the field should
+  also be deprecated.
+- if a super initializing formal parameter is marked deprecated, the parameter
+  it forwards to in the superclass should also be deprecated.
```
bwilkerson, line 23: I was originally thinking that this last case should be allowed, but I think I've convinced myself that if you want to remove the parameter from the subclass' constructor while leaving it in the superclass' constructor that it would be better to change the parameter from being a super initializing formal parameter to being a normal formal parameter that's just ignored.

But what about the other direction:
```dart
class A {
  A({@deprecated int x});
}
class B {
  B({super.x}); // lint?
}
```
My inclination is that this should be caught by the hint that says that you're passing a value to a deprecated parameter, and not by this lint. (Investigating that idea led to https://github.com/dart-lang/sdk/issues/48206.)

bwilkerson (approved):
I had some random thoughts, but I'm happy with it as is.

srawlins:
I love this. You just have an open question, @pq

lib/src/rules/deprecated_consistency.dart
pq: Agreed.  Advice removed.  Thanks!

PR #3177, opened 2022-01-24
add super initializer examples

pq:
See: #3131 

/cc @bwilkerson

### DIFF
--- a/lib/src/rules/prefer_asserts_in_initializer_lists.dart
+++ b/lib/src/rules/prefer_asserts_in_initializer_lists.dart
@@ -12,7 +12,7 @@ import '../ast.dart';
 const _desc = r'Prefer putting asserts in initializer list.';
 
 const _details = r'''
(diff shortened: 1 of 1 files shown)


lib/src/rules/prefer_asserts_in_initializer_lists.dart
```diff
@@ -31,6 +31,31 @@ class A {
...
+**BAD:**
+```dart
+class A {
+  final int a;
+  A(this.a);
+}
+
+class B extends A {
+  B(super.a) {
+    assert(a != 0);
```
bwilkerson, line 55: I remember seeing some discussion about the scope of `a`, but can't remember now whether it's visible in the body of the constructor. We should check (if you haven't already) because the two examples might not be equivalent (that is, the `a` in the body _might_ be a reference to the getter rather than the parameter).

bwilkerson (approved):
Assuming the example is correct, lgtm.

PR #3178, opened 2022-01-24
add super initializer example

pq:
See: #3131 

/cc @bwilkerson

### DIFF
--- a/lib/src/rules/type_init_formals.dart
+++ b/lib/src/rules/type_init_formals.dart
@@ -16,8 +16,8 @@ From the [style guide](https://dart.dev/guides/language/effective-dart/style/):
 
 **DON'T** type annotate initializing formals.
 
-If a constructor parameter is using `this.x` to initialize a field, then the
-type of the parameter is understood to be the same type as the field.
+If a constructor parameter is using `this.x` or `super.x` to initialize a field,
+then the type of the parameter is understood to be the same type as the field.
 
 **GOOD:**
 ```dart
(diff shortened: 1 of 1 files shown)


lib/src/rules/type_init_formals.dart
```diff
@@ -16,8 +16,8 @@ From the [style guide](https://dart.dev/guides/language/effective-dart/style/):
 
 **DON'T** type annotate initializing formals.
 
-If a constructor parameter is using `this.x` to initialize a field, then the
-type of the parameter is understood to be the same type as the field.
+If a constructor parameter is using `this.x` or `super.x` to initialize a field,
```
bwilkerson, line 19: Kind of a nit, but `super.x` doesn't initialize a field, so this might be confusing.

bwilkerson (approved):
The example is fine, but I have a small concern with the wording. Your choice whether to change it before landing.

lib/src/rules/type_init_formals.dart
pq: Ah, let's get it right.

How about:

    If a constructor parameter is using `this.x` to initialize a field or `super.x` to forward to a super constructor,

bwilkerson: That sounds good for the first part. The "same type as the field" on the next line will also need to be updated. Once you do that it might be clearer if you just split it into two sentences, but I don't know for sure because I haven't tried writing it.

pq: I see your point.  I gave it a shot in two sentences.  PTAL.

bwilkerson (approved):
That lgtm.

PR #3180, opened 2022-01-24
super param support for `library_private_types_in_public_api`

pq:
See: #3131

/cc @bwilkerson


bwilkerson (commented):
What about the following case:
```dart
class _O extends Object {}
class _A {
  _A(_O o);
}
class B extends _A {
  B(super.o); // lint?
}
```
My expectation is that this would also be flagged, even though the type of the parameter is implicit rather than explicit, but I don't think the code as written will catch it.

lib/src/rules/library_private_types_in_public_api.dart
```diff
@@ -209,6 +209,27 @@ class Validator extends SimpleAstVisitor<void> {
...
+    }
+
+    // Check for a declared type.
+    var type = node.type;
+    if (type != null) {
+      type.accept(this);
+      return;
+    }
+
+    // Check implicit type.
```
bwilkerson, line 225: Do we also need to check the implicit type in a `FieldFormalParameter`? For example:

```dart
class C {
  _O _x;

  C(this._x);
```

bwilkerson (approved):

lib/src/rules/library_private_types_in_public_api.dart
pq: Funny, I thought about that and then convinced myself that the declaration would be checked but that's not right, so yes!

Issue #3095, opened 2021-12-06
enhanced enums support for `avoid_renaming_method_parameters`


pq:

pq:
```dart
  test_rename() async {
    await assertDiagnostics(r'''
class C {
  int f(int f) => f;
}
enum A extends C {
  a,b,c;
  @override
  int f(int x) => x;
}
''', [
      lint('avoid_renaming_method_parameters', 28, 8), // todo(pq): update index
    ]);
  }
```

Tip-of-trunk fails w/:

```
===== asynchronous gap ===========================
test/rules/avoid_renaming_method_parameters.dart 26:5  AvoidRenamingMethodParametersTest.test_rename

Expected but did not find:
  avoid_renaming_method_parameters [28, 8]

Found but did not expect:
  HintCode.UNUSED_ELEMENT [79, 1, The declaration 'f' isn't referenced.]
  ParserErrorCode.MISSING_IDENTIFIER [57, 1, Expected an identifier.]
  CompileTimeErrorCode.DUPLICATE_DEFINITION [48, 1, The name 'C' is already defined.]
  ParserErrorCode.MISSING_IDENTIFIER [55, 1, Expected an identifier.]
  ParserErrorCode.MISSING_ENUM_BODY [40, 7, An enum definition must have a body with at least one constant name.]
  ParserErrorCode.EXPECTED_TOKEN [40, 7, Expected to find ';'.]
  ParserErrorCode.MISSING_CONST_FINAL_VAR_OR_TYPE [40, 7, Variables must be declared using the keywords 'const', 'final', 'var' or a type name.]
  CompileTimeErrorCode.UNDEFINED_IDENTIFIER [58, 1, Undefined name 'c'.]
  ParserErrorCode.UNEXPECTED_TOKEN [55, 1, Unexpected text ';'.]
  ParserErrorCode.EXPECTED_TOKEN [54, 1, Expected to find ';'.]
  CompileTimeErrorCode.UNDEFINED_IDENTIFIER [56, 1, Undefined name 'b'.]
  ParserErrorCode.MISSING_FUNCTION_PARAMETERS [48, 1, Functions must have an explicit list of parameters.]
  ParserErrorCode.EXPECTED_IDENTIFIER_BUT_GOT_KEYWORD [40, 7, 'extends' can't be used as an identifier because it's a keyword.]
  CompileTimeErrorCode.UNDEFINED_IDENTIFIER [54, 1, Undefined name 'a'.]
[...]

PR #3188, opened 2022-01-27
enhanced enum tests for `avoid_renaming_method_parameters`

pq:
See: #3095 

/cc @bwilkerson

### DIFF
--- a/test/rules/avoid_renaming_method_parameters.dart
+++ b/test/rules/avoid_renaming_method_parameters.dart
@@ -28,7 +28,7 @@ class AvoidRenamingMethodParametersTest extends LintRuleTest {
 class C {
   int f(int f) => f;
 }
-enum A extends C {
+enum A implements C {
   a,b,c;
   @override
   int f(int x) => x;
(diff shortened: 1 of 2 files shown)


test/rules/avoid_renaming_method_parameters.dart
```diff
@@ -0,0 +1,40 @@
...
+  @override
+  String get lintRule => 'avoid_renaming_method_parameters';
+
+  @FailingTest(issue: 'https://github.com/dart-lang/linter/issues/3095')
+  test_rename() async {
+    await assertDiagnostics(r'''
+class C {
+  int f(int f) => f;
+}
+enum A extends C {
```
bwilkerson, line 31: The [enhanced enum spec](https://github.com/dart-lang/language/blob/master/working/0158%20-%20Enhanced%20Enum/proposal.md) allows an enum to implement an interface, but doesn't allow it to extend another class (because they all implicitly extend `Enum`). So this test case is invalid.

pq: Aha!  Thanks for that clarification.  Updating to use `implements`.

pq:
Test updated (but still failing due to an unexpected `HintCode.OVERRIDE_ON_NON_OVERRIDING_METHOD` diagnostic).

/fyi @scheglov

bwilkerson (approved):

PR #3215, opened 2022-02-02
Add improved text for type_init_formals

### LINKED ISSUE #3214: The `type_init_formals` documentation fails to mention that a parameter type that differs from the inferred one is not flagged
The `type_init_formals` lint is [documented](https://dart-lang.github.io/linter/lints/type_init_formals.html) to flag any initializing formal that has an explicitly declared type, for instance:

```dart
class A {
  int x;
  A(int this.x); // Lint, we'd get the same type by inference anyway.
}
```

However, this is not helpful in the case where the declared type differs from the one that would be inferred. For example:

```dart
class Either {
  // Invariant: one of these is null, the other is non-null.
  final int? x, y;

  // Constructors enforce the invariant.
  Either.x(int this.x) : y = null;
  Either.y(int this.y) : x = null;

  // Various methods may now be written under the assumption that the invariant holds.
  ...
}
```

The lint does not flag the latter example, and I believe this is exactly as it should be. But the documentation should be adjusted to say so.

srawlins:
# Description

Add text to address #3214

Fixes #3214

It appears that the spelling "OK" is the most common in this package.

### DIFF
--- a/lib/src/rules/type_init_formals.dart
+++ b/lib/src/rules/type_init_formals.dart
@@ -22,6 +22,9 @@ a constructor parameter is using `super.x` to forward to a super constructor,
 then the type of the parameter is understood to be the same as the super
 constructor parameter.
 
+Type annotating an initializing formal with a different type than that of the
+field is OK.
+
 **GOOD:**
 ```dart
 class Point {


bwilkerson (approved):

pq (approved):
Thanks!

PR #3225, opened 2022-02-10
Check full resolved paths for avoid_relative_lib_imports

### LINKED ISSUE #2419: avoid_relative_lib_imports should check resolved URIs (not literal uri content)
This check is too limited.

```dart
      final uri = Uri.parse(node.uriContent);
      if (uri.scheme.isEmpty) {
        return uri.path.contains('/lib/');
      }
```

and only flags cases like `import '../lib/baz.dart'` and misses cases where 'lib' is not in the import uri content but is in the file's URI.

DanTup:
@pq @bwilkerson while testing a server fix, I bumped into #2419 and thought it might be a simple thing to try to fix. I've ended up with some questions though:

- should we be concerned about the number of new lint warnings this change would introduce? (I have no data, but it seems like it was very easy to accidentally not follow this lint)
- did I correctly move the test from integration_tests to rules?
- is it correct/valid to use `node.uriContent` as a fallback if `node.uriSource` is null (which I presume occurs when the specified file doesn't exist)
- why do I get `URI_DOES_NOT_EXIST` for some of these tests but not others (for example `test_relativePath_resolvesToLib` does not raise it suggesting the creation of `other.dart` was successful, but `test_relativePath_explicitLib` basically does the same, but using `../lib/other.dat` instead of just `other.dart` and triggers this error)
- [ ] changelog will need updating if this does go ahead 

Fixes #2419.


test/rules/avoid_relative_lib_imports.dart
```diff
@@ -0,0 +1,74 @@
...
+    var filePath = '$testPackageRootPath/lib/other.dart';
+    newFile(filePath);
+    await assertDiagnostics(
+      '''
+// ignore: unused_import
+import '$filePath';
+''',
+      [
+        lint('avoid_relative_lib_imports', 32, filePath.length + 2),
+        // TODO(dantup): Why???
```
bwilkerson, line 40: I'm not sure, but I will point out that the file path is an absolute path, so it's possible that we aren't handling that case correctly. @scheglov would probably know.

test/rules/avoid_relative_lib_imports.dart
```diff
@@ -0,0 +1,74 @@
...
+    var filePath = '$testPackageRootPath/lib/other.dart';
+    newFile(filePath);
+    await assertDiagnostics(
+      '''
+// ignore: unused_import
+import '../lib/other.dart';
+''',
+      [
+        lint('avoid_relative_lib_imports', 32, 19),
+        // TODO(dantup): Why???
```
bwilkerson, line 56: This is another odd case. I don't remember the justification for it, but I believe that analyzer rejects paths inside `lib` that escape `lib`, even if they re-enter `lib`. @scheglov would remember.

bwilkerson (commented):
> should we be concerned about the number of new lint warnings this change would introduce?

Good question. I don't have any idea how many users are using this lint, so I don't know what the impact is likely to be. But given how poor the checking was before, and how few issues I've seen opened against it, I'm going to guess that there are very few users.

> did I correctly move the test from integration_tests to rules?

lgtm

> is it correct/valid to use node.uriContent as a fallback if node.uriSource is null (which I presume occurs when the specified file doesn't exist)

Sometimes. It should catch relative imports in libraries outside of `lib`, but won't catch relative imports in libraries inside `lib`. IIRC, we want to catch both, in which case we'll probably need to have separate logic for the two cases. (For libraries outside of `lib`, checking the `uriContent` should be sufficient, without needing to look at the `uriSource` first.)

> why do I get URI_DOES_NOT_EXIST for some of these tests ...

I've added comments in-line.

test/rules/avoid_relative_lib_imports.dart
scheglov: We don't do this intentionally, but using `Uri.resolveUri()` does following:
1. When we resolve `package:foo/../lib/bar.dart`, we lose the package name, and get `package:./lib/bar.dart`, which cannot be resolved.
2. When we resolve `../lib/bar.dart` in `package:foo/foo.dart`, we get `package:foo/lib/bar.dart`, which might exist (if you have `lib` in `lib`), but usually is not.

DanTup: Ah, that explains this one. What should I do here? Should I add an `// ignore` for it, and a comment along those lines?

Any ideas why the same occurs for `test_absolutePath` above which is a full path (`/home/test/lib/other.dart`)?

DanTup:
@bwilkerson 

> It should catch relative imports in libraries outside of `lib`, but won't catch relative imports in libraries inside `lib`.

Isn't that what test `test_relativePath_resolvesToLib` is doing? It imports `"other.dart"` (from inside the lib folder). Or do you mean things that specifically do `../lib/other.dart`?

bwilkerson:
I was thinking there might still a hole if we have an import like this:
```dart
import '../other.dart';
```
but now I'm thinking that must be fine, or we would have gotten reports about it.

lib/src/rules/avoid_relative_lib_imports.dart
```diff
@@ -67,7 +68,8 @@ class _Visitor extends SimpleAstVisitor<void> {
     if (uriContent != null) {
       var uri = Uri.tryParse(uriContent);
       if (uri != null && uri.scheme.isEmpty) {
-        return uri.path.contains('/lib/');
+        var fullPath = node.uriSource?.fullName ?? uri.path;
+        return path.split(fullPath).contains('lib');
```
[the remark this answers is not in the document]
bwilkerson: I believe that the linter is always given a resolved AST, so it shouldn't be `null`. However, it might not be worth the risk.

@scheglov Why does the `LinterContext` expose `LinterContextUnit` instead of just exposing the `ResolvedLibraryResult`? It would solve this problem, and lint authors would learn the public API of the analyzer rather than this linter-specific API.

DanTup:
@pq @bwilkerson unrelated to this change - is there a reason the files in `test/rules/` do not end with `_test.dart`? In VS Code we use `_test.dart` as a signal that the files are tests - without that I can't just press `F5` to run the tests in the current file (nor will they show test-level CodeLens or be discovered in the test explorer). I'm now wondering if that rule is correct? 🤔

test/rules/avoid_relative_lib_imports.dart
DanTup: I haven't figured out where to pause to track down what's happening here properly, but this looks very strange. The Import string is an absolute path, but the source it's looking for looks like it's been interpreted relative to the package root or something:

Screenshot 2022-02-10 at 18 32 27

That makes me wonder whether this test is valid - do we expected to support absolute paths like this? (I thought we did, but upon reflection, it seems like a very odd thing for someone to write).

lib/src/rules/avoid_relative_lib_imports.dart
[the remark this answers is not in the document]
DanTup: I'm not very familiar with this - is adding `resourceProvider` to `LinterContext` something that needs to be done in the SDK (in package:analyzer) first?

DanTup:
> 🖐️ This is a core recommended lint so I think the impact may be significant. (When I did an analysis of pub packages recently, I saw it defined in 95% of projects that had options enabled; moreover it's used for pub package scoring.) TL;DR we should tread carefully!

Oof, that sounds like it complicate things. Given the example in the lint is `../lib` I think it could be interpreted to only mean from outside of the lib folder (and that's how it seemed to behave). Would it be better to have a separate lint for within lib (`avoid_relative_imports_in_lib` 🙃)?

Hangon.. I was led here while making a fix for [`always_use_package_imports`](https://dart-lang.github.io/linter/lints/always_use_package_imports.html)... that sounds a lot like the lint I just described.  Is it certain #2419 is an issue to be solved by this lint, and not that one?

lib/src/rules/avoid_relative_lib_imports.dart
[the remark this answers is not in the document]
bwilkerson: Yes.

DanTup:
@pq @bwilkerson any thoughts on [this comment above](https://github.com/dart-lang/linter/pull/3225#issuecomment-1035351198). I checked the implementation of `always_use_package_imports` and I think it handles what I was trying to add here:

- `always_use_package_imports` makes sure all imports inside `lib` are `package:`
- `avoid_relative_lib_imports` makes sure no imports contain `/lib/`

It seems like together (without these changes), they may cover everything. The former is for things _inside_ `lib` and the latter for things _outside_ of `lib`.

pq:
💯 to  @bwilkerson's proposal!

DanTup:
sgtm - do tweaks are at https://github.com/dart-lang/linter/pull/3231.

PR #3228, opened 2022-02-11
add `_test` suffix to rule tests

pq:
/cc @bwilkerson 

/fyi @DanTup


bwilkerson (approved):
Interesting that we weren't being consistent. Good to see that we are now.

DanTup:
Cool, thanks! 🙂

PR #3229, opened 2022-02-11
skip `main` in `avoid_void_async`

### LINKED ISSUE #3141: proposal: update `avoid_void_async` to ignore main
## Proposal: update `avoid_void_async` to ignore `main`

As per https://github.com/dart-lang/site-www/pull/3739#issuecomment-1009295663, update `avoid_void_async` to ignore main.

## Details

As a general rule, `main` does not need to be annotated as returning a `Future<void>` since it's generally not intended to be invoked and awaited.  (Test suites are a common counter-example but still not that common.)  Requiring the annotation is just noise and contradicts idiomatic Dart.

## Kind

**Style**

## Good Examples 

```dart
void main() async {
  await printOrderMessage();
}

Future<void> printOrderMessage() async {
   ...
}
```

## Bad Examples

```dart
Future<void> main() async {
  await printOrderMessage();
}

Future<void> printOrderMessage() async {
   ...
}
```

pq:
Fixes #3141

/cc @bwilkerson


lib/src/rules/avoid_void_async.dart
```diff
@@ -29,6 +29,19 @@ Future<void> f() async {}
 Future<void> f2() async => null;
 ```
 
+**EXCEPTION**
+
+An exception is made for `main`, where the `Future` annotation *can* (and
+generally should) be dropped in favor of `void`.
```
parlough, line 35: Maybe somehow clarify this only applies to top-level `main` functions?

```suggestion
An exception is made for top-level `main` functions, 
where the `Future` annotation *can* (and generally should) 
be dropped in favor of `void`.
```

pq: Good catch.  Will do!

bwilkerson (approved):

PR #3231, opened 2022-02-14
Add a note above how always_use_package_imports and avoid_relative_lib_imports complement each other

DanTup:
See https://github.com/dart-lang/linter/pull/3225#issuecomment-1036306676.

@pq @bwilkerson

### DIFF
--- a/lib/src/rules/always_use_package_imports.dart
+++ b/lib/src/rules/always_use_package_imports.dart
@@ -18,8 +18,9 @@ that is to ensure you consistently use absolute imports for files withing the
 `lib/` directory.
 
 This is the opposite of 'prefer_relative_imports'.
-Might be used with 'avoid_relative_lib_imports' to avoid relative imports of
-files within `lib/` directory outside of it. (for example `test/`)
+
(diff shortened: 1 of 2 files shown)


lib/src/rules/always_use_package_imports.dart
```diff
@@ -18,8 +18,9 @@ that is to ensure you consistently use absolute imports for files withing the
 `lib/` directory.
 
 This is the opposite of 'prefer_relative_imports'.
-Might be used with 'avoid_relative_lib_imports' to avoid relative imports of
-files within `lib/` directory outside of it. (for example `test/`)
+
+You can also use 'avoid_relative_lib_imports' to disallow relative imports of
+files within `lib/` directory outside of it (for example `test/`).
```
DanTup, line 23: This one already referenced the other, though I tweaked the text slightly as the wording seemed a bit ambiguous ("Might be used with B to do A" sounded like A was a description of what this lint would do, not what B does), though I'm happy to tweak if someone can word it better.

bwilkerson (approved):

Issue #3094, opened 2021-12-06
enhanced enums support for `avoid_equals_and_hash_code_on_mutable_classes`


pq:

pq:
Enums are constant so we'll just want the negative test.

PR #3233, opened 2022-02-15
enhanced enum support for `avoid_equals_and_hash_code_on_mutable_classes`

pq:
See: https://github.com/dart-lang/linter/issues/3094

/cc @bwilkerson


pq:
Updated to add a test to ensure we don't start reporting false positives down the road...

/cc @bwilkerson

bwilkerson (approved):

PR #3234, opened 2022-02-15
enhanced enum support for `avoid_setters_without_getters`

### LINKED ISSUE #3205: enhanced enum support for `avoid_setters_without_getters`

pq:
Fixes #3205

/cc @bwilkerson

### DIFF
--- a/lib/src/rules/avoid_setters_without_getters.dart
+++ b/lib/src/rules/avoid_setters_without_getters.dart
@@ -63,6 +63,7 @@ class AvoidSettersWithoutGetters extends LintRule {
       NodeLintRegistry registry, LinterContext context) {
     var visitor = _Visitor(this);
     registry.addClassDeclaration(this, visitor);
+    registry.addEnumDeclaration(this, visitor);
   }
 }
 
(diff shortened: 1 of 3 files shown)


lib/src/rules/avoid_setters_without_getters.dart
```diff
@@ -63,6 +63,7 @@ class AvoidSettersWithoutGetters extends LintRule {
       NodeLintRegistry registry, LinterContext context) {
     var visitor = _Visitor(this);
     registry.addClassDeclaration(this, visitor);
+    registry.addEnumDeclaration(this, visitor);
```
bwilkerson, line 66: Should we also be looking at mixin declarations?

bwilkerson (approved):
The idea of having a setter on an enum is very strange. I haven't yet thought of a good use case for it. But given that you can, I think this change makes sense.

lib/src/rules/avoid_setters_without_getters.dart
pq: Maybe?  I added a todo since this could introduce new diagnostics and we'd want to do a little vetting of impact first.

Issue #3201, opened 2022-01-27
enhanced enums support for `use_late_for_private_fields_and_variables`


pq:

pq:
Invalid.

PR #3235, opened 2022-02-15
enum test for `use_late_for_private_fields_and_variables`

pq:
See: #3201

/cc @bwilkerson


bwilkerson (commented):
As discussed, I don't think this is valid.

PR #3241, opened 2022-02-22
Improve description & example for avoid_print

### LINKED ISSUE #2787: `avoid_print` should have an explanation
Usually lints provide an explanation why particular code should be avoided.

https://dart-lang.github.io/linter/lints/avoid_print.html `avoid_print` should have such an explanation.
Plus perhaps an alternative, i.e. `good` section, for flutter it could contain `debugPrint` instead

Merrit:
# Description

The existing message gave no hint on what to do *instead* of using
print calls, so this gives an explanation and a few examples.

Fixes #2787

### DIFF
--- a/lib/src/rules/avoid_print.dart
+++ b/lib/src/rules/avoid_print.dart
@@ -14,10 +14,42 @@ const _desc = r'Avoid `print` calls in production code.';
 const _details = r'''
 **DO** avoid `print` calls in production code.
 
+For production code use `debugPrint`, a logging framework, or surround `print`
+calls with a check for `kDebugMode`.
+
 **BAD:**
 ```dart
 void f(int x) {
   print('debug: $x');
+  ...
+}
+```
+
+
+**GOOD:**
+```dart
+void f(int x) {
+  debugPrint('debug: $x');
+  ...
+}
+```
+
+
+**GOOD:**
+```dart
+void f(int x) {
+  log('log: $x');
+  ...
+}
+```
+
+
+**GOOD:**
+```dart
+void f(int x) {
+  if (kDebugMode) {
+      print('debug: $x');
+  }
   ...
 }
 ```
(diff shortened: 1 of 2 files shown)


Merrit:
The instructions say to ["name a reviewer"](https://github.com/dart-lang/linter/blob/master/CONTRIBUTING.md?plain=1#L62), but without further context I wouldn't know how to do so..

lib/src/rules/avoid_print.dart
```diff
@@ -14,13 +14,45 @@ const _desc = r'Avoid `print` calls in production code.';
 const _details = r'''
 **DO** avoid `print` calls in production code.
 
+For production code use `debugPrint`, a logging framework, or surround `print`
+calls with a check for `kDebugMode`.
```
parlough, line 18: I like the idea of mentioning these alternatives, but these only exist for Flutter users and a Dart developer may be confused when they cannot find these.

Maybe we can make that explicit somehow?

```suggestion
For production code, consider using a logging framework.
If you are using Flutter, you can also use use `debugPrint`
or surround `print` calls with a check for `kDebugMode`
```

parlough:
> The instructions say to ["name a reviewer"](https://github.com/dart-lang/linter/blob/master/CONTRIBUTING.md?plain=1#L62), but without further context I wouldn't know how to do so..

It's okay to not mention a review if you don't need a specific reviewer. @pq will review the change or assign someone :)

lib/src/rules/avoid_print.dart
Merrit: Very good point, looks better too! :+1:

bwilkerson (approved):

pq:
This looks great.  Could you add yourself to [AUTHORS](https://github.com/dart-lang/linter/blob/master/AUTHORS) while you're at it?

Merrit:
> This looks great. Could you add yourself to [AUTHORS](https://github.com/dart-lang/linter/blob/master/AUTHORS?rgh-link-date=2022-02-22T23%3A04%3A02Z) while you're at it?

Done :smile:

pq (changes requested):
thanks!

💎

pq (approved):

Merrit:
Thank you! :tada:

pq:
Of course!  Thanks for your contribution!

PR #3247, opened 2022-02-24
enhanced enum support for `sort_constructors_first`

### LINKED ISSUE #3207: enhanced enum support for `sort_constructors_first`

pq:
Fixes #3207

/cc @bwilkerson

### DIFF
--- a/lib/src/rules/sort_constructors_first.dart
+++ b/lib/src/rules/sort_constructors_first.dart
@@ -46,6 +46,7 @@ class SortConstructorsFirst extends LintRule {
       NodeLintRegistry registry, LinterContext context) {
     var visitor = _Visitor(this);
     registry.addClassDeclaration(this, visitor);
+    registry.addEnumDeclaration(this, visitor);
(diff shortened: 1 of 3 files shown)


lib/src/rules/sort_constructors_first.dart
```diff
@@ -71,4 +71,14 @@ class _Visitor extends SimpleAstVisitor<void> {
...
   }
+
+  @override
+  void visitClassDeclaration(ClassDeclaration node) {
+    check(node.members);
+  }
+
+  @override
+  void visitEnumDeclaration(EnumDeclaration node) {
+    check(node.members);
```
bwilkerson, line 82: I'm guessing that this means that enum constants aren't included in `members`.

bwilkerson (approved):

lib/src/rules/sort_constructors_first.dart
pq: Yep.

PR #3248, opened 2022-02-24
enhanced enum support for `sort_unnamed_constructors_first`

### LINKED ISSUE #3208: enhanced enums support for `sort_unnamed_constructors_first`

pq:
Fixes #3208

/cc @bwilkerson

### DIFF
--- a/lib/src/rules/sort_unnamed_constructors_first.dart
+++ b/lib/src/rules/sort_unnamed_constructors_first.dart
@@ -46,6 +46,7 @@ class SortUnnamedConstructorsFirst extends LintRule {
       NodeLintRegistry registry, LinterContext context) {
     var visitor = _Visitor(this);
     registry.addClassDeclaration(this, visitor);
+    registry.addEnumDeclaration(this, visitor);
   }
 }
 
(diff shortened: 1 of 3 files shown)


lib/src/rules/sort_unnamed_constructors_first.dart
```diff
@@ -54,10 +55,9 @@ class _Visitor extends SimpleAstVisitor<void> {
 
   _Visitor(this.rule);
 
-  @override
-  void visitClassDeclaration(ClassDeclaration node) {
+  void check(NodeList<ClassMember> classMembers) {
     // Sort members by offset.
-    var members = node.members.toList()
+    var members = classMembers.toList()
       ..sort((ClassMember m1, ClassMember m2) => m1.offset - m2.offset);
```
bwilkerson, line 61: I didn't think of this on the last PR, but I'm kind of surprised that we need to sort the members. I thought we already returned them in source order. Not critical, but it would improve performance to not make a pass confirming that everything is already sorted.

bwilkerson (approved):

lib/src/rules/sort_unnamed_constructors_first.dart
pq: Interesting!  I've made a note to loop back and check on this.  Thanks!

albertms10: Do you think it may be interesting to use a `SplayTreeSet` instead to keep always the order within the data structure?

PR #3253, opened 2022-02-26
+ use_enums

### LINKED ISSUE #3173: consider a `prefer_enums` lint

pq:
I'd consider this still a WIP but ready for initial feedback.

/cc @bwilkerson 

Fixes #3173

### DIFF
--- a/lib/src/rules/prefer_enums.dart
+++ b/lib/src/rules/prefer_enums.dart
@@ -1,0 +1,254 @@
+// Copyright (c) 2022, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/analysis/features.dart';
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/element.dart';
+
+import '../analyzer.dart';
+import '../ast.dart';
+import '../util/dart_type_utilities.dart';
+
+const _desc = r'Prefer enums over classes that behave like enums.';
+
+const _details = r'''
+Classes that look like enumerations should be declared as `enum`s.
+
+**DO** use enums where appropriate.
+
+Candidates for enums are classes that:
+  * are concrete,
+  * are private or have only private generative constructors,
+  * have two or more static const fields with the same type as the class,
+  * have generative constructors that are only invoked at the top-level of the
+    initialization expression of these static fields,
+  * do not define `hashCode`, `==` or `index`,
+  * do not extend any class beside Object, and
+  * have no subclasses declared in the defining library.
+
+**BAD:**
+```dart
+class LogPriority {
+  static const error = LogPriority._(1, 'Error');
+  static const warning = LogPriority._(2, 'Warning');
+  static const unknown = LogPriority._unknown(2, 'Log');
+
+  final String prefix;
+  final int priorty;
+  const LogPriority._(this.priority, this.prefix);
+  const LogPriority._unknown(String prefix) : this._(-1, prefix);
+}
+```
+
+**GOOD:**
+```dart
+enum LogPriority {
+  error(1, 'Error'),
+  warning(2, 'Warning'),
+  log.unknown('Log');
+ 
+  final String prefix;
+  final int priorty;
+  LogPriority(this.priority, this.prefix);
+  LogPriority.unknown(String prefix) : this(-1, prefix);
+}
+```
+''';
+
+class PreferEnums extends LintRule {
+  PreferEnums()
+      : super(
+            name: 'prefer_enums',
+            description: _desc,
+            details: _details,
+            group: Group.style);
+
+  @override
+  void registerNodeProcessors(
+      NodeLintRegistry registry, LinterContext context) {
+    if (!context.isEnabled(Feature.enhanced_enums)) return;
+
+    // Skip contexts w/ parts.
+    if (context.allUnits.length > 1) return;
+
+    var visitor = _Visitor(this, context);
+    registry.addClassDeclaration(this, visitor);
+  }
+}
+
+/// A superclass for the [_EnumVisitor] and [_NonEnumVisitor].
+class _BaseVisitor extends RecursiveAstVisitor<void> {
+  /// The element representing the enum declaration that's being visited.
+  final ClassElement classElement;
+
+  _BaseVisitor(this.classElement);
+
+  /// Return `true` if the given [node] is an invocation of a generative
+  /// constructor from the class being converted.
+  bool invokesGenerativeConstructor(InstanceCreationExpression node) {
+    var constructorElement = node.constructorName.staticElement;
+    return constructorElement != null &&
+        !constructorElement.isFactory &&
+        constructorElement.enclosingElement == classElement;
+  }
+}
+
+/// A visitor used to visit the class being linted. This visitor throws an
+/// exception if a constructor for the class is invoked anywhere other than the
+/// top-level expression of an initializer for one of the fields being converted.
+///
+/// Note: copied and modified from server.
+class _EnumVisitor extends _BaseVisitor {
+  /// A flag indicating whether we are currently visiting the children of a
+  /// field declaration that will be converted to be a constant.
+  bool inConstantDeclaration = false;
+
+  List<VariableDeclaration> variableDeclarations;
+
+  _EnumVisitor(ClassElement classElement, this.variableDeclarations)
+      : super(classElement);
+
+  @override
+  void visitInstanceCreationExpression(InstanceCreationExpression node) {
+    if (!inConstantDeclaration) {
+      if (invokesGenerativeConstructor(node)) {
+        throw _InvalidEnumException();
+      }
+    }
+    inConstantDeclaration = false;
+    super.visitInstanceCreationExpression(node);
+  }
+
+  @override
+  void visitVariableDeclaration(VariableDeclaration node) {
+    if (variableDeclarations.contains(node)) {
+      inConstantDeclaration = true;
+    }
+    super.visitVariableDeclaration(node);
+    inConstantDeclaration = false;
+  }
+}
+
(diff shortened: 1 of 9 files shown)


lib/src/rules/prefer_enums.dart
```diff
@@ -0,0 +1,254 @@
...
+    }
+
+    if (candidateConstants.length < 2) return;
+    if (DartTypeUtilities.hasSubclassInDefiningCompilationUnit(classElement)) {
+      return;
+    }
+
+    try {
+      node.accept(_EnumVisitor(classElement, candidateConstants));
+    } on _InvalidEnumException {
```
pq, line 242: I preserved this approach from the assist but expect that it would be more performant to use a flag rather than a thrown exception to short-circuit the common negative case.

@bwilkerson: wdyt?  Should I go with my gut and refactor?

lib/src/rules/prefer_enums.dart
```diff
@@ -0,0 +1,254 @@
...
+import 'package:analyzer/dart/analysis/features.dart';
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/element.dart';
+
+import '../analyzer.dart';
+import '../ast.dart';
+import '../util/dart_type_utilities.dart';
+
+const _desc = r'Prefer enums over classes that behave like enums.';
```
pq, line 14: Wordsmithing here and below most welcome.  (I consider this text to be a placeholder at best.)

lib/src/rules/prefer_enums.dart
```diff
@@ -0,0 +1,254 @@
...
+
+Candidates for enums are classes that:
+  * are concrete,
+  * are private or have only private generative constructors,
+  * have two or more static const fields with the same type as the class,
+  * have generative constructors that are only invoked at the top-level of the
+    initialization expression of these static fields,
+  * do not define `hashCode`, `==` or `index`,
+  * do not extend any class beside Object, and
+  * have no subclasses declared in the defining library.
```
pq, line 29: It occurs to me that it'd probably be good to explain *why* we're restricting cases to things we can verify without whole-world analysis.  Maybe a footnote?  (Thoughts on how best to say this? I wonder if we shouldn't have some boilerplate for this or an FAQ or something since we make similar compromises elsewhere that could benefit from documentation...)

lib/src/rules/prefer_enums.dart
```diff
@@ -0,0 +1,254 @@
...
+  error(1, 'Error'),
+  warning(2, 'Warning'),
+  log.unknown('Log');
+ 
+  final String prefix;
+  final int priorty;
+  LogPriority(this.priority, this.prefix);
+  LogPriority.unknown(String prefix) : this(-1, prefix);
+}
+```
```
pq, line 57: More real motivating examples would be cool here I think.

@eernstg?  @lrhn? @munificent?

pq:
@bwilkerson: one place this lint is less smart than the assist is in ignoring the case where `index` is defined and used in a migratable way.  I'm not sure how important that case is?  Thoughts welcome!

lib/src/rules/prefer_enums.dart
```diff
@@ -0,0 +1,254 @@
...
+
+**DO** use enums where appropriate.
+
+Candidates for enums are classes that:
+  * are concrete,
+  * are private or have only private generative constructors,
+  * have two or more static const fields with the same type as the class,
+  * have generative constructors that are only invoked at the top-level of the
+    initialization expression of these static fields,
+  * do not define `hashCode`, `==` or `index`,
```
pq, line 27: So... technically `index` might be OK if it's being used in a migratable way.  The open-question is whether this is something the lint should be aware of.

lib/src/rules/prefer_enums.dart
```diff
@@ -0,0 +1,254 @@
...
+import '../ast.dart';
+import '../util/dart_type_utilities.dart';
+
+const _desc = r'Prefer enums over classes that behave like enums.';
+
+const _details = r'''
+Classes that look like enumerations should be declared as `enum`s.
+
+**DO** use enums where appropriate.
+
```
pq, line 20: Maybe a bit explaining the enhanced-enums feature here?

lib/src/rules/prefer_enums.dart
```diff
@@ -0,0 +1,254 @@
...
+      : super(classElement);
+
+  @override
+  void visitInstanceCreationExpression(InstanceCreationExpression node) {
+    if (!inConstantDeclaration) {
+      if (invokesGenerativeConstructor(node)) {
+        throw _InvalidEnumException();
+      }
+    }
+    inConstantDeclaration = false;
```
bwilkerson, line 120: If there are multiple fields that could be constants in the same variable declaration, such as
```
static const E a = E('a'), b = E('b');
```
then this will set the flag `false` for the first field and not restore it prior to visiting the second field.

This isn't a problem in the assist because I decided to simplify the initial implementation so that we only convert the class if the fields to be converted to constants are in separate field declarations. That determination is done before we get to the visitor, so I don't have to handle it there.

This needs to be rewritten in order to support the case above to something like:
```
var wasInConstantDeclaration = inConstantDeclaration;
inConstantDeclaration = false;
super.visitInstanceCreationExpression(node);
inConstantDeclaration = wasInConstantDeclaration;
```

bwilkerson (commented):
I haven't looked at the tests, but I thought I'd send out some initial thoughts.

lib/src/rules/prefer_enums.dart
pq: Yeah.  I like `use_enums` a bit more.  Updated.  Thanks!

lib/src/rules/prefer_enums.dart
pq: I'll add a TODO.

lib/src/rules/prefer_enums.dart
```diff
@@ -0,0 +1,254 @@
...
+  * do not define `hashCode`, `==` or `index`,
+  * do not extend any class beside Object, and
+  * have no subclasses declared in the defining library.
+
+**BAD:**
+```dart
+class LogPriority {
+  static const error = LogPriority._(1, 'Error');
+  static const warning = LogPriority._(2, 'Warning');
+  static const unknown = LogPriority._unknown(2, 'Log');
```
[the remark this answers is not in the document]
pq: Good catch.  Fixed.

lib/src/rules/prefer_enums.dart
```diff
@@ -0,0 +1,254 @@
...
+        var name = member.name?.name;
+        if (classElement.isPublic &&
+            (name == null || !Identifier.isPrivateName(name))) {
+          return;
+        }
+      }
+    }
+
+    if (candidateConstants.length < 2) return;
+    if (DartTypeUtilities.hasSubclassInDefiningCompilationUnit(classElement)) {
```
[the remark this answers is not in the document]
pq: Good catch.

lib/src/rules/prefer_enums.dart
pq: I added a test case for this but it passes as is.

```dart
class A {
  static const A a = A._(), b = A._();
  const A._();
}
```

What am I missing? 🤔

lib/src/rules/prefer_enums.dart
```diff
@@ -0,0 +1,254 @@
...
+            details: _details,
+            group: Group.style);
+
+  @override
+  void registerNodeProcessors(
+      NodeLintRegistry registry, LinterContext context) {
+    if (!context.isEnabled(Feature.enhanced_enums)) return;
+
+    // Skip contexts w/ parts.
+    if (context.allUnits.length > 1) return;
```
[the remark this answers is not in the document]
pq: Right.  Good call.  I added a test for parts and removed this restriction.

lib/src/rules/use_enums.dart
```diff
@@ -0,0 +1,244 @@
...
+      : super(classElement);
+
+  @override
+  void visitInstanceCreationExpression(InstanceCreationExpression node) {
+    if (!inConstantDeclaration) {
+      if (invokesGenerativeConstructor(node)) {
+        throw _InvalidEnumException();
+      }
+    }
+    inConstantDeclaration = false;
```
bwilkerson, line 118: I can't figure out how to respond to the existing thread.

You're not missing anything, I was. Because we're visiting a `VariableDeclaration` rather than the enclosing `FieldDeclaration`, the flag is getting reset for every field. Sorry for the confusion.

pq:
Doing a bit of poking around for cases that justify allowing migratable `index` I've only found one (`FontWeight` in flutter) so far but it's an interesting datapoint:

https://github.com/flutter/engine/blob/64f1252a97a5ecbf8aabd39f583f86bc7cc1dd4d/lib/ui/text.dart#L18-L104

bwilkerson:
We might want to reject that one anyway because of the definition of `values`. I don't think we currently do, but it's an error to have a definition of `values` in an enum. (values_declaration_in_enum)

pq:
> We might want to reject that one anyway because of the definition of values. 

Check added.

@bwilkerson: as per our conversation this morning, I'm inclined to land this and revisit `index` support later so we can pivot to looking at super initializers.

wdyt?

test/rules/use_enums_test.dart
```diff
@@ -0,0 +1,387 @@
...
+class A {
+  static const A a = A._();
+  static const A b = A._();
+  const A._();
+  static List<A> values() => [];
+}
+''');
+  }
+
+  test_no_lint_nonConstConstructor() async {
```
bwilkerson, line 259: This is another case that we might want to relax in the lint. If there's a non-const constructor but all of the fields are final, then it might still be reasonable to flag the class.

bwilkerson (approved):

lib/src/rules/prefer_enums.dart
munificent: I don't have any off-hand (I struggled with this too when writing the formatter tests for enhanced enums), but we can always circle back later when we see more uses in the wild.

PR #3254, opened 2022-02-26
Add `use_colored_box`

minhqdao:
## Description

Use `ColoredBox` over `Container` when only the `Color` is specified.

Addresses https://github.com/flutter/flutter/issues/90722 and https://github.com/dart-lang/linter/issues/3255

## Kind

Style advice.

## Bad Example

```dart
Widget buildArea() {
  return Container(
    color: Colors.blue,
    child: const Text('...'),
  );
}
```

## Good Example

```dart
Widget buildArea() {
  return const ColoredBox(
    color: Colors.blue,
    child: Text('...'),
  );
}
```

### DIFF
--- a/example/all.yaml
+++ b/example/all.yaml
@@ -186,6 +186,7 @@ linter:
     - unrelated_type_equality_checks
     - unsafe_html
     - use_build_context_synchronously
+    - use_colored_box
     - use_decorated_box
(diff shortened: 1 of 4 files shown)


bwilkerson (approved):

pq (approved):
💎

Issue #3239, opened 2022-02-18
False positive for `use_decorated_box`


ookami-kb:
**Describe the issue**

`Container` has nullable `decoration` parameter, so construction like this is totally valid:

```dart
Container(
  decoration: _counter.isEven
      ? const BoxDecoration(color: Colors.green)
      : null,
  child: const Text(
    'You have pushed the button this many times:',
  ),
),
```

But with `use_decorated_box` lint rule enable, it suggests using `DecoratedBox` instead. But it's not possible since `DecoratedBox` doesn't accept nullable `Decoration?`.

**To Reproduce**

```dart
import 'package:flutter/material.dart';

void main() {
  runApp(const MyApp());
}

const Decoration? decoration = null;

class MyApp extends StatelessWidget {
  const MyApp({Key? key}) : super(key: key);

  // This widget is the root of your application.
  @override
  Widget build(BuildContext context) {
    return MaterialApp(
      title: 'Flutter Demo',
      theme: ThemeData(
        primarySwatch: Colors.blue,
      ),
      home: Container(
        decoration: decoration,
        child: const Center(
          child: Text('Hello World'),
        ),
      ),
    );
  }
}
```

**Expected behavior**

`use_decorated_box` shouldn't trigger if nullable `Decoration?` is passed as a parameter to `Container`.

minhqdao:
You're right, a `DecoratedBox` doesn't accept a nullable `Decoration?`, which is why you're using a `Container` as a fallback, that does. In my opinion, the linting rule is actually correct, but a `DecoratedBox` should accept a `Decoration?` instead of a `Decoration` to be able to deal with cases such as yours.

As I also implemented the original rule, I can also try to fix it, if that is generally desired. But again, I find it a bit strange (actually, I find it wrong) to loosen the rule on `Container` just because a `DecoratedBox` cannot handle `null`.

ookami-kb:
Well, as for me, `DecoratedBox` without `decoration` looks really weird and could lead to even more non-obvious behavior. E.g. construction like this would definitely be weird:

```dart
DecoratedBox(child: Text('Hello World'))
```

From that perspective, I would say that fixing `use_decorated_box` makes more sense: it shouldn't advise switching to `DecoratedBox` if you **cannot** switch to it, whatever the actual reason is.

minhqdao:
[...]

PR #3271, opened 2022-03-05
Fix false positive `use_decorated_box` and `use_colored_box`

minhqdao:
This PR alters `use_decorated_box` so it doesn't advise switching from a `Container` to a `DecoratedBox` when a nullable `Decoration?` expression is used because a `DecoratedBox` only accepts a non-null `Decoration`.

The same is done for `use_colored_box`.

See discussion: https://github.com/dart-lang/linter/issues/3239

### DIFF
--- a/lib/src/rules/use_decorated_box.dart
+++ b/lib/src/rules/use_decorated_box.dart
@@ -6,6 +6,7 @@ import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
 
 import '../analyzer.dart';
+import '../util/dart_type_utilities.dart';
 import '../util/flutter_utils.dart';
 
 const _desc = r'Use `DecoratedBox`.';
@@ -59,16 +60,17 @@ class UseDecoratedBox extends LintRule {
   @override
   void registerNodeProcessors(
       NodeLintRegistry registry, LinterContext context) {
-    var visitor = _Visitor(this);
+    var visitor = _Visitor(this, context);
 
     registry.addInstanceCreationExpression(this, visitor);
   }
 }
 
 class _Visitor extends SimpleAstVisitor {
-  final LintRule rule;
+  _Visitor(this.rule, this.context);
 
-  _Visitor(this.rule);
+  final LintRule rule;
+  final LinterContext context;
 
   @override
   void visitInstanceCreationExpression(InstanceCreationExpression node) {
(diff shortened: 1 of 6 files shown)


lib/src/rules/use_decorated_box.dart
```diff
@@ -91,20 +93,25 @@ class _Visitor extends SimpleAstVisitor {
...
-      var label = argument.name.label;
-      if (label.name == 'decoration') {
-        hasDecoration = true;
-      } else if (label.name == 'child') {
+      var name = argument.name.label.name;
+      if (name == 'decoration') {
+        // Not lint if decoration is null or nullable because a DecoratedBox
+        // only accepts a non-null Decoration
+        if (DartTypeUtilities.isNonNullable(
+            context, argument.expression.staticType)) {
```
bwilkerson, line 109: Checking for nullability is good, but we're missing the more important test to ensure that the type of the expression is assignable to `BoxDecoration`.

If you wouldn't mind fixing that while you're at it, then also add a test to ensure that we don't lint when the `decoration` argument  is some other kind of decoration (maybe a `ShapeDecoration`).

test_data/rules/use_colored_box.dart
```diff
@@ -7,9 +7,7 @@
 import 'package:flutter/foundation.dart';
 import 'package:flutter/widgets.dart';
 
-class Color {
-  Color(int value);
-}
+class Color {}
```
bwilkerson, line 10: You can't do so in this PR because the change has to be made to the analyzer package, but this class really ought to be part of the mock SDK we use for testing. When we do that we'll make the class match the published version of `Color`, including making the constructor take an argument.

Once it's been added, then this test will need to be converted to using the definition from the mock SDK, and that will require restoring all of the arguments. It would ease that future conversion if you could revert the change to the definition of `Color` and the invocations of the constructor.

test_data/rules/use_decorated_box.dart
```diff
@@ -20,7 +20,7 @@ Widget containerWithKey() {
 
 Widget containerWithDecoration() {
   return Container( // LINT
-    decoration: BoxDecoration(),
+    decoration: Decoration(),
```
bwilkerson, line 23: I believe that these changes (from `BoxDecoration` to `Decoration`) also need to be reverted. It's a false positive to lint here because this code can't use `DecoratedBox`.

srawlins:
Sorry for the year+ delay, @minhqdao . @bwilkerson left some comments, are you able to review them and upload another commit? Thanks!

lib/src/rules/use_decorated_box.dart
minhqdao: A `Container` accepts a `Decoration?`, which looks like a deliberate design decision to me. This also doesn't throw an error:

```dart
child: Container(decoration: const ShapeDecoration(shape: CircleBorder()))
```

If a Container is supposed to only accept a `BoxDecoration`, why not just allow the constructor to accept a `BoxDecoration` instead of a `Decoration`?

PR #3279, opened 2022-03-08
For `always_specify_types`, use NamedType.name.staticElement in favor of DartType.alias.

### LINKED ISSUE #3275: Apparent regression in `always_specify_types` with `typedef`
**Describe the issue**
https://github.com/dart-lang/linter/issues/2692 seems to have (partially?) regressed recently on Flutter master. We have an auto-roller that rolls flutter/flutter into flutter/packages, and it started failing sometime in the last week or so (I still need to pinpoint the exact regression point) due to new issues with the `analyze` step we run in CI ([example failure](https://github.com/flutter/packages/pull/1127/checks?check_run_id=5467528658)).

The failures are:
```
   info - lib/src/dart/binary.dart:354:12 - Specify type annotations. - always_specify_types
   info - lib/src/dart/binary.dart:401:16 - Specify type annotations. - always_specify_types
   info - lib/src/dart/text.dart:2200:32 - Specify type annotations. - always_specify_types
   info - lib/src/dart/text.dart:2223:33 - Specify type annotations. - always_specify_types
   info - lib/src/flutter/content.dart:198:66 - Specify type annotations. - always_specify_types
   info - lib/src/flutter/runtime.dart:555:25 - Specify type annotations. - always_specify_types
   info - lib/src/flutter/runtime.dart:666:16 - Specify type annotations. - always_specify_types
   info - lib/src/flutter/runtime.dart:668:16 - Specify type annotations. - always_specify_types
   info - lib/src/flutter/runtime.dart:955:29 - Specify type annotations. - always_specify_types
   info - lib/src/flutter/runtime.dart:997:12 - Specify type annotations. - always_specify_types
```

An example failing line is:
```dart
final DynamicMap results = DynamicMap();
```

where the failure shows as being the `DynamicMap()` part. But `DynamicMap` is
```
typedef DynamicMap = Map<String, Object?>;
```

Interestingly it doesn't have an issue with the declaration of `results`, just the constructor call.

Other failures are `DynamicMap.fromEntries` and `DynamicList.generate` (where `DynamicList` is a similar typedef for `List<Object?>`).

**To Reproduce**

Minimal example:
```
typedef Foo = Map<String, Object>;
final Foo foo = Foo();
```

**Expected behavior**
No `always_specify_types` lint.

**Additional context**
Changing the second line to
```
final Foo foo = Map<String, Object>();
```

makes the lint message go away, so it's definitely the typedef.

stereotype441:
In some circumstances, the analyzer doesn't reliably set
DartType.alias to point to the type alias that led to a type.  (This
is the ultimate cause of #3275).  I'm currently researching whether
it's possible to improve the analyzer behavior in this regard.  In the
meantime, we can fix #3275 by changing the linter to grab the type
alias element directly from NamedType.name.staticElement.

Fixes #3275.

### DIFF
--- a/lib/src/rules/always_specify_types.dart
+++ b/lib/src/rules/always_specify_types.dart
@@ -4,6 +4,7 @@
 
 import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/element.dart';
 import 'package:analyzer/dart/element/type.dart';
 
 import '../analyzer.dart';
@@ -110,8 +111,9 @@ class _Visitor extends SimpleAstVisitor<void> {
   void visitNamedType(NamedType namedType) {
     var type = namedType.type;
     if (type is InterfaceType) {
-      var element = type.alias?.element ?? type.element;
-      if (element.typeParameters.isNotEmpty &&
+      var element = namedType.name.staticElement;
+      if (element is TypeParameterizedElement &&
+          element.typeParameters.isNotEmpty &&
           namedType.typeArguments == null &&
           namedType.parent is! IsExpression &&
           !element.hasOptionalTypeArgs) {


pq (approved):
👍

pq:
Thanks!

PR #3287, opened 2022-03-14
Add rationale & example to prefer_for_elements_to_map_fromIterable

### LINKED ISSUE #3259: Reasoning for: prefer_for_elements_to_map_fromIterable
[prefer_for_elements_to_map_fromIterable](https://github.com/dart-lang/linter/blob/658f497eef358e989e3c41d60bfcc30bfaea217b/lib/src/rules/prefer_for_elements_to_map_fromIterable.dart) says not to use Map.fromIterable:

> When building maps from iterables, it is preferable to use for elements.
> 
> **BAD:**
> ```dart
> Map<String, WidgetBuilder>.fromIterable(
>   kAllGalleryDemos,
>   key: (demo) => '${demo.routeName}',
>   value: (demo) => demo.buildRoute,
> );
> ```
> **GOOD:**
> ```dart
> return {
>   for (var demo in kAllGalleryDemos)
>     '${demo.routeName}': demo.buildRoute,
> };
> ```

No reasoning is provided.

It seems this lint would prefer me to do something like:

```dart
enum Trees {
  maple,
  oak,
  willow,
}

final Map<Trees, List<String>> fromForLoop = {
  for (var tree in Trees.values) tree: []
};
```

Instead of something like this:

```dart
final Map<Trees, List<String>> fromIterable = Map.fromIterable(
  Trees.values,
  key: (tree) => tree,
  value: (_) => [],
);
```

The `fromIterable` option seems to me easier to write and cleaner to read. So if there is some concrete reason for the lint, I think we should include that reasoning - such as performance concerns or what-have-you.

I also notice the provided examples seem to be Flutter, so perhaps the lint is intended only for widget-based code? I looked for an original PR regarding this lint to find the reasoning, but I couldn't find one.

Merrit:
# Description

Added the rationale behind this lint. Also another example that doesn't rely on Flutter, and attempts to show how the alternative can be used to good effect.

Fixes #3259

### DIFF
--- a/lib/src/rules/prefer_for_elements_to_map_fromIterable.dart
+++ b/lib/src/rules/prefer_for_elements_to_map_fromIterable.dart
@@ -8,12 +8,12 @@ import 'package:analyzer/dart/ast/visitor.dart';
 
 import '../analyzer.dart';
 
-const _desc = r'Prefer for elements when building maps from iterables.';
+const _desc = r"Prefer 'for' elements when building maps from iterables.";
 
 const _details = r'''
-When building maps from iterables, it is preferable to use for elements.
+When building maps from iterables, it is preferable to use 'for' elements.
 
-Using for elements brings several benefits including:
+Using 'for' elements brings several benefits including:
 
 - Performance
 - Flexibility
--- a/lib/src/rules/prefer_for_elements_to_map_fromIterable.dart
+++ b/lib/src/rules/prefer_for_elements_to_map_fromIterable.dart
@@ -13,6 +13,15 @@ const _desc = r'Prefer for elements when building maps from iterables.';
 const _details = r'''
 When building maps from iterables, it is preferable to use for elements.
 
+Using for elements brings several benefits including:
+
+- Performance
+- Flexibility
+- Readability
+- Improved type inference
+- Improved interaction with null safety
+
+
 **BAD:**
 ```dart
 Map<String, WidgetBuilder>.fromIterable(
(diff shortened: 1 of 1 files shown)


pq:
Thank you!

/fyi @lrhn @munificent

lib/src/rules/prefer_for_elements_to_map_fromIterable.dart
```diff
@@ -13,6 +13,15 @@ const _desc = r'Prefer for elements when building maps from iterables.';
 const _details = r'''
 When building maps from iterables, it is preferable to use for elements.
 
+Using for elements brings several benefits including:
```
lrhn, line 16: The "for" in "for elements" here reads a little too much as a separate word, not part of the "for elements". Makes me thing "use *what* for *which* elements?".

Consider quoting ("for elements") or dashing (for-elements) or even code-quoting the "for" (`for`-elements), both here and above.

Merrit: Agree, I was following the existing phrasing but it doesn't sound great to me either.

"for elements" itself sounds a bit odd to me, but maybe others are used to that.. I would've thought something like "Prefer for-loops when building maps from iterables.", or is "elements" preferred?

bwilkerson:
"for elements" is preferred over "for loop" to distinguish the collection literal element use from the use as a statement.

Single quotes around the "for" keyword is preferred for consistency with all of the other diagnostic messages.

pq (approved):
LGTM. Thanks for all the back-and-forth.

@lrhn?

lrhn (approved):
LGTM!

Merrit:
Thanks for the help all :grinning:

PR #3293, opened 2022-03-16
fix `use_super_parameters` false positive w/ field formal params

pq:
See: https://dart-review.googlesource.com/c/sdk/+/237600

/cc @bwilkerson

### DIFF
--- a/lib/src/rules/use_super_parameters.dart
+++ b/lib/src/rules/use_super_parameters.dart
@@ -172,6 +172,7 @@ class _Visitor extends SimpleAstVisitor {
       bool match = false;
       for (var i = 0; i < constructorParams.length && !match; ++i) {
         var constructorParam = constructorParams[i];
+        if (constructorParam is FieldFormalParameter) continue;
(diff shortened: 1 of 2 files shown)


lib/src/rules/use_super_parameters.dart
```diff
@@ -172,6 +172,7 @@ class _Visitor extends SimpleAstVisitor {
       bool match = false;
       for (var i = 0; i < constructorParams.length && !match; ++i) {
         var constructorParam = constructorParams[i];
+        if (constructorParam is FieldFormalParameter) continue;
```
bwilkerson, line 175: We should also skip `SuperFormalParameter`. We won't run into any yet, but don't want false positives after the feature has started being adopted.

bwilkerson (approved):

lib/src/rules/use_super_parameters.dart
pq: Good catch.

PR #3316, opened 2022-03-22
Add tests corresponding to a bug in the fix

bwilkerson:
There was a bug in the fix, and I added tests to confirm that there isn't a corresponding bug in the lint.


pq (approved):
LGTM!

PR #3320, opened 2022-03-28
update to analyzer 3.4 APIs

pq:
/cc @scheglov @bwilkerson


test/rule_test_support.dart
```diff
@@ -388,7 +388,7 @@ abstract class _ContextResolutionTest with ResourceProviderMixin {
       throw StateError('Only dart files can be changed after analysis.');
     }
 
-    return super.newFile(path, content: content);
+    return super.newFile2(path, content);
```
scheglov, line 391: I would recommend to rename the containing method to `newFile2`, with corresponding making `content` a required positional parameter.

scheglov (approved):

pq:
Sorry for the noise here.  I hadn't run the tests and am looking at the support I added for `unnecessary_getters_setters`.  Given that you can't have a non-final field in an enum, it seems like the lint should just not fire.

WDYT?

test/rule_test_support.dart
pq: Done!

bwilkerson:
> Given that you can't have a non-final field in an enum, it seems like the lint should just not fire.

I don't understand why not. If the user writes
```
enum E {
  a = E(1),
  b = E(2);
  final int _value;
  const E(this._value);
  int get value => _value;
}
```
Wouldn't it be valid to suggest changing the field's name to be that of the getter and removing the getter?

bwilkerson (approved):

PR #3324, opened 2022-03-30
[unsafe_html] Allow assigning to img.src.

vrana:
This used to be dangerous because of img.src = 'javascript:hack' but browsers don't allow this anymore. We allow it in all other languages.

### DIFF
--- a/lib/src/rules/unsafe_html.dart
+++ b/lib/src/rules/unsafe_html.dart
@@ -18,8 +18,8 @@ const _details = r'''
 **AVOID**
 
 * assigning directly to the `href` field of an AnchorElement
-* assigning directly to the `src` field of an EmbedElement, IFrameElement,
-  ImageElement, or ScriptElement
+* assigning directly to the `src` field of an EmbedElement, IFrameElement, or
+  ScriptElement
 * assigning directly to the `srcdoc` field of an IFrameElement
 * calling the `createFragment` method of Element
 * calling the `open` method of Window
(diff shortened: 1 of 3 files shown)


srujzs (approved):

pq (approved):

jamesderlin:
[The Dart 2.17 release notes](https://github.com/dart-lang/sdk/blob/master/CHANGELOG.md#2170) indicate that this was supposed to be included, but I'm still seeing `unsafe_html` reported against:

```dart
import 'dart:html';

void f(ImageElement image) {
  image.src = 'path/to/file';
}
```
with `unsafe_html` enabled in my `analysis_options.yaml` and with `pubspec.yaml` requiring Dart 2.17.0.

Any ideas?

jamesderlin:
Hm, I'm also not seeing the change in behavior to `no_leading_underscores_for_local_identifiers` to ignore `_`.  I've tried uninstalling and reinstalling the `apt` package for Dart on my Linux x64 machine and also tried running `dart` from the 2.17.0 `.zip` archive, but I'm still not seeing the new linter changes.  I don't have any other copies of `dart` in `PATH`.

jamesderlin:
Okay, I tried installing 2.18.0-99.0.dev and *am* seeing the linter changes, so I guess the 2.17 release notes are wrong.

parlough:
@pq I think when you updated the SDK to linter 1.23 you accidentally added those changes to the 2.17 changelog: https://github.com/dart-lang/sdk/commit/226fceaef2341f294e1c80a536ce07420ab65189

pq:
>  I think when you updated the SDK to linter 1.23 you accidentally added those changes to the 2.17 changelog

What confuses me is that that change also includes the bump in `DEPS` that brings in 1.23.  IOW: I'd expect if we are pulling in my changelog changes at that commit, we should get my updated DEPS too?

parlough:
Yeah those changes are successfully present on the `main` branch. James mentioned they worked in the latest dev build. Only problem is the section of the changelog those changes were included in.

jamesderlin:
I see that the changelog has been fixed. (Thanks, @parlough!)

(And sorry for hijacking this issue.)

pq:
> (And sorry for hijacking this issue.)

Not at all.  Thanks for catching this!

PR #3328, opened 2022-04-01
handle functions in asserts for require_trailing_commas

a14n:
# Description

This PR prevents `require_trailing_commas` to trigger on asserts with function like the following one:

```dart
    assert(() {
      _debugIsRenderViewInitialized = true;
      return true;
    }());
```

### DIFF
--- a/lib/src/rules/require_trailing_commas.dart
+++ b/lib/src/rules/require_trailing_commas.dart
@@ -163,6 +163,15 @@ class _Visitor extends SimpleAstVisitor<void> {
       return true;
     }
 
+    // Exception is allowed if the last parameter is a anonymous function call.
+    // This case arises a lot in asserts.
+    if (lastNode is FunctionExpressionInvocation &&
+        lastNode.function is FunctionExpression &&
+        _isSameLine(lastNode.argumentList.leftParenthesis,
+            lastNode.argumentList.rightParenthesis)) {
+      return true;
+    }
+
     // Exception is allowed if the last parameter is a set, map or list literal.
     if (lastNode is SetOrMapLiteral || lastNode is ListLiteral) return true;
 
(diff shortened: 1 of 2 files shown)


lib/src/rules/require_trailing_commas.dart
```diff
@@ -163,6 +163,15 @@ class _Visitor extends SimpleAstVisitor<void> {
       return true;
     }
 
+    // Exception is allowed if the last parameter is a anonymous function call.
```
bwilkerson, line 166: nit: "parameter" --> "argument".

It's wrong in a couple of previous cases as well, but fixing them is even more optional.

lib/src/rules/require_trailing_commas.dart
```diff
@@ -163,6 +163,15 @@ class _Visitor extends SimpleAstVisitor<void> {
...
+    // This case arises a lot in asserts.
+    if (lastNode is FunctionExpressionInvocation &&
+        lastNode.function is FunctionExpression &&
+        _isSameLine(lastNode.argumentList.leftParenthesis,
```
bwilkerson, line 170: I'm not sure why this is an interesting condition. If they aren't on the same line then I'd expect the comma would want to go inside the argument list, not outside:
```
  f((a, b, c) {} (
    a,
    b,
    c,
  ));
```

But I did notice that there aren't any tests that cover the case where the parens aren't on the same line.

lib/src/rules/require_trailing_commas.dart
a14n: Done

lib/src/rules/require_trailing_commas.dart
a14n: This kind of code looks similar to a function call with an object creation:

```dart
f(MyObject(
  a,
  b,
));
```
And in this case the current version of the lint ([that is really controversal](https://github.com/dart-lang/linter/pull/2566)) is to required a trailing comma to end up to:
```dart
f(
  MyObject(
    a,
    b,
  ),
);
```

So I didn't add anything special here.

> But I did notice that there aren't any tests that cover the case where the parens aren't on the same line.

Added.

bwilkerson (approved):
Thanks.

pq (approved):

PR #3329, opened 2022-04-04
bump markdown dep

pq:
It would probably be safest to update the SDK DEP before landing this (so the linter doesn't get ahead of what it will integrate w/ in the SDK).

@srawlins: any reason we can't bump `markdown` in the SDK?  Will that cause issues for `dartdoc`?

/cc @bwilkerson  @srawlins

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -28,7 +28,7 @@ dev_dependencies:
   github: ^9.0.0
   grinder: ^0.9.0
   lints: ^2.0.0
-  markdown: ^4.0.0
+  markdown: ^5.0.0
   matcher: ^0.12.10
   pub_semver: ^2.0.0
   test: ^1.16.1


bwilkerson (approved):
I'm fine with this change, but we might want to hold off on landing it until we can safely bump the version in the SDK. This does, however, verify that `linter` won't be broken if we bump the version in the SDK before we've updated the `linter`'s constraints.

pq:
> we might want to hold off on landing it until we can safely bump the version in the SDK.

💯 See my comment above.  I'm inclined to wait and am happy to experiment w/ bumping the DEP in the SDK (assuming @srawlins doesn't have reservations or hasn't done it already!)

srawlins:
> @srawlins: any reason we can't bump markdown in the SDK? Will that cause issues for dartdoc?

I haven't looked yet. I think the only breaking change was making a public API parameter non-nullable.

pq:
Blocked behind https://github.com/dart-lang/sdk/issues/48967.

PR #3333, opened 2022-04-06
handle multiline-strings in require_trailing_commas

a14n:
# Description

This PR prevents `require_trailing_commas` to trigger on multiline-strings like the following one:

```dart
  print('''
    --help      Print a usage message.
    --temp-dir  A location where temporary files may be written. Defaults to a
                directory in the system temp folder. If a temp_dir is not
                specified, then the default temp_dir will be created, used, and
                removed automatically.
    ''');
```


pq (approved):
Nice.

👍

PR #3340, opened 2022-04-08
check trailing commas in list/set/map literals

a14n:
# Description

This PR checks trailing commas in list/set/map literals.

### DIFF
--- a/test_data/rules/require_trailing_commas.dart
+++ b/test_data/rules/require_trailing_commas.dart
@@ -39,7 +39,7 @@ class RequireTrailingCommasExample {
       : assert(true,
             'A very very very very very very very very long string'); // LINT
 
-  void method1(Object param1, Object param2, {Object? param3, Object? param4}) {}
+  void method1(Object p1, Object p2, {Object? param3, Object? param4}) {}
 
   void method2(
     Object param1,
@@ -275,3 +275,48 @@ A multiline string
 }
 
 class AnExtremelyLongClassNameOneTwoThreeFourFiveSixSevenEightNineTen {}
+
+lists() {
+  var x;
+  x = []; //OK
+  x = [1]; //OK
+  x = [1, if (true) 2]; //OK
+  x = [
+    1, //
+    2
+  ]; // LINT
+  x = [
+    1, //
+    if (true) 2
+  ]; // LINT
+}
+
+sets() {
+  var x;
+  x = <dynamic>{}; //OK
+  x = {1}; //OK
+  x = {1, if (true) 2}; //OK
+  x = {
+    1, //
+    2
+  }; // LINT
+  x = {
+    1, //
+    if (true) 2
+  }; // LINT
+}
+
+maps() {
+  var x;
+  x = {}; //OK
+  x = {1: 1}; //OK
+  x = {1: 1, if (true) 2: 2}; //OK
+  x = {
+    1: 1, //
+    2: 2
+  }; // LINT
+  x = {
+    1: 1, //
+    if (true) 2: 2
+  }; // LINT
+}
--- a/lib/src/rules/require_trailing_commas.dart
+++ b/lib/src/rules/require_trailing_commas.dart
@@ -61,11 +61,13 @@ class RequireTrailingCommas extends LintRule {
   ) {
     var visitor = _Visitor(this);
     registry
-      ..addCompilationUnit(this, visitor)
       ..addArgumentList(this, visitor)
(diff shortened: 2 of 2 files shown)


test_data/rules/require_trailing_commas.dart
```diff
@@ -39,7 +39,7 @@ class RequireTrailingCommasExample {
       : assert(true,
             'A very very very very very very very very long string'); // LINT
 
-  void method1(Object param1, Object param2, {Object? param3, Object? param4}) {}
+  void method1(Object p1, Object p2, {Object? param3, Object? param4}) {}
```
bwilkerson, line 42: Curious to know why you changed the names of half the parameters (rather than either none or all). It doesn't change the semantics of the test, just seems inconsistent.

bwilkerson (commented):
The implementation looks good.

I can't speak to whether the change in semantics is what users want, but this is likely to be a breaking change for at least some users.

test_data/rules/require_trailing_commas.dart
a14n: Before the change the line length was 81 and the formatting (assumed by this rule) splitted the line.

a14n:
> but this is likely to be a breaking change for at least some users.

I agree but I don't think the creation of a new rule `require_trailing_commas_for_list_set_map` is worth it. This addition of list/set/maps looks consistant to the initial intent of this lint imho.

goderbauer:
Overall this looks reasonable to me. Have you by any chance already applied this to the Flutter code base to see if there are any surprising cases on a larger code base like that?

a14n:
Correct me if I'm wrong but `require_trailing_commas` is not enabled on Flutter code base.

I will take a look among the 12K+ problems reported by VS code when I enable the current version of the lint.

goderbauer:
> Correct me if I'm wrong but require_trailing_commas is not enabled on Flutter code base.

You are right, we never did enable it it looks like.

a14n:
I tested this PR on flutter and there are 700+ diagnostics about missing trailing commas in set/list/map literals. You can take a look at https://github.com/flutter/flutter/pull/102585

What should we do with the current PR?

goderbauer:
I scrolled through https://github.com/flutter/flutter/pull/102585. There are no surprises in there for me. So extending the trailing comma to collections still sounds reasonable to me.

bwilkerson (approved):

pq:
My biggest concern on this is the time to migrate internal code that will break when we try and roll with this landed.

I'm guessing that the `AddTrailingComma` correction producer should get updated in server so we can do cleanup with `dart fix`?

If anyone wants to pitch in there, tests live here:

https://github.com/dart-lang/sdk/blob/main/pkg/analysis_server/test/src/services/correction/fix/add_trailing_comma_test.dart

a14n:
`AddTrailingComma` correction updated in https://github.com/dart-lang/sdk/pull/48968 to support this PR

pq:
This is another one that will likely require a bunch of internal cleanup before landing in the SDK.  I'm a bit oversubscribed at the moment but will put the feelers out for volunteers...  Thanks @a14n!

a14n:
> This is another one that will likely require a bunch of internal cleanup before landing in the SDK.

@pq is there a lot of internal cleanup to do?

srawlins:
This rule is enabled internally, so I'll kick off a job to test it out.

srawlins:
There is much to clean up internally, haha. I've started the process. May take a few days. Thanks a million for your patience!!

srawlins:
Update, I have an in-progress change to fix the internal code but there are a lot of hoops to jump though, so I'm going to pause that. No ETA for when this can land.

srawlins:
@a14n this is now making great progress internally, would you mind making one more pass over the flutter repos before I land? I'm very excited for this one 😁

a14n:
AFAICT this lint rule is [disable in flutter/flutter](https://github.com/flutter/flutter/blob/159ebf736acd9af684dac033ec1a94f9b92d28b1/analysis_options.yaml#L187).

srawlins:
Oh I see, woohoo! Thanks!

@pq any concerns with me landing then? Tracking for internal compliance is b/279356486.

srawlins (approved):

srawlins:
Quick note that this does not cover ListPatterns or MapPatterns. Mentioned in [an open issue about records](https://github.com/dart-lang/linter/issues/3650#issuecomment-1561653166).

PR #3346, opened 2022-04-11
mismatched type false positive fix for `prefer_initializing_formals`

### LINKED ISSUE #3345: `prefer_initializing_formals` false positive
In the analyzer code base:

```dart

  final Element element;

  /// Creates an [ElementWalker] which walks the child elements of a class
  /// element.
  ElementWalker.forClass(ClassElement element)
      : element = element, // FALSE POSITIVE
        _accessors = element.accessors.where(_isNotSynthetic).toList(),
        _constructors = element.isMixinApplication
            ? null
            : element.constructors.where(_isNotSynthetic).toList(),
        _functions = element.methods,
        _typeParameters = element.typeParameters,
        _variables = element.fields.where(_isNotSynthetic).toList();

```

pq:
Fixes #3345

/cc @bwilkerson

### DIFF
--- a/test/rules/prefer_initializing_formals_test.dart
+++ b/test/rules/prefer_initializing_formals_test.dart
@@ -1,0 +1,31 @@
+// Copyright (c) 2022, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:test_reflective_loader/test_reflective_loader.dart';
+
+import '../rule_test_support.dart';
+
+main() {
+  defineReflectiveSuite(() {
+    defineReflectiveTests(PreferInitializingFormalsTest);
+  });
+}
+
+@reflectiveTest
+class PreferInitializingFormalsTest extends LintRuleTest {
+  @override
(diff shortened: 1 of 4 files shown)


test/rules/prefer_initializing_formals_test.dart
```diff
@@ -0,0 +1,31 @@
...
+  @override
+  String get lintRule => 'prefer_initializing_formals';
+
+  /// https://github.com/dart-lang/linter/issues/3345
+  test_noLint() async {
+    await assertNoDiagnostics(r'''
+class C {
+  final Object name;
+  final int length;
+  C.forName(String name) : name = name,
```
bwilkerson, line 26: As I said in another review (which wasn't until after you'd already sent this out), Why wouldn't we want this to be rewritten as
```
C.forName(String this.name): length = name.length;
```

pq: Of course.  You're totally spot on.

It looks like `dart fix` needs these smarts actually.  I'll open a tracking bug and then close this out.

pq: https://github.com/dart-lang/sdk/issues/48804

PR #3351, opened 2022-04-13
fix `avoid_init_to_null` w/ super initializing non-null defaults

pq:
Fixes: #3349 

I don't love reaching into `Impl` classes so if there's a better way, I'd welcome it!

/cc @bwilkerson

### DIFF
--- a/lib/src/rules/avoid_init_to_null.dart
+++ b/lib/src/rules/avoid_init_to_null.dart
@@ -7,8 +7,6 @@ import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
 import 'package:analyzer/dart/element/element.dart';
 import 'package:analyzer/dart/element/type.dart';
-// ignore: implementation_imports
-import 'package:analyzer/src/dart/element/element.dart';
 
 import '../analyzer.dart';
 import '../util/dart_type_utilities.dart';
--- a/lib/src/rules/avoid_init_to_null.dart
+++ b/lib/src/rules/avoid_init_to_null.dart
@@ -6,6 +6,8 @@ import 'package:analyzer/dart/analysis/features.dart';
 import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
 import 'package:analyzer/dart/element/type.dart';
+// ignore: implementation_imports
+import 'package:analyzer/src/dart/element/element.dart';
 
 import '../analyzer.dart';
 import '../util/dart_type_utilities.dart';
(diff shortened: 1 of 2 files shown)


lib/src/rules/avoid_init_to_null.dart
```diff
@@ -88,8 +90,13 @@ class _Visitor extends SimpleAstVisitor<void> {
   @override
   void visitDefaultFormalParameter(DefaultFormalParameter node) {
     var declaredElement = node.declaredElement;
-    if (declaredElement == null) {
-      return;
+    if (declaredElement == null) return;
+    if (declaredElement is DefaultSuperFormalParameterElementImpl) {
```
asashour, line 94: I guess there is no need to access the implementation, we can use `SuperFormalParameterElement` and `FieldFormalParameterElement`

bwilkerson: I can't easily see which APIs you need from `DefaultSuperFormalParameterElementImpl` and `DefaultFieldFormalParameterElementImpl` that aren't public APIs, but you might not need to reference them anymore.

You might consider combining the `if`s on lines 93 and 94 as
```
if (declaredElement is! DefaultSuperFormalParameterElementImpl) return;
```
just to make the code a little shorter.

lib/src/rules/avoid_init_to_null.dart
```diff
@@ -88,8 +90,13 @@ class _Visitor extends SimpleAstVisitor<void> {
...
+      var superConstructorParameter = declaredElement.superConstructorParameter;
+      if (superConstructorParameter is DefaultFieldFormalParameterElementImpl) {
+        var defaultValue = superConstructorParameter.defaultValueCode;
+        if (defaultValue != null) return;
```
bwilkerson, line 98: It looks like this is too broad. To confirm or deny, consider adding a test with the following code:
```
class A {
  String? a;
  A({this.a = null});
}
class B extends A {
  B({super.a = null}); // LINT
}
```

bwilkerson (approved):

lib/src/rules/avoid_init_to_null.dart
pq: Thanks!

lib/src/rules/avoid_init_to_null.dart
pq: I think the test case you maybe meant was:

```dart
class A {
  String? a;
  A({this.a = ''});
}
class B extends A {
  B({super.a = null});
}
```

and indeed, this is not working as written!

pq: (Deleted a mistyped example.)

PR #3356, opened 2022-04-21
allow just underscores in `no_leading_underscores_for_local_identifiers`

### LINKED ISSUE #3126: no_leading_underscores_for_local_identifiers reports variables named `_`
**Describe the issue**

no_leading_underscores_for_local_identifiers reports variables named `_`, which I think should be allowed (perhaps only for parameters).

**To Reproduce**

```dart
try {
  // ...
} catch (_) {
  // _ is not used.
}
```

**Expected behavior**

This should be allowed. A catch clause _must_ accept 1 or 2 parameters, not 0.

pq:
Fixes #3126

/cc @bwilkerson

### DIFF
--- a/test_data/rules/no_leading_underscores_for_local_identifiers.dart
+++ b/test_data/rules/no_leading_underscores_for_local_identifiers.dart
@@ -9,7 +9,7 @@
 void fn0() {
   for (var _ in []) { } // OK
   var _ = g(); //OK
-  for (var _ = 0; _ < 1; ++_) { } // OK
+  for (var __ = 0; __ < 1; ++__) { } // OK
 }
 
 int g() => 0;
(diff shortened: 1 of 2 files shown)


test_data/rules/no_leading_underscores_for_local_identifiers.dart
```diff
@@ -5,6 +5,15 @@
 // test w/ `dart test -N no_leading_underscores_for_local_identifiers`
 
 
+///https://github.com/dart-lang/linter/issues/3126
+void fn0() {
+  for (var _ in []) { } // OK
+  var _ = g(); //OK
+  for (var _ = 0; _ < 1; ++_) { } // OK
```
bwilkerson, line 12: It would be good to have at least one test with a name that is multiple underscores.

bwilkerson (approved):

test_data/rules/no_leading_underscores_for_local_identifiers.dart
pq: Done. Thanks.

PR #3361, opened 2022-04-25
fix local functions in `no_leading_underscores_for_local_identifiers`

### LINKED ISSUE #3360: no_leading_underscores_for_local_identifiers doesn't complain about local functions
The following code does not produce a `no_leading_underscores_for_local_identifiers` lint, even though I would have expected it to complain about `_bar`:

```dart
int foo() {
  int _bar() { // Why no no_leading_underscores_for_local_identifiers lint here?
    return 10;
  }
  return _bar();
}
```

Complaining here would also seem to be in line with "Effective Dart" [1] (emphasize mine):

> There is no concept of “private” for local variables, parameters, **local functions**, or library prefixes. When one of those has a name that starts with an underscore, it sends a confusing signal to the reader. To avoid that, **don’t use leading underscores** in those names.

[1] https://dart.dev/guides/language/effective-dart/style#dont-use-a-leading-underscore-for-identifiers-that-arent-private

pq:
Fixes #3360

/cc @bwilkerson

### DIFF
--- a/lib/src/rules/no_leading_underscores_for_local_identifiers.dart
+++ b/lib/src/rules/no_leading_underscores_for_local_identifiers.dart
@@ -66,6 +66,7 @@ class NoLeadingUnderscoresForLocalIdentifiers extends LintRule
     registry.addFormalParameterList(this, visitor);
     registry.addForPartsWithDeclarations(this, visitor);
     registry.addFunctionDeclaration(this, visitor);
+    registry.addFunctionDeclarationStatement(this, visitor);
     registry.addVariableDeclarationStatement(this, visitor);
   }
 }
@@ -115,10 +116,8 @@ class _Visitor extends SimpleAstVisitor<void> {
   }
 
   @override
-  void visitFunctionDeclaration(FunctionDeclaration node) {
-    if (node.parent is FunctionDeclarationStatement) {
-      checkIdentifier(node.name);
-    }
+  void visitFunctionDeclarationStatement(FunctionDeclarationStatement node) {
+    checkIdentifier(node.functionDeclaration.name);
   }
 
   @override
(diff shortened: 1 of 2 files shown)


lib/src/rules/no_leading_underscores_for_local_identifiers.dart
```diff
@@ -113,6 +114,13 @@ class _Visitor extends SimpleAstVisitor<void> {
     }
   }
 
+  @override
+  void visitFunctionDeclaration(FunctionDeclaration node) {
+    if (node.parent is FunctionDeclarationStatement) {
```
bwilkerson, line 119: Consider visiting `FunctionDeclarationStatement` and unconditionally invoking `checkIdentifier(node.functionDeclaration.name)`. It should be more performant because it will avoid a type check.

bwilkerson (approved):

lib/src/rules/no_leading_underscores_for_local_identifiers.dart
pq: Excellent.  Thanks!

lib/src/rules/no_leading_underscores_for_local_identifiers.dart
```diff
@@ -65,6 +65,8 @@ class NoLeadingUnderscoresForLocalIdentifiers extends LintRule
     registry.addDeclaredIdentifier(this, visitor);
     registry.addFormalParameterList(this, visitor);
     registry.addForPartsWithDeclarations(this, visitor);
+    registry.addFunctionDeclaration(this, visitor);
```
albertms10, line 68: Could this registry entry be safely removed, as we are not using it?

pq: Oh dear.  Great catch.  Thanks!

https://github.com/dart-lang/linter/pull/3364

PR #3370, opened 2022-04-29
`use_key_in_widget_constructors` to report at the constructor `name` or `returnType`

### LINKED ISSUE #3368: `use_key_in_widget_constructors` should be reported at the identifier of the constant constructors
With a case like:

```dart
import 'package:flutter/material.dart';

class MyWidget extends StatelessWidget {
  const MyWidget();

  @override
    return Container();
  }
}
```

there is no suggested fix for the lint.

The reason seems to be that the lint is reported at the `ConstructorDeclaration`, however the fix is expecting the `parent` to be a `ConstructorDeclaration`.

https://github.com/dart-lang/sdk/blob/dd007c2c71978996502dee1c582e277389b33969/pkg/analysis_server/lib/src/services/correction/dart/add_key_to_constructors.dart#L52-L54

I think the lint should be changed in this case, instead of the fix. In which the lint reports at the `SimpleIdentifier`, not the `ConstructorDeclaration`.

asashour:
Fixes #3368

### DIFF
--- a/lib/src/rules/use_key_in_widget_constructors.dart
+++ b/lib/src/rules/use_key_in_widget_constructors.dart
@@ -94,7 +94,7 @@ class _Visitor extends SimpleAstVisitor<void> {
           }
           return false;
         })) {
-      rule.reportLintForToken(node.firstTokenAfterCommentAndMetadata);
+      rule.reportLint(node.name ?? node.returnType);
     }
     super.visitConstructorDeclaration(node);
   }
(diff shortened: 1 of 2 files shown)


bwilkerson (approved):

pq (approved):

pq:
FYI, I'm in the process of finalizing the 1.23.0 release and want to hold off on landing this until that lands.  Hopefully later today...  Thanks!

pq:
@asashour: still waiting on https://github.com/flutter/flutter/pull/102833 to make it before I can land https://dart-review.googlesource.com/c/sdk/+/242391.  Getting there!

Issue #3334, opened 2022-04-07
`noop_primitive_operations` enhancements


asashour:
Update:

as communicated below:

>Would it make sense to expand it to covering additional cases where a non-String will automatically be converted, such as in string interpolations ('aaa ${x.toString()} zzz') and asserts, or in SDK APIs that take any object and convert it to a string using toString, such as StringBuffer.write.

~~## unnecessary_toString~~

~~## Description~~

~~Avoid calling `toString` on strings.~~

~~## Details~~

~~Avoid calling `toString` on strings.  The target is already a `String`.~~

~~## Kind~~

~~Unneeded code.~~

~~## Good Examples ~~

~~String s1 = '';~~
~~String s2 = s1;~~

~~## Bad Examples~~

~~String s1 = '';~~
~~String s2 = s1.toString();~~

~~## Discussion~~

~~`avoid_type_to_string` is related.~~

bwilkerson:
How often does this occur in real code?

Would it make sense to expand it to covering additional cases where a non-`String` will automatically be converted, such as in string interpolations (`'aaa ${x.toString()} zzz'`) and asserts, or in SDK APIs that take any object and convert it to a string using `toString`, such as `StringBuffer.write`.

pq:
[`noop_primitive_operations`](https://dart-lang.github.io/linter/lints/noop_primitive_operations.html) might fit this bill already?

asashour:
> [`noop_primitive_operations`](https://dart-lang.github.io/linter/lints/noop_primitive_operations.html) might fit this bill already?

Indeed, this is it.

> ... to expand it to covering additional cases ..

I guess yes, why not. The more value is added to the user, the better.

`'aaa ${x.toString()} zzz'` is already covered. Not sure about the rest.

a14n:
As #3379 has been merged can we close this issue?

PR #3379, opened 2022-05-06
catch '${x.toString()}' with noop_primitive_operations

a14n:
# Description

Improve` noop_primitive_operations` to catch `.toString()` in string interpolations.

Related to #3334

### DIFF
--- a/lib/src/rules/noop_primitive_operations.dart
+++ b/lib/src/rules/noop_primitive_operations.dart
@@ -80,7 +80,8 @@ class _Visitor extends SimpleAstVisitor<void> {
     var type = node.realTarget?.staticType;
     if (type == null) {
       // print(xxx.toString())
-      if (node.methodName.staticElement?.isDartCorePrint ?? false) {
+      if (node.methodName.staticElement.isDartCorePrint &&
+          node.argumentList.arguments.length == 1) {
         _checkToStringInvocation(node.argumentList.arguments.first);
       }
       return;
--- a/lib/src/rules/noop_primitive_operations.dart
+++ b/lib/src/rules/noop_primitive_operations.dart
@@ -28,6 +28,8 @@ string.toString();
 string = 'hello\n'
     'world\n'
     ''; // useless empty string
+
+'string with ${x.toString()}';
 ```
 ''';
 
@@ -47,6 +49,7 @@ class NoopPrimitiveOperations extends LintRule {
   ) {
     var visitor = _Visitor(this, context);
     registry.addAdjacentStrings(this, visitor);
+    registry.addInterpolationExpression(this, visitor);
     registry.addMethodInvocation(this, visitor);
   }
 }
@@ -63,6 +66,16 @@ class _Visitor extends SimpleAstVisitor<void> {
       if (literal.stringValue?.isEmpty ?? false) {
         rule.reportLint(literal);
       }
+    }
+  }
+
+  @override
+  void visitInterpolationExpression(InterpolationExpression node) {
+    var expression = node.expression;
+    if (expression is MethodInvocation &&
+        expression.methodName.name == 'toString' &&
+        expression.argumentList.arguments.isEmpty) {
+      rule.reportLint(expression.methodName);
     }
   }
 
--- a/lib/src/rules/noop_primitive_operations.dart
+++ b/lib/src/rules/noop_primitive_operations.dart
@@ -86,7 +86,13 @@ class _Visitor extends SimpleAstVisitor<void> {
       // print(xxx.toString())
       if (node.methodName.name == 'print' &&
           node.methodName.staticElement?.library?.name == 'dart.core') {
-        rule.reportLint(node.methodName);
+        var arg = node.argumentList.arguments.first;
+        if (arg is MethodInvocation &&
+            arg.methodName.name == 'toString' &&
+            arg.argumentList.arguments.isEmpty) {
+          rule.reportLint(node.methodName);
+          return;
+        }
       }
       return;
     }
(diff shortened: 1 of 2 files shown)


bwilkerson (approved):
The implementation is good, but again this is a breaking change for anyone using it. I think it's worth doing, but might require some additional work to land.

Also, we don't have a fix for this lint, but this case, at least, is an easy fix. (That doesn't need to block merging this PR.)

pq:
> The implementation is good, but again this is a breaking change for anyone using it.

This lint is enabled for (at least) a project internally so yeah, we'll have to see if any issues crop up.

> Also, we don't have a fix for this lint, but this case, at least, is an easy fix. 

Which is just to say, it'd be great to get this done before we publish a new linter with this improvement. 


PS: This looks great!

pq (approved):
👍

a14n:
Not sure if I should have created another PR but I added the case of `print(xxx.toString())`

Let me know if you prefer another PR (I understood you were not going to merge it quickly so I made this  change here to avoid too much concurrent pending branchs)

a14n:
I will add the corresponding fixes.

lib/src/rules/noop_primitive_operations.dart
```diff
@@ -66,10 +69,27 @@ class _Visitor extends SimpleAstVisitor<void> {
...
+  }
+
   @override
   void visitMethodInvocation(MethodInvocation node) {
     var type = node.realTarget?.staticType;
-    if (type == null) return;
+    if (type == null) {
+      // print(xxx.toString())
+      if (node.methodName.name == 'print' &&
+          node.methodName.staticElement?.library?.name == 'dart.core') {
```
pq, line 88: I wonder if we shouldn't prefer:

```dart
node.methodName.staticElement?.library?.isDartCore ?? false
```

(and if so, update `AvoidPrint` similarly).

Dunno.

@scheglov: do you all have comventions in analyzer?

pq:
> Not sure if I should have created another PR but I added the case of print(xxx.toString())

Nah.  Here is fine.

And thanks!

pq:
> I will add the corresponding fixes.

Thank you!

a14n:
> I will add the corresponding fixes.

Actually I don't find existing corrections for diagnostics `noop_primitive_operations`. (I tested on `1.2.toDouble()`)

lib/src/rules/noop_primitive_operations.dart
scheglov: `isDartCore` is definitely better.

Also, I would have added an extension like `isDartCorePrint` on `Element?` and checked that it is `FunctionElement` with the required name and library.

pq:
> Actually I don't find existing corrections for diagnostics noop_primitive_operations. (I tested on 1.2.toDouble())

Hmmmm.  Right.  It looks like that hasn't been added.  Looking in `error_fix_status.yaml` I see:

```yaml
    LintCode.noop_primitive_operations:
      status: needsEvaluation
```

lib/src/rules/noop_primitive_operations.dart
pq: Playing around with that now.  Stay tuned! 😄

pq: Landed.  Thanks @scheglov!

@a14n: could you migrate to using `isDartCorePrint` as introduced by https://github.com/dart-lang/linter/pull/3382?

lib/src/rules/noop_primitive_operations.dart
```diff
@@ -66,10 +70,21 @@ class _Visitor extends SimpleAstVisitor<void> {
...
+    _checkToStringInvocation(node.expression);
+  }
+
   @override
   void visitMethodInvocation(MethodInvocation node) {
     var type = node.realTarget?.staticType;
-    if (type == null) return;
+    if (type == null) {
+      // print(xxx.toString())
+      if (node.methodName.staticElement?.isDartCorePrint ?? false) {
```
scheglov, line 83: You don't need `?.` here - the extension already works on `Element?`.

Which itself might be a good lint? :-)

lib/src/rules/noop_primitive_operations.dart
```diff
@@ -66,10 +70,21 @@ class _Visitor extends SimpleAstVisitor<void> {
...
+        _checkToStringInvocation(node.argumentList.arguments.first);
```
scheglov, line 84: The code under analysis might be incomplete, and have `print();` without any arguments.

lib/src/rules/noop_primitive_operations.dart
a14n: Should it go into `invalid_null_aware_operator` ?

a14n:
Flutter will be diagnostic free once https://github.com/flutter/flutter/pull/103226 has landed.

lib/src/rules/noop_primitive_operations.dart
a14n: https://github.com/dart-lang/sdk/pull/48980 is an attempt to cover this case.

a14n:
ping

pq:
Apologies, @a14n, I wasn't able to make progress this week and now out for 2 of vacation.  I'll pick it up when I'm back.  Sorry for the delay!

a14n:
@pq Friendly reminder ;)

pq:
Thanks for the nudge and apologies for the slowness.  I've just kicked off an internal run.

pq:
Looks good.  Thanks again for your patience.  And the fix! 🎉

PR #3380, opened 2022-05-06
rev the min. dep on package:args

devoncarew:
- rev the min. dep on package:args
- add an additional markdown badge to the readme

`package:args` `2.1.0` introduced a new `mandatory` named param to one of it's APIs; it looks like linter is using this param; here's the error I saw against package:args 2.0.0 locally:

```
  error • tool/rule.dart:21:7 • The named parameter 'mandatory' isn't defined. Try correcting the name to an
          existing named parameter's name, or defining a named parameter with the name 'mandatory'. •
          undefined_named_parameter
```


pq (approved):
👍

PR #3381, opened 2022-05-06
remove stale dep on `package:glob`

pq:
See: https://dart-review.googlesource.com/c/sdk/+/243922

/cc @devoncarew @bwilkerson


devoncarew (approved):
👍

PR #3382, opened 2022-05-06
add isDartCorePrint extension method

pq:
Follow-up from: https://github.com/dart-lang/linter/pull/3379#discussion_r867170225

/cc @scheglov @a14n

### DIFF
--- a/lib/src/ast.dart
+++ b/lib/src/ast.dart
@@ -438,7 +438,8 @@ class _ElementVisitorAdapter extends GeneralizingElementVisitor {
 extension ElementExtension on Element? {
   bool get isDartCorePrint {
     var element = this;
-    if (element is! FunctionElement) return false;
-    return element.name == 'print' && element.library.isDartCore;
+    return element is FunctionElement &&
+        element.name == 'print' &&
+        element.library.isDartCore;
   }
 }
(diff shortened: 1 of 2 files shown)


lib/src/ast.dart
```diff
@@ -434,3 +434,11 @@ class _ElementVisitorAdapter extends GeneralizingElementVisitor {
     }
   }
 }
+
+extension ElementExtension on Element? {
+  bool get isDartCorePrint {
+    var element = this;
```
scheglov, line 440: In analyzer we prefer making this `final self = this;`.

lib/src/ast.dart
```diff
@@ -434,3 +434,11 @@ class _ElementVisitorAdapter extends GeneralizingElementVisitor {
...
+    if (element is! FunctionElement) return false;
```
scheglov, line 441: Even less important, but just note that `element is FunctionElement` would fit well into the following `&&` expression. Although there will be multiple lines :-)

scheglov (approved):

lib/src/ast.dart
pq: I do actually like that better.  Updated.  Thanks!

lib/src/ast.dart
pq: Ah, nice convention.

PR #3383, opened 2022-05-08
chore: Set permissions for GitHub actions

naveensrinivasan:
Restrict the GitHub token permissions only to the required ones; this way, even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

[Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

Signed-off-by: naveen <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -19,7 +19,6 @@ jobs:
   build:
     permissions:
       checks: write  # for coverallsapp/github-action to create new checks
-      contents: read  # for actions/checkout to fetch code
     name: ${{ matrix.linter-bot }}
     runs-on: ubuntu-latest
     strategy:
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -12,8 +12,14 @@ on:
 env:
   PUB_ENVIRONMENT: bot.github
 
+permissions:
+  contents: read
+
 jobs:
   build:
+    permissions:
+      checks: write  # for coverallsapp/github-action to create new checks
+      contents: read  # for actions/checkout to fetch code
     name: ${{ matrix.linter-bot }}
     runs-on: ubuntu-latest
     strategy:


devoncarew:
Thanks! I don't have context for scoping github actions permissions. If it is something we want to do, we'd likely want a policy for dart-lang/ (or at least some consistently between repos). cc @athomas and @godofredoc to see if they have familiarity with restricting github actions permissions.

pq:
Thanks @devoncarew!  And thanks @naveensrinivasan.  This seems reasonable but I'll want to hold off to get some input from @athomas and @godofredoc.

.github/workflows/build.yaml
```diff
@@ -12,8 +12,14 @@ on:
...
   PUB_ENVIRONMENT: bot.github
 
+permissions:
+  contents: read
+
 jobs:
   build:
+    permissions:
+      checks: write  # for coverallsapp/github-action to create new checks
+      contents: read  # for actions/checkout to fetch code
```
godofredoc, line 22: If we are setting default `contents: read` above, do we need this line here?

godofredoc:
> Thanks @devoncarew! And thanks @naveensrinivasan. This seems reasonable but I'll want to hold off to get some input from @athomas and @godofredoc.

Hi Phil,

This change LGTM. The PR is actually explicitly setting read permissions by default and explicitly setting the minimum write permissions for the workflow to work.

.github/workflows/build.yaml
naveensrinivasan: That wasn't required. I removed it. Thanks

pq:
Awesome.  Thanks for taking a look @godofredoc.

Thanks for the contribution @naveensrinivasan!

pq (approved):
💎

PR #3392, opened 2022-05-10
add unnecessary_null_aware_operator_on_extension_on_nullable

a14n:
Following up on https://github.com/dart-lang/linter/pull/3379#discussion_r867192583 to add a warning when using null aware operator with a member defined in an extension on nullable type.

```dart
extension on String? {
  bool get isNullOrEmpty {
    var self = this;
    return self == null || self.isEmpty;
  }
}

main() {
  String? s = null;
  s?.isNullOrEmpty; // WARN here
  s.isNullOrEmpty; // OK
}
```

Migrated from https://github.com/dart-lang/sdk/pull/48980 into linter


bwilkerson (approved):

lib/src/rules/unnecessary_null_aware_operator_on_extension_on_nullable.dart
```diff
@@ -0,0 +1,100 @@
...
+import 'package:analyzer/dart/element/element.dart';
+
+import '../analyzer.dart';
+
+const _desc =
+    r'Unnecessary null aware operator on extension on a nullable type';
+
+const _details = r'''
+
+Avoid null aware operator for member defined in an extension on a nullable type.
```
pq, line 16: ? => 

Avoid null aware operators for members defined in an extension on a nullable type.

pq (approved):
💎

Issue #3395, opened 2022-05-12
☂️  [task] remove `DartProject`


pq:
`DartProject` has never worked completely w/ server (see https://github.com/dart-lang/linter/issues/154) and since it's creation, `LinterContext` has been introduced which, moving forward, is the right way to answer the kinds of questions it was designed to address.  Since the lints that are currently using it are failing fast, my thinking is that we can safely remove all references to it from the existing lints as a non-breaking change.  We can then follow-up with a new set of changes that updates the (already) broken lints to use `LinterContext` methods.  These changes will produce a bunch of new diagnostics which though (now) correct may have ecosystem impact so we'll want to carefully vet their roll-out.

---

## Plan

- [x] remove `DartProject` lint references
  - [x] `directives_ordering` (https://github.com/dart-lang/linter/pull/3397)
  - [x] `package_api_docs` (https://github.com/dart-lang/linter/pull/3398)
  - [x] `package_prefixed_library_names` (https://github.com/dart-lang/linter/pull/3399)
- [x] deprecate `DartProject` (and visitor) in analyzer (https://dart-review.googlesource.com/c/sdk/+/244760)
- [x] remove `DartProject` from analyzer (in a breaking change release) (https://dart-review.googlesource.com/c/sdk/+/245240)
- [ ] fix functionality in:
  - [ ] `directives_ordering`
  - [ ] `package_api_docs`
  - [ ] `package_prefixed_library_names`

pq:
This is at a natural stopping point.  If the functionality additions get picked up, they can get tracked in their own issues.

PR #3399, opened 2022-05-12
remove `DartProject` references

pq:
See: https://github.com/dart-lang/linter/issues/3395

/cc @bwilkerson

### DIFF
--- a/lib/src/rules/package_prefixed_library_names.dart
+++ b/lib/src/rules/package_prefixed_library_names.dart
@@ -51,10 +51,7 @@ library my_package.src.private;
 bool matchesOrIsPrefixedBy(String name, String prefix) =>
     name == prefix || name.startsWith('$prefix.');
 
-class PackagePrefixedLibraryNames extends LintRule
-    implements ProjectVisitor, NodeLintRule {
-  DartProject? project;
-
+class PackagePrefixedLibraryNames extends LintRule {
   PackagePrefixedLibraryNames()
       : super(
             name: 'package_prefixed_library_names',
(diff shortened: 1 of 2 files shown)


lib/src/rules/package_prefixed_library_names.dart
```diff
@@ -84,28 +73,33 @@ class _Visitor extends SimpleAstVisitor<void> {
   _Visitor(this.rule);
 
   @override
+  // ignore: prefer_expression_function_bodies
```
bwilkerson, line 76: It seems wrong to recommend using an expression function body when there's no value to return. Somehow `void f() => null;` feels wrong.

test_data/rules/package_prefixed_library_names.dart
```diff
@@ -4,4 +4,5 @@
 
 // test w/ `dart test -N package_prefixed_library_names`
 
-library linter.not_where_it_should_be; //LINT
+// See: https://github.com/dart-lang/linter/issues/3395
+library linter.not_where_it_should_be; //FAILING
```
bwilkerson, line 8: Why would commenting out code that isn't being run cause the test to start failing? How confident are we that the commented out code really isn't being run?

bwilkerson (approved):

test_data/rules/package_prefixed_library_names.dart
pq: It *is* being run in the linter tests since they do take the code path that initializes project instances.  That's however the only path that does and so the tests here are out of sync w/ how server and the analyzer CLIs work.  Obviously, not good!

pq: Related: https://github.com/dart-lang/linter/issues/2969

PR #3400, opened 2022-05-12
fix coverage

pq:
Coverage is borked.

https://github.com/dart-lang/linter/runs/6414611200?check_suite_focus=true


pq:
Not working...

PR #3403, opened 2022-05-13
remove unnecessary interface

pq:
(This is implementing already by `LintRule`.)

/cc @bwilkerson @scheglov

### DIFF
--- a/lib/src/rules/no_leading_underscores_for_local_identifiers.dart
+++ b/lib/src/rules/no_leading_underscores_for_local_identifiers.dart
@@ -48,8 +48,7 @@ void print(String name) {
 ```
 ''';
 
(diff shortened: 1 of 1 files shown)


scheglov (approved):
Do we need this interface at all, if we have `Linter.registerNodeProcessors`?

Also, there might be a logic that I don't understand, but why do we have both `LintRule` and `Linter`? Why not one `LintRule` that has all of it?

pq:
> Do we need this interface at all, if we have Linter.registerNodeProcessors?

Maybe not!

> Why not one LintRule that has all of it?

It's a good question.  I can't recall why there's an extra abstract base class.  Might be something to tidy up.

Thanks!

PR #3404, opened 2022-05-13
`unawaited_futures` to handle string interpolation

### LINKED ISSUE #3377: `unawaited_futures` is not handling string interpolation
```dart
Future<String> someFuture() async => 'asdf';

Future<void> f() async {
  someFuture();
  var y = '${someFuture()}';
}
```

**Expected behavior**
Both invocations report the lint.

**Actual behavior**
Only the first invocation reports the lint.

asashour:
Fixes #3377

### DIFF
--- a/lib/src/rules/unawaited_futures.dart
+++ b/lib/src/rules/unawaited_futures.dart
@@ -68,7 +68,10 @@ class _Visitor extends SimpleAstVisitor<void> {
 
   @override
   void visitCascadeExpression(CascadeExpression node) {
-    node.cascadeSections.forEach(_visit);
+    var sections = node.cascadeSections;
+    for (int i = 0; i < sections.length; i++) {
+      _visit(sections[i]);
+    }
   }
 
   @override
--- a/lib/src/rules/unawaited_futures.dart
+++ b/lib/src/rules/unawaited_futures.dart
@@ -57,6 +57,7 @@ class UnawaitedFutures extends LintRule {
     var visitor = _Visitor(this);
     registry.addExpressionStatement(this, visitor);
     registry.addCascadeExpression(this, visitor);
+    registry.addInterpolationExpression(this, visitor);
   }
 }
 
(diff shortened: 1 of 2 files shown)


test_data/rules/unawaited_futures.dart
```diff
@@ -71,6 +71,10 @@ foo12() async {
   x..[0] = fut();
 }
 
+foo13() async {
+  var y = '${fut()}'; //LINT
```
bwilkerson, line 75: An uncommon corner case that we don't need to address in this PR, but should the lint flag in the following location:
```
var a;
var b = '${a = fut()}';
```

lib/src/rules/unawaited_futures.dart
```diff
@@ -67,13 +68,7 @@ class _Visitor extends SimpleAstVisitor<void> {
...
   @override
   void visitCascadeExpression(CascadeExpression node) {
-    for (var expr in node.cascadeSections) {
-      if ((expr.staticType?.isDartAsyncFuture ?? false) &&
-          _isEnclosedInAsyncFunctionBody(expr) &&
-          expr is! AssignmentExpression) {
-        rule.reportLint(expr);
-      }
-    }
+    node.cascadeSections.forEach(_visit);
```
bwilkerson, line 71: Although it's textually cleaner in this case, `forEach` is really slow. A for-in loop is a little better, but a normal for loop is by far the fastest way to iterate:
```
var sections = node.cascadeSections;
for (int i = 0; i < sections.length; i++) {
  _visit(sections[i]);
}
```
I don't know how important the performance difference is in this case, but I'm starting to think that we need to be more aware of performance in lint code.

bwilkerson (approved):

test_data/rules/unawaited_futures.dart
asashour: I think there are test cases against this, e.g. the below doesn't lint

```dart
foo4() async {
  var x = fut();
  return x;
}
```

and the code in

https://github.com/dart-lang/linter/blob/bee643570ba670720b479f0248e28e788faac1a5/lib/src/rules/unawaited_futures.dart#L82

lib/src/rules/unawaited_futures.dart
asashour: Thanks, done.

pq (approved):

PR #3415, opened 2022-05-17
`prefer_final_parameters` to handle `DefaultFormalParameter`

### LINKED ISSUE #3407: `prefer_final_parameters` and super parameters
If you turn on `prefer_final_parameters`, you would get complaints about not using `final` before super parameters. But I think it should be in alignment with self parameters, for which the linter doesn't complain about in a similar situation.

The code that the linter currently prefers now with `prefer_final_parameters: true`:

```dart
class A {
  final String? a;

  const A({this.a});
}

class B extends A {
  final String? b;

  const B({this.b, final super.a});
}
```

The code that I think the linter should prefer with `prefer_final_parameters: true`:

```dart
class A {
  final String? a;

  const A({this.a});
}

class B extends A {
  final String? b;

  const B({this.b, super.a});
}
```

asashour:
Fixes #3407

### DIFF
--- a/lib/src/rules/prefer_final_parameters.dart
+++ b/lib/src/rules/prefer_final_parameters.dart
@@ -96,6 +96,9 @@ class _Visitor extends SimpleAstVisitor<void> {
       FormalParameterList? parameters, FunctionBody body) {
     if (parameters != null) {
       for (var param in parameters.parameters) {
+        if (param is DefaultFormalParameter) {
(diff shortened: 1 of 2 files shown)


bwilkerson (approved):

pq (approved):
💎

parlough:
Thanks @asashour!

PR #3422, opened 2022-05-20
`prefer_initializing_formals` iteration optimizations

pq:
Our [benchmark bot](https://github.com/dart-lang/linter/runs/6532600308?check_suite_focus=true), while not rigorous, does suggest a significant improvement.

Before:

image

After:

image

Note that this is a recommended lint so *any* improvement will be felt by lots of folks.

/cc @srawlins @scheglov @bwilkerson


scheglov (approved):
It looks that this is objectively an improvement.
And there are no any semantic mistakes as far as I can see.
So, LGTM.

Subjectively it looks less fancy, but I'm not going to argues against performance numbers :-)

bwilkerson (approved):

PR #3423, opened 2022-05-20
`prefer_const_constructors_in_immutables`: optimize out sync* and extra iteration

pq:
Another wee optimization on a `flutter` recommended lint.  (Some context in https://github.com/dart-lang/sdk/issues/32102.)

[BEFORE](https://github.com/dart-lang/linter/runs/6497647704?check_suite_focus=true):

image

[AFTER](https://github.com/dart-lang/linter/runs/6532900300?check_suite_focus=true):

image

A pretty significant improvement!

/cc @srawlins @scheglov @bwilkerson 

/fyi @goderbauer

### DIFF
--- a/lib/src/rules/prefer_const_constructors_in_immutables.dart
+++ b/lib/src/rules/prefer_const_constructors_in_immutables.dart
@@ -120,20 +120,23 @@ class _Visitor extends SimpleAstVisitor<void> {
 
   bool _hasImmutableAnnotation(ClassElement clazz) {
     var selfAndInheritedClasses = _getSelfAndInheritedClasses(clazz);
-    var selfAndInheritedAnnotations =
-        selfAndInheritedClasses.expand((c) => c.metadata).map((m) => m.element);
-    return selfAndInheritedAnnotations.any(_isImmutable);
+    for (var cls in selfAndInheritedClasses) {
+      if (cls.metadata.any((m) => _isImmutable(m.element))) return true;
+    }
+    return false;
   }
 
   bool _hasMixin(ClassElement clazz) => clazz.mixins.isNotEmpty;
 
   static Iterable<ClassElement> _getSelfAndInheritedClasses(
-      ClassElement self) sync* {
+      ClassElement self) {
+    var elements = <ClassElement>[];
     ClassElement? current = self;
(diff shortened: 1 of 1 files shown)


lib/src/rules/prefer_const_constructors_in_immutables.dart
```diff
@@ -120,20 +120,22 @@ class _Visitor extends SimpleAstVisitor<void> {
...
+      if (cls.metadata.any((m) => _isImmutable(m.element))) return true;
+    }
+    return false;
   }
 
   bool _hasMixin(ClassElement clazz) => clazz.mixins.isNotEmpty;
 
-  static Iterable<ClassElement> _getSelfAndInheritedClasses(
-      ClassElement self) sync* {
+  static Iterable<ClassElement> _getSelfAndInheritedClasses(ClassElement self) {
```
scheglov, line 131: You can change the return type to `List`, so that consumers will know that it is not necessary to call `toList()` in case if they iterate multiple times. Or not, if you know that the4 client does exactly one iteration.

scheglov: The name is not very specific. It says "inherited", but actually wants "super".

lib/src/rules/prefer_const_constructors_in_immutables.dart
```diff
@@ -120,20 +120,22 @@ class _Visitor extends SimpleAstVisitor<void> {
...
+    var elements = <ClassElement>[];
```
scheglov, line 132: You don't actually need `elements` because `seenElements` is `LinkedHashSet`, so it preserves the order (if you need it at all), and you could `return seenElements.toList()` instead.

scheglov (approved):

lib/src/rules/prefer_const_constructors_in_immutables.dart
pq: Updated.  Thanks!

lib/src/rules/prefer_const_constructors_in_immutables.dart
pq: Ah!  Great catch.  Thanks.

PR #3424, opened 2022-05-20
migrate from `sync*`

pq:
WIP


bwilkerson (commented):
What makes this a questionable change? It might not be as much as could be done, but it's a step in the right direction.

scheglov:
> What makes this a questionable change? It might not be as much as could be done, but it's a step in the right direction.

I cannot argue with numbers - the timings are better now.
So, I think we can land this.

But the difference is just `3 ms` on `28 ms`, now sure what is the target codebase, but in general it does not seem very much.

bwilkerson (approved):

PR #3425, opened 2022-05-25
remove unnecessary registry.addCompilationUnit

a14n:

Issue #2081, opened 2020-04-30
tooling to sanity check lints


pq:
A good lint has a few properties that would be easy to check for statically.  We should consider some tooling.

A few quick thoughts:

* ensure `register*` calls have corresponding `visit*` methods implemented
* flag uses of `RecursiveAstVisitor`s
* flag naming convention violations
* warn against (certain) internal analyzer API uses
* <your idea here>

(Note that some of this is currently done in the build which is great for lints in this repo but ones hosted elsewhere could benefit from similar validation.)

/fyi @bwilkerson

srawlins:
But who lints the linters? 😛

bwilkerson:
They're self-linting.

pq:
Turtles all the way down!

pq:
#3426 introduces a basic framework for self-checking via lints.  I started with a check to ensure that we have visit methods that correspond to all node registrations.

PR #3426, opened 2022-05-25
automate validation of node visitor registrations

pq:
Registrations of nodes without corresponding visit method implementations are no-ops.  Removing these unnecessary registrations removes a needless runtime call to visit and makes the code easier to understand.  This code introduces automated validation as a build step.  As recently as today, we've had PRs to fix these issues manually (see for example #3425.) 

See also: https://github.com/dart-lang/linter/issues/2081.

/cc @bwilkerson 

/fyi @a14n


bwilkerson (commented):
The code looks good.

I still wonder whether making this a test wouldn't be easier for lint contributors so that they don't have to learn to run a tool in addition to the normal tests. But maybe catching problems on the CI is good enough.

pq:
I'm nominally inclined to run this in the CI and am kind of imagining these checks becoming part of a presubmit step which could potentially emit warnings too (which would just get lost in the test output).

I don't feel super strongly though...

bwilkerson (approved):

PR #3431, opened 2022-05-27
implement `discarded_futures`

### LINKED ISSUE #3429: proposal: `discarded_futures`
## Description

Don't discard futures in async functions.

## Details

Making asynchronous calls in non-`async` functions is usually the sign of a programming error.  In general these functions should be marked `async` and such futures should likely be awaited (as enforced by `unawaited_futures`).

In case you _really do_ want to discard a Future in a non-`async` function, the recommended way is to use `unawaited` from `dart:async`. The `// ignore` and `// ignore_for_file` comments also work.

## Kind

Error

## Good Examples 

```dart
Future<void> recreateDir(String path) async {
  await deleteDir(path);
  await createDir(path);
}

Future<void> deleteDir(String path) async {}

Future<void> createDir(String path) async {}
```

## Bad Examples

```dart
void recreateDir(String path) {
  deleteDir(path);
  createDir(path);
}

Future<void> deleteDir(String path) async {}

Future<void> createDir(String path) async {}
```

## Discussion

See #836, #2953 and [`unawaited_futures`](https://dart-lang.github.io/linter/lints/unawaited_futures.html).

/fyi @eernstg @diegovar @dowski @bsutton

### Discussion checklist

- [x] List any existing rules this proposal modifies, complements, overlaps or conflicts with.
- [x] List any relevant issues (reported here, the [SDK Tracker], or elsewhere).
- [x] If there's any prior art (e.g., in other linters), please add references here.
- [x] If this proposal corresponds to [Effective Dart] or [Flutter Style Guide] advice, please call it out. (If there isn’t any corresponding advice, should there be?)
- [x] If this proposal is motivated by real-world examples, please provide as many details as you can.  Demonstrating potential impact is _especially valuable_.

[Writing Lints]: https://github.com/dart-lang/linter/blob/master/doc/WritingLints.MD
[Effective Dart]: https://dart.dev/guides/language/effective-dart
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[SDK Tracker]: https://github.com/dart-lang/sdk/issues

pq:
Fixes #3429

/cc @bwilkerson

### DIFF
--- a/lib/src/rules/discarded_futures.dart
+++ b/lib/src/rules/discarded_futures.dart
@@ -52,10 +52,30 @@ class DiscardedFutures extends LintRule {
             group: Group.style);
 
   @override
-  void registerNodeProcessors(NodeLintRegistry registry, LinterContext context) {
+  void registerNodeProcessors(
+      NodeLintRegistry registry, LinterContext context) {
     var visitor = _Visitor(this);
     registry.addFunctionDeclaration(this, visitor);
     registry.addMethodDeclaration(this, visitor);
+  }
+}
+
+class _InvocationVisitor extends RecursiveAstVisitor<void> {
+  bool foundFuture = false;
+
+  @override
+  void visitFunctionExpressionInvocation(FunctionExpressionInvocation node) {
+    if (foundFuture) return;
+    foundFuture = node.staticInvokeType.isFuture;
+    super.visitFunctionExpressionInvocation(node);
+  }
+
+  @override
+  void visitMethodInvocation(MethodInvocation node) {
+    if (foundFuture) return;
+    if (node.methodName.staticElement.isDartAsyncUnawaited) return;
+    foundFuture = node.staticInvokeType.isFuture;
+    super.visitMethodInvocation(node);
   }
 }
 
@@ -63,6 +83,13 @@ class _Visitor extends SimpleAstVisitor {
   final LintRule rule;
 
   _Visitor(this.rule);
+
+  bool discardsFuture(FunctionBody body) {
+    if (body.isAsynchronous) return false;
+    var visitor = _InvocationVisitor();
+    body.accept(visitor);
+    return visitor.foundFuture;
+  }
 
   @override
   visitFunctionDeclaration(FunctionDeclaration node) {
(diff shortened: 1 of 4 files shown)


lib/src/rules/discarded_futures.dart
```diff
@@ -0,0 +1,125 @@
...
+    foundFuture = node.staticInvokeType.isFuture;
+    super.visitFunctionExpressionInvocation(node);
+  }
+
+  @override
+  void visitMethodInvocation(MethodInvocation node) {
+    if (foundFuture) return;
+    if (node.methodName.staticElement.isDartAsyncUnawaited) return;
+    foundFuture = node.staticInvokeType.isFuture;
+    super.visitMethodInvocation(node);
```
bwilkerson, line 78: We might need to stop recursing when we get to a local function or closure that is async:
```
void f(String command) {
  handler[command] = () async {
    await g();
    g();
  }
}
Future<int> g() async => 0;
```
If it's awaited inside then there shouldn't be any error; if it isn't awaited then it's a different error.

lib/src/rules/discarded_futures.dart
```diff
@@ -0,0 +1,125 @@
...
+    if (body.isAsynchronous) return false;
+    var visitor = _InvocationVisitor();
+    body.accept(visitor);
+    return visitor.foundFuture;
+  }
+
+  @override
+  visitFunctionDeclaration(FunctionDeclaration node) {
+    if (discardsFuture(node.functionExpression.body)) {
+      rule.reportLint(node.name);
```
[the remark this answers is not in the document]
pq: I agree!  For some reason I was sort of stuck on the solution (which I think would generally be to make the function `async`) but even if that is what folks want to do, knowing what invocation got them there would be a good thing.  Thanks!

lib/src/rules/discarded_futures.dart
pq: Oh, nice one!  Thanks.

lib/src/rules/discarded_futures.dart
```diff
@@ -0,0 +1,125 @@
...
+    if (foundFuture) return;
+    foundFuture = node.staticInvokeType.isFuture;
+    super.visitFunctionExpressionInvocation(node);
+  }
+
+  @override
+  void visitMethodInvocation(MethodInvocation node) {
+    if (foundFuture) return;
+    if (node.methodName.staticElement.isDartAsyncUnawaited) return;
+    foundFuture = node.staticInvokeType.isFuture;
```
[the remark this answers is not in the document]
pq: Great catches.  Done.

lib/src/rules/discarded_futures.dart
```diff
@@ -0,0 +1,166 @@
...
+    if (body.isAsynchronous) return;
+    var visitor = _InvocationVisitor(rule);
+    body.accept(visitor);
+  }
+
+  void checkVariables(VariableDeclarationList variables) {
+    if (variables.type?.type?.isFuture ?? false) return;
+    for (var variable in variables.variables) {
+      var initializer = variable.initializer;
+      if (initializer is FunctionExpression) {
```
bwilkerson, line 109: Do we want to check when the function expression isn't at the top level? Consider
```
var x = someFunction((e) => asyncFunction(e));
```

test/rules/discarded_futures_test.dart
```diff
@@ -0,0 +1,198 @@
...
+
+Future<void> deleteDir(String path) async {}
+Future<void> createDir(String path) async {}
+''', [
+      lint('discarded_futures', 34, 9),
+      lint('discarded_futures', 53, 9),
+    ]);
+  }
+
+  test_function_closure() async {
```
bwilkerson, line 77: This is an interesting case. A slight variation would also be interesting:
```
Future<void> f() async {
  () {
    createDir('.');
  }();
}
Future<void> createDir(String path) async {}
```

bwilkerson (approved):

PR #3435, opened 2022-05-30
Avoid `depend_on_referenced_packages` false positive in colocated test directory

rorystephenson:
# Description

When test files are colocated (in a `test/` subdirectory of `lib/src/`) the `depend_on_referenced_packages` lint does not consider dev dependencies as part of the file's dependencies and thus gives false positives for all referenced dev dependencies.

### DIFF
--- a/lib/src/rules/depend_on_referenced_packages.dart
+++ b/lib/src/rules/depend_on_referenced_packages.dart
@@ -72,7 +72,8 @@ class DependOnReferencedPackages extends LintRule {
         for (var dep in dependencies)
           if (dep.name?.text != null) dep.name!.text!,
       if (devDependencies != null &&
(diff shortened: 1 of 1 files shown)


pq:
fyi @jakemac53

jakemac53:
All files under `lib` are considered public, anybody could import them with a `package:` import.

rorystephenson:
> All files under `lib` are considered public, anybody could import them with a `package:` import. So this is working as intended. I don't think we want to special case directories called `test` or anything like that.

The proposal is to use the existing detection for test directories (see the PR code), not a special case for directories called "test/".

srawlins:
I agree. (Sorry for the delay.) I think we can only support the package conventions laid out [by convention](https://dart.dev/tools/pub/package-layout). I'll close this for now.

PR #3453, opened 2022-06-14
add no_top_public_members_in_executable_libraries

### LINKED ISSUE #1258: Ban public top-level members in a test
### Proposal

Given a library with URI with pattern `/test/.*_test.dart`, report any public top-level members except `main`. "Top-level members in a test should be private (or relocated to within `main`), to prevent unused members."

### Secondary proposal

Given a library which does not live in it's declaring package's public API (i.e. files in `bin/`, `test/`, and `tool/`), report any public top-level members except `main`.

### Justification

This lint helps prevent code bloat. Without full-program analysis, we cannot tell if a public member is used anywhere, so it is not reported by the analyzer as unused_variable or something similar. For top-level members, only _private_ ones can be rigorously checked for usage. But in _tests_ (and `bin/` and `tool/` scripts), which are typically their own entry points and never imported by another library, we want to be able to report unused top-level members.

This includes top-level variables, constants, functions, typedefs, enums, mixins, and classes.

This is a counter-request to #685. I think this is better choice for a few reasons. #685 proposes either a change to the analyzer's unused analysis to take into account the library URI ("does it end with _test.dart?"), leading to possible false positives. I don't think that's going to go well. The alternative is to implement a new unused analysis in the linter. Unused analysis is expensive though, and adding a second one in linter (duplicating a ton of code) will make it _more_ expensive.

### Downsides

One downside to recommending private top-level members is that it might be counter to what users typically do today. I think a lot of people just write regular, public, top-level consts and helper functions. Personally, I always write private top-level members, so that I know when something becomes unused. Perhaps a study should be done on what most people do. #685 does not have this downside.

a14n:
# Description

Top-level members in an executable library should be private (or relocated to within main), to prevent unused members.

**BAD:**

```dart
main() {}
void f() {}
```

**GOOD:**

```dart
main() {}
void _f() {}
```

Fixes #1258

### DIFF
--- a/example/all.yaml
+++ b/example/all.yaml
@@ -94,6 +94,7 @@ linter:
     - no_leading_underscores_for_local_identifiers
     - no_logic_in_create_state
     - no_runtimeType_toString
+    - no_top_public_members_in_executable_libraries
     - non_constant_identifier_names
     - noop_primitive_operations
     - null_check_on_nullable_type_parameter
--- a/lib/src/rules.dart
+++ b/lib/src/rules.dart
@@ -96,6 +96,7 @@ import 'rules/no_leading_underscores_for_library_prefixes.dart';
 import 'rules/no_leading_underscores_for_local_identifiers.dart';
 import 'rules/no_logic_in_create_state.dart';
 import 'rules/no_runtimeType_toString.dart';
+import 'rules/no_top_public_members_in_executable_libraries.dart';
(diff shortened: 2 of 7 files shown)


a14n:
PTAL

bwilkerson (commented):
The implementation looks fine; I can't speak to the question of whether this is a lint that we want to add (and that question should be discussed in the issue, not here).

PR #3454, opened 2022-06-14
use_late_for_private_fields_and_variables: handling of enum

### LINKED ISSUE #3421: use_late_for_private_fields_and_variables on final nullable field should not trigger on enum
**Describe the issue**

use_late_for_private_fields_and_variables on final nullable field should not trigger on enum if unused internally.

**To Reproduce**
This way it shows two lints info, use_late_for_private_fields_and_variables and unused_field.
```dart
enum A {
  a('');

  const A([this._private]);
  /// Lint -> Use late for private members with a non-nullable type.dart[use_late_for_private_fields_and_variables]
  ///
  /// The value of the field '_private' isn't used.
  /// Try removing the field, or using it.dart[unused_field](https://dart.dev/diagnostics/unused_field)
  final String? _private;
}

class B {
  const B(this.public, [this._private]);
  final String? _private;
  final A public;

  @override
  String toString() => _private ?? '';
}
```

If I use the field inside the same library/file the unused lint stops showing.
```dart
enum A {
  a('');

  const A([this._private]);
  /// Lint -> Use late for private members with a non-nullable type.dart[use_late_for_private_fields_and_variables]
  final String? _private; 
}

class B {
  const B(this.public, [this._private]);
  final String? _private;
  final A public;

  @override
  String toString() => (public._private ?? '') + (_private ?? ''); /// Using the field here to see the unused lint go away.
}
```

If I use the field inside the enum, both linters stop showing.
```dart
enum A {
  a('');

  const A([this._private]);
  final String? _private;  /// Not linting because it's being used in [toString]

  @override
  String toString() => _private ?? '';
}

class B {
  const B(this.public, [this._private]);
  final String? _private;
  final A public;

  @override
  String toString() => _private ?? '';  /// Doesn't have to be here anymore.
}
```

The lint doesn't activate on classes, just enums, from my testing.

**Expected behavior**
The linter should not warn if the field is final inside enums as well.

a14n:
Fixes #3421

### DIFF
--- a/lib/src/rules/use_late_for_private_fields_and_variables.dart
+++ b/lib/src/rules/use_late_for_private_fields_and_variables.dart
@@ -129,6 +129,11 @@ class _Visitor extends UnifyingAstVisitor<void> {
   }
 
   @override
+  void visitEnumDeclaration(EnumDeclaration node) {
+    // See: https://dart.dev/tools/diagnostic-messages#late_final_field_with_const_constructor
+  }
+
+  @override
   void visitFieldDeclaration(FieldDeclaration node) {
     for (var variable in node.fields.variables) {
       var parent = node.parent;
--- a/test_data/rules/use_late_for_private_fields_and_variables.dart
+++ b/test_data/rules/use_late_for_private_fields_and_variables.dart
@@ -80,3 +80,11 @@ class ContentChannelException {
 
   Exception get exception => _exception!;
 }
+
+enum MyEnum {
+  a('');
+
+  const MyEnum([this._private]);
+
+  final String? _private; // OK (enum constructor must be const)
+}


a14n:
PTAL

pq (approved):
👍

bwilkerson (approved):
I believe that this fixes the problem, but I don't like the way the lint was originally implemented. It seems entirely possible for there to be a memory leak related to `latables`.

a14n:
Could we merge this PR?

> I believe that this fixes the problem, but I don't like the way the lint was originally implemented. It seems entirely possible for there to be a memory leak related to `latables`.

We could address that in a another PR. Could you create an issue about that and give some advices how to improve the code?

bwilkerson:
Sorry, yes. I didn't mean that this PR needed to be changed before it could land.

PR #3459, opened 2022-06-15
prefer_final_locals: handling of declaration list

a14n:
# Description

`prefer_final_locals` should trigger only if all variables in the declaration list can be final.

Related to #3444

### DIFF
--- a/lib/src/rules/prefer_final_locals.dart
+++ b/lib/src/rules/prefer_final_locals.dart
@@ -87,6 +87,10 @@ class _Visitor extends SimpleAstVisitor<void> {
         return;
       }
     }
-    rule.reportLint(node);
+    if (node.keyword != null) {
+      rule.reportLintForToken(node.keyword);
+    } else if (node.type != null) {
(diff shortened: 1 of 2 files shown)


lib/src/rules/prefer_final_locals.dart
```diff
@@ -71,20 +71,22 @@ class _Visitor extends SimpleAstVisitor<void> {
...
+        return;
+      }
+      var declaredElement = variable.declaredElement;
+      if (function != null &&
+          declaredElement != null &&
+          function.isPotentiallyMutatedInScope(declaredElement)) {
+        return;
+      }
     }
+    rule.reportLint(node);
```
bwilkerson, line 90: It would be good to reduce the highlight range here. If we can't get it smaller, consider highlighting the `var` or type rather than the whole node.

bwilkerson (approved):

pq (approved):
💎

pq:
Fantastic.  Thanks!

PR #3470, opened 2022-06-23
prefer_final_locals: no lint on fields

a14n:
# Description

#3459 has introduced a bug where class fields are no subject to diagnostics. This PR fixes the problem.


bwilkerson (approved):

pq (approved):
💎

PR #3482, opened 2022-06-30
Allow a simple override to make a @protected method public

### LINKED ISSUE #3475: unnecessary_override reports method M when super method is @protected and M is not
**Describe the issue**
A method override might exist to make a `@protected` method public, which does not seem to violate the existing [protected contract](https://pub.dev/documentation/meta/latest/meta/protected-constant.html). Such methods should not be reported as unnecessary. I feel like they're similar to an override which changes the types of parameters.

**To Reproduce**

```dart
// lib1.dart

class C {
  @protected
  void m() {}
}

class D extends C {
  @override
  void m() => super.m();
}

// lib2.dart

import 'lib1.dart';

void f() => D().m();
```

**Expected behavior**

No `unnecessary_overrides` report.

srawlins:
# Description

Allow a simple override to make a @protected method public.

Fixes #3475


lib/src/rules/unnecessary_overrides.dart
```diff
@@ -127,36 +140,55 @@ abstract class _AbstractUnnecessaryOverrideVisitor extends SimpleAstVisitor {
...
     var metadata = declaration.declaredElement?.metadata;
     if (metadata != null) {
       for (var annotation in metadata) {
-        if (!annotation.isOverride) {
-          return true;
-        }
+        if (annotation.isOverride) continue;
+        if (annotation.isProtected && _inheritedMethod.hasProtected) continue;
+        // TODO(srawlins): The above check probably holds for any annotation
+        // which has a private constructor. That is, overriding
```
bwilkerson, line 152: It isn't particularly important, but I think the criteria ought to be "zero arguments" or "identical arguments" instead of "private constructor", though that probably only applies reliably to our current set of annotations. I could see an annotation that was defined in such a way that it was not inherited and hence needed to be repeated at different layers in the hierarchy. Generalizing in this situation is probably error-prone, so this might not need to be a TODO.

bwilkerson (approved):

lib/src/rules/unnecessary_overrides.dart
srawlins: Good point. I agree that there are subtle bugs that lie this way. The `protected` request comes from google internal, so I know it has an effect, and that people are expanding visibility this way.

bwilkerson: Yeah, my comment was purely about the TODO, not about the code.

One question: Is `@protected` inherited, and if so does overriding actually stop the inheritance?

Issue #3485, opened 2022-06-30
integration_tests depend on a 'packages' file


srawlins:
**Describe the issue**

The integration tests seem to depend on a `--packages` flag, which is now failing with analyzer 4.2.0.

See CI failures here: https://github.com/dart-lang/linter/runs/7123357781?check_suite_focus=true

Reproduce with `dart run test test/integration_test.dart`.

srawlins:
@scheglov I'm sort of guessing here, and I know work was done recently to remove some legacy support for 'packages' files. Can you confirm this landed recently, and may only be showing up in 4.2.0?

srawlins:
Looks like linter sets 'packages' files via analyzer's `DriverOptions.packageConfigPath`.

srawlins:
Seems fixable by adding several `.dart_tool/package_config.json` files.

scheglov:
Yes, we removed support for `.packages` just before `analyzer 4.2.0`.

PR #3486, opened 2022-06-30
Remove all use of --packages flag in tests

srawlins:
# Description

Removes all use of the `--packages` flag. This flag seems to not work any longer, perhaps starting with analyzer 4.2.0.

This is just to get CI back up; I do not remove the existence of the flag in this PR.

Fixes https://github.com/dart-lang/linter/issues/3485


scheglov (approved):

lib/src/cli.dart
```diff
@@ -68,6 +68,7 @@ Future runLinter(List<String> args, LinterOptions initialLintOptions) async {
     ..addMultiOption('rules',
         help: 'A list of lint rules to run. For example: '
             'avoid_as,annotate_overrides')
+    // TODO(srawlins): Remove this flag; it does not work any more.
```
bwilkerson, line 71: As a first step, consider marking the option as hidden and updating the documentation to indicate that it is ignored.

Issue #2555, opened 2021-03-23
Hint and/or lint for changing `List.from` to `List.of`


rakudrama:
Many uses of `List.from` can be changed to `List.of`.
The reasons to do so:

- `List.of` is nicer to use because the extra constraint helps type inference and editor suggestions
- `List.of` is more efficient
- There is an education problem that since `List.of` is newer, there are lots of poor examples of `List.from` that could have been migrated to `List.of`.

Examples that could be hints (i.e. no false positives):

- The explicit or inferred type argument _`E`_ in `List<`_`E`_`>.from(iterable)` matches the static type of the iterable argument
- The explicit type argument _`E`_ in `List<`_`E`_`>.from(iterable)` is a supertype of the static type of the iterable argument

More complicated is where `dynamic` creeps in.
I have seen examples like:
```dart
void foo(Map<String, String> map) {
  for (final key in List.from(map.keys)) {
    print(key.isEmpty);
...
```
If we change `List.from` to `List.of`, the inferred type of `key` moves from `dynamic` to `String`. This would be helpful to the developer for completion suggestion, since `key.isE•` can now only complete to `key.isEmpty`, and not `key.isEven`.
We would want to suggest the replacement if the new inferred types do not lead to new or different errors or warnings or a different resolution of elements (e.g. a dynamic instance method call becoming a static extension method call).

pq:
Does the same advice extend to `Set.of`?

a14n:
Currious : is `List<E>.of(iterableOfE)` better than `iterableOfE.toList()` ?

If you look at https://api.dart.dev/stable/2.8.4/dart-core/Iterable/toList.html you can see that the implementation relies on `List.from`. Should it be changed to use `List.of` ?

a14n:
> Does the same advice extend to `Set.of`?

And `Map.of`?

rakudrama:
The advice extends to `Set.of` and `Map.of`.

@a14n: Yes, we should change that `List.from` to `List.of`. That code dates back to Dart 1 before there was a `List.of`, and the proposed hint would help find it.

---

@a14n: Regarding `iterableOfE`, less clear cut, so I would not suggest a similar hint.

They are different in that the run time type value `E` for `toList` comes from the object, and for `List<E>.of` from the context of the call.
I would use `List.of` if I wanted to update an element of the list; for lists that are not modified, `toList` is fine.

Reasons to use `iterableOfE.toList()`:

- It is more concise
- Plays well with type infererence
[...]

PR #3513, opened 2022-07-12
add unused_top_members_in_executable_libraries

### LINKED ISSUE #2395: Report unused public elements in an entrypoint library

If a library has an entrypoint, I'd like to report any publicly accessible elements which are not used within the library as unused.

An entrypoint is a top-level `main` function, or a function annotated with `@pragma(vm:entry-point)`. These files are typically:

* a VM script in `bin/`,
* a web app in `web/`,
* a VM tool in `tool/`,
* a test like `test/**/*_test.dart`,

I think there will be very very few false positives. And perhaps any false positive indicates a code smell in library arrangement (the analyzer codebase I believe has some test files which reference another test file, each of which has a `main`).

Any element which is _not_ publicly visible should already be reported with analyzer's `unused_local_variable`, `unused_element`, or `unused_field` diagnostics. This lint would cover publicly visible elements:

* Any public class instance members (fields, methods, constructors), even those on private classes and mixins.
* Any public top-level functions, variables, classes, enums, mixins, extensions, and typedefs.

**Examples**

```dart
// foo_test.dart

void main() {
  test(...);
}

// Unused class
class MockStream extends Mock implements Stream {
  // Unused field
  int y = 7;

  // Unused method
  void m() {}
}

// Unused top-level function
void helper() {}

// Unused top-level variable
const x = 1;
```

**Additional context**

This is an alternative route to #685 and #1258.

a14n:
# Description

Report unused top-level members in executable libraries.

Fixes #2395

### DIFF
--- a/lib/src/rules/unreachable_from_main.dart
+++ b/lib/src/rules/unreachable_from_main.dart
@@ -17,8 +17,10 @@ const _desc = 'Unreachable top-level members in executable libraries.';
 const _details = r'''
 
 Top-level members in an executable library should be used directly inside this
-library.  Executable libraries are usually never imported and it's better to
-avoid defining unused members.
+library.  An executable library is a library that contains a `main` top-level
+function or that contains a top-level function annotated with
+`@pragma('vm:entry-point')`).  Executable libraries are usually never imported
+and it's better to avoid defining unused members.
 
 This rule assumes that an executable library isn't imported by other files
 except to execute its `main` function.
@@ -100,6 +102,8 @@ class _Visitor extends SimpleAstVisitor<void> {
       }
     }
 
+    // The following map contains for every declaration the set of the
+    // declarations it references.
     var dependencies = Map<Declaration, Set<Declaration>>.fromIterable(
       topDeclarations,
       value: (declaration) =>
@@ -113,13 +117,9 @@ class _Visitor extends SimpleAstVisitor<void> {
                     ],
                   ])
               .whereNotNull()
-              .map((e) {
-                while (e.enclosingElement2 != null &&
-                    e.enclosingElement2 is! CompilationUnitElement) {
-                  e = e.enclosingElement2!;
-                }
-                return e;
-              })
+              .map((e) => e.thisOrAncestorMatching((a) =>
+                  a.enclosingElement2 == null ||
+                  a.enclosingElement2 is CompilationUnitElement))
               .map((e) => declarationByElement[e])
               .whereNotNull()
               .where((e) => e != declaration)
@@ -127,7 +127,12 @@ class _Visitor extends SimpleAstVisitor<void> {
     );
 
     var usedMembers = entryPoints.toSet();
-    var toTraverse = Queue.from(usedMembers);
+    // The following variable will be used to visit every reachable declaration
+    // starting from entry-points. At every loop an element is removed. This
+    // element is marked as used and we add its dependencies in the declaration
+    // list to traverse. Once this list is empty `usedMembers` contains every
+    // declarations reachable from an entry-point.
+    var toTraverse = Queue.of(usedMembers);
     while (toTraverse.isNotEmpty) {
       var declaration = toTraverse.removeLast();
       for (var dep in dependencies[declaration]!) {
@@ -163,26 +168,17 @@ class _Visitor extends SimpleAstVisitor<void> {
       (e.name.name == 'main' || e.metadata.any(_isPragmaVmEntry));
 
   bool _isPragmaVmEntry(Annotation annotation) {
-    var elementAnnotation = annotation.elementAnnotation;
-    if (elementAnnotation != null) {
-      var value = elementAnnotation.computeConstantValue();
-      if (value != null) {
-        var type = value.type;
-        if (type != null) {
-          var element = type.element;
-          if (element != null) {
-            var library = element.library;
-            if (library != null && library.isDartCore ||
-                element.name == 'pragma') {
-              var name = value.getField('name');
-              return name != null &&
-                  name.hasKnownValue &&
-                  name.toStringValue() == 'vm:entry-point';
-            }
-          }
-        }
-      }
-    }
-    return false;
+    var value = annotation.elementAnnotation?.computeConstantValue();
+    if (value == null) return false;
+    var element = value.type?.element;
+    if (element == null || !element.isPragma) return false;
+    var name = value.getField('name');
+    return name != null &&
+        name.hasKnownValue &&
+        name.toStringValue() == 'vm:entry-point';
   }
+}
+
+extension on Element {
+  bool get isPragma => (library?.isDartCore ?? false) && name == 'pragma';
 }
--- a/lib/src/rules/unreachable_from_main.dart
+++ b/lib/src/rules/unreachable_from_main.dart
@@ -119,8 +119,8 @@ class _Visitor extends SimpleAstVisitor<void> {
                   ])
               .whereNotNull()
               .map((e) => e.thisOrAncestorMatching((a) =>
-                  a.enclosingElement2 == null ||
-                  a.enclosingElement2 is CompilationUnitElement))
+                  a.enclosingElement3 == null ||
+                  a.enclosingElement3 is CompilationUnitElement))
               .map((e) => declarationByElement[e])
               .whereNotNull()
               .where((e) => e != declaration)
@@ -152,12 +152,12 @@ class _Visitor extends SimpleAstVisitor<void> {
 
     for (var member in unusedMembers) {
       if (member is NamedCompilationUnitMember) {
-        rule.reportLint(member.name);
+        rule.reportLintForToken(member.name2);
       } else if (member is VariableDeclaration) {
-        rule.reportLint(member.name);
+        rule.reportLintForToken(member.name2);
       } else if (member is ExtensionDeclaration) {
         rule.reportLintForToken(
-            member.name?.token ?? member.firstTokenAfterCommentAndMetadata);
+            member.name2 ?? member.firstTokenAfterCommentAndMetadata);
       } else {
         rule.reportLintForToken(member.firstTokenAfterCommentAndMetadata);
       }
@@ -166,7 +166,7 @@ class _Visitor extends SimpleAstVisitor<void> {
 
   bool _isEntryPoint(Declaration e) =>
       e is FunctionDeclaration &&
-      (e.name.name == 'main' || e.metadata.any(_isPragmaVmEntry));
+      (e.name2.lexeme == 'main' || e.metadata.any(_isPragmaVmEntry));
 
   bool _isPragmaVmEntry(Annotation annotation) {
     if (!annotation.isPragma) return false;
@@ -195,6 +195,6 @@ extension on Annotation {
       // Dunno what this is.
       return false;
     }
-    return type.element?.isPragma ?? false;
+    return type is InterfaceType && type.element2.isPragma;
   }
 }
--- a/lib/src/rules/unreachable_from_main.dart
+++ b/lib/src/rules/unreachable_from_main.dart
@@ -12,7 +12,7 @@ import 'package:collection/collection.dart';
 import '../analyzer.dart';
 import '../util/dart_type_utilities.dart';
 
-const _desc = 'Unused top-level members in executable libraries.';
+const _desc = 'Unreachable top-level members in executable libraries.';
 
 const _details = r'''
 
@@ -41,10 +41,10 @@ void f() {}
 
 ''';
 
-class UnusedTopMembersInExecutableLibraries extends LintRule {
-  UnusedTopMembersInExecutableLibraries()
+class UnreachableFromMain extends LintRule {
+  UnreachableFromMain()
       : super(
-          name: 'unused_top_members_in_executable_libraries',
+          name: 'unreachable_from_main',
           description: _desc,
           details: _details,
           group: Group.style,
--- a/lib/src/rules/unreachable_from_main.dart
+++ b/lib/src/rules/unreachable_from_main.dart
@@ -7,6 +7,7 @@ import 'dart:collection';
 import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
 import 'package:analyzer/dart/element/element.dart';
+import 'package:analyzer/dart/element/type.dart';
 import 'package:collection/collection.dart';
 
 import '../analyzer.dart';
@@ -19,7 +20,7 @@ const _details = r'''
 Top-level members in an executable library should be used directly inside this
 library.  An executable library is a library that contains a `main` top-level
 function or that contains a top-level function annotated with
-`@pragma('vm:entry-point')`).  Executable libraries are usually never imported
+`@pragma('vm:entry-point')`).  Executable libraries are not usually imported
 and it's better to avoid defining unused members.
 
 This rule assumes that an executable library isn't imported by other files
@@ -168,10 +169,9 @@ class _Visitor extends SimpleAstVisitor<void> {
       (e.name.name == 'main' || e.metadata.any(_isPragmaVmEntry));
 
   bool _isPragmaVmEntry(Annotation annotation) {
+    if (!annotation.isPragma) return false;
     var value = annotation.elementAnnotation?.computeConstantValue();
     if (value == null) return false;
-    var element = value.type?.element;
-    if (element == null || !element.isPragma) return false;
     var name = value.getField('name');
     return name != null &&
         name.hasKnownValue &&
@@ -181,4 +181,20 @@ class _Visitor extends SimpleAstVisitor<void> {
 
 extension on Element {
   bool get isPragma => (library?.isDartCore ?? false) && name == 'pragma';
+}
+
+extension on Annotation {
+  bool get isPragma {
+    var element = elementAnnotation?.element;
+    DartType type;
+    if (element is ConstructorElement) {
+      type = element.returnType;
+    } else if (element is PropertyAccessorElement && element.isGetter) {
+      type = element.returnType;
+    } else {
+      // Dunno what this is.
+      return false;
+    }
+    return type.element?.isPragma ?? false;
+  }
 }
--- a/lib/src/rules/unreachable_from_main.dart
+++ b/lib/src/rules/unreachable_from_main.dart
@@ -143,7 +143,19 @@ class _Visitor extends SimpleAstVisitor<void> {
           element.isPublic &&
           !element.hasVisibleForTesting;
     });
-    unusedMembers.forEach(rule.reportLint);
+
+    for (var member in unusedMembers) {
+      if (member is NamedCompilationUnitMember) {
+        rule.reportLint(member.name);
+      } else if (member is VariableDeclaration) {
+        rule.reportLint(member.name);
+      } else if (member is ExtensionDeclaration) {
+        rule.reportLintForToken(
+            member.name?.token ?? member.firstTokenAfterCommentAndMetadata);
+      } else {
+        rule.reportLintForToken(member.firstTokenAfterCommentAndMetadata);
+      }
+    }
   }
 
   bool _isEntryPoint(Declaration e) =>
--- a/lib/src/rules/unused_top_members_in_executable_libraries.dart
+++ b/lib/src/rules/unused_top_members_in_executable_libraries.dart
@@ -15,8 +15,11 @@ const _desc = 'Unused top-level members in executable libraries.';
 const _details = r'''
 
 Top-level members in an executable library should be used directly inside this
-library. Executable library are usually never used as dependency and it's better
-to avoid defining unused members.
+library. Executable libraries are usually never used as dependency and it's
+better to avoid defining unused members.
+
+This rule assumes that an executable library shouldn't be imported by other
+files except to execute its `main` function.
 
 **BAD:**
 
@@ -32,7 +35,7 @@ main() {
   f();
 }
 void f() {}
-``
+```
 
 ''';
 
@@ -66,8 +69,6 @@ class _Visitor extends SimpleAstVisitor<void> {
     if (node.directives.whereType<PartOfDirective>().isNotEmpty) return;
     if (node.directives.whereType<PartDirective>().isNotEmpty) return;
 
-    if (!_isInsideExecutableLibrary(node)) return;
-
     var topDeclarations = node.declarations
         .expand((e) => [
               if (e is TopLevelVariableDeclaration)
@@ -76,6 +77,8 @@ class _Visitor extends SimpleAstVisitor<void> {
                 e,
             ])
         .toSet();
+
+    if (!topDeclarations.any(_isEntryPoint)) return;
 
     var declarationByElement = <Element, Declaration>{};
     for (var declaration in topDeclarations) {
@@ -118,9 +121,8 @@ class _Visitor extends SimpleAstVisitor<void> {
     while (toTraverse.isNotEmpty) {
       var declaration = toTraverse.removeLast();
       for (var dep in dependencies[declaration]!) {
-        if (!usedMembers.contains(dep)) {
-          usedMembers.add(dep);
-          if (!toTraverse.contains(dep)) toTraverse.add(dep);
+        if (usedMembers.add(dep)) {
+          toTraverse.add(dep);
         }
       }
     }
@@ -160,13 +162,5 @@ class _Visitor extends SimpleAstVisitor<void> {
       }
     }
     return false;
-  }
-
-  bool _isInsideExecutableLibrary(AstNode node) {
-    var root = node.root;
-    if (root is! CompilationUnit) return false;
-    var library = root.declaredElement?.library;
-    return library != null &&
-        library.exportNamespace.definedNames.containsKey('main');
   }
 }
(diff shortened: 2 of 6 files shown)


bwilkerson (commented):
It isn't clear to me that this is a rule we want to have.

As noted below, it will have false positives unless a convention is followed, and there's no support for enforcing that convention. If we think that this is a convention that we should support, then we should probably have a lint that ensures that it isn't violate. It seems to me that without that enforcement we run a risk of having a high rate of false positives from this rule.

lib/src/rules/unused_top_members_in_executable_libraries.dart
```diff
@@ -0,0 +1,172 @@
...
+
+import '../analyzer.dart';
+import '../util/dart_type_utilities.dart';
+
+const _desc = 'Unused top-level members in executable libraries.';
+
+const _details = r'''
+
+Top-level members in an executable library should be used directly inside this
+library. Executable library are usually never used as dependency and it's better
```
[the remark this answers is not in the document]
a14n: I added a comment regarding the assumption.

lib/src/rules/unused_top_members_in_executable_libraries.dart
```diff
@@ -0,0 +1,172 @@
...
+```
+
+**GOOD:**
+
+```dart
+main() {
+  f();
+}
+void f() {}
+``
```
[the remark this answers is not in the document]
a14n: done

lib/src/rules/unused_top_members_in_executable_libraries.dart
```diff
@@ -0,0 +1,172 @@
...
+
+    var entryPoints = topDeclarations.where(_isEntryPoint);
+    if (entryPoints.isEmpty) return;
+
+    var usedMembers = entryPoints.toSet();
+    var toTraverse = usedMembers.toList();
+    while (toTraverse.isNotEmpty) {
+      var declaration = toTraverse.removeLast();
+      for (var dep in dependencies[declaration]!) {
+        if (!usedMembers.contains(dep)) {
```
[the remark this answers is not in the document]
a14n: good catch. Done

lib/src/rules/unused_top_members_in_executable_libraries.dart
```diff
@@ -0,0 +1,172 @@
...
+          usedMembers.add(dep);
+          if (!toTraverse.contains(dep)) toTraverse.add(dep);
```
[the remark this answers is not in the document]
a14n: Correct. Done.

lib/src/rules/unused_top_members_in_executable_libraries.dart
```diff
@@ -0,0 +1,172 @@
...
+                  name.toStringValue() == 'vm:entry-point';
+            }
+          }
+        }
+      }
+    }
+    return false;
+  }
+
+  bool _isInsideExecutableLibrary(AstNode node) {
```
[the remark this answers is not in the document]
a14n: Actually I changed the test to check if there is a declaration conforming the `_isEntryPoint` function (inlined).

a14n (commented):
I ran this lint on flutter (see https://github.com/flutter/flutter/pull/107464) and I didn't face false positives.

a14n:
PTAL

test_data/rules/unused_top_members_in_executable_libraries.dart
```diff
@@ -0,0 +1,56 @@
+// Copyright (c) 2022, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// test w/ `dart test -N unused_top_members_in_executable_libraries`
```
srawlins, line 5: Because I'm not familiar with `traverseNodesInDFS`, and therefore don't trust it implicitly :), I'd love to see some interesting cases like:

```dart
class C {
  const C();
}

void f<T extends C> {} // Usage in type bound

void Function(C c) f2() {} // Usage in Function type

void f3([Object p = C()]) {} // Usage in default value

@C() // Usage in annotation
void f4() {}

extension on Future<C> {} // Usage in type parameter in extension `on` clause.

File: test_data/rules/unused_top_members_in_executable_libraries.dart [hunk 7]
```diff
@@ -0,0 +1,56 @@
...
+
```
srawlins, line 6: Here's a case where we _do_ report an element as unused, even though it is _referenced_: anywhere in the right operand of `is`, _except_ for type aliases:

```dart
class C {}

typedef ListInt = List<int>;

void f(Object p) {
  if (p is C || p is Future<C>) {} // NOT usage
  if (p is ListInt) {} // YES usage
}
```

srawlins, line 6: If a variable declaration uses a type but not in an initializer, that is _not_ use:

```dart
class C {}

C? c; // NOT usage
```

test_data/rules/unused_top_members_in_executable_libraries.dart
```diff
@@ -0,0 +1,56 @@
...
+}
+
+const a = 1; // LINT
+const b = 1; // OK
+
+int v = 1; // LINT
+
+typedef A = String; // LINT
+
+class C {} // LINT
```
srawlins, line 26: Could you add two constructors here to show that the presence of constructors (which refer to the name of the class) does not count the class as "used"

```
class C { // LINT
  C();
  C.named();
}
```

test_data/rules/unused_top_members_in_executable_libraries.dart
```diff
@@ -0,0 +1,56 @@
...
+
+void f2() // OK
+{
+  f1();
+}
+
+void f3(Function f) {} //OK
+void f4(int p) {} //OK
+
+int id = 0; // OK
```
srawlins, line 50: Could you add tests showing that assignment is not "usage." Something like:

```dart
int? id2 = 0;
void f() {
  id2 = 1; // NOT usage
  id2 ??= 2; // NOT usage
}
```

test_data/rules/unused_top_members_in_executable_libraries.dart
srawlins, line 6: How do you want to treat doc comments? I think they are probably _not_ use:

```
class C {}

/// I love [C].
void main() {}
```

srawlins, line 6: Could you add tests for both a named an an unnamed extension?

test_data/rules/unused_top_members_in_executable_libraries.dart
srawlins, line 50: Could you add tests similar to this `id` field one but with setters and getters?

```dart
int get id3 => 0;
void set id3(int value) {};

void f() {
  id3 += 4; // usage
}

**Review by srawlins — CHANGES_REQUESTED** [review]

**Comment by srawlins** [comment]
@pq is there a way to benchmark this before it is landed? I worry about performance on this one because of traverseNodesInDFS.

**Comment by pq** [comment]
> is there a way to benchmark this before it is landed?

Great question!  The good news is each PR triggers the benchmark bot and that gives us a feel.  If you expand the checks, you'll see it:

image


Click on details and you can look at the output.

In this case, you're right that as implemented, this is relatively costly:

image

File: lib/src/rules/unused_top_members_in_executable_libraries.dart [hunk 10]
```diff
@@ -0,0 +1,166 @@
...
+
+import '../analyzer.dart';
+import '../util/dart_type_utilities.dart';
+
+const _desc = 'Unused top-level members in executable libraries.';
+
+const _details = r'''
+
+Top-level members in an executable library should be used directly inside this
+library. Executable libraries are usually never used as dependency and it's
```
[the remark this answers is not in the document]
a14n: done

lib/src/rules/unused_top_members_in_executable_libraries.dart
```diff
@@ -0,0 +1,166 @@
...
+better to avoid defining unused members.
+
+This rule assumes that an executable library shouldn't be imported by other
```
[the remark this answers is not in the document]
a14n: done

lib/src/rules/unused_top_members_in_executable_libraries.dart
```diff
@@ -0,0 +1,166 @@
...
+}
+
+class _Visitor extends SimpleAstVisitor<void> {
+  _Visitor(this.rule);
+
+  final LintRule rule;
+
+  @override
+  void visitCompilationUnit(CompilationUnit node) {
+    // TODO(a14n): add support of libs with parts
```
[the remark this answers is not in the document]
a14n: done

lib/src/rules/unused_top_members_in_executable_libraries.dart
```diff
@@ -0,0 +1,166 @@
...
+          if (getter != null) declarationByElement[getter] = declaration;
+          var setter = element.setter;
+          if (setter != null) declarationByElement[setter] = declaration;
+        } else {
+          declarationByElement[element] = declaration;
+        }
+      }
+    }
+
+    var dependencies = Map<Declaration, Set<Declaration>>.fromIterable(
```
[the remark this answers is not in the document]
a14n: Without type annotations it becomes a `Map<dynamic, Set<Declaration>>` instead of a `Map<Declaration, Set<Declaration>>`

lib/src/rules/unused_top_members_in_executable_libraries.dart
```diff
@@ -0,0 +1,166 @@
...
+      topDeclarations,
+      value: (e) => DartTypeUtilities.traverseNodesInDFS(e as Declaration)
+          .expand((e) => [
```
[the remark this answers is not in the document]
a14n: done

lib/src/rules/unused_top_members_in_executable_libraries.dart
```diff
@@ -0,0 +1,166 @@
...
+                  e.writeElement,
+                ],
+              ])
+          .whereNotNull()
+          .map((e) => declarationByElement[e])
+          .whereNotNull()
+          .toSet(),
+    );
+
+    var entryPoints = topDeclarations.where(_isEntryPoint);
```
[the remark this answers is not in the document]
a14n: I move this code on the first check.

lib/src/rules/unused_top_members_in_executable_libraries.dart
```diff
@@ -0,0 +1,166 @@
...
+    if (entryPoints.isEmpty) return;
+
+    var usedMembers = entryPoints.toSet();
+    var toTraverse = usedMembers.toList();
```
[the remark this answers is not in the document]
a14n: done

lib/src/rules/unused_top_members_in_executable_libraries.dart
```diff
@@ -0,0 +1,166 @@
...
+  }
+
+  bool _isEntryPoint(Declaration e) =>
+      e is FunctionDeclaration &&
+      (e.name.name == 'main' || e.metadata.any(_isPragmaVmEntry));
+
+  bool _isPragmaVmEntry(Annotation annotation) {
+    var elementAnnotation = annotation.elementAnnotation;
+    if (elementAnnotation != null) {
+      var value = elementAnnotation.computeConstantValue();
```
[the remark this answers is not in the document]
a14n: > we can perhaps delay this if we know that `elementAnnotation.element` is not the `pragma` class?

`elementAnnotation.element` would point to the ConstructorElement and it will work most of the time. Unfortunatelly I think this wouldn't support code like:

```dart
const entryPoint = pragma('vm:entry-point');

@entryPoint
f(){}
```

test_data/rules/unused_top_members_in_executable_libraries.dart
```diff
@@ -0,0 +1,56 @@
...
+{
+  f2();
+}
+
+void f2() // OK
+{
+  f1();
+}
+
+void f3(Function f) {} //OK
```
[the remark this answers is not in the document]
a14n: done

test_data/rules/unused_top_members_in_executable_libraries.dart
a14n: I added those tests

test_data/rules/unused_top_members_in_executable_libraries.dart
a14n: done

test_data/rules/unused_top_members_in_executable_libraries.dart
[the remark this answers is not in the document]
a14n: I may be wrong but some of the comments you did was not really related to this lint (warn on **public** top level members that are not accessible directly or transitivelly from an entry-point ) and so I didn't add test case for those comments I left unanswered.

srawlins: 🎉 Thanks!

test_data/rules/unused_top_members_in_executable_libraries.dart
srawlins: Er, I guess not unnamed, because that is handled by the `unused_element` hint.

srawlins: Is this one of the cases you thought would be unrelated? I'm not sure why. In this case, I would want C reported as unused, and I would not want ListInt reported as unused.

a14n: TBH I don't understand why `C` should be reported here (assuming `f` is reachable from an entrypoint).

The rule doesn't really analyze the execution flow, it only visit the AST children of the `f` node.

a14n: Named extension is now covered by https://github.com/dart-lang/linter/pull/3513/files#diff-4f5160f7cce759c5604eb650859c439c3084aee4efd0b42cdd07d2bb8d3f69f1R109

a14n: Perhaps the term _unused_ is a little misleading. It should be understood as _unaccessible from any entry point_.

srawlins: I don't think execution flow is important either; just syntax tree.

In this example, C would be considered _used_, right?

```dart
class C {}

void f() {
  var c = C();
}
```

or even

```dart
class C {}

void f(C c) {}
```

I think those would be "usages" of C. However, in this code:

```dart
class C {}

void f(Object p) {
  if (p is C || p is Future<C>) {}
}
```

I would not count C as being "used." (This is how unused_element works.) Any time `C` is found in the right side of an `is ` expression does not count as usage.

a14n: > I would not count C as being "used." (This is how unused_element works.) Any time `C` is found in the right side of an `is ` expression does not count as usage.

Thanks for the explanation. 

As mentioned in my in-between comment above my lint don't really mimics the `unused_element` behaviour. It simply visit recursivelly the whole AST (without exception like in the example you provided with right side of `is`) starting from entrypoints and warns if a top public declaration is not reachable.

srawlins: Would `traverseNodesInDFS`'s `excludeCriteria` parameter allow this? I can't tell offhand because I can't find any tests for `traverseNodesInDFS`.

a14n: I could add this kind of rules but I don't know where to find a list of all the exceptions to mimic the behaviour of `unused_elements`. 
That said I'm not sure the cost of maintaining this logic in 2 places is worth it. If some changes land on `unused_element` we will have to redo the job on linter side.
(Side note: I'm a little surprise that `C` is not considered used because if you remove `class C {}` the code doesn't compile)

bwilkerson: > Perhaps the term _unused_ is a little misleading. It should be understood as _unaccessible from any entry point_.

If what you're really trying to lint are declarations that are not reachable from `main`, then yes, the terminology is getting in the way. Consider changing the name of the lint to `unreachable_from_main`.

And if that's the case, then we can re-review this PR from that perspective (because it changes the kinds of possible bugs that we should be looking for).

bwilkerson: > I'm a little surprise that `C` is not considered used because if you remove `class C {}` the code doesn't compile

I have to agree, that does seem a bit odd.

From a tree-shaking perspective, you couldn't remove `C` unless you also removed the then-branch of the `if`. If we think that there can't be any instances of `C` or any subclass of `C`, then it seems like the `if` should be marked as dead code, but `C` shouldn't be reported until the `if` is removed. (But maybe I'm misunderstanding your example.)

a14n: Sorry for the bad naming :-/ I used this name because of #2395 that has this wording.

> Consider changing the name of the lint to `unreachable_from_main`.

:+1: I will update the PR

a14n: Renamed. PTAL

lib/src/rules/unused_top_members_in_executable_libraries.dart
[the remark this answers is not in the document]
srawlins: Ack

lib/src/rules/unreachable_from_main.dart
```diff
@@ -0,0 +1,176 @@
...
+      }
+    }
+
+    var unusedMembers = topDeclarations.difference(usedMembers).where((e) {
+      var element = e.declaredElement;
+      return element != null &&
+          element.isPublic &&
+          !element.hasVisibleForTesting;
+    });
+    unusedMembers.forEach(rule.reportLint);
```
bwilkerson, line 146: Given that these are declarations, it would be better if we could report the lint on the name of the declaration, or the `extension` keyword in the case of an unnamed extension.

a14n: done

test_data/rules/unused_top_members_in_executable_libraries.dart
srawlins: > I could add this kind of rules but I don't know where to find a list of all the exceptions to mimic the behaviour of unused_elements.

I've just enumerated the exceptions in the docs: https://github.com/dart-lang/sdk/commit/2bedf1e55d593ed88cb63bb9321354ccb6e0bc15

> If we think that there can't be any instances of C or any subclass of C, then it seems like the if should be marked as dead code, ...

Filed https://github.com/dart-lang/sdk/issues/49584

> ... but C shouldn't be reported until the if is removed.

That's fair. We report it today and I don't think we should stop reporting it unless we complete #49584 as it provides value.

lib/src/rules/unreachable_from_main.dart
```diff
@@ -0,0 +1,188 @@
...
+import 'package:collection/collection.dart';
+
+import '../analyzer.dart';
+import '../util/dart_type_utilities.dart';
+
+const _desc = 'Unreachable top-level members in executable libraries.';
+
+const _details = r'''
+
+Top-level members in an executable library should be used directly inside this
```
srawlins, line 19: I think it would be good to define "executable library" in the description.

test_data/rules/unreachable_from_main.dart
```diff
@@ -0,0 +1,119 @@
...
+  usageInFunctionType();
+  usageInDefaultValue();
+  usageInAnnotation();
+  Future<C5>.value(C5()).extensionUsage();
+  accessors();
+}
+
+class Comment {} // LINT
+
+const a = 1; // LINT
```
srawlins, line 28: Can you add a test for multiple variables declared in a single declaration? I guess something like at least one of them is reachable, and at least one is not.

lib/src/rules/unreachable_from_main.dart
```diff
@@ -0,0 +1,188 @@
...
+              .expand((e) => [
+                    if (e is SimpleIdentifier) e.staticElement,
+                    // with `id++` staticElement of `id` is null
+                    if (e is CompoundAssignmentExpression) ...[
+                      e.readElement,
+                      e.writeElement,
+                    ],
+                  ])
+              .whereNotNull()
+              .map((e) {
```
srawlins, line 116: I think this `map` callback is equivalent to some invocation of `Element.thisOrAncestorMatching`, which would make it more readable. Something like

```dart
.map((e) =>
    e.thisOrAncestorMatching(
        (a) => a.enclosingElement2 == null || a.enclosingElement2 is CompilationUnitElement))
```

maybe? WDYT?

lib/src/rules/unreachable_from_main.dart
```diff
@@ -0,0 +1,188 @@
...
+                return e;
+              })
+              .map((e) => declarationByElement[e])
+              .whereNotNull()
+              .where((e) => e != declaration)
+              .toSet(),
+    );
+
+    var usedMembers = entryPoints.toSet();
+    var toTraverse = Queue.from(usedMembers);
```
srawlins, line 130: I think this constructs a `Queue<dynamic>`. We want `Queue.of`.

srawlins, line 130: Could you add a comment above this declaration explaining what `toTraverse` is? Or perhaps what the algorithm in the while-loop is.

lib/src/rules/unreachable_from_main.dart
```diff
@@ -0,0 +1,188 @@
...
+    }
+  }
+
+  bool _isEntryPoint(Declaration e) =>
+      e is FunctionDeclaration &&
+      (e.name.name == 'main' || e.metadata.any(_isPragmaVmEntry));
+
+  bool _isPragmaVmEntry(Annotation annotation) {
+    var elementAnnotation = annotation.elementAnnotation;
+    if (elementAnnotation != null) {
```
srawlins, line 167: 5 levels of if-statements makes this code pretty hard for me to read. Can we instead short circuit `return false` when we hit a condition that makes this _not_ a pragma, and flatten out the code?

test_data/rules/unused_top_members_in_executable_libraries.dart
srawlins: This comment still stands. I want to make sure that `traverseNodeInDFS` and your algorithm do not count `C.named();` as a legitimate reachability reference to `C`. `C` should be marked unreachable.

lib/src/rules/unreachable_from_main.dart
```diff
@@ -0,0 +1,188 @@
...
+          if (getter != null) declarationByElement[getter] = declaration;
+          var setter = element.setter;
+          if (setter != null) declarationByElement[setter] = declaration;
+        } else {
+          declarationByElement[element] = declaration;
+        }
+      }
+    }
+
+    var dependencies = Map<Declaration, Set<Declaration>>.fromIterable(
```
srawlins, line 103: I think it would be good to have a comment above this variable. It takes a lot of logic to figure out what these 25 lines do.

srawlins (changes requested):

lib/src/rules/unreachable_from_main.dart
a14n: Description added.

test_data/rules/unreachable_from_main.dart
a14n: done

lib/src/rules/unreachable_from_main.dart
a14n: done

lib/src/rules/unreachable_from_main.dart
a14n: Good call! Isn't there a lint for that :) ?

a14n: done

lib/src/rules/unreachable_from_main.dart
a14n: done

test_data/rules/unused_top_members_in_executable_libraries.dart
a14n: done

a14n:
PTAL

lib/src/rules/unreachable_from_main.dart
srawlins: I had the same question, haha. https://github.com/dart-lang/linter/issues/2555

a14n:
I perhaps missed some comments. Is there something blocking on this PR?

srawlins:
I forgot to click Send on my last 3 comments; sorry about that! I think my open comments amount to:

* How do we treat references in comments?
* Can we avoid evaluating the `@pragma` constant until we know its a `@pragma` annotation?

a14n:
> * How do we treat references in comments?

There was already a test to trigger a diagnostic on elements only ref in comment (see `Comment` in test file)

> * Can we avoid evaluating the `@pragma` constant until we know its a `@pragma` annotation?

Done. Thanks for the code snippet.

a14n:
After rebasing on `main` (and upgrading the analyzer) references in comments are now visited. I updated the test to not mark `Comment` as unreachable.

PTAL

lib/src/rules/unreachable_from_main.dart
```diff
@@ -0,0 +1,200 @@
...
+import '../util/dart_type_utilities.dart';
+
+const _desc = 'Unreachable top-level members in executable libraries.';
+
+const _details = r'''
+
+Top-level members in an executable library should be used directly inside this
+library.  An executable library is a library that contains a `main` top-level
+function or that contains a top-level function annotated with
+`@pragma('vm:entry-point')`).  Executable libraries are not usually imported
```
srawlins, line 23: I think there is a stray `)` here, at the end of the sentence.

srawlins (approved):
Looks great! Thanks! 🎉

PR #3524, opened 2022-07-13
enable `combinators_ordering`

pq:
(And remove deprecated `invariant_booleans`.)

/cc @bwilkerson

/fyi @a14n

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -36,11 +36,11 @@ linter:
     - cancel_subscriptions
 #    - cascade_invocations
     - cast_nullable_to_non_nullable
+    - combinators_ordering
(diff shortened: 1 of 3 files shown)


pq:
Ah.  It looks like `combinators_ordering` isn't in a dev build yet.  We'll have to wait on this until it is.

bwilkerson (approved):

bwilkerson:
Do we have a fix for this? We really ought to. It's mechanical and tedious to fix by hand.

pq:
> Do we have a fix for this? We really ought to. It's mechanical and tedious to fix by hand.

We do not.

/fyi @jcollins-g

Issue #2650, opened 2021-05-13
document what lints have quick-fixes


pq:
See also: https://github.com/dart-lang/lints/issues/12.

In the generated docs, it would be great to document if a given lint has an associated quick-fix.  (And maybe bulk-fix?)

@parlough, @kwalrath: ideas for how to present this information most welcome!

bwilkerson:
I made this comment on the other thread, but I'll repeat it here because it's a different audience.

I've wondered a few time whether we might want to add something to the documentation for the lints (and other diagnostics) to indicate that there are fixes available for them. Perhaps a generated line at the end for the form

Fixable using dart fix or in the IDE using '<fix label>'.

with support for a list if there's more than one fix.

bernaferrari:
I'm having the same problem you just described @bwilkerson. I added flutter_lints and now half my codebase begs for const in widgets. However, dart fix doesn't handle these (even though the IDE has the yellow lamp). I also expected dart fix to handle unused imports, but it doesn't. I believe you could make dart fix more powerful, even if it required an `--aggressive` command. I prefer to fix 200 issues even if it creates 5 than having to manually fix 200 things (and potentially creating more than 5).

bwilkerson:
There is definitely room to improve the tool.

> ... begs for const in widgets. However, dart fix doesn't handle these (even though the IDE has the yellow lamp).

The lints would probably be because of the `prefer_const_constructors` rule. The reason that particular fix isn't enabled is because the fix isn't clever enough to notice when a `const` doesn't need to be added in one place because an outer constructor invocation had a `const` added to it, so it adds unnecessary code (which turns into lints if you also have the `unnecessary_const` rule enabled). We debated about whether it would be better to have fewer fixes enabled or to require users to run `dart fix` multiple times, and we decided to enable fewer fixes. Sounds like you might have preferred the other option.

Of course, the better solution (which we didn't have time to implement earlier) would be to fix the fix generator so that the problem doesn't arise. That's definitely something we're hoping to get back to at some point.

> I also expected dart fix to handle unused imports, but it doesn't.

[...]

PR #3527, opened 2022-07-14
generate fix status documentation

pq:
See https://github.com/dart-lang/linter/issues/2650.

image

image

Sample machine.json output:

```json
  {
    "name": "always_use_package_imports",
    "description": "Avoid relative imports for files in `lib/`.",
    "group": "errors",
    "maturity": "stable",
    "incompatible": [],
    "sets": [],
    "fixStatus": "hasFix",
    "details": "*DO* avoid relative imports for files in `lib/`.\n\nWhen mixing relative and absolute imports it's possible to create confusion\nwhere the same member gets imported in two different ways. One way to avoid\nthat is to ensure you consistently use absolute imports for files withing the\n`lib/` directory.\n\nThis is the opposite of 'prefer_relative_imports'.\n\nYou can also use 'avoid_relative_lib_imports' to disallow relative imports of\nfiles within `lib/` directory outside of it (for example `test/`).\n\n**GOOD:**\n\n```dart\nimport 'package:foo/bar.dart';\n\nimport 'package:foo/baz.dart';\n\nimport 'package:foo/src/baz.dart';\n...\n```\n\n**BAD:**\n\n```dart\nimport 'baz.dart';\n\nimport 'src/bag.dart'\n\nimport '../lib/baz.dart';\n\n...\n```\n\n"
  },
  {
    "name": "avoid_dynamic_calls",
    "description": "Avoid method calls or property accesses on a \"dynamic\" target.",
    "group": "errors",
    "maturity": "stable",
    "incompatible": [],
    "sets": [],
    "fixStatus": "needsEvaluation",
    "details": "\n**DO** avoid method calls or accessing properties on an object that is either\nexplicitly or implicitly statically typed \"dynamic\". Dynamic calls are treated\nslightly different in every runtime environment and compiler, but most\nproduction modes (and even some development modes) have both compile size and\nruntime performance penalties associated with dynamic calls.\n\nAdditionally, targets typed \"dynamic\" disables most static analysis, meaning it\nis easier to lead to a runtime \"NoSuchMethodError\" or \"NullError\" than properly\nstatically typed Dart code.\n\nThere is an exception to methods and properties that exist on \"Object?\":\n- a.hashCode\n- a.runtimeType\n- a.noSuchMethod(someInvocation)\n- a.toString()\n\n... these members are dynamically dispatched in the web-based runtimes, but not\nin the VM-based ones. Additionally, they are so common that it would be very\npunishing to disallow `any.toString()` or `any == true`, for example.\n\nNote that despite \"Function\" being a type, the semantics are close to identical\nto \"dynamic\", and calls to an object that is typed \"Function\" will also trigger\nthis lint.\n\n**BAD:**\n```dart\nvoid explicitDynamicType(dynamic object) {\n  print(object.foo());\n}\n\nvoid implicitDynamicType(object) {\n  print(object.foo());\n}\n\nabstract class SomeWrapper {\n  T doSomething<T>();\n}\n\nvoid inferredDynamicType(SomeWrapper wrapper) {\n  var object = wrapper.doSomething();\n  print(object.foo());\n}\n\nvoid callDynamic(dynamic function) {\n  function();\n}\n\nvoid functionType(Function function) {\n  function();\n}\n```\n\n**GOOD:**\n```dart\nvoid explicitType(Fooable object) {\n  object.foo();\n}\n\nvoid castedType(dynamic object) {\n  (object as Fooable).foo();\n}\n\nabstract class SomeWrapper {\n  T doSomething<T>();\n}\n\nvoid inferredType(SomeWrapper wrapper) {\n  var object = wrapper.doSomething<Fooable>();\n  object.foo();\n}\n\nvoid functionTypeWithParameters(Function() function) {\n  function();\n}\n```\n\n"
  },
```

I do wish I had a nice place to link out to describe quick-fixes.  Is there one I'm missing?

/cc @bwilkerson @parlough @srawlins

### DIFF
--- a/test/doc_test.dart
+++ b/test/doc_test.dart
@@ -1,0 +1,19 @@
+// Copyright (c) 2022, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:test/test.dart';
+
+import '../tool/doc.dart';
+
+void main() {
+  group('doc generation', () {
+    test('fixStatus (sanity)', () async {
+      var fixStatusMap = await fetchFixStatusMap();
+      // Doc generation reads the fix status map to associate fix status
+      // badges with rule documentation.  Here we check one for sanity.
+      // If the file moves or format changes, we'd expect this to fail.
+      expect(fixStatusMap['always_declare_return_types'], 'hasFix');
+    });
+  });
+}
--- a/tool/doc.dart
+++ b/tool/doc.dart
@@ -13,6 +13,7 @@ import 'package:http/http.dart' as http;
 import 'package:linter/src/analyzer.dart';
 import 'package:linter/src/rules.dart';
 import 'package:markdown/markdown.dart';
+import 'package:yaml/yaml.dart';
 
 import 'machine.dart';
 import 'since.dart';
@@ -43,7 +44,6 @@ void main(List<String> args) async {
   var auth = token is String ? Authentication.withToken(token) : null;
 
   var createDirectories = options['create-dirs'] == true;
-
   var enableMarkdown = options['markdown'] == true;
 
   await generateDocs(outDir,
@@ -92,20 +92,6 @@ final List<LintRule> rules =
     List<LintRule>.of(Registry.ruleRegistry, growable: false)..sort();
 
 late Map<String, SinceInfo> sinceInfo;
-
-String get enumerateErrorRules =>
-    rules.where((r) => r.group == Group.errors).map(toDescription).join('\n\n');
-
-String get enumerateGroups => Group.builtin
-    .map((Group g) =>
-        '<li><strong>${g.name} -</strong> ${markdownToHtml(g.description)}</li>')
-    .join('\n');
-
-String get enumeratePubRules =>
-    rules.where((r) => r.group == Group.pub).map(toDescription).join('\n\n');
-
-String get enumerateStyleRules =>
-    rules.where((r) => r.group == Group.style).map(toDescription).join('\n\n');
 
 Future<String> get pedanticLatestVersion async {
   var url =
@@ -164,6 +150,26 @@ Future<LintConfig?> fetchConfig(String url) async {
   return processAnalysisOptionsFile(req.body);
 }
 
+final Map<String, String> _fixStatusMap = <String, String>{};
+
+Future<Map<String, String>> fetchFixStatusMap() async {
+  if (_fixStatusMap.isNotEmpty) return _fixStatusMap;
+  var url =
+      'https://raw.githubusercontent.com/dart-lang/sdk/main/pkg/analysis_server/lib/src/services/correction/error_fix_status.yaml';
+  var client = http.Client();
+  print('loading $url...');
+  var req = await client.get(Uri.parse(url));
+  var yaml = loadYamlNode(req.body) as YamlMap;
+  for (var entry in yaml.entries) {
+    var code = entry.key as String;
+    if (code.startsWith('LintCode.')) {
+      _fixStatusMap[code.substring(9)] =
+          (entry.value as YamlMap)['status'] as String;
+    }
+  }
+  return _fixStatusMap;
+}
+
 Future<void> fetchSinceInfo(Authentication? auth) async {
   sinceInfo = await getSinceMap(auth);
 }
@@ -208,16 +214,20 @@ Future<void> generateDocs(String? dir,
   // Fetch since info.
   await fetchSinceInfo(auth);
 
+  var fixStatusMap = await fetchFixStatusMap();
+  print(fixStatusMap);
+
   // Generate rule files.
   for (var l in rules) {
-    RuleHtmlGenerator(l).generate(outDir);
+    RuleHtmlGenerator(l, fixStatusMap[l.name] ?? 'unregistered')
+        .generate(outDir);
     if (enableMarkdown) {
       RuleMarkdownGenerator(l).generate(filePath: outDir);
     }
   }
 
   // Generate index.
-  HtmlIndexer(Registry.ruleRegistry).generate(outDir);
+  HtmlIndexer(Registry.ruleRegistry, fixStatusMap).generate(outDir);
 
   if (enableMarkdown) {
     MarkdownIndexer(Registry.ruleRegistry).generate(filePath: outDir);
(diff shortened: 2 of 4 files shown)


tool/doc.dart
```diff
@@ -208,16 +214,20 @@ Future<void> generateDocs(String? dir,
   // Fetch since info.
   await fetchSinceInfo(auth);
 
+  var fixStatusMap = await fetchFixStatusMap();
+  print(fixStatusMap);
```
srawlins, line 218: Is the map long? Is this for debugging or should this be included in the submitted PR?

pq: Good catch!  That was a stray debugging call.  (Removed.)

pq:
> Is there one I'm missing?

I wonder if @bwilkerson's medium article shouldn't be hosted on dart.dev?

https://medium.com/dartlang/quick-fixes-for-analysis-issues-c10df084971a

parlough:
Thanks for working on this. As for a documentation destination, I was thinking to add a small section to https://dart.dev/tools/linter-rules (https://github.com/dart-lang/site-www/issues/4143) which I could PR use of to the linter when I finish.

We could also adopt the article for usage on dart.dev.

parlough:
I was thinking of adding a small section to https://dart.dev/tools/linter-rules (https://github.com/dart-lang/site-www/issues/4143) which we could link to. Then I could make a PR to the linter adding these lints.

It would also be nice to adapt the article for use on dart.dev for general use, but that would take a larger overhaul.

tool/doc.dart
```diff
@@ -164,6 +152,24 @@ Future<LintConfig?> fetchConfig(String url) async {
   return processAnalysisOptionsFile(req.body);
 }
 
+Future<Map<String, String>> fetchFixStatusMap() async {
+  if (_fixStatusMap.isNotEmpty) return _fixStatusMap;
+  var url =
+      'https://raw.githubusercontent.com/dart-lang/sdk/main/pkg/analysis_server/lib/src/services/correction/error_fix_status.yaml';
```
bwilkerson, line 158: I have to ask, but would it be better to just load the `FixProcessor` and ask whether each lint's name is a key in the map? Seems like this approach would make it hard to change the structure of the `error_fix_status.yaml` file, and that might cause problems.

bwilkerson:
> I do wish I had a nice place to link out to describe quick-fixes. Is there one I'm missing?

No, we don't have documentation for the fixes (or assists, or refactorings for that matter).

> I wonder if @bwilkerson's medium article shouldn't be hosted on dart.dev?

I don't really care either way, but as far as I know we don't host any of the Medium articles on either `dart.dev` or `flutter.dev`, so I think you're asking a bigger question.

> I was thinking of adding a small section to https://dart.dev/tools/linter-rules (dart-lang/site-www#4143) which we could link to.

I added a comment to that issue.

tool/doc.dart
pq: Great question!

The issue with this approach is that I think we'd have to introduce a dependency on `analysis_server` and we don't want to do this for obvious reasons.  To get around this in the past, I pulled the `FixProcessor` source down and scraped it for this mapping but my sense is that that is at least as brittle as this.

A thought about structural changes: maybe a simple test to sanity-check would be good?  (I'll go ahead and add one to show you what I have in mind.)

test/doc_test.dart
```diff
@@ -0,0 +1,19 @@
...
+import '../tool/doc.dart';
+
+void main() {
+  group('doc generation', () {
+    test('fixStatus (sanity)', () async {
+      var fixStatusMap = await fetchFixStatusMap();
+      // Doc generation reads the fix status map to associate fix status
+      // badges with rule documentation.  Here we check one for sanity.
+      // If the file moves or format changes, we'd expect this to fail.
+      expect(fixStatusMap['always_declare_return_types'], 'hasFix');
```
pq, line 16: @bwilkerson : here's the kind of sanity check I was proposing

tool/doc.dart
pq: > I will, however, point out that this is a rather subtle dependency on the package in the sense that it's still depending on a file in that package, it's just that the dependency isn't explicit and it's pinned to HEAD rather than a published version. 

Right.  I agree.

> Of course, I probably wouldn't be bothered about it at all if the linter were in the sdk repo, so maybe that's a signal that I should worry about it now either.

Another good observation.  We have a lot of subtle dependencies already and it's a little uncomfortable to add more.  Having said that, my gut says *this* file is pretty unlikely to change (often) and with a sanity check in place we're sure to notice any breakage before we try and publish new docs that are less complete because this data is no longer available.

pq:
> I don't really care either way, but as far as I know we don't host any of the Medium articles on either `dart.dev` or `flutter.dev`, so I think you're asking a bigger question.

Right.  I meant more along the lines of hosting the _content_ which IMO is too good to be overlooked on medium!

bwilkerson:
I'll leave the question of whether `dart.dev` and/or `flutter.dev` ought to have something like a directory of links to Medium articles and other external content, organized by topic area, to @sfshaza2. And I'm not even going to touch the question of where such content should be hosted.

But thanks for the positive review.

sfshaza2:
I love being pulled in on s/w PRs! (Nope, not joking.)

Yes, what I'd like to do in the short term is to add links to Brian's articles wherever appropriate. I JUST added links to all 3 articles from a Flutter page, but I should do the same (or @parlough can) to these articles. I just did a quick search on the Dart site and we could add links to:

* [Linter rules](https://dart.dev/tools/linter-rules)
* [Customizing static analysis](https://dart.dev/guides/language/analysis-options)
* [dart fix](https://dart.dev/tools/dart-fix)
* [Dart: the command-line tool](https://dart.dev/tools/dart-tool)

I guess what we should ask ourselves, Brian, for any future docs you want to write is: is the info better served on the blog or on the website? In retrospect, these articles should have gone to the website. Website docs have far more views *and* we are committed to maintaining them. That is not true on the blog.

bwilkerson (approved):

pq:
Yay @sfshaza2!

Since none of these docs cover quick-fixes in real detail, would it be weird to link out to the medium article for now?

To be clear, it would only be a link from the "has fix" badge:

image

sfshaza2:
No, I think linking to the Medium article is fine!

sfshaza2:
I know this PR has landed and all, but YES writing pages for the website counts as community involvement. This is different than writing API docs, for example, or doc proposals. We can give you detailed analytics for perf time as to how viewed your pages are, etc.

I'm happy to duke this out with Jacob, VJ, or anyone else. In a friendly, work appropriate way, of course.  :D

PR #3537, opened 2022-07-19
validate that rule descriptions are not hardcoded in tests

pq:
Follow-up from: #3536.

/cc @bwilkerson


test/validate_no_rule_description_references.dart
```diff
@@ -0,0 +1,23 @@
...
+import 'package:linter/src/analyzer.dart';
+import 'package:linter/src/rules.dart';
+import 'package:test/test.dart';
+
+void main() {
+  group('check for rule message descriptions in tests:', () {
+    registerLintRules();
+    for (var rule in Analyzer.facade.registeredRules) {
+      test(rule.name, () async {
+        var result = Process.runSync('grep', ['-R', rule.description, 'test']);
```
bwilkerson, line 16: I think this won't work for the lints that have messages encoded in custom instances of `LintCode`, nor will it work for messages that have placeholders (like `{0}`) in them.

pq: Good point!  This caught a bunch of bad integration tests so I'm inclined to think the coverage is useful (if incomplete).  I'll do a bit of looking at the reflective tests and see if there's sanity checking we can do there...  Thanks!

pq: (FWIW I see one maybe problematic case in a reflective test but not sure it's worth removing.)

pq: Another argument in favor of landing this is that it would have flagged all the cases in #3536.

bwilkerson: > I'll do a bit of looking at the reflective tests ...

My comment wasn't related to reflective tests, so I'm confused.

I just meant that (a) instead of using `rule.description` we probably need to iterate over `rule.lintCodes` and grab the `problemMessage` from each code, and (b) replace placeholders with a regexp like `.*`.

> Another argument in favor of landing this ...

Landing it without fixing these issues is fine, but we should add a comment so that our future selves can remember what's going on here. :-)

PR #3541, opened 2022-07-20
ensure reflective tests don't enable default on experiments

pq:
A number of reflective tests enable experiments explicitly that are now on by default in our minimum supported SDK.  This change removes those redundant declarations and adds validation to ensure that we don't accrue similar technical debt in the future.

/cc @bwilkerson

### DIFF
--- a/test/rule_test_support.dart
+++ b/test/rule_test_support.dart
@@ -2,6 +2,8 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:io' as io;
+
 import 'package:analyzer/dart/analysis/results.dart';
 import 'package:analyzer/error/error.dart';
 import 'package:analyzer/file_system/file_system.dart';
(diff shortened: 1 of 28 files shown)


test/rule_test_support.dart
```diff
@@ -254,6 +267,17 @@ class PubPackageResolutionTest extends _ContextResolutionTest {
   @mustCallSuper
   void setUp() {
     super.setUp();
+    // Check for any needlessly enabled experiments.
+    for (var experiment in experiments) {
+      var feature = ExperimentStatus.knownFeatures[experiment];
+      var releaseVersion = feature?.releaseVersion;
+      if (releaseVersion != null && releaseVersion >= minSdkVersion) {
```
bwilkerson, line 274: I'm not sure why we need to compare the `releaseVersion` with the `minSdkVersion`. I would have expected that it's sufficient to ask whether the feature is enabled by default. What am I missing?

bwilkerson (approved):

test/rule_test_support.dart
pq: I may be overthinking it but what I'm trying to guard against is the case where someone is testing locally with an unreleased SDK that enables a feature that is *not* yet enabled by default in a published SDK.  In that case, there would be no way to get the test to pass locally and not have the CI break.

bwilkerson: Check my reasoning.

When we're running a test, the code under test is being analyzed by the version of the `analyzer` package that was selected by pub (usually that latest published version). That same version of the `analyzer` package is the one being queried to see whether the feature has been enabled by default. As a result, asking the feature should be all you need to do because it's the same test that will be used by the analyzer when analyzing the code under test to find the default features.

In other words, I don't think the SDK running the test ever has any bearing on the state of enablement of the features.

pq: As discussed.  I agree!  Thanks for the sanity check. 🙏

PR #3545, opened 2022-07-21
Insert reasoning for `avoid_function_literals_in_foreach_calls` lint

domesticmouse:
# Description

Starts workshopping a reasoning for why `avoid_function_literals_in_foreach_calls` is a good idea for Dart developers.
I'm happy to take corrections in both the intent of the lint and for my explanation prose =)

Fixes: https://github.com/dart-lang/linter/issues/3544


lib/src/rules/avoid_function_literals_in_foreach_calls.dart
```diff
@@ -16,6 +16,9 @@ const _details = r'''
 
 **AVOID** using `forEach` with a function literal.
 
+`forEach` requires a lambda, which creates a new scope, which increases code size and run time.
```
kevmoo, line 19: At a minimum, a new call stack. A new scope *if* an anonymous function is used.

@munificent likely has the right wording here...

kevmoo: er, new stack *frame*

munificent: It tends to be a fool's errand to rely on performance arguments for style choices because who know what the optimizing compiler is able to eliminate? In practice, I don't even know if `forEach()` is actually slower.

There are some subtle semantic reasons to think about:

* `return` inside the closure only returns from that iteration and not the surrounding function. That might be what you want, or might be confusing.
* If you assign to other variables inside the closure, it might break their ability to be type promoted where using a for-in loop is more visible to flow analysis.
* You can't `await` inside the closure.

But, for the most part, the main reason for this guideline is just consistency. You could mostly use `forEach()` without any problems, but there's no reason to prefer that over for-in and if there are two mostly equal choices, the style guide tends to just pick one because consistency is always valuable.

domesticmouse: I've had a go at summarising @munificent's exposition. I'm unsure if I've done it justice, or just butchered it.

On the flipside, as pointed out by @puf, the **good** code sample actually contained code from the **bad** code, which was causing confusion

domesticmouse:
PTAL @pq @bwilkerson 

I'd like to get some version of this fix in =)

bwilkerson (approved):

lib/src/rules/avoid_function_literals_in_foreach_calls.dart
```diff
@@ -16,6 +16,12 @@ const _details = r'''
 
 **AVOID** using `forEach` with a function literal.
 
+The `for` loop enables a developer to be clear and explicit as to their intent.
+A return in the body of the `for` loop returns from the body of function,
```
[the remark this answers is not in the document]
domesticmouse: Suggestion adopted, thanks!

pq (approved):

pq:
Thanks @domesticmouse!

PR #3547, opened 2022-07-22
update to analyzer 4.3.0 APIs

pq:
/cc @scheglov

### DIFF
--- a/lib/src/rules/unnecessary_lambdas.dart
+++ b/lib/src/rules/unnecessary_lambdas.dart
@@ -207,7 +207,7 @@ class _Visitor extends SimpleAstVisitor<void> {
       if (target is SimpleIdentifier) {
         var element = target.staticElement;
         if (element is PrefixElement) {
-          for (var import in element.imports) {
+          for (var import in element.imports2) {
             if (import.isDeferred) {
               return;
             }
@@ -240,4 +240,8 @@ class _Visitor extends SimpleAstVisitor<void> {
       }
     }
   }
+}
+
+extension LibraryImportElementExtension on LibraryImportElement {
+  bool get isDeferred => prefix is DeferredImportElementPrefix;
 }
--- a/lib/src/ast.dart
+++ b/lib/src/ast.dart
@@ -276,10 +276,10 @@ bool _checkForSimpleGetter(MethodDeclaration getter, Expression? expression) {
   if (expression is SimpleIdentifier) {
     var staticElement = expression.staticElement;
     if (staticElement is PropertyAccessorElement) {
-      var enclosingElement = getter.declaredElement?.enclosingElement;
+      var enclosingElement = getter.declaredElement?.enclosingElement2;
       // Skipping library level getters, test that the enclosing element is
       // the same
-      if (staticElement.enclosingElement == enclosingElement) {
+      if (staticElement.enclosingElement2 == enclosingElement) {
         return staticElement.isSynthetic && staticElement.variable.isPrivate;
       }
     }
--- a/lib/src/rules/avoid_classes_with_only_static_members.dart
+++ b/lib/src/rules/avoid_classes_with_only_static_members.dart
@@ -86,7 +86,7 @@ class _Visitor extends SimpleAstVisitor<void> {
     var interface = context.inheritanceManager.getInterface(declaredElement);
     var map = interface.map;
     for (var member in map.values) {
-      var enclosingElement = member.enclosingElement;
+      var enclosingElement = member.enclosingElement2;
       if (enclosingElement is ClassElement &&
           !enclosingElement.isDartCoreObject) {
         return;
--- a/lib/src/rules/avoid_implementing_value_types.dart
+++ b/lib/src/rules/avoid_implementing_value_types.dart
@@ -123,7 +123,7 @@ class _Visitor extends SimpleAstVisitor<void> {
 
   static bool _overridesEquals(ClassElement element) {
     var method = element.lookUpConcreteMethod('==', element.library);
-    var enclosing = method?.enclosingElement;
+    var enclosing = method?.enclosingElement2;
     return enclosing is ClassElement && !enclosing.isDartCoreObject;
   }
 }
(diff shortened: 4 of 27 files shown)


lib/src/rules/unnecessary_lambdas.dart
```diff
@@ -241,3 +241,7 @@ class _Visitor extends SimpleAstVisitor<void> {
     }
   }
 }
+
+extension LibraryImportElementExtension on LibraryImportElement {
```
pq, line 245: FWIW: I used to get `isDeferred` from `ImportElement` itself.  Not sure if `LibraryImportElement` should have it too?

scheglov: `prefix is DeferredImportElementPrefix`

scheglov: Ah, you already did it.

No, I don't think we need `isDeferred`. The type check on the `prefix` works just a well, and in theory we could have more data, e.g. the offset of the `deferred` keyword.

bwilkerson: If I were designing the API I would avoid needing to use type checks. Using a getter like `isDeferred` is
- easier to read,
- easier to change in the future, and
- generally more performant.

I'd like to see `isDeferred` added to the base API, even if it's only on the prefix.

bwilkerson (approved):

lib/src/rules/unnecessary_lambdas.dart
scheglov: Are you saying that we should add `bool get isDeferred` to `ImportElementPrefix`?

Let's say that later we wanted to know `deferredKeywordOffset`, which has any meaning only when `ImportElementPrefix` is deferred. With `DeferredImportElementPrefix` this is easy - add a getter that always has meaningful value `int get deferredKeywordOffset`. Without `DeferredImportElementPrefix` we would either make it return `int?`, or return a special value like `-1` (which is worse than nullable, because nullable forces a check).

I think if there are data dependencies, they should be expressed using types.

bwilkerson: > Are you saying that we should add `bool get isDeferred` to `ImportElementPrefix`?

I'm saying that I would, yes.

> I think if there are data dependencies, they should be expressed using types.

And I can't disagree with that. I don't like the impact it's having on the API, but I haven't yet identified why. If I figure it out, and if it's a valid issue, I'll let you know.

But that doesn't mean that we shouldn't provide a convenient getter in cases where clients aren't trying to access any of the dependent data. It's still a better API. (And I forgot to mention that it's also more discoverable.)

scheglov: We say that something is better API when it has some properties. Just saying that something is better does not seem convincing. More discoverable is one such property that we could say makes API better.

I have an issue with more discoverable though. It depends. In extreme we could design `class Element {}` that has all properties that are present in `Element` subtypes. And type checks as well. Do we want `bool get isClass`, `bool get isEnum` and `bool get isMethod`? Just in case if users don't know that there are subtypes of `Element`?

I don't agree about a convenience getter in this case. For `LibraryImportElement` I decided not to be an extremist, and added `LibraryElement? get importedLibrary`, because it is too much to add `final uri = import.uri; final importedLibrary = uri is DirectiveUriWithLibrary ? uri.library : null;`. But just `import.prefix is DeferredImportElementPrefix`? I think this is convenient enough.

bwilkerson: Yep, we clearly disagree. I suspect it's related to the as-yet-unidentified reason why I don't like the current approach. Without knowing what that reason is, it's hard to argue my case. So I won't persist, but I will think out loud.

I suspect that the disagreement is related to the same issue you raised around discoverability. It is perfectly reasonable to expect users to figure out that a class named `Vehicle` won't have behavior / state for every conceivable kind of vehicle, but that there will be subclasses for the supported kinds of vehicles. I think that `Element` is similar. `Element` is a broad category of objects, just like vehicle is a broad category of objects.

But at some point that all breaks down. If we're modeling vehicles, and we have a class named `Car` with a getter `List<Seat> get seats`, do we really want a subclass named `HeatedSeat`? It's certainly the case that not all seats are heated, but then not all seats recline, or can be moved forward and backward, or up and down, or have adjustable lumbar support. At some point we stop thinking of all of these options as defining separate categories. That's the case here. We're less likely to think of categories of seats and just think of them as attributes of the things in the category `Seat`. (There are exceptions, but I think the general principle applies.)

I would expect a developer to know to look for the class `Car`, but I wouldn't expect them to look for the class `SeatWithHeatVerticalAndLumbarAdjustments`. I'd expect them to assume that they could ask a `Seat` whether it's heated, and even to ask it for the heater when it might not have one (despite the data dependency).

I can't accurately express the general principle yet, but I'm sure there is one, and it feels like the current hierarchy is violating that principle.

scheglov: I agree, I would not go with `SeatWithHeatVerticalAndLumbarAdjustments` too.

I would design this as:
```dart
abstract class Seat {
  SeatHeating? get heating;
  SeatReclining? get reclining;
  SeatLumbarAdjustment? get lumbarAdjustment;
}

abstract class SeatHeating {
  double get currentTemperature;
  double requestedTemperature;
  double get maxTemperature;
}

/// TODO(scheglov) Should we separate `ContinuousSeatReclining` and `DiscreteSeatReclining`?
abstract class SeatReclining {
  double angle;
  double get minAngle;
  double get maxAngle;
}

abstract class SeatLumbarAdjustment {
  LumbarAdjustmentPosition position;
  double juttingOut; // TODO(scheglov) Should we use Range here?
}

enum LumbarAdjustmentPosition {
  low, middle, high
}
```

So, if we expect that there are independent forks possible, we should make these forking points their own attributes. But when there is just one, then we can build an inheritance tree for the enclosing object itself, like `ImportElementPrefix` and `DeferredImportElementPrefix`. When we see that some forking can be reused between objects, like `DirectiveUri` is reused for `LibraryImportElement`, `LibraryExportElement`, `AugmentationImportElement`; we still want to make it an attribute, even though there is just one forking point.

PR #3550, opened 2022-07-25
fix doc example syntax

pq:
/cc @bwilkerson

### DIFF
--- a/lib/src/rules/prefer_inlined_adds.dart
+++ b/lib/src/rules/prefer_inlined_adds.dart
@@ -17,13 +17,13 @@ Declare elements in list literals inline, rather than using `add` and
 **BAD:**
 ```dart
 var l = ['a']..add('b')..add('c');
-var l2 = ['a']..addAll(['b', 'c'])
+var l2 = ['a']..addAll(['b', 'c']);
 ```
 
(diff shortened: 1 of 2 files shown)


bwilkerson (approved):

lib/src/rules/prefer_inlined_adds.dart
```diff
@@ -17,13 +17,13 @@ Declare elements in list literals inline, rather than using `add` and
...
 var l = ['a']..add('b')..add('c');
-var l2 = ['a']..addAll(['b', 'c'])
+var l2 = ['a']..addAll(['b', 'c']);
 ```
 
 **GOOD:**
 ```dart
 var l = ['a', 'b', 'c'];
-var 2 = ['a', 'b', 'c'];
+var 12 = ['a', 'b', 'c'];
```
parlough, line 26: Should this still be `l2` to match the bad example?

```suggestion
var l2 = ['a', 'b', 'c'];
```

pq: Ha.  Absolutely.

pq: Thanks!

PR #3551, opened 2022-07-25
`depend_on_referenced_packages`: cache available dependencies

pq:
It looks like a simple cache of the calculated packages (avoiding needless extra trips to process the pubspec) stand to make `depend_on_referenced_packages` twice as fast in our benchmarks.

NOTE that I generally think mutable `static` anything in a lint is an anti-pattern but in this case safe and probably worth it.  Very open to discussion!

/cc @bwilkerson @jakemac53 

---

Current run:

image

Previous run:

image

### DIFF
--- a/lib/src/rules/depend_on_referenced_packages.dart
+++ b/lib/src/rules/depend_on_referenced_packages.dart
@@ -70,7 +70,8 @@ class DependOnReferencedPackages extends LintRule {
       NodeLintRegistry registry, LinterContext context) {
     var package = context.package;
     var packageHash = package.hashCode;
-    // Only calculate deps if we've got a new package hash.
+    // Only calculate `availableDeps` if we've got a new package hash, otherwise
+    // fall back on our cache.
     if (packageHash != lastPackageHash) {
       lastPackageHash = packageHash;
       if (package is! PubWorkspacePackage) return;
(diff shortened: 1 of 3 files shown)


lib/src/rules/depend_on_referenced_packages.dart
```diff
@@ -64,26 +68,32 @@ class DependOnReferencedPackages extends LintRule {
...
-      name,
-      if (dependencies != null)
-        for (var dep in dependencies)
-          if (dep.name?.text != null) dep.name!.text!,
-      if (devDependencies != null &&
-          !isInPublicDir(context.currentUnit.unit, context.package))
-        for (var dep in devDependencies)
-          if (dep.name?.text != null) dep.name!.text!,
-    ];
+    var packageHash = package.hashCode;
```
bwilkerson, line 72: `hashCode` seems like a questionable choice. AFAICT the `package` is inheriting the implementation from `Object`. I don't remember what the lifetime of a `WorkspacePackage` is, but it means that we might get a different hash code for the same set of dependencies (which is inefficient, but no worse than today) and might be able to get the same hash code for a different set of dependencies (which would be breaking).

If you're certain that can't happen (and that no-one will override `hashCode` in those classes in the future), then I guess it's OK. Otherwise, consider implementing a dedicated method.

pq: Right.  I was just second-guessing this actually.  An even worse case would be getting the *same* `WorkspacePackage` after it's been changed and so an inconsistent cache.  I'll take a closer look.

Thanks!

jakemac53:
SGTM

srawlins:
Good to land, @pq?

pq:
Hmmmm. I seem to recall an issue with this one. I'm inclined to just close it for now and re-open if/when someone can investigate.

PR #3554, opened 2022-07-26
update test expectations

pq:
Should fix the build...

/cc @srawlins @bwilkerson

### DIFF
--- a/test/rules/prefer_contains_test.dart
+++ b/test/rules/prefer_contains_test.dart
@@ -27,6 +27,7 @@ condition() {
 ''', [
       // No lint
       error(HintCode.UNUSED_LOCAL_VARIABLE, 41, 4),
+      error(CompileTimeErrorCode.ARGUMENT_TYPE_NOT_ASSIGNABLE, 77, 3),
     ]);
   }
 
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -12,7 +12,7 @@ environment:
   sdk: '>=2.17.0 <3.0.0'
 
 dependencies:
-  analyzer: ^4.3.0
+  analyzer: ^4.3.1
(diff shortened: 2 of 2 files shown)


srawlins:
Did you mean to CC Brandon Wang?

test/rules/prefer_contains_test.dart
```diff
@@ -27,6 +27,7 @@ condition() {
 ''', [
       // No lint
       error(HintCode.UNUSED_LOCAL_VARIABLE, 41, 4),
+      error(CompileTimeErrorCode.ARGUMENT_TYPE_NOT_ASSIGNABLE, 77, 3),
```
srawlins, line 30: I don't understand where this is coming from. I think the code is valid... `var next;` makes a dynamic, so the assignment should be valid.

pq: The issue is that the list is of type `int` and we're passing in a String to `indexOf`.

image

pq:
> Did you mean to CC Brandon Wang?

Whoops.  Nope!  Apologies for the noise, Brandon!

test/rules/prefer_contains_test.dart
srawlins: Ah I see. So strange that this wasn't picked up before...

srawlins (approved):

test/rules/prefer_contains_test.dart
pq: Yeah.  That's why I bumped the analyzer lower bound.  I'm guessing this got fixed in `4.3.1`.  (@scheglov?)

scheglov: I don't remember fixing anything related recently.

PR #3567, opened 2022-08-01
add lint use_string_in_part_of_directives

### LINKED ISSUE #3487: "Strings in `part of` directives" Effective Dart guideline does not have a lint
Most of the "usage" guidelines in [Effective Dart][1] have a lint, but ["DO use strings in `part of` directives"][2] do not.

Should it be added?

[1]: https://dart.dev/guides/language/effective-dart/usage
[2]: https://dart.dev/guides/language/effective-dart/usage#do-use-strings-in-part-of-directives

a14n:
# Description

From [effective dart](https://dart.dev/guides/language/effective-dart/usage#do-use-strings-in-part-of-directives):

**DO** use strings in `part of` directives.

**BAD:**

```dart
part of my_library;
```

**GOOD:**

```dart
part of '../../my_library.dart';
```

Fixes #3487


test/integration/use_string_in_part_of_directives.dart
```diff
@@ -0,0 +1,68 @@
+// Copyright (c) 2020, the Dart project authors. Please see the AUTHORS file
```
pq, line 1: Would you consider migrating this test to the `reflective_test_loader` style?  These run faster and are easier to debug.

FWIW `ConditionalUriDoesNotExistTest` and `PreferConstConstructorsTest` are examples of tests that uses this style for tests that require multiple files.

(Aside: we should really write up some docs for this...  😬 )

a14n: For this rule I need several files. I don't find example dealing with multiple files. Do you have an example of such case?

pq: Sure!  

In this case `newFile2` creates `a.dart` which is imported by the test file in `assertNoDiagnostics`.  

```dart

  test_deferred_arg() async {
    newFile2('$testPackageLibPath/a.dart', '''
class A {
  const A();
}

const aa = A();
''');

    await assertNoDiagnostics(r'''
import 'a.dart' deferred as a;

class B {
  const B(Object a);
}

main() {
  var b = B(a.aa);
  print(b);
}   
''');
  }

```


You can call `newFile2` multiple times with different library names (`$testPackageLibPath/b.dart` etc.) to get more libraries to import into your test file if needed.

I need to run out for a but but if you get stuck, feel free to share how far you get and I'm happy to take a look.

Thanks!

a14n: Thanks for your help. Tests are now migrated. PTAL

pq (approved):
Super. Thanks!

PR #3568, opened 2022-08-01
rework avoid_types_on_closure_parameters to allow not-redundant types

### LINKED ISSUE #1099: avoid_types_on_closure_parameters should only apply for type repeating inferred type
```dart
dynamic x;
x = (String a) => a.length; // should be OK
```

### LINKED ISSUE #2131: False positive for "avoid_types_on_closure_parameters"
When contravariance applies

```
class Unused {
  const Unused();
}

const unused = Unused();

@immutable
class User {
  final String name;

  const User({this.name});

  User Function({String name}) get copyWith => ({
        Object name = unused, // False positive
      }) =>
          User(
            name: name == unused ? this.name : name as String,
          );
}
```

### LINKED ISSUE #3330: avoid_types_on_closure_parameters false positive with implicit-dynamic
**Describe the issue**
Some SDK functions type their error callback parameter as `Function` to accept both `Function(Object)` and `Function(Object, StackTrace)`:

```dart
Future future;
future.then(..., onError: <this callback>);
```

As such, writing the following will trigger implicit-dynamic:

```dart
Future future;
future.then(..., onError: (err, stack) {
  // err and stack are both dynamic
});
```

The solution should be to explicitly write the type:

```dart
Future future;
future.then(..., onError: (Object err, StackTrace stack) {
  // fixed implicit-dynamic
});
```

but doing this now causes `avoid_types_on_closure_parameters` on both parameters

**Expected behavior**

I would expect avoid_types_on_closure_parameters to detect that Object/StackTrace are types more specific than `dynamic`, and therefore allow it.

**Additional context**

Possibly related to https://github.com/dart-lang/linter/issues/695 and https://github.com/dart-lang/linter/issues/2131

a14n:
# Description

rework avoid_types_on_closure_parameters to allow not-redundant types 

Fixes #3330
Fixes #2131
Fixes #1099

### DIFF
--- a/lib/src/rules/avoid_types_on_closure_parameters.dart
+++ b/lib/src/rules/avoid_types_on_closure_parameters.dart
@@ -4,6 +4,7 @@
 
 import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/type.dart';
 
 import '../analyzer.dart';
 
@@ -11,11 +12,9 @@ const _desc = r'Avoid annotating types for function expression parameters.';
 
 const _details = r'''
 
-**AVOID** annotating types for function expression parameters.
+From [effective dart](https://dart.dev/guides/language/effective-dart/design#dont-annotate-inferred-parameter-types-on-function-expressions):
 
-Annotating types for function expression parameters is usually unnecessary
-because the parameter types can almost always be inferred from the context,
-thus making the practice redundant.
+**DON’T** annotate inferred parameter types on function expressions.
 
 **BAD:**
 ```dart
@@ -43,12 +42,12 @@ class AvoidTypesOnClosureParameters extends LintRule {
   @override
   void registerNodeProcessors(
       NodeLintRegistry registry, LinterContext context) {
-    var visitor = _Visitor(this);
+    var visitor = _Visitor(this, context);
     registry.addFunctionExpression(this, visitor);
   }
 }
 
-class AvoidTypesOnClosureParametersVisitor extends SimpleAstVisitor {
+class AvoidTypesOnClosureParametersVisitor extends SimpleAstVisitor<void> {
   LintRule rule;
 
   AvoidTypesOnClosureParametersVisitor(this.rule);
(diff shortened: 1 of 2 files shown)


pq:
Fantastic.  We'll need to do an internal dry-run regardless , but if I understand the change correctly, this just addresses false-positives?

a14n:
> I understand the change correctly, this just addresses false-positives?

Yes it should relax the original version. So it should add new diagnostics.

lib/src/rules/avoid_types_on_closure_parameters.dart
```diff
@@ -86,12 +85,42 @@ class AvoidTypesOnClosureParametersVisitor extends SimpleAstVisitor {
...
+        }
+      }
+      expectedType = parent.declaredElement?.type;
+    } else if (parent is ReturnStatement || parent is ExpressionFunctionBody) {
+      expectedType =
+          node.thisOrAncestorOfType<FunctionDeclaration>()?.returnType?.type;
+    } else if (staticParameterElement != null) {
+      expectedType = staticParameterElement.type;
+    } else if (parent is NamedExpression) {
+      expectedType = parent.element?.type;
```
a14n, line 117: @bwilkerson I was surprise to hava to add this line (to catch the `Future.then(..., onError: ...)` case). I thought `staticParameterElement` wouldn't have been null here. Is it the normal behaviour or a bug?

bwilkerson: It does seem inconsistent that an expression used as an argument wouldn't have a `staticParameterElement`. I'm not sure whether there's a reason for the current behavior, but I'd like to explore changing it. We should probably open an issue in the sdk repo and loop Konstantin into the conversation.

pq:
> Yes it should relax the original version. So it shouldn't add new diagnostics.

Add?  If these are false positives, I'd expect there to be **_fewer_**...

Sorry if I'm being thick!

a14n:
@pq that's what I tried to say : fewer lints

lib/src/rules/avoid_types_on_closure_parameters.dart
a14n: Looks like this bug was already reported (by me :D) in https://github.com/dart-lang/sdk/issues/45964

a14n:
> We'll need to do an internal dry-run

Did you have a chance to run it internally?

pq:
> Did you have a chance to run it internally?

Sorry, no.  It may be a bit --- I'm still trying to land the last linter release in the SDK (https://dart-review.googlesource.com/c/sdk/+/253501) and have a backlog of other tasks. 

Will definitely queue this up for our next release though.

Thanks for your patience!

pq:
(And also for the nudge!)

devoncarew:

PR #3570, opened 2022-08-01
Move 5 static functions out of DartTypeUtilities

srawlins:
# Description

Move 5 static functions out of DartTypeUtilities. Helps linter comply with [AVOID defining a class that contains only static members.](https://dart.dev/guides/language/effective-dart/design#avoid-defining-a-class-that-contains-only-static-members) and makes the code much more readable.

* `canonicalElementsAreEqual` can be a top-level function
* `canonicalElementsFromIdentifiersAreEqual` can be a top-level function
* `getCanonicalElement` can be an extension on `Element` and be shortened to `canonicalElement` 🎉 
* `getCanonicalElementFromIdentifier` can be an extension on `AstNode` and renamed to be much shorter: `canonicalElement` 🎉 
* `isNullLiteral` can be an extension on `Expression`

### DIFF
--- a/lib/src/util/dart_type_utilities.dart
+++ b/lib/src/util/dart_type_utilities.dart
@@ -146,6 +146,10 @@ class DartTypeUtilities {
           DartType? type, String? className, String? library) =>
       _extendsClass(type, <ClassElement>{}, className, library);
 
+  @Deprecated('Replace with `rawNode.canonicalElement`')
+  static Element? getCanonicalElementFromIdentifier(AstNode? rawNode) =>
+      rawNode.canonicalElement;
+
   static Iterable<InterfaceType> getImplementedInterfaces(InterfaceType type) {
     void recursiveCall(InterfaceType? type, Set<ClassElement> alreadyVisited,
         List<InterfaceType> interfaceTypes) {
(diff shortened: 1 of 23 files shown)


pq:
From a style perspective, I'm totally into this (and there are other changes we could make too).  Having said that, this is kinda-sorta API insofar as Google3-internal lints may get broken by a move.  Maybe consider a quick codesearch to see?

A more polite approach might be to deprecate these methods with a threat to remove them real soon.  (In practice this may be nearly just as breaking though...  Let's discuss!)

srawlins:
I'll replace with deprecated forwarders for anything used in google3.

srawlins:
* `canonicalElementsAreEqual`, `canonicalElementsFromIdentifiersAreEqual`, `getCanonicalElement`: 0 uses
* `getCanonicalElementFromIdentifier`, `isNullLiteral`: 2 uses each

srawlins:
Added deprecated replacements for the two with references in google3.

bwilkerson (approved):

lib/src/util/dart_type_utilities.dart
```diff
@@ -595,6 +558,48 @@ class InterfaceTypeDefinition {
   }
 }
 
+extension ElementExtensions on Element {
```
pq, line 568: Consider moving these into `ast.dart`?  Or maybe introducing a new `extensions.dart` like we have in analyzer?

pq (approved):

lib/src/util/dart_type_utilities.dart
srawlins: Ah ha I hadn't seen that file. Done!

PR #3572, opened 2022-08-02
Move two more DartTypeUtilities static methods into extensions

srawlins:
# Description

Move 2 static functions out of DartTypeUtilities. Helps linter comply with [AVOID defining a class that contains only static members.](https://dart.dev/guides/language/effective-dart/design#avoid-defining-a-class-that-contains-only-static-members) and makes the code much more readable.

* `asEnumLikeClass` and `hasSubclassInDefiningCompilationUnit`
* Also moved `isEnumLikeClass` into this extension, as it seems equally utility-like as the others.

### DIFF
--- a/lib/src/extensions.dart
+++ b/lib/src/extensions.dart
@@ -3,7 +3,9 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'package:analyzer/dart/ast/ast.dart';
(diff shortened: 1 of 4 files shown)


pq (approved):
👍 

Did you happen to do an internal codesearch?

srawlins:
Neither is used in google3 😁

PR #3574, opened 2022-08-02
Move 4 static DartTypeUtilities methods to extensions

srawlins:
# Description

* `isClassElement` can be `isClass` on ClassElementExtension; added documentation
* `getLastStatementInBlock` can be `get lastStatement` on BlockExtension; added documentation
* `extendsClass` can be `extendsClass` on DartTypeExtension
* `getImplementedInterfaces` can be `get implementedInterfaces` on InterfaceTypeExtension

Only `extendsClass` is used internally, so kept that with a Deprecation notice.

I think readability is improved in each case.

### DIFF
--- a/lib/src/rules/omit_local_variable_types.dart
+++ b/lib/src/rules/omit_local_variable_types.dart
@@ -8,6 +8,7 @@ import 'package:analyzer/dart/element/element.dart';
 import 'package:analyzer/dart/element/type.dart';
 
 import '../analyzer.dart';
+import '../extensions.dart';
 import '../util/dart_type_utilities.dart';
 
 const _desc = r'Omit type annotations for local variables.';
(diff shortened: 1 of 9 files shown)


lib/src/rules/omit_local_variable_types.dart
```diff
@@ -102,9 +103,8 @@ class _Visitor extends SimpleAstVisitor<void> {
       }
       var iterableType = loopParts.iterable.staticType;
       if (iterableType is InterfaceType) {
-        var iterableInterfaces = DartTypeUtilities.getImplementedInterfaces(
-                iterableType)
-            .where((type) =>
+        var iterableInterfaces = iterableType.implementedInterfaces.where(
```
bwilkerson, line 106: I wonder whether this might not be cleaner and even more correct to use `DartType.asInstanceOf`.

bwilkerson (approved):

pq (approved):
💎

lib/src/rules/omit_local_variable_types.dart
srawlins: I was wondering the same thing, haha. I've added a TODO.

PR #3575, opened 2022-08-02
Move DartTypeUtilities.overridesMethod to an extension

srawlins:
# Description

Move `DartTypeUtilities.overridesMethod` to an extension, and rename it `get isOverride` (sticks to convention of other bool-returning getters). I also refactored the end of the method body to be more readable.

### DIFF
--- a/lib/src/extensions.dart
+++ b/lib/src/extensions.dart
@@ -138,3 +138,36 @@ extension AstNodeExtensions on AstNode? {
 extension ExpressionExtensions on Expression? {
   bool get isNullLiteral => this?.unParenthesized is NullLiteral;
 }
+
+extension MethodDeclarationExtension on MethodDeclaration {
+  /// Returns whether this method is an override of a method in any supertype.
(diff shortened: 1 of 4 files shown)


lib/src/extensions.dart
```diff
@@ -138,3 +138,36 @@ extension AstNodeExtensions on AstNode? {
...
+    if (parentElement == null) {
+      return false;
+    }
+    var parentLibrary = parentElement.library;
+
+    if (isGetter) {
+      // Search supertypes for a getter of the same name.
+      return parentElement.allSupertypes
+          .any((t) => t.lookUpGetter2(name, parentLibrary) != null);
+    } else if (isSetter) {
```
pq, line 163: Much clearer.  Thanks!

pq (approved):

lib/src/extensions.dart
bwilkerson: Agreed.

PR #3578, opened 2022-08-04
New lint: Avoid unstable final fields

eernstg:
This PR implements the lint `avoid_unstable_final_fields`, cf. issue #3440.

The basic idea is that a final field declaration is taken to indicate that the given property is immutable, and hence every override which is not guaranteed to return the same value each time it is invoked will be linted.

```dart
class A {
  final int i; // Taken to be immutable, because it's a final variable.
  A(this.i);
}

class B1 implements A {
  static _i = 0;
  int get i => ++_i; // LINT: Does not return the same object every time.
}

class B2 implements A {
  final A a = A(42);
  int get i => a.i; // OK.
}
```

The PR is intended to serve as a point of reference, facilitating discussion and feedback on the lint and the underlying ruleset, as well as on the implementation.

[Edit: Removed remarks about landing this PR.]

### DIFF
--- a/lib/src/rules/avoid_unstable_final_fields.dart
+++ b/lib/src/rules/avoid_unstable_final_fields.dart
@@ -6,7 +6,6 @@ import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/token.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
 import 'package:analyzer/dart/element/element.dart';
-import 'package:analyzer/dart/element/type.dart';
 
 import '../analyzer.dart';
 
--- a/lib/src/rules/avoid_unstable_final_fields.dart
+++ b/lib/src/rules/avoid_unstable_final_fields.dart
@@ -505,7 +505,7 @@ abstract class _AbstractVisitor extends ThrowingAstVisitor<void> {
     if (prefixDeclaration is PrefixElement) {
       var declaredElement = node.identifier.staticElement?.declaration;
       if (!_isStable(declaredElement)) isStable = false;
-    } else if (prefixDeclaration is ClassElement) {
+    } else if (prefixDeclaration is InterfaceElement) {
       var declaredElement = node.identifier.staticElement?.declaration;
       if (!_isStable(declaredElement)) isStable = false;
     }
--- a/lib/src/rules/avoid_unstable_final_fields.dart
+++ b/lib/src/rules/avoid_unstable_final_fields.dart
@@ -286,7 +286,7 @@ abstract class _AbstractVisitor extends ThrowingAstVisitor<void> {
 
   @override
   void visitInterpolationExpression(InterpolationExpression node) {
-    // TODO(eernst): Wc eould handle several cases here.
+    // TODO(eernst): Wc could handle several cases here.
     isStable = false;
   }
 
@@ -319,6 +319,7 @@ abstract class _AbstractVisitor extends ThrowingAstVisitor<void> {
 
   @override
   void visitPostfixExpression(PostfixExpression node) {
+    print('>>> PostFixExpression, $node');
     // TODO(eernst): `x.y?.z` is property accesses, so always unstable here?
     isStable = false;
   }
--- a/lib/src/rules/avoid_unstable_final_fields.dart
+++ b/lib/src/rules/avoid_unstable_final_fields.dart
@@ -105,7 +105,7 @@ bool _isLocallyStable(Element element) {
       for (var elementAnnotation in metadata) {
         var metadataElement = elementAnnotation.element;
         if (metadataElement is ConstructorElement) {
-          var metadataOwner = metadataElement.declaration.enclosingElement3;
+          var metadataOwner = metadataElement.declaration.enclosingElement;
           if (metadataOwner is ClassElement && metadataOwner.isDartCoreObject) {
             // A declaration with `@Object()` is not considered stable.
             return false;
@@ -166,7 +166,7 @@ abstract class _AbstractVisitor extends ThrowingAstVisitor<void> {
 
   bool _isStable(Element? element) {
     if (element == null) return false; // This would be an error in the program.
-    var enclosingElement = element.enclosingElement3;
+    var enclosingElement = element.enclosingElement;
     if (_isLocallyStable(element)) return true;
     if (element is PropertyAccessorElement) {
       if (element.isStatic) return false;
@@ -630,18 +630,18 @@ class _FieldVisitor extends _AbstractVisitor {
     Name? name;
     InterfaceElement? interfaceElement;
     for (var variable in node.fields.variables) {
-      var declaredElement = variable.declaredElement2;
+      var declaredElement = variable.declaredElement;
       if (declaredElement is FieldElement) {
         // A final instance variable can never violate stability.
         if (declaredElement.isFinal) continue;
         // A non-final instance variable is always a violation of stability.
         // Check if stability is required.
         interfaceElement ??=
-            declaredElement.enclosingElement3 as InterfaceElement;
+            declaredElement.enclosingElement as InterfaceElement;
         libraryUri ??= declaredElement.library.source.uri;
         name ??= Name(libraryUri, declaredElement.name);
         if (_inheritsStability(interfaceElement, name)) {
-          doReportLint(variable.name2);
+          doReportLint(variable.name);
         }
       }
     }
@@ -655,15 +655,15 @@ class _MethodVisitor extends _AbstractVisitor {
   void visitMethodDeclaration(MethodDeclaration node) {
     if (!node.isGetter) return;
     declaration = node;
-    var declaredElement = node.declaredElement2;
+    var declaredElement = node.declaredElement;
     if (declaredElement != null) {
-      var enclosingElement = declaredElement.enclosingElement3;
+      var enclosingElement = declaredElement.enclosingElement;
       if (enclosingElement is InterfaceElement) {
         var libraryUri = declaredElement.library.source.uri;
         var name = Name(libraryUri, declaredElement.name);
         if (!_inheritsStability(enclosingElement, name)) return;
         node.body.accept(this);
-        if (!isStable) doReportLint(node.name2);
+        if (!isStable) doReportLint(node.name);
       } else {
         // Extensions cannot override anything.
       }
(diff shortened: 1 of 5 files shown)


lib/src/rules/avoid_unstable_final_fields.dart
```diff
@@ -0,0 +1,685 @@
...
+    var visitor = _Visitor(this, context);
+    registry.addFieldDeclaration(this, visitor);
+    registry.addMethodDeclaration(this, visitor);
+  }
+}
+
+bool _isLocallyStable(Element element) {
+  if (element is PropertyAccessorElement &&
+      element.isSynthetic &&
+      element.correspondingSetter == null) {
```
bwilkerson, line 93: `correspondingSetter` will also return `null` if `element` is a setter. You probably want to first test `element.isGetter`.

lib/src/rules/avoid_unstable_final_fields.dart
```diff
@@ -0,0 +1,685 @@
...
+  } else if (element is FunctionElement) {
+    // A tear-off of a top-level function or static method is stable,
+    // local functions and function literals are not.
+    return element.isStatic;
+  } else if (element is ClassElement) {
+    // A reified type of a class is stable.
+    return true;
+  } else if (element is MethodElement) {
+    // An instance method tear-off is never stable.
+    return false;
```
bwilkerson, line 118: I believe that this needs to be `return element.isStatic;` because static methods are also represented by `MethodElement`s.

lib/src/rules/avoid_unstable_final_fields.dart
```diff
@@ -0,0 +1,685 @@
...
+  }
+  // TODO(eernst): Any cases still missing?
```
bwilkerson, line 120: The latest version of the analyze package (a PR to move to it is in flight as I type) uses the new classes `EnumElement` and `MixinElement` rather than using `ClassElement` for all of the cases. You'll probably want to add those cases the next time you rebase your branch.

lib/src/rules/avoid_unstable_final_fields.dart
```diff
@@ -0,0 +1,685 @@
...
+  return false;
+}
+
+abstract class _AbstractVisitor extends ThrowingAstVisitor<void> {
+  final LintRule rule;
+  final LinterContext context;
+
+  // Will be true initially when a getter body is traversed. Will be made
+  // true if the getter body turns out to be unstable. Is checked after the
```
bwilkerson, line 129: 'true' --> 'false'

lib/src/rules/avoid_unstable_final_fields.dart
```diff
@@ -0,0 +1,685 @@
...
+  // traversal of the body, ot emit a lint if it is still false at that time.
```
bwilkerson, line 130: 'ot'?

lib/src/rules/avoid_unstable_final_fields.dart
```diff
@@ -0,0 +1,685 @@
...
+  var causes = <PropertyAccessorElement>{};
+
+  // Initialized in `visitMethodDeclaration` if a lint might be emitted.
+  // It is then guaranteed that `declaration.isGetter` is true.
+  late final MethodDeclaration declaration;
+
+  _AbstractVisitor(this.rule, this.context);
+
+  bool _inheritsStability(
+      ClassElement classElement, Name name, LinterContext context) {
```
bwilkerson, line 144: You probably don't need to pass the `LinterContext`, here and below, given that it's visible as an instance field.

lib/src/rules/avoid_unstable_final_fields.dart
```diff
@@ -0,0 +1,685 @@
...
+    if (declaredElement != null) {
+      var enclosingElement = declaredElement.enclosingElement;
+      if (enclosingElement is ClassElement) {
+        var libraryUri = declaredElement.library.source.uri;
+        var name = Name(libraryUri, declaredElement.name);
+        if (!_inheritsStability(enclosingElement, name, context)) return;
+        node.body.accept(this);
+        if (!isStable) doReportLint(node, node.name);
+      } else {
+        // Extensions cannot override anything.
```
bwilkerson, line 658: You'll need to add branches for `EnumElement` and `MixinElement` when you rebase.

bwilkerson (commented):
This looks good. Just a couple of nits and forward pointers to changes that will land soon.

lib/src/rules/avoid_unstable_final_fields.dart
eernstg: Indeed!

lib/src/rules/avoid_unstable_final_fields.dart
eernstg: Done.

eernstg (commented):
Got it, thanks!

lib/src/rules/avoid_unstable_final_fields.dart
eernstg: Having rebased, I get a lot of github action failures, in libraries that I haven't touched: https://github.com/dart-lang/linter/runs/7767039216?check_suite_focus=true.

eernstg: The github actions fail, reporting 'info - lib/src/rules/avoid_unstable_final_fields.dart:641:39 - 'name' is deprecated and shouldn't be used. Use name2 instead. Try replacing the use of the deprecated member with the replacement. - deprecated_member_use', in spite of the fact that `name2` does not exist.

bwilkerson: The class `MethodDeclaration` defines both `name` and `name2`:
```
  /// Return the name of the method.
  @Deprecated('Use name2 instead')
  SimpleIdentifier get name;

  /// Return the name of the method.
  Token get name2;
```
It's possible that it was in a bad state at one point, or that `node` has been promoted to some other type. If you update and it still isn't right, let me know. (Or if the "Got it" below means this issue has been resolved, ignore this comment.)

eernstg: OK, I now get no error that `name2` does not exist, but it is a `Token`, and `reportLint` wants an `AstNode` (which is the `SimpleIdentifier` that contains the name for other lint reports). How do I get hold of an `AstNode` for the name when I only have a `Token`?

bwilkerson: > How do I get hold of an `AstNode` for the name when I only have a `Token`?

You can't. You can use `reportLintForToken` for those cases.

> Are lint messages not supposed to have context information? Is there some other mechanism which is more appropriate?

Yes, lints can have context messages.

The `reportLint*` methods take an optional `contextMessages:` argument that will associate the context messages with the diagnostic.

eernstg: Ah, great, `reportLintForToken` was just what I needed!

For the context messages I'm already using that optional parameter, but I couldn't find any other way to _create_ a diagnostic message than via the constructor of `DiagnosticMessageImpl`, and that's where I need to import a `src` library. Is there another way?

bwilkerson: No, there isn't, though there ought to be. I guess the answer is that you have to have a src import for now.

eernstg: OK, thanks a lot! I've added `// ignore:implementation_imports` just before that import, hoping that this will suppress the failure in the github actions.

eernstg: Worked!

srawlins:
Now that Dart 3 has landed, I'd love to hear your thoughts on continuing down this road, @eernstg . The description above says this is not intended to be landed at this point.

I think I'd be very interested in hearing whether stable getters (https://github.com/dart-lang/language/issues/1518) is still on the table for any upcoming release. It is probably my favorite possibly upcoming feature :)

That being said, this rule would still provide great value even if we never land that feature, yes?

Thanks!

eernstg:
Thanks for the kind words! The 'final getters'/'stable getters' feature does come up in discussions in the language team repeatedly, and I'm not alone in pushing it. But most of the language team is worried about the potential breakage (if we change the meaning of the existing modifier `final` on non-local variable declarations) or the potentially small impact (if we require a new keyword like `stable` in order to make any getter stable/final). So it's not a given that any version of this feature will be added to any release of Dart.

Nevertheless, I think it would be great if we could land the lint: It flags situations where a final instance variable is overridden by a getter which is not known to return the same object on repeated invocations, and that seems to be a somewhat delicate situation anyway. It would hardly hurt anyone if the lint exists, as long as it isn't part of a set like 'recommended' or 'core'.

One thing needs to be settled, however: There must be a way to indicate that a given `final` instance variable induces a getter which is not final/stable. In the current version of the lint I'm using `@Object()` to indicate that, just because this was something that I could do without changing anything in package meta or dart:core (and also because this annotation is basically never used today). We could use a random value like `const unstableGetter = '2d1ae4a0241b54c8ab4f4b55fc546079a1fe4191';` which can be declared anywhere it is needed, and then `@unstableGetter` can be used to play this role; however, that's probably too brittle to be acceptable.

srawlins:
> Nevertheless, I think it would be great if we could land the lint

Yeah I think this is a good path forward (to enable the language feature, or as its own end).

> One thing needs to be settled, however: There must be a way to indicate that a given final instance variable induces a getter which is not final/stable.

I think we'd be very open to adding an annotation to `package:meta`. `@unstableGetter` perhaps. I like the idea of the magic, equal but distant constants that any developer can declare and then use, but I think the UX there is pretty weird. Fine to stay conventional with `package:meta`. I can start that conversation.

bwilkerson:
It shouldn't be a problem to add an annotation to mark an unstable getter. Just to double check though, are we sure we want a lint with an annotation to opt out rather than no lint and an annotation to opt in (that is, assume every getter is unstable unless it's marked as being `@stable`)? As I understand it, without the `@stable` annotation users can't mark getters as being stable, only `final` fields get that designation.

Assuming we want the lint, we'll probably need a fix for cases where the lint breaks the code. Is there any remediation other than marking all overriding getters as being unstable?

If we have the notion of a stable getter, is there anything we can/should do at the invocation sites? I could, for example, imagine suggesting that invocations within a loop could be hoisted (as long as the receiver doesn't change), or suggesting that a single invocation would be more efficient when the same getter is invoked multiple times. These are, of course, things that it would be nice for the compiler to optimize automatically, but if this isn't a language feature the compiler probably can't depend on the semantics.

bwilkerson:
Since the last comment on this PR we've moved the linter sources into the `sdk` repo and stopped accepting PRs against this repo. Would you mind converting this into a Gerrit CL?

eernstg:
Sure! I'll do that tomorrow.

eernstg:
@bwilkerson, I've created https://dart-review.googlesource.com/c/sdk/+/326342.

eernstg:
Closing: This effort is now handled via https://dart-review.googlesource.com/c/sdk/+/326342.

PR #3585, opened 2022-08-04
Speed up use_late_for_private_fields_and_variables

srawlins:
# Description

The gist is that UnifyingAstVisitor was a little expensive, as we ultimately only do any work if we're looking at (1) the left side of an AssignmentExpression, or (2) an Identifier or a PropertyAccess. So it wasn't too hard to just visit those nodes with a RecursiveAstVisitor.

Also in `visitFieldDeclaration`, I moved the `parent != null` check to be outside the for loop

I also tidied up a few comments to be [formatted like sentences](https://dart.dev/guides/language/effective-dart/documentation#do-format-comments-like-sentences).

Fixes https://github.com/dart-lang/linter/issues/3584

Does not address the P1 memory leak or the false positive bug.

### DIFF
--- a/lib/src/rules/use_late_for_private_fields_and_variables.dart
+++ b/lib/src/rules/use_late_for_private_fields_and_variables.dart
@@ -74,7 +74,7 @@ class UseLateForPrivateFieldsAndVariables extends LintRule {
   }
 }
 
-class _Visitor extends UnifyingAstVisitor<void> {
+class _Visitor extends RecursiveAstVisitor<void> {
   static final lateables =
       <CompilationUnitElement, List<VariableDeclaration>>{};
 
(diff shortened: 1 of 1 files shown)


pq:
Whoa.  Checkout these benchmark comparisons.

Before:

image


After:

image

That's a big improvement!

@a14n: I'm curious about the rationale for using a unifying visitor...  Maybe this made more sense with a previous iteration of the source?

a14n:
> @a14n: I'm curious about the rationale for using a unifying visitor... Maybe this made more sense with a previous iteration of the source?

TBH I don't remember and TBH2 I didn't think there was this perf diff between UnifyingAstVisitor and RecursiveAstVisitor

srawlins:
On my machine I also noticed a 50% improvement 🎉

pq (approved):
💎

lib/src/rules/use_late_for_private_fields_and_variables.dart
```diff
@@ -74,7 +74,7 @@ class UseLateForPrivateFieldsAndVariables extends LintRule {
   }
 }
 
-class _Visitor extends UnifyingAstVisitor<void> {
+class _Visitor extends RecursiveAstVisitor<void> {
```
bwilkerson, line 77: This is still concerning. The visitors added to the registry (line 73) should implement `SimpleAstVisitor`. With this particular code pattern we're visiting the AST twice, which is less efficient than just registering for the visit methods implemented in the visitor. (I didn't look in depth to understand if there's a reason that wouldn't work here.)

PR #3586, opened 2022-08-05
Use package_config.json format for mock packages files.

scheglov:
We are removing `.packages` format, and the analyzer stopped supporting it.


scheglov:
Was already done in a wrong branch.

PR #3592, opened 2022-08-10
Add lint implicit_call_tearoffs

natebosch:
In an effort to simplify the language we may consider removing the
implicit `call` tearoff which can coerce an instance of any class which
defines a `call` method into a `Function` type. Authors should
explicitly add the `.call` so there is less magic.


natebosch:
I'm having trouble with this lint. I don't see any of the `ImplicitCallReference` nodes that I expect.

Testing this locally requires a dependency override on `analyzer` with https://dart-review.googlesource.com/c/sdk/+/254465


I can see that the `_insertImplicitCallReference` code path is being hit, and the `NodeReplacer.replace` call returns `true`.

bwilkerson:
@srawlins @scheglov 

Is it possible that the `ImplicitCallReference` node is being dropped by either writing or reading the summary file?

lib/src/rules/avoid_implicit_call_tearoffs.dart
```diff
@@ -0,0 +1,71 @@
...
+            description: _desc,
+            details: _details,
+            group: Group.style,
+            maturity: Maturity.experimental);
+
+  @override
+  void registerNodeProcessors(
+      NodeLintRegistry registry, LinterContext context) {
+    var visitor = _Visitor(this);
+    registry.addSimpleIdentifier(this, visitor);
```
stereotype441, line 58: This needs to be `registry.addImplicitCallReference(this, visitor);`.  This is the reason your lint isn't firing when you test it.

stereotype441 (commented):
It turns out that the utility function that dumps the AST in the event of a test failure does a fresh parse of the AST (see https://github.com/dart-lang/sdk/blob/f45353eafd1a28b5b33ff227a5c46c657c0b550c/pkg/analyzer/lib/src/lint/util.dart#L57), so that's why you weren't seeing `ImplicitCallReference` nodes in your AST dump.

natebosch:
@bwilkerson - do you expect that an expression like `Callable() as Function` should get a `ImplicitCallReference` node in the AST?

I can see that we [check whether to insert an implicit call reference](https://github.com/dart-lang/sdk/blob/d0a60d67c9820d0a146f48adf0c1111b87e840f8/pkg/analyzer/lib/src/generated/resolver.dart#L989), however the `contextType` argument is always `null`, and the parent won't be an `AssignmentExpression`, so it will [bail before inserting an implicit call reference](https://github.com/dart-lang/sdk/blob/d0a60d67c9820d0a146f48adf0c1111b87e840f8/pkg/analyzer/lib/src/generated/resolver.dart#L2615)

Is it possible there is a bug that is causing some implicit call tearoffs to not be represented?

bwilkerson:
I don't know. I think @srawlins added that node, so he's likely to be a better source for answers than I am.

srawlins:
> I can see that we [check whether to insert an implicit call reference](https://github.com/dart-lang/sdk/blob/d0a60d67c9820d0a146f48adf0c1111b87e840f8/pkg/analyzer/lib/src/generated/resolver.dart#L989), however the contextType argument is always null, and the parent won't be an AssignmentExpression, so it will [bail before inserting an implicit call reference](https://github.com/dart-lang/sdk/blob/d0a60d67c9820d0a146f48adf0c1111b87e840f8/pkg/analyzer/lib/src/generated/resolver.dart#L2615)

Hmm I doubt that the `contextType` passed in to `visitAsExpression` on line 983 is buggy, as it is used in flow analysis, etc. But maybe we should treat `as Function` as an (explicit) implicit tear-off of `.call`? Since subclassing Function is not allowed. I don't think I accounted for `as Function` when writing the feature.

srawlins:
> Is it possible there is a bug that is causing some implicit call tearoffs to not be represented?

Most definitely, since it's not used a lot in the wild.

natebosch:
> Hmm I doubt that the `contextType` passed in to `visitAsExpression` on line 983 is buggy, as it is used in flow analysis, etc.

In what situation would you expect it to be non-null? When I see this method hit in the code I'm parsing I only have seen `null`.

> But maybe we should treat `as Function` as an (explicit) implicit tear-off of `.call`? Since subclassing Function is not allowed. I don't think I accounted for `as Function` when writing the feature.

I had assumed that you did account for `as Function` base on the handling of implicit call references in `visitAsExpession`. Can you help me understand the intent of that call if we _aren't_ accounting for `as Function` with implicit call references?

srawlins:
> In what situation would you expect it to be non-null? When I see this method hit in the code I'm parsing I only have seen null.

I would expect it to be non-null for something like the AsExpression here:

```dart
T f(Callable callable) => callable as Function;
```

I think the contextType there should be T (like `int` or `Function` or something).

I had assumed that you did account for as Function base on the handling of implicit call references in visitAsExpession. Can you help me understand the intent of that call if we aren't accounting for as Function with implicit call references?

No they're just sort of accounted for in most expressions. I'm not sure the list, but for example in [this test case](https://github.com/dart-lang/sdk/blob/main/pkg/analyzer/test/src/dart/resolution/ast_rewrite_test.dart#L126) we rewrite `c1 ?? c2` to be an ImplicitCallReference, which I think takes place in [`visitBinaryExpression`](https://github.com/dart-lang/sdk/blob/d0a60d67c9820d0a146f48adf0c1111b87e840f8/pkg/analyzer/lib/src/generated/resolver.dart#L1063)

There are more tests below that which show which expressions we'll rewrite as an implicit call tear-off.

I actually can't get a similar test for an AsExpression to generate an ImplicitCallReference node; it might be because inference is never used when computing the static type of an as-expression (the static type is just always the right side if the  as-expression).

natebosch:
Ah, I see now that `as Function` does _not_ have an implicit call tearoff like I thought it did. It throws at runtime instead.

natebosch:
@pq - this lint is ready for review.

kevmoo:
@natebosch – we should add this to lints/core.yaml – right? Do we have an issue open to track that?

natebosch:
> we should add this to lints/core.yaml – right?

Yes

> Do we have an issue open to track that?

https://github.com/dart-lang/language/issues/2399 tracks the language change.  I noted there that we'll also want to include this line in the core set.

natebosch:
@pq - do you have an opinion on whether this should be named `avoid_implicit_call_tearoffs` instead? I had that originally and renamed it, IIRC because I saw something that indicated we wanted to move away from that naming pattern, but now I can't find the reference.

pq:
> @pq - do you have an opinion on whether this should be named avoid_implicit_call_tearoffs instead? 

I don't feel super strongly to be honest but we've been trending away from prefixes like `avoid` so `implicit_call_tearoffs` feels more consistent.

lib/src/rules/implicit_call_tearoff.dart
```diff
@@ -0,0 +1,71 @@
...
+import '../analyzer.dart';
+
+const _desc =
+    r'Explicitly tear-off `call` methods when using an object as a Function.';
+
+const _details = r'''
+**DO** ...
+Explicitly tear off `.call` methods from objects when assigning to a Function
+type.
+
```
pq, line 17: Consider a sentence of rationale here?  (Maybe cribbed from the PR description which I found super edifying!)

bwilkerson:
+1 to not using "avoid".

pq (changes requested):
This looks great.  Could we rename it to `implicit_call_tearoffs` (plural)?

test_data/rules/implicit_call_tearoff.dart
```diff
@@ -0,0 +1,67 @@
+// Copyright (c) 2022, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// test w/ `dart test -N implicit_call_tearoff`
+
+class C {
+  void call() {}
```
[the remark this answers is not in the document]
srawlins: Actually you can even _explicitly_ instantiate the _implicit_ tear-off (\*cringe*):

```dart
class C {
  T call<T>(T t) => t;
}

foo() {
  C()<int>;
}
```

witchcraft.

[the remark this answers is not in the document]
natebosch: Done.

lib/src/rules/implicit_call_tearoff.dart
natebosch: Done.

example/all.yaml
```diff
@@ -78,6 +78,7 @@ linter:
     - flutter_style_todos
     - hash_and_equals
     - implementation_imports
+    - implicit_call_tearoff
```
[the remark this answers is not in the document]
natebosch: Done.

lib/src/rules/implicit_call_tearoffs.dart
```diff
@@ -0,0 +1,72 @@
...
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+
+import '../analyzer.dart';
+
+const _desc =
+    r'Explicitly tear-off `call` methods when using an object as a Function.';
+
+const _details = r'''
+**DO** ...
```
srawlins, line 14: Hmm, is this ellipsis intentional?

srawlins (approved):
I think it's great. Love the tests.

lib/src/rules/implicit_call_tearoffs.dart
natebosch: Not intentional. I missed removing it from the template.

pq (approved):
Awesome.  Thanks!

PR #3594, opened 2022-08-10
Move traverseNodesInDFS to extension

srawlins:
# Description

Move `DartTypeUtilities.traverseNodesInDFS` to be an extension method on AstNode.

I also refactored it to be non-recursive. This appears to save ~3% in the big benchmark, but I don't know what the error window is... it might be a negligible performance change.

It is used internally, so I've kept the old static method with a deprecation.

### DIFF
--- a/lib/src/extensions.dart
+++ b/lib/src/extensions.dart
@@ -2,6 +2,8 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:collection';
+
 import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/constant/value.dart';
 import 'package:analyzer/dart/element/element.dart';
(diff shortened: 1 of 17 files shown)


lib/src/extensions.dart
```diff
@@ -148,6 +148,44 @@ extension AstNodeExtension on AstNode? {
...
+  ///
+  /// It excludes the nodes for which the [excludeCriteria] returns true. If
+  /// [excludeCriteria] is not provided, all nodes are included.
+  Iterable<AstNode> traverseNodesInDFS({AstNodePredicate? excludeCriteria}) {
+    var self = this;
+    if (self == null) {
+      return [];
+    }
+    var nodes = <AstNode>{};
+    var nodesToVisit = List.of(self.childEntities.whereType<AstNode>());
```
bwilkerson, line 163: Also not necessary.

Consider adding a private extension method on `AstNode` like
```
List<AstNode> get childNodes => self.childEntities.whereType<AstNode>();
```

lib/src/extensions.dart
```diff
@@ -148,6 +148,44 @@ extension AstNodeExtension on AstNode? {
...
+    if (excludeCriteria == null) {
+      while (nodesToVisit.isNotEmpty) {
+        var node = nodesToVisit.removeAt(0);
+        nodes.add(node);
+        var i = 0;
+        // Queue up child nodes in the order they appear in `childEntities`.
+        for (var child in node.childEntities) {
```
bwilkerson, line 170: Also not necessary.

Consider using `insertAll` rather than a loop.

lib/src/extensions.dart
```diff
@@ -148,6 +148,44 @@ extension AstNodeExtension on AstNode? {
     }
     return null;
   }
+
+  /// Builds the list resulting from traversing the node in DFS and does not
+  /// include the node itself.
+  ///
+  /// It excludes the nodes for which the [excludeCriteria] returns true. If
+  /// [excludeCriteria] is not provided, all nodes are included.
+  Iterable<AstNode> traverseNodesInDFS({AstNodePredicate? excludeCriteria}) {
```
bwilkerson, line 157: Not necessary, because cleaning up this method wasn't the goal, but ...

Given the potential size of the list being returned, I wonder whether this would be a place where it would be worth the overhead to use `sync*`.

bwilkerson (approved):

lib/src/extensions.dart
srawlins: Good call, done.

lib/src/extensions.dart
srawlins: 🤦 I knew I was missing something. Thanks!

lib/src/extensions.dart
srawlins: Yeah I wondered that too.

I think I'll leave this as is, and look at the use cases in linter, to see how many of them immediately call something like `firstWhere`, etc. I don't think it would be out of bounds to maybe add a new method here for different use cases.

PR #3596, opened 2022-08-11
prefer_contains: Move constant evaluation to be later

### LINKED ISSUE #3549: improve performance for `prefer_contains` (by delaying const eval)
While looking at #3546, I noticed that if we moved constant evaluation of the compared int value until after we've established that it's a function call we care about, we might get a little perf boost.

srawlins:
# Description

prefer_contains: Move constant evaluation to be later

For performance reasons

Fixes #3549


bwilkerson (approved):

pq:
Fantastic!  Curious if you happened to look at benchmarks before and after? (Not that our benchmarks are great but I'm curious if it made a measurable difference.)

srawlins:
I didn't, haha. But I think it's a better paradigm anyhow.

PR #3600, opened 2022-08-12
Move last methods out from DartTypeUtilities

srawlins:
# Description

`DartTypeUtilities.implementsAnyInterface`, `DartTypeUtilities.implementsAnyInterface`, `DartTypeUtilities.isSameAs` become extension methods on DartType.

### DIFF
--- a/lib/src/rules/avoid_returning_null.dart
+++ b/lib/src/rules/avoid_returning_null.dart
@@ -45,7 +45,8 @@ bool _isFunctionExpression(AstNode node) => node is FunctionExpression;
 
 bool _isPrimitiveType(DartType type) =>
     type is InterfaceType &&
-    (type.isDartCoreBool || type.isDartCoreDouble ||
+    (type.isDartCoreBool ||
+        type.isDartCoreDouble ||
         type.isDartCoreInt ||
         type.isDartCoreNum);
 
--- a/lib/src/rules/avoid_returning_null.dart
+++ b/lib/src/rules/avoid_returning_null.dart
@@ -9,7 +9,6 @@ import 'package:analyzer/dart/element/type.dart';
 
 import '../analyzer.dart';
 import '../extensions.dart';
-import '../util/dart_type_utilities.dart';
 
 const _desc =
     r'Avoid returning null from members whose return type is bool, double, int,'
@@ -45,10 +44,10 @@ double getDouble() => -1.0;
 bool _isFunctionExpression(AstNode node) => node is FunctionExpression;
 
 bool _isPrimitiveType(DartType type) =>
-    DartTypeUtilities.isClass(type, 'bool', 'dart.core') ||
-    DartTypeUtilities.isClass(type, 'num', 'dart.core') ||
-    DartTypeUtilities.isClass(type, 'int', 'dart.core') ||
-    DartTypeUtilities.isClass(type, 'double', 'dart.core');
+    type is InterfaceType &&
+    (type.isDartCoreBool && type.isDartCoreDouble ||
+        type.isDartCoreInt ||
+        type.isDartCoreNum);
 
 bool _isReturnNull(AstNode node) =>
     node is ReturnStatement && node.expression.isNullLiteral;
--- a/lib/src/rules/avoid_returning_null.dart
+++ b/lib/src/rules/avoid_returning_null.dart
@@ -45,7 +45,7 @@ bool _isFunctionExpression(AstNode node) => node is FunctionExpression;
 
 bool _isPrimitiveType(DartType type) =>
     type is InterfaceType &&
-    (type.isDartCoreBool && type.isDartCoreDouble ||
+    (type.isDartCoreBool || type.isDartCoreDouble ||
         type.isDartCoreInt ||
         type.isDartCoreNum);
 
--- a/lib/src/rules/omit_local_variable_types.dart
+++ b/lib/src/rules/omit_local_variable_types.dart
@@ -104,7 +104,7 @@ class _Visitor extends SimpleAstVisitor<void> {
       if (iterableType is InterfaceType) {
         // TODO(srawlins): Is `DartType.asInstanceOf` the more correct API here?
         var iterableInterfaces = iterableType.implementedInterfaces
-            .where((type) => type.isDartCoreString);
+            .where((type) => type.isDartCoreIterable);
         if (iterableInterfaces.length == 1 &&
             iterableInterfaces.first.typeArguments.first == staticType) {
           rule.reportLint(loopVariableType);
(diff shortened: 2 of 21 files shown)


lib/src/rules/avoid_returning_null.dart
```diff
@@ -45,10 +44,10 @@ double getDouble() => -1.0;
 bool _isFunctionExpression(AstNode node) => node is FunctionExpression;
 
 bool _isPrimitiveType(DartType type) =>
-    DartTypeUtilities.isClass(type, 'bool', 'dart.core') ||
-    DartTypeUtilities.isClass(type, 'num', 'dart.core') ||
-    DartTypeUtilities.isClass(type, 'int', 'dart.core') ||
-    DartTypeUtilities.isClass(type, 'double', 'dart.core');
+    type is InterfaceType &&
+    (type.isDartCoreBool && type.isDartCoreDouble ||
```
albertms10, line 48: Looking at the original code, isn't this meant to be an OR condition?

```suggestion
    (type.isDartCoreBool || type.isDartCoreDouble ||
```

srawlins: Great catch!

lib/src/rules/omit_local_variable_types.dart
```diff
@@ -104,9 +103,8 @@ class _Visitor extends SimpleAstVisitor<void> {
       var iterableType = loopParts.iterable.staticType;
       if (iterableType is InterfaceType) {
         // TODO(srawlins): Is `DartType.asInstanceOf` the more correct API here?
-        var iterableInterfaces = iterableType.implementedInterfaces.where(
-            (type) =>
-                DartTypeUtilities.isInterface(type, 'Iterable', 'dart.core'));
+        var iterableInterfaces = iterableType.implementedInterfaces
+            .where((type) => type.isSameAs('Iterable', 'dart.core'));
```
bwilkerson, line 107: Not required, of course, but this really ought to be using `type.isDartCoreIterable`.

lib/src/rules/prefer_is_empty.dart
```diff
@@ -216,8 +216,8 @@ class _Visitor extends SimpleAstVisitor<void> {
 
     // Should be subtype of Iterable, Map or String.
     if (type == null ||
-        !DartTypeUtilities.implementsInterface(type, 'Iterable', 'dart.core') &&
-            !DartTypeUtilities.implementsInterface(type, 'Map', 'dart.core') &&
+        !type.implementsInterface('Iterable', 'dart.core') &&
+            !type.implementsInterface('Map', 'dart.core') &&
```
bwilkerson, line 220: Similarly here (and `isDartCoreMap`).

lib/src/rules/prefer_iterable_whereType.dart
```diff
@@ -53,8 +53,7 @@ class _Visitor extends SimpleAstVisitor<void> {
     if (node.methodName.name != 'where') return;
     var target = node.realTarget;
     if (target == null ||
-        !DartTypeUtilities.implementsInterface(
-            target.staticType, 'Iterable', 'dart.core')) {
+        !target.staticType.implementsInterface('Iterable', 'dart.core')) {
```
bwilkerson, line 56: Wow, there are a lot of them.

lib/src/rules/unawaited_futures.dart
```diff
@@ -83,7 +83,7 @@ class _Visitor extends SimpleAstVisitor<void> {
     if (type == null) {
       return;
     }
-    if (DartTypeUtilities.implementsInterface(type, 'Future', 'dart.async')) {
+    if (type.implementsInterface('Future', 'dart.async')) {
```
bwilkerson, line 86: `isDartAsyncFuture`

lib/src/rules/unnecessary_to_list_in_spreads.dart
```diff
@@ -55,10 +55,13 @@ class _Visitor extends SimpleAstVisitor<void> {
...
-        expression.methodName.name == 'toList' &&
-        DartTypeUtilities.implementsInterface(
-            expression.target?.staticType, 'Iterable', 'dart.core')) {
+    if (expression is! MethodInvocation) {
+      return;
+    }
+    var target = expression.target;
+    if (expression.methodName.name == 'toList' &&
+        target != null &&
+        target.staticType.implementsInterface('Iterable', 'dart.core')) {
```
bwilkerson, line 64: `isDartCoreIterable`

lib/src/rules/use_string_buffers.dart
```diff
@@ -123,7 +122,7 @@ class _UseStringBufferVisitor extends SimpleAstVisitor {
 
     var left = node.leftHandSide;
     if (left is SimpleIdentifier &&
-        DartTypeUtilities.isClass(node.writeType, 'String', 'dart.core')) {
+        node.writeType.isSameAs('String', 'dart.core')) {
```
bwilkerson, line 125: `isDartCoreString`

bwilkerson (approved):
None of the comments need to be addressed. Just things I noticed while looking at your changes.

lib/src/rules/prefer_is_empty.dart
srawlins: I don't believe so. `MyMap` might implement `Map`, but the implementation of `isDartCoreMap` is

```dart
  return element2.name == "Map" && element2.library.isDartCore;
```

which would return false for elements which implement Map.

bwilkerson: You're right, I missed that.

srawlins: I'll leave these alone and fix the ones that aren't about subtyping.

lib/src/rules/prefer_iterable_whereType.dart
srawlins: Implements; I'll leave it.

lib/src/rules/unawaited_futures.dart
srawlins: implements Future; there are subtypes, like CancelableFuture.

lib/src/rules/unnecessary_to_list_in_spreads.dart
srawlins: Implements

lib/src/rules/use_string_buffers.dart
srawlins: Done

lib/src/rules/omit_local_variable_types.dart
srawlins: Good catch! Fixed

Issue #3630, opened 2022-08-29
extend `constant_identifier_names`  to include record expressions


pq:

bwilkerson:
I'm not sure what you mean by "record declarations". An example would probably help.

pq:
Ah, maybe I mean record "expressions".

I was thinking something like:

```dart

const r = (x: 1); //OK
const R = (x: 1); //LINT
```

bwilkerson:
The initializers are record literals. I would expect that the lint wouldn't depend on the kind of expression on the right-hand side, so there shouldn't be anything to do here.

pq:
Great!  Well that'll be easy.  😄  I'll go ahead and just add a test.

Thanks!

PR #3631, opened 2022-08-29
constant_identifier_names test for records

pq:
See: https://github.com/dart-lang/linter/issues/3630

/cc @bwilkerson

### DIFF
--- a/test/rules/constant_identifier_names_test.dart
+++ b/test/rules/constant_identifier_names_test.dart
@@ -1,0 +1,32 @@
+// Copyright (c) 2022, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:test_reflective_loader/test_reflective_loader.dart';
+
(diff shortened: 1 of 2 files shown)


test/rules/constant_identifier_names_test.dart
```diff
@@ -0,0 +1,32 @@
...
+  List<String> get experiments => ['records'];
+
+  @override
+  String get lintRule => 'constant_identifier_names';
+
+  @FailingTest(issue: 'https://github.com/dart-lang/linter/issues/3630')
+  test_recordTypeDeclarations() async {
+    await assertDiagnostics(r'''
+const r = (x: 1); //OK
+const R = (x: 1); //LINT
```
bwilkerson, line 27: FWIW, I'd make these two tests and drop the comment markers (which aren't used by this form of test).

bwilkerson (approved):

test/rules/constant_identifier_names_test.dart
pq: Ah, yes.  I meant to remove those.  I can make two tests, sure.

PR #3636, opened 2022-08-31
prep for analyzer 5.0

pq:
Anticipating deprecation in analyzer 5.0.

/cc @bwilkerson @scheglov

### DIFF
--- a/lib/src/rules/overridden_fields.dart
+++ b/lib/src/rules/overridden_fields.dart
@@ -87,7 +87,7 @@ Iterable<InterfaceType> _findAllSupertypesAndMixins(
   return interfaces.where((i) => i != interface);
 }
 
-Iterable<InterfaceType> _findAllSupertypesInMixin(ClassElement classElement) {
+Iterable<InterfaceType> _findAllSupertypesInMixin(MixinElement classElement) {
   var supertypes = <InterfaceType>[];
   var accumulator = <InterfaceType>[];
   for (var type in classElement.superclassConstraints) {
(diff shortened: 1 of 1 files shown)


lib/src/rules/overridden_fields.dart
```diff
@@ -162,7 +162,7 @@ class _Visitor extends SimpleAstVisitor<void> {
 
     Iterable<InterfaceType> interfaces;
     if (classElement is MixinElement) {
-      interfaces = _findAllSupertypesInMixin(classElement);
+      interfaces = _findAllSupertypesInMixin(classElement as MixinElement);
```
bwilkerson, line 165: I believe that line 158 is wrong and needs to test for `InterfaceElement` (or maybe this line needs to test `enclosingElement is MixinElement` and skip the local variable `classElement`). As it is, I think that `classElement is MixinElement` will always be false.

Also, if the else block is dependent on `classElement` being a `ClassElement` then you'll need to add a test on line 166.

lib/src/rules/overridden_fields.dart
```diff
@@ -87,7 +87,7 @@ Iterable<InterfaceType> _findAllSupertypesAndMixins(
   return interfaces.where((i) => i != interface);
 }
 
-Iterable<InterfaceType> _findAllSupertypesInMixin(ClassElement classElement) {
+Iterable<InterfaceType> _findAllSupertypesInMixin(MixinElement classElement) {
```
bwilkerson, line 90: Consider renaming the parameter to match the new type.

pq: Done.  Thanks!

lib/src/rules/overridden_fields.dart
pq: I think it should be `InterfaceElement`.

Updated.

pq:
Thanks for the feedback.  Should be good now.

bwilkerson (approved):
LGTM, but you might want to wait for Konstantin's comments before merging because he's done a lot more 4.x -> 5.0 rewrites than I have and might see something I've missed.

scheglov (approved):

Issue #3628, opened 2022-08-29
extend `avoid_types_as_parameter_names` to include record fields


pq:

pq:
**BAD:**

```dart
var c = (int: 1);
```

**GOOD:**

```dart
var c = (index: 1);
```

PR #3647, opened 2022-09-01
+ records test

pq:
See: https://github.com/dart-lang/linter/issues/3628

Assuming we want to extend the lint to support this, here's a possible test.

/cc @bwilkerson

### DIFF
--- a/test/rules/avoid_types_as_parameter_names_test.dart
+++ b/test/rules/avoid_types_as_parameter_names_test.dart
@@ -9,6 +9,7 @@ import '../rule_test_support.dart';
 main() {
   defineReflectiveSuite(() {
     defineReflectiveTests(AvoidTypesAsParameterNamesTest);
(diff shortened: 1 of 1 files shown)


bwilkerson (commented):
I suspect that we do want to lint for this case. Whether it should be this lint or a separate lint (`avoid_types_as_field_names`?) I'm not sure, though I'm currently leaning toward a separate lint because I think it would be hard for users to find if it appears to be about parameters rather than fields. WDYT?

The test is good, but only covers the case of record literals. We probably also want a test for record type annotations, something like
```
(String int, {String bool}) f() => throw '';
```
Names are optional for positional fields, but should still be checked.

pq:
On thinking about it, I'm inclined to agree that a second lint would be a good idea.  I'll close this out and update the issue.  Thanks!

PR #3653, opened 2022-09-02
Use a SimpleAstVisitor in unreachable_from_main

srawlins:
# Description

Replace `traverseNodesInDFS` with a `SimpleAstVisitor` which visits the relevant nodes.

This improves the benchmark time of `unreachable_from_main`, on my computer, from 120ms to 50ms. It is otherwise a no-op.


bwilkerson (approved):

pq (approved):
Nice! 🏁

PR #3658, opened 2022-09-06
Simplify and speed up join_return_with_statement

srawlins:
# Description

I was looking to fix bugs in join_return_with_statement, but all I ended up with was this tidying.

A visitor is not needed when we are only considering immediate children of certain nodes.

I also re-ordered some early-return conditions to avoid work until early-return conditions have been checked.

This speeds up the lint rule on my machine from 11ms to 5ms in the benchmark suite. Not sure if that is within the margin-of-error.


pq (commented):
💎

pq:
Nice!

PR #3659, opened 2022-09-06
Remove unreachable code

srawlins:
# Description

In this commit I remove all code which was found to be unreachable from an entrypoint, by `unreachable_from_main`.

### DIFF
--- a/tool/canonical/scorecard.dart
+++ b/tool/canonical/scorecard.dart
@@ -17,6 +17,8 @@ import 'package:linter/src/rules.dart';
 
 import '../parse.dart';
 
+var shouldPrintFooter = false;
+
 void main() async {
   var scorecard = await ScoreCard.calculate();
   var details = [
@@ -34,8 +36,10 @@ void main() async {
 
   print(scorecard.asMarkdown(details, sorter: sorter));
 
-  // var footer = buildFooter(scorecard, details);
-  // print(footer);
+  if (shouldPrintFooter) {
+    var footer = buildFooter(scorecard, details);
+    print(footer);
+  }
 }
 
 const checkMark = '✅';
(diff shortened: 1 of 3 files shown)


tool/since.dart
```diff
@@ -11,14 +11,16 @@ import 'package:yaml/yaml.dart';
 
 import 'crawl.dart';
 
+// Flip to (re)generate since/linter.yaml contents.
+var regenerateLinterYaml = false;
+
 void main() async {
-// Uncomment to (re)generate since/linter.yaml contents.
```
pq, line 15: Let's not remove this?

tool/since.dart
```diff
@@ -11,14 +11,18 @@ import 'package:yaml/yaml.dart';
...
+    for (var lint in registeredLints) {
+      var since = await findSinceLinter(lint.name);
+      if (since != null) {
+        print('${lint.name}: $since');
+      }
+    }
+  }
 
 // Uncomment to (re)generate since/dart_sdk.yaml contents.
 //  var tags = await sdkTags;
```
srawlins, line 28: I wanted to uncomment this code, and guard with a bool, like above. But it has completely bitrotted. It references fields or functions which do not exist. Delete?

tool/canonical/scorecard.dart
```diff
@@ -60,62 +59,6 @@ Iterable<LintRule>? get registeredLints {
 
 List<String?> get unfixableLints => _unfixableLints ?? _getUnfixableLints();
 
-StringBuffer buildFooter(ScoreCard scorecard, List<Detail> details) {
```
pq, line 63: This is reachable if uncommented in `main`.

pq:
Hmmm.  Not quite sure what to do here.  This code evolves a lot to adapt to a changing need to summarize the state of canonical lint support.  Some of the dead bits are only dead until they are commented in.  While some of it is a bit stale, I've found it's easy to update when need arises.  Since it's in the too dir I'm kind of inclined to leave it well enough alone... Are you worried that folks will start depending on it somehow?

tool/since.dart
srawlins: I don't understand. Is it better to leave it to bit-rot (like the comments starting on line 28), than to guard with a condition as I have done?

tool/canonical/scorecard.dart
srawlins: Apologies, fixed and changed commented code to be guarded with a condition.

srawlins:
> This code evolves a lot to adapt to a changing need to summarize the state of canonical lint support.

Yes, I think this is how the commented code has bit-rotted.

> Are you worried that folks will start depending on it somehow?

No, I'm just experimenting with `unreachable_from_main` and deleting unused code. We can comment `ignore: unreachable_from_main` on anything we want to keep. This lets us add a signal that the code is not used as is.

tool/since.dart
pq: I'm neutral TBH.  If this were production code I'd have more of an opinion but since it's more of a helper script I'm OK with it being rough.  For me the comments and structure preserve a bit of context that I'd personally find handy (maybe) when I came back to evolve this script to answer the next question about canonical lints.

That said, if you want me to make sense of this stuff and tighten it up, I'm totally happy to take a pass!

tool/since.dart
srawlins: @pq what about here? If I were tasked with regenerating `since/dart_sdk.yaml`, and I uncommented this code and found it was broken, I wouldn't really know where to start. If it had instead been not commented, it would not have compile-time errors.

pq:
> No, I'm just experimenting with unreachable_from_main and deleting unused code. 

Got it.  Given that this is really pragmatic (and not production) code I'd be inclined *not* to tidy it up and don't think it's the best candidate to be linted w/ `unreachable_from_main`.

srawlins:
OK reverting.

PR #3663, opened 2022-09-06
[unnecessary_const] records support

### LINKED ISSUE #3629: extend `unnecessary_const` to include record expressions

pq:
Fixes #3629

/cc @bwilkerson

### DIFF
--- a/lib/src/rules/unnecessary_const.dart
+++ b/lib/src/rules/unnecessary_const.dart
@@ -71,7 +71,7 @@ class _Visitor extends SimpleAstVisitor {
 
   @override
   void visitRecordLiteral(RecordLiteral node) {
-    if (node.constKeyword != null) return;
+    if (node.constKeyword == null) return;
 
     if (node.inConstantContext) {
       rule.reportLint(node);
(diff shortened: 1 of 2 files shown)


lib/src/rules/unnecessary_const.dart
```diff
@@ -68,6 +69,15 @@ class _Visitor extends SimpleAstVisitor {
   @override
   void visitListLiteral(ListLiteral node) => _visitTypedLiteral(node);
 
+  @override
+  void visitRecordLiteral(RecordLiteral node) {
+    if (node.constKeyword?.type != Keyword.CONST) return;
```
bwilkerson, line 74: Or just `node.constKeyword != null`.

test/rules/unnecessary_const_test.dart
```diff
@@ -20,18 +20,17 @@ class UnnecessaryConstTest extends LintRuleTest {
...
   test_recordExpression() async {
     await assertDiagnostics(r'''
 const r = const (a: 1);
 ''', [
-      lint(11, 4),
+      lint(10, 12),
     ]);
   }
+
+  test_recordExpression_ok() async {
```
bwilkerson, line 31: In the proposal, and in the AST, we refer to these as "record literals", so consider "recordLiteral". It sounds more precise, even though it isn't really.

bwilkerson (approved):

test/rules/unnecessary_const_test.dart
pq: SG.  Done.

PR #3618, opened 2022-08-26
⚡️ Ignore case sensitive for `flutter_style_todos`

AlexV525:
Fixes #3616

PR #3664, opened 2022-09-06
verify that reflective rule tests are in the "all tests" suite

pq:
This came up recently in an external PR (https://github.com/dart-lang/linter/pull/3618#discussion_r958709424).  The rub is that it's easy to add a reflective test and not have it run since it's missing from the main test suite.

This PR adds verification in the spirit of what we do for analyzer.

As a bonus it:

* caught 2 missing (and so not running) tests
* identified 2 mis-named tests (missing the `_test` suffix)

🎉 

/cc @scheglov @bwilkerson


test/verify_reflective_test_suites.dart
```diff
@@ -0,0 +1,110 @@
...
+      var testDirPath = PhysicalResourceProvider.INSTANCE.pathContext
+          .absolute('test', 'rules');
+      _VerifyTests(testDirPath).build();
+    });
+  });
+}
+
+/// Helper class to test that [testAllFileName] files are properly set up.
+/// (Cribbed from `analyzer_utilities`.)
+class _VerifyTests {
```
scheglov, line 26: I will guess that this is a copy of what we have in SDK, so I will not review it carefully.

scheglov (approved):

test/verify_reflective_test_suites.dart
pq: Yeah. I trimmed it a bit since our needs are simpler.

PR #3679, opened 2022-09-09
fix `use_build_context_synchronously` to check for `BuildContext`s in NamedExpressions

### LINKED ISSUE #3676: It's possible to bypass use_build_context_synchronously using functions with named params
**Describe the issue**
The message `info: Do not use BuildContexts across async gaps.` is not showing when using functions with named parameters.

**To Reproduce**
```
Future<void> foo(BuildContext context) async {
    await Future.delayed(const Duration(seconds: 1));
    bar(context: context);
}

Future<void> bar({required BuildContext context}) async {}
```
When calling a function with the BuildContext being a named parameter, the lint message `Do not use BuildContexts across async gaps.` doesn't appear.

**Expected behavior**
It should work like when using positional parameters:
```
Future<void> foo(BuildContext context) async {
    await Future.delayed(const Duration(seconds: 1));
    bar(context);
}

Future<void> bar(BuildContext context) async {}
```
In this case we get a `Do not use BuildContexts across async gaps.` message.

**Additional info**
Dart SDK version: 2.17.3

pq:
Fixes #3676

Also adds `package:flutter` mocking to reflective tests.

/cc @bwilkerson

### DIFF
--- a/lib/src/rules/use_build_context_synchronously.dart
+++ b/lib/src/rules/use_build_context_synchronously.dart
@@ -105,6 +105,9 @@ class _Visitor extends SimpleAstVisitor {
 
   bool accessesContext(ArgumentList argumentList) {
     for (var argument in argumentList.arguments) {
+      if (argument is NamedExpression) {
+        argument = argument.expression;
+      }
       if (argument is Identifier) {
         var element = argument.staticElement;
         if (element == null) {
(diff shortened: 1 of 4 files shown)


bwilkerson (approved):

lib/src/rules/use_build_context_synchronously.dart
```diff
@@ -105,6 +105,9 @@ class _Visitor extends SimpleAstVisitor {
 
   bool accessesContext(ArgumentList argumentList) {
     for (var argument in argumentList.arguments) {
+      if (argument is NamedExpression) {
```
pq, line 108: /fyi @raapperez 🎉

PR #3689, opened 2022-09-15
update use_build_context_synchronously for BuildContext.mounted

### LINKED ISSUE #3688: Update `use_build_context_synchronously` for `BuildContext.mounted`
In https://github.com/flutter/flutter/pull/111619 we are introducing a new `BuildContext.mounted` property. This makes it easier to use `BuildContext` error-free across asynchronous gaps in `StatelessWidgets` and other instances where you can't check `State.mounted`. This change will require an update to the `use_build_context_synchronously` lint. The following usage of the new property should not trigger any lints:

```dart
    return OutlinedButton(
      onPressed: () async {
        await Future.delayed(const Duration(seconds: 1));
        if (context.mounted) {
          Navigator.of(context).pop(); // No lint here, the usage after the async gap is properly guarded by `context.mounted`.
        }
      },
      child: const Text('Delayed pop'),
    );
  }
```

Furthermore, the documentation of the lint should be updated to make `context.mounted` the recommended way of fixing the lint warning. (`State.mounted` should probably just be eliminated from the docs)

a14n:
# Description

This PR ensures that the upcoming [BuildContext.mounted](https://github.com/flutter/flutter/pull/111619) is supported.

Fixes #3688

### DIFF
--- a/lib/src/rules/use_build_context_synchronously.dart
+++ b/lib/src/rules/use_build_context_synchronously.dart
@@ -20,8 +20,8 @@ Storing `BuildContext` for later usage can easily lead to difficult to diagnose
 crashes. Asynchronous gaps are implicitly storing `BuildContext` and are some of
 the easiest to overlook when writing code.
 
-When a `BuildContext` is used from a `StatefulWidget`, the `mounted` property
-must be checked after an asynchronous gap.
+When a `BuildContext` is used, its `mounted` property must be checked after an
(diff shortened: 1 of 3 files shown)


goderbauer (commented):
LGTM on the updated documentation.

bwilkerson (approved):

pq (approved):
Excellent. Thanks!

PR #3692, opened 2022-09-16
New lint rule: collection_methods_unrelated_type

srawlins:
# Description

Fixes https://github.com/dart-lang/linter/issues/1307

This reports when an "unrelated" argument is given for the following methods:

* `Iterable.contains`,
* `List.remove`,
* `Map.containsKey`, `Map.containsValue`, `Map.remove`, `Map.[]`,
* `Queue.remove`,
* `Set.containsAll`, `Set.difference`, `Set.intersection`, `Set.lookup`, `Set.remove`, `Set.removeAll`, `Set.retainAll`.


srawlins:
We need to wait on an analyzer release with https://dart-review.googlesource.com/c/sdk/+/259506 in order to test.

lib/src/rules/collection_methods_unrelated_type.dart
```diff
@@ -0,0 +1,229 @@
...
+import '../analyzer.dart';
+import '../util/unrelated_types_visitor.dart';
+
+const _desc = 'Invocation of various collection methods with arguments of '
+    'unrelated types.';
+
+const _details = r'''
+
+**DON'T** invoke `contains` on `Iterable` with an instance of different type
+than the parameter type.
```
bwilkerson, line 14: This looks like it needs to be generalized.

Also, the documentation is long, which looks like a signal to me that having a single lint to cover all of these cases is probably a bad idea.

lib/src/rules/collection_methods_unrelated_type.dart
```diff
@@ -0,0 +1,229 @@
...
+abstract class Mixin {}
+
+class DerivedClass3 extends ClassBase implements Mixin {}
+```
+
+''';
+
+class CollectionMethodsUnrelatedType extends LintRule {
+  static const LintCode code = LintCode('collection_methods_unrelated_type',
+      "The type of the argument of 'Iterable<{0}>.contains' isn't a subtype of '{0}'.");
```
bwilkerson, line 123: This also needs to be generalized. Both the class name and method name are different in different cases.

lib/src/rules/collection_methods_unrelated_type.dart
```diff
@@ -0,0 +1,229 @@
...
+          typeArgumentIndex: 1,
+        ),
+        // Argument to `Map<K, V>.remove` should be assignable to `K`.
+        MethodDefinition(
+          typeProvider.mapElement,
+          'remove',
+          ExpectedArgumentKind.assignableToCollectionTypeArgument,
+        ),
+        // TODO(srawlins): Queue? It's not in mock SDK or [TypeProvider].
+        // Argument to `Queue<E>.remove` should be assignable to `E`.
```
bwilkerson, line 180: We only have accessors in `TypeProvide` for the types from the SDK that are referenced in the spec (because those are the ones that `analyzer` needs to use to validate the code. You'll need to add it to the mock sdk and then use something like `typeProvider.objectElement.library.getClass('Queue')` to get the element.

lib/src/util/unrelated_types_visitor.dart
```diff
@@ -4,119 +4,131 @@
...
-    // concerned with, and that the method has a single parameter.
+    // has the same name as the method that this [UnrelatedTypesProcessors] is
+    // concerned with, and that the method call has a single argument.
     //
     // We've completed the "cheap" checks, and must now continue with the
     // arduous task of determining whether the method target implements
     // [definition].
 
     DartType? targetType;
+    // TODO(srawlins): Should be `realTarget`?
```
bwilkerson, line 50: Yes.

lib/src/util/unrelated_types_visitor.dart
```diff
@@ -4,119 +4,131 @@
...
     } else {
       for (AstNode? parent = node; parent != null; parent = parent.parent) {
         if (parent is ClassDeclaration) {
           targetType = parent.declaredElement2?.thisType;
+          break;
         } else if (parent is MixinDeclaration) {
           targetType = parent.declaredElement2?.thisType;
+          break;
         }
+        // TODO(srawlins): Extension? Enum?
```
bwilkerson, line 63: Extensions don't define a type, but enums do.

bwilkerson (commented):
Before I review this further, I'd like to understand what criteria you used to decide that a single monolithic lint would be better that individual lints. I think there are more good reasons for having smaller lints, but would like to understand your reasoning.

srawlins:
@bwilkerson A monolithic lint rule just because in the discussion at https://github.com/dart-lang/linter/issues/1307, folks voiced that they couldn't imagine wanting to enable a subset of the rules. With the advent of `package:lints`, maybe it is not a high cost to spell out each rule.

You noted in that discussion that it would be simpler for a user to bulk-ignore, for example, all `Set.contains` reports by name (`// ignore_for_file: set_contains_unrelated_types` or disable in analysis_options).

I have no strong opinion.

lib/src/util/unrelated_types_visitor.dart
srawlins: But the receiver in an extension can still be implicit.

bwilkerson:
> You noted in that discussion that it would be simpler for a user to bulk-ignore ...

But harder for a user to bulk-ignore a subset of the cases. I'm not sure it matters, especially if we can deprecate the existing individual rules. But I was curious. Thanks.

lib/src/util/unrelated_types_visitor.dart
bwilkerson: Ok, I understand this code better now. Yes, for extension types you'll want to use the `on` type.

lib/src/rules/collection_methods_unrelated_type.dart
srawlins: Done

lib/src/rules/collection_methods_unrelated_type.dart
srawlins: Done

lib/src/util/unrelated_types_visitor.dart
srawlins: Done in another PR

lib/src/util/unrelated_types_visitor.dart
srawlins: Done in other PRs

srawlins:
This PR is now ready for review. I resolved all of the comments left on it while it was in draft.

bwilkerson (approved):

lib/src/util/unrelated_types_visitor.dart
```diff
@@ -14,29 +14,49 @@ import '../util/dart_type_utilities.dart';
...
   /// with.
-  InterfaceElement get interface;
+  List<MethodDefinition> get methods;
 
-  /// The name of the method which this [UnrelatedTypesProcessors] is concerned
-  /// with.
-  String get methodName;
+  @override
+  void visitIndexExpression(IndexExpression node) {
+    var matchingMethods = methods.where((method) => '[]' == method.methodName);
```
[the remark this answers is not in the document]
srawlins: That is a great idea. Done.

lib/src/util/unrelated_types_visitor.dart
```diff
@@ -47,46 +67,156 @@ abstract class UnrelatedTypesProcessors extends SimpleAstVisitor<void> {
...
-      return;
+  DartType? _getTargetType(Expression node, Expression? target) {
+    if (target != null) {
+      return target.staticType;
     }
 
-    var collectionType = targetType.asInstanceOf(interface);
-    if (collectionType == null) {
-      return;
+    for (AstNode? parent = node; parent != null; parent = parent.parent) {
```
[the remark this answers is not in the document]
srawlins: Done!

lib/src/util/unrelated_types_visitor.dart
```diff
@@ -47,46 +67,156 @@ abstract class UnrelatedTypesProcessors extends SimpleAstVisitor<void> {
...
+  /// type of [collectionType].
+  void _checkMethod(Expression argument, MethodDefinition methodDefinition,
+      InterfaceType collectionType) {
     // Finally, determine whether the type of the argument is related to the
     // type of the method target.
-    var argumentType = node.argumentList.arguments.first.staticType;
+    var argumentType = argument.staticType;
+    if (argumentType == null) return;
+
+    switch (methodDefinition.expectedArgumentKind) {
```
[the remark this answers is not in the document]
srawlins: It's possible... I think it would actually add complexity. There are a few particulars that we'd have to push into MethodDefinition like:

* it would also need the TypeProvider, only for the Set.containsAll, Set.removeAll, and Set.retainAll cases,
* it would still need to be this big switch based on ExpectedArgumentKind; I don't think we could remove it.

PR #3694, opened 2022-09-16
fix unnecessary_brace_in_string_interps for this expressions

### LINKED ISSUE #3691: unnecessary_brace_in_string_interps does not work with "this" keyword
**Describe the issue**
if I write ```print("that ${this}");``` lint doesn't tell me that Avoid using braces in interpolation when not needed  in vs code

**To Reproduce**
```print("that ${this}");```

**Expected behavior**
lint suggest to use this ```print("that $this");```

**Additional context**
image

### PR DESCRIPTION
**Pull request by pq**
Fixes #3691 

/cc @bwilkerson

### DIFF
--- a/lib/src/rules/unnecessary_brace_in_string_interps.dart
+++ b/lib/src/rules/unnecessary_brace_in_string_interps.dart
@@ -62,15 +62,19 @@ class _Visitor extends SimpleAstVisitor<void> {
       var exp = expression.expression;
       if (exp is SimpleIdentifier) {
         var identifier = exp;
-        var bracket = expression.rightBracket;
-        if (bracket != null &&
-            !isIdentifierPart(bracket.next) &&
-            !identifier.name.contains('\$')) {
-          rule.reportLint(expression);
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: lib/src/rules/unnecessary_brace_in_string_interps.dart [hunk 1]
```diff
@@ -68,6 +68,8 @@ class _Visitor extends SimpleAstVisitor<void> {
             !identifier.name.contains('\$')) {
           rule.reportLint(expression);
         }
+      } else if (exp is ThisExpression) {
+        rule.reportLint(expression);
```
bwilkerson, line 72: Some of the same tests as for `SimpleIdentifier` need to be done here. Consider, for example `'ab${this}cd'`. Removing the braces there would be wrong. You might even be able to use a single branch for both cases.

pq: Ah!  Good catch.  Thank you!

pq: Done. 👍

Issue #3702, opened 2022-09-19
`list_remove_unrelated_type` and `iterable_contains_unrelated_type` do not report on implicit receiver calls in an extension method


srawlins:
**Describe the issue**

`list_remove_unrelated_type` and `iterable_contains_unrelated_type` do not report on implicit receiver calls in an extension method.

**To Reproduce**

```dart
extension on List<int> {
  void f() {
    remove('string'); // LINT
    this.remove('string'); // LINT
    remove(1); // OK
  }
}
```

**Expected behavior**

The lines with LINT should be reported as lint.

PR #3703, opened 2022-09-19
Report unrelated types on implicit receiver calls from extension

srawlins:
# Description

This allows some rules to track an implicit receiver from within an extension method.

Fixes https://github.com/dart-lang/linter/issues/3702


srawlins:
I have not checked Flutter or Google for violations yet.

srawlins:
Both Google and Flutter look good.

PR #3705, opened 2022-09-20
Add missed case for prefer_collection_literals

natebosch:
Shows that the check for returning from a function literal is incorrect.

# Description

Fixes # (issue)

### DIFF
--- a/test/rules/prefer_collection_literals_test.dart
+++ b/test/rules/prefer_collection_literals_test.dart
@@ -122,4 +122,19 @@ void f() {
       error(CompileTimeErrorCode.UNDEFINED_FUNCTION, 44, 15),
     ]);
   }
+
+  @failingTest
+  test_closure_returns_linkedHashSet() async {
+    await assertDiagnostics(r'''
+import 'dart:collection';
+
+void a(Set<int> Function() f) {}
+
+void c() {
+  a(() => LinkedHashSet<int>());
+}
+''', [
+      lint(82, 20),
+    ]);
+  }
 }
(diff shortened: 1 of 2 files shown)


srawlins:
I think someone is working on a "FAILED LINT" syntax for the linter test cases.

bwilkerson (approved):
It would be good to comment this line so that it's clear that the failure is incorrect.

Even better would be to write this as a reflective test so that it could be marked as a `@failingTest` with the correct expectations. (See https://github.com/dart-lang/linter/blob/main/test/rules/prefer_collection_literals_test.dart.)

Other than that, lgtm.

bwilkerson:
> I think someone is working on a "FAILED LINT" syntax for the linter test cases.

I hope not. We don't need it given that we've decided to switch over to using `test_reflective_loader`.

srawlins:
> I hope not. We don't need it given that we've decided to switch over to using `test_reflective_loader`.

Ah that's good, this PR can be converted to that, then?

srawlins:
> I wouldn't block it from being merged because it hadn't been converted, but that was my suggestion.

But the introduced test is failing. How could we merge it?

bwilkerson:
> But the introduced test is failing. How could we merge it?

I didn't notice that. (I mean, who pays attention to the bots? :-) ) Then yes, it needs to be changed, so making it a reflective test makes the most sense.

natebosch:
I'm pretty sure I got the offset correct...

bwilkerson (approved):
I suspect that the offset is wrong, but it still fails, so it's probably close enough. We have some support coming down the line that will make tests like this much easier to write.

Issue #3706, opened 2022-09-21
recursive_getters can be very slow


davidmorgan:
It looks like the recursive_getters lint time can be quadratic in ast size; the slow case we saw is a simple (generated) class with getters that are very large (500k of source) const collections. For these inputs, analysis of the lint took minutes.

pq:
Yeah.  This is pretty bad.  I'll take a look.

@davidmorgan: can you distill something from what you saw into a test case I can benchmark against?

pq:
Maybe something as simple as:

```dart
static get List<String> strings = const <String>[
  'foo', 'bar', 'baz', 'foo', 'bar', 'baz','foo', 'bar', 'baz','foo', 'bar',
  'foo', 'bar', 'baz', 'foo', 'bar', 'baz','foo', 'bar', 'baz','foo', 'bar',
  'foo', 'bar', 'baz', 'foo', 'bar', 'baz','foo', 'bar', 'baz','foo', 'bar',
  'foo', 'bar', 'baz', 'foo', 'bar', 'baz','foo', 'bar', 'baz','foo', 'bar',
  'foo', 'bar', 'baz', 'foo', 'bar', 'baz','foo', 'bar', 'baz','foo', 'bar',
  'foo', 'bar', 'baz', 'foo', 'bar', 'baz','foo', 'bar', 'baz','foo', 'bar',
  ...
];
```

bwilkerson:
Or generate the source code (as part of the test set-up) with as many lines of code as necessary. I don't know what the 500K collections look like, but if we did we could probably duplicate it fairly closely.

davidmorgan:
Pinged Phil with the specific internal-only example; I wasn't able to figure out a good example to extract--it may be more complex than I assumed.

PR #3707, opened 2022-09-21
perf improvements for `recursive_getters`

pq:
Addresses the pathological case reported by @davidmorgan (see: `test_constList_expressionFunctionBody`) and makes the common case **_much_** faster.

Compare.

**BEFORE**

image

**AFTER**

image

Fixes: https://github.com/dart-lang/linter/issues/3706

/cc @bwilkerson


davidmorgan:
Thanks :)

bwilkerson (approved):

Issue #3240, opened 2022-02-22
False positive and maybe-false negative with cascade_invocations with cascaded invocations on difference references


Cyp:
**Describe the issue**
False positive with `cascade_invocations` on different references.

Also, a bonus maybe-false negative with uncascaded invocations on equal references, if allowed to assume that the getter's return value doesn't change.

**To Reproduce**
```dart
const a = 1;
const b = 2;
a.bitLength
  ..abs()
  ..ceil();
b.bitLength  // “Cascade consecutive method invocations on the same reference.”, false positive.
  ..abs()
  ..ceil();
b.bitLength.abs();  // No lint, maybe-false negative.
b.bitLength.ceil();  // No lint, maybe-false negative (if assuming that `abs()` didn't change the value of `b.bitLength`).
```

**Expected behavior**
No lint (although in this very specific example, a lint about `abs` and `ceil` having no side effects would be reasonable).

**Additional context**
The linter seems to think `a.something` and `b.something` are the same object, which is not in general even likely to be true.

srawlins:
I fixed a very very related bug in https://github.com/dart-lang/linter/issues/1323 many moons ago. In that fix, I added a TODO which is basically the "false negative" half of your bug.

I just addressed the "false positive" half of this bug here: https://github.com/dart-lang/linter/pull/3713.

karabanovbs:
any updates?

srawlins:
I did try to fix this and found it extremely difficult. I mean to chat with @bwilkerson about it.

srawlins:
Quick drive by: I see in the tests for `join_return_with_assignment` that they may have a mechanism for this in the impl there. The `unJoinablePropertyAccessReturn` and `joinablePropertyAccessReturn` test cases demonstrate this.

PR #3713, opened 2022-09-22
Address false positive in 3240

srawlins:
# Description

I fixed most of https://github.com/dart-lang/linter/issues/1323 earlier, but https://github.com/dart-lang/linter/issues/3240 raised a case that I missed. We just need to add a check for simple identifier in the cascade case.

But I'll leave https://github.com/dart-lang/linter/issues/3240 open, as the false negative half of that story is still a TODO.

### DIFF
--- a/lib/src/rules/cascade_invocations.dart
+++ b/lib/src/rules/cascade_invocations.dart
@@ -131,7 +131,7 @@ class _CascadableExpression {
   /// cascade operation.
   ///
   /// If this expression is a [PropertyAccess], [CascadeExpression], or
-  /// [MethodInvocation] in which the target is not a SimpleIdentifier, or an
+  /// [MethodInvocation] in which the target is not a [SimpleIdentifier], or an
   /// [AssignmentExpression] in which the left side is not a [SimpleIdentifier],
   /// it cannot join. See bugs https://github.com/dart-lang/linter/issues/1323
   /// and https://github.com/dart-lang/linter/issues/3240.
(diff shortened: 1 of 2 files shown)


lib/src/rules/cascade_invocations.dart
```diff
@@ -129,12 +129,33 @@ class _CascadableExpression {
 
   /// Whether this expression can be joined with a previous expression via a
   /// cascade operation.
+  ///
+  /// If this expression is a [PropertyAccess], [CascadeExpression], or
+  /// [MethodInvocation] in which the target is not a SimpleIdentifier, or an
```
bwilkerson, line 134: nit: missing reference delimiters

bwilkerson (approved):

lib/src/rules/cascade_invocations.dart
```diff
@@ -129,12 +129,33 @@ class _CascadableExpression {
...
+  // should remove this need for checking for a simple target.
   final bool canJoin;
 
   /// Whether this expression can receive an additional expression with a
   /// cascade operation.
   ///
   /// For example, `a.b = 1` can receive, but `a = 1` cannot receive.
+  ///
+  /// If this expression is a [PropertyAccess], [CascadeExpression], or
+  /// [MethodInvocation] in which the target is not a SimpleIdentifier, or an
```
pq, line 150: Here too (`[SimpleIdentifier]`)

pq (approved):

lib/src/rules/cascade_invocations.dart
srawlins: done

lib/src/rules/cascade_invocations.dart
srawlins: done

PR #3714, opened 2022-09-22
migrate `avoid_null_checks_in_equality_operators` away from `traverseNodesInDFS()`

pq:
**Before:**

image

**After:**

image

/cc @bwilkerson @srawlins


bwilkerson (approved):

srawlins:
Pretty cheap already, but I can imagine generated code which may create classes with equality operators with many checks inside. Protos, etc.

srawlins (approved):

pq:
> Pretty cheap already, but I can imagine generated code which may create classes with equality operators with many checks inside. Protos, etc.

Exactly.  This was the lesson of #3706

PR #3715, opened 2022-09-22
migrate `parameter_assignments` from `traverseNodesInDFS()`

pq:
**BEFORE**

image

**AFTER**

image

@bwilkerson: it's still a bit awkward but on the right track I think.  Comments welcome!

### DIFF
--- a/lib/src/rules/parameter_assignments.dart
+++ b/lib/src/rules/parameter_assignments.dart
@@ -3,11 +3,9 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'package:analyzer/dart/ast/ast.dart';
-import 'package:analyzer/dart/ast/token.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
 
 import '../analyzer.dart';
-import '../extensions.dart';
 
 const _desc =
     r"Don't reassign references to parameters of functions or methods.";
--- a/lib/src/rules/parameter_assignments.dart
+++ b/lib/src/rules/parameter_assignments.dart
@@ -113,11 +113,9 @@ class _DeclarationVisitor extends RecursiveAstVisitor {
   final bool paramDefaultsToNull;
   bool hasBeenAssigned = false;
 
-  _DeclarationVisitor(this.parameter, this.rule)
-      : paramIsNotNullByDefault = parameter is SimpleFormalParameter ||
-            _isDefaultFormalParameterWithDefaultValue(parameter),
-        paramDefaultsToNull = parameter is DefaultFormalParameter &&
-            parameter.defaultValue == null;
+  _DeclarationVisitor(this.parameter, this.rule,
+      {required this.paramIsNotNullByDefault,
+      required this.paramDefaultsToNull});
 
   @override
   visitAssignmentExpression(AssignmentExpression node) {
@@ -125,14 +123,12 @@ class _DeclarationVisitor extends RecursiveAstVisitor {
       if (_isFormalParameterReassigned(parameter, node)) {
         rule.reportLint(node);
       }
-    } else {
-      if (paramDefaultsToNull) {
-        if (_isFormalParameterReassigned(parameter, node)) {
-          if (hasBeenAssigned) {
-            rule.reportLint(node);
-          }
-          hasBeenAssigned = true;
+    } else if (paramDefaultsToNull) {
+      if (_isFormalParameterReassigned(parameter, node)) {
+        if (hasBeenAssigned) {
+          rule.reportLint(node);
         }
+        hasBeenAssigned = true;
       }
     }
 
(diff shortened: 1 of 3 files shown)


lib/src/rules/parameter_assignments.dart
```diff
@@ -124,75 +106,95 @@ class ParameterAssignments extends LintRule {
...
-            node.functionExpression.body
-                .isPotentiallyMutatedInScope(declaredElement)) {
-          _reportIfSimpleParameterOrWithDefaultValue(e, node);
+  visitAssignmentExpression(AssignmentExpression node) {
+    if (paramIsNotNullByDefault) {
+      if (_isFormalParameterReassigned(parameter, node)) {
+        rule.reportLint(node);
+      }
+    } else {
+      if (paramDefaultsToNull) {
```
bwilkerson, line 129: nit: I'd make this an `else if` in order to show that the condition is on the same level as (or parallel to) `paramIsNotNullByDefault`; that is, the outer conditions are testing the nature of the parameter whereas inner conditions are testing the assignment.

lib/src/rules/parameter_assignments.dart
```diff
@@ -124,75 +106,95 @@ class ParameterAssignments extends LintRule {
...
+  }
+}
+
+class _Visitor extends SimpleAstVisitor<void> {
+  final LintRule rule;
+
+  _Visitor(this.rule);
+
+  @override
+  void visitFunctionDeclaration(FunctionDeclaration node) {
```
bwilkerson, line 175: A `FunctionDeclaration` is a top-level function. We wrap it in a `FunctionDeclarationStatement` for local functions, so this method takes care of both of those cases. What it misses is closures, which are represented by `FunctionExpression`. If we should produce a diagnostic in the following case:
```
list.where((e) {
  e = 3; // Lint here?
});
```
then I think we need to visit `FunctionExpression` instead.

lib/src/rules/parameter_assignments.dart
```diff
@@ -124,75 +106,95 @@ class ParameterAssignments extends LintRule {
...
+    var parameterList = node.functionExpression.parameters;
+    if (parameterList == null) return;
+    for (var e in parameterList.parameters) {
+      var declaredElement = e.declaredElement;
+      if (declaredElement != null &&
+          node.functionExpression.body
+              .isPotentiallyMutatedInScope(declaredElement)) {
+        node.accept(_DeclarationVisitor(e, rule));
```
bwilkerson, line 183: We only need to visit the body, both here and below. Probably not a measurable improvement in our measurement code base, but it means that we could unify these two visit methods so that they looked something like
```
void visitFunctionDeclaration(FunctionDeclaration node) {
  _checkParameters(
    node.functionExpression.parameters,
    node.functionExpression.body,
  );
}
```

lib/src/rules/parameter_assignments.dart
```diff
@@ -124,75 +106,95 @@ class ParameterAssignments extends LintRule {
...
+  final bool paramIsNotNullByDefault;
+  final bool paramDefaultsToNull;
+  bool hasBeenAssigned = false;
 
-  _Visitor(this.rule);
+  _DeclarationVisitor(this.parameter, this.rule)
+      : paramIsNotNullByDefault = parameter is SimpleFormalParameter ||
+            _isDefaultFormalParameterWithDefaultValue(parameter),
+        paramDefaultsToNull = parameter is DefaultFormalParameter &&
+            parameter.defaultValue == null;
```
bwilkerson, line 120: If both `paramIsNotNullByDefault` and `paramDefaultsToNull` are `false`, then there's no reason to visit the body of the function because we'll never report a lint. Consider computing these values outside the constructor and passing them in if we need to create the visitor. Doing so would avoid visiting bodies unnecessarily and creating unused objects, though the latter is presumably cheap because they'd be short-lived.

test/rules/parameter_assignments_test.dart
```diff
@@ -0,0 +1,166 @@
...
+
+  test_function_named_optional_ok() async {
+    await assertNoDiagnostics(r'''
+void f({int? optional}) {
+  optional ??= 8;
+}
+''');
+  }
+
+  test_function_ok() async {
```
bwilkerson, line 71: nit: rather than 'ok' I'd say why it was ok. In this case perhaps something like 'noAssignment'.

test/rules/parameter_assignments_test.dart
```diff
@@ -0,0 +1,166 @@
...
+import '../rule_test_support.dart';
+
+main() {
+  defineReflectiveSuite(() {
+    defineReflectiveTests(ParameterAssignmentsTest);
+  });
+}
+
+@reflectiveTest
+class ParameterAssignmentsTest extends LintRuleTest {
```
bwilkerson, line 16: There should be one test for assignment of parameters of local functions.

lib/src/rules/parameter_assignments.dart
```diff
@@ -124,75 +106,95 @@ class ParameterAssignments extends LintRule {
   }
 }
 
-class _Visitor extends SimpleAstVisitor<void> {
+class _DeclarationVisitor extends RecursiveAstVisitor {
```
bwilkerson, line 109: Do we want to locate assignments within nested functions? For example:
```
void f(int p) {
  void g() {
    p = 3;
  }
  g();
}
```
If so, we should have a test for that case. If not, you'll need to override `visitFunctionExpression` to stop the recursive visitor from going inside.

bwilkerson (commented):
This is good, but we still visit the function body once for every parameter. Given how expensive visitors are, consider rewriting the logic to something like:
```
for each parameter,
  if the parameter should be checked, add it and information about it to a map
if there are any parameters to check
  visit the body looking for assignments to any of those parameters
```

test/rules/parameter_assignments_test.dart
pq: Added.  Thanks!

lib/src/rules/parameter_assignments.dart
pq: Right.  Closures are not currently supported and adding them would be breaking so I'll add a TODO.

lib/src/rules/parameter_assignments.dart
pq: Good call.  Thanks!

lib/src/rules/parameter_assignments.dart
pq: Updated.  Thanks!

lib/src/rules/parameter_assignments.dart
pq: Yep.  Test added.

Thanks!

pq:
@bwilkerson: comments addressed (thank you!!!)  would you mind taking a fresh look?

bwilkerson (approved):

Issue #1843, opened 2019-11-15
unnecessary_parenthesis shouldn't apply on `(a - b) as A`


a14n:
The following code triggers a lint `unnecessary_parenthesis`

```dart
final Offset delta = (b - a) as Offset;
```

without parenthesis it's not obvious the cast applies on `b - a` instead of `a`

bwilkerson:
I agree.

The question, though, is whether the rule is suppose to catch
- places where the use of parentheses doesn't change the semantics of the code (and hence are unnecessary in a strict sense), or
- places where they are both unnecessary in a strict sense _and_ don't improve readability.

The first is uncontroversial; the grammar defines unnecessary. The second is subject to personal preference. I don't know whether the rule is already following the second option, but I am concerned about adding special cases. If we're considering adding a special case, then I think we should at least actively request community feedback.

@davidmorgan Interested in your thoughts.

a14n:
AFAICT there are already several exceptions. See the [rule history](https://github.com/dart-lang/linter/commits/master/lib/src/rules/unnecessary_parenthesis.dart).

srawlins:
> The question, though, is whether the rule is suppose to catch

> * places where the use of parentheses doesn't change the semantics of the code (and hence are unnecessary in a strict sense), or
> * places where they are both unnecessary in a strict sense and don't improve readability.

It's the latter.

davidmorgan:
Armed with a (google internal only, sorry) list of pre-existing violations of the lint as it stands today

https://docs.google.com/spreadsheets/d/1hA38kTTAzHANsISIxjYO2fCsr6Sg6JXjSQb185u0-P8/view#gid=1710344775

I think we should probably strip this lint down to a small, easily explained list of common cases of unnecessary parentheses.

sigurdm:
Here is a similar example we hit:
```
main() {
  ((b) => b.a) as dynamic Function(dynamic);
}
```
Without the parenthesis b.a would be casted, not the function expresssion.

PR #3716, opened 2022-09-22
Fix unnecessary_parenthesis with as-expression

srawlins:
# Description

Fixes unnecessary_parenthesis handling of an `as` expression and an `is` expression. Also adds documentation explaining the heuristics of the rule.

Also changes 'parenthesis' to 'parentheses' where grammatically necessary.

Fixes https://github.com/dart-lang/linter/issues/1843

### DIFF
--- a/lib/src/rules/unnecessary_parenthesis.dart
+++ b/lib/src/rules/unnecessary_parenthesis.dart
@@ -7,10 +7,10 @@ import 'package:analyzer/dart/ast/visitor.dart';
 
 import '../analyzer.dart';
 
-const _desc = r'Unnecessary parenthesis can be removed.';
+const _desc = r'Unnecessary parentheses can be removed.';
 
 const _details = r'''
-**AVOID** using parenthesis when not needed.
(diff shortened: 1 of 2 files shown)


lib/src/rules/unnecessary_parenthesis.dart
```diff
@@ -22,6 +22,21 @@ a = b;
 a = (b);
 ```
 
+Parentheses are considered unnecessary if they do not change the meaning of the
+code and they do not improve the readability of the code. The goal is not to
+force all developers to maintain the expression precedence table in their heads,
+which is why the second condition is included. Examples of this condition
+include:
+
+* cascade expressions - it is sometimes not clear what the target of a cascase
```
bwilkerson, line 31: "cascase" --> "cascade"

bwilkerson (approved):

lib/src/rules/unnecessary_parenthesis.dart
srawlins: Thanks! Done

PR #3718, opened 2022-09-23
migrate `cascade_invocations` from `traverseNodesInDFS()`

pq:
Modest gains, but I'd like to deprecate `traverseNodesInDFS()` since it performs poorly in general (and terribly in pathological cases).

/cc @srawlins

### DIFF
--- a/lib/src/rules/cascade_invocations.dart
+++ b/lib/src/rules/cascade_invocations.dart
@@ -266,12 +266,41 @@ class _CascadableExpression {
       !_hasCriticalDependencies(expressionBox);
 
   bool _hasCriticalDependencies(_CascadableExpression expressionBox) {
-    bool isCriticalNode(AstNode node) =>
-        node.canonicalElement == expressionBox.element;
-    return expressionBox.isCritical &&
-        criticalNodes.any((node) =>
-            isCriticalNode(node) ||
-            node.traverseNodesInDFS().any(isCriticalNode));
+    if (!expressionBox.isCritical) return false;
+
+    for (var node in criticalNodes) {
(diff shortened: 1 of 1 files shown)


srawlins (approved):

lib/src/rules/cascade_invocations.dart
```diff
@@ -266,12 +266,41 @@ class _CascadableExpression {
...
+      if (_NodeVisitor(expressionBox).isOrHasCriticalNode(node)) {
+        return true;
+      }
+    }
+
+    return false;
+  }
+}
+
+class _NodeVisitor extends GeneralizingAstVisitor {
```
bwilkerson, line 281: This should use `UnifyingAstVisitor`.

lib/src/rules/cascade_invocations.dart
```diff
@@ -266,12 +266,41 @@ class _CascadableExpression {
...
+  final _CascadableExpression expressionBox;
+
+  bool foundCriticalNode = false;
+  _NodeVisitor(this.expressionBox);
+
+  bool isCriticalNode(AstNode node) =>
+      node.canonicalElement == expressionBox.element;
+
+  bool isOrHasCriticalNode(AstNode node) {
+    if (isCriticalNode(node)) return true;
```
bwilkerson, line 291: As per a comment on a different PR, when we visit the `node` we'll repeat this test. Unless we gain a performance benefit from skipping the visit, we should remove this line to avoid duplication.

lib/src/rules/cascade_invocations.dart
pq: Thanks!

PR #3719, opened 2022-09-23
migrate `prefer_foreach` from `traverseNodesInDFS()`

pq:
Surprisingly big gain in the general case (and we all know this can get quadratic in the pathological one). 🏁 

**BEFORE**

image

**AFTER**

image

/cc @srawlins @bwilkerson

### DIFF
--- a/lib/src/rules/prefer_foreach.dart
+++ b/lib/src/rules/prefer_foreach.dart
@@ -116,7 +116,7 @@ class _PreferForEachVisitor extends SimpleAstVisitor {
   }
 }
 
-class _ReferenceFinder extends GeneralizingAstVisitor {
+class _ReferenceFinder extends UnifyingAstVisitor {
   bool found = false;
   final LocalVariableElement? element;
   _ReferenceFinder(this.element);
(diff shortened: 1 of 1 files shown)


srawlins (approved):
🎉

lib/src/rules/prefer_foreach.dart
```diff
@@ -121,6 +116,29 @@ class _PreferForEachVisitor extends SimpleAstVisitor {
   }
 }
 
+class _ReferenceFinder extends GeneralizingAstVisitor {
```
bwilkerson, line 119: This should use `UnifyingAstVisitor`.

lib/src/rules/prefer_foreach.dart
```diff
@@ -121,6 +116,29 @@ class _PreferForEachVisitor extends SimpleAstVisitor {
...
+  bool found = false;
+  final LocalVariableElement? element;
+  _ReferenceFinder(this.element);
+
+  bool references(Expression target) {
+    if (target.canonicalElement == element) return true;
```
bwilkerson, line 125: This test will be repeated in `visitNode`. If there isn't a performance benefit from not visiting the node (which there might well be if we expect this to return immediately often enough), then consider removing it.

pq: It's a good point. I'm inclined to short-circuit visiting wherever possible but I haven't measured this...  I'll look a bit closer.  Thanks!

pq:
(Landing as-is and will investigate the need for AST visit short-circuiting.)

PR #3724, opened 2022-09-23
Report avoid_returning_this on the return statement

### LINKED ISSUE #780: avoid_returning_this complaints should highlight the return not the function
Just a nit

image

(generated code from built_value)

I found this confusing because the getter name contains `this` as well.
If it's a longer body with multiple returns, it can be cumbersome to locate the issue.

srawlins:
# Description

This just changes the lint report position to be the first `return this` statement.

Fixes #780

### DIFF
--- a/lib/src/rules/avoid_returning_this.dart
+++ b/lib/src/rules/avoid_returning_this.dart
@@ -98,7 +98,7 @@ class _Visitor extends SimpleAstVisitor<void> {
           .traverseNodesInDFS(excludeCriteria: _isFunctionExpression)
           .whereType<ReturnStatement>();
       if (returnStatements.isNotEmpty && returnStatements.every(_returnsThis)) {
-        rule.reportLint(returnStatements.first);
+        rule.reportLint(returnStatements.first.expression);
       }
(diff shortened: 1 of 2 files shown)


lib/src/rules/avoid_returning_this.dart
```diff
@@ -99,9 +96,9 @@ class _Visitor extends SimpleAstVisitor<void> {
     if (body is BlockFunctionBody) {
       var returnStatements = body.block
           .traverseNodesInDFS(excludeCriteria: _isFunctionExpression)
-          .where(_isReturnStatement);
+          .whereType<ReturnStatement>();
       if (returnStatements.isNotEmpty && returnStatements.every(_returnsThis)) {
-        rule.reportLintForToken(node.name2);
+        rule.reportLint(returnStatements.first);
```
bwilkerson, line 101: I'd prefer narrowing it further to reporting the diagnostic on the `this`.

srawlins: Done!

srawlins:
PTAL @bwilkerson

bwilkerson (approved):

PR #3725, opened 2022-09-23
Add unnecessary_stateful_widgets

### LINKED ISSUE #3683: lint: StatefulWidget could be a StatelessWidget
Occasionally, I see Flutter code where people use StatefulWidgets even though a StatelessWidget would have been sufficient. Example:

```dart
class MyWidget extends StatefulWidget {
  const MyWidget({super.key});

  @override
  State<MyWidget> createState() => _MyWidgetState();
}

class _MyWidgetState extends State<MyWidget> {
  @override
    return Container();
  }
}
```

The `State` class doesn't store any state in member variables and it also doesn't use any of the properties provided by `Sate` (e.g. `mounted`). This could easily be refactored to be a more concise `StatelessWidget`:

```dart
class MyWidget extends StatelessWidget {
  const MyWidget({super.key});

  @override
    return Container();
  }
}
```

I wonder if it would be possible to have a lint for this?

Of course, from a functional perspective using a StatefulWidget here is totally fine, the additional performance overhead of a StatefulWidget over a StatelessWidget is probably neglectable. However, using a StatelessWidget where appropriate leads to more compact, concise, and readable code and is more idiomatic, IMO.

### LINKED ISSUE #58868: lint: StatefulWidget could be a StatelessWidget
Occasionally, I see Flutter code where people use StatefulWidgets even though a StatelessWidget would have been sufficient. Example:

```dart
class MyWidget extends StatefulWidget {
  const MyWidget({super.key});

  @override
  State<MyWidget> createState() => _MyWidgetState();
}

class _MyWidgetState extends State<MyWidget> {
  @override
    return Container();
  }
}
```

The `State` class doesn't store any state in member variables and it also doesn't use any of the properties provided by `Sate` (e.g. `mounted`). This could easily be refactored to be a more concise `StatelessWidget`:

```dart
class MyWidget extends StatelessWidget {
  const MyWidget({super.key});

  @override
    return Container();
  }
}
```

I wonder if it would be possible to have a lint for this?

Of course, from a functional perspective using a StatefulWidget here is totally fine, the additional performance overhead of a StatefulWidget over a StatelessWidget is probably neglectable. However, using a StatelessWidget where appropriate leads to more compact, concise, and readable code and is more idiomatic, IMO.

a14n:
Fixes dart-lang/sdk#58868

### DIFF
--- a/lib/src/rules/unnecessary_stateful_widgets.dart
+++ b/lib/src/rules/unnecessary_stateful_widgets.dart
@@ -80,8 +80,8 @@ class _Visitor extends SimpleAstVisitor<void> {
   void visitCompilationUnit(CompilationUnit node) {
     var classes = node.declarations.whereType<ClassDeclaration>().toList();
     widgetLoop:
-    for (var statefulWidget
-        in classes.where((e) => isStatefulWidget(e.declaredElement2))) {
+    for (var statefulWidget in classes
+        .where((e) => isExactStatefulWidget(e.declaredElement2?.supertype))) {
       var createState = statefulWidget.members
           .whereType<MethodDeclaration>()
           .firstWhereOrNull((e) => e.name2.lexeme == 'createState');
--- a/lib/src/rules/unnecessary_stateful_widgets.dart
+++ b/lib/src/rules/unnecessary_stateful_widgets.dart
@@ -13,7 +13,6 @@ import '../util/flutter_utils.dart';
 const _desc = r'Unnecessary StatefulWidget.';
 
 const _details = r'''
-
 Don't use `StatefulWidget` when a `StatelessWidget` is sufficient. Using a
 `StatelessWidget` where appropriate leads to more compact, concise, and readable
 code and is more idiomatic.
--- a/lib/src/rules/unnecessary_stateful_widgets.dart
+++ b/lib/src/rules/unnecessary_stateful_widgets.dart
@@ -101,6 +101,8 @@ class _Visitor extends SimpleAstVisitor<void> {
       if (extendsClause == null) continue;
       var parent = extendsClause.superclass.name.staticElement;
       if (parent is! ClassElement || !isExactState(parent)) continue;
+      // check no field
+      if (stateDeclaration.fields.isNotEmpty) continue;
       // check no overriden methods except `build`
       for (var method in stateDeclaration.methods) {
         if (method.name2.lexeme == 'build') continue;
@@ -130,6 +132,8 @@ class _StateUsageVisitor extends RecursiveAstVisitor<void> {
 }
 
 extension on ClassDeclaration {
+  Iterable<FieldDeclaration> get fields =>
+      members.whereType<FieldDeclaration>();
   Iterable<MethodDeclaration> get methods =>
       members.whereType<MethodDeclaration>();
 }
--- a/lib/src/rules/unnecessary_stateful_widgets.dart
+++ b/lib/src/rules/unnecessary_stateful_widgets.dart
@@ -81,6 +81,9 @@ class _Visitor extends SimpleAstVisitor<void> {
     widgetLoop:
     for (var statefulWidget in classes
         .where((e) => isExactStatefulWidget(e.declaredElement2?.supertype))) {
+      // if Stateful is used in name, we ignore it
+      if (statefulWidget.name2.lexeme.contains('Stateful')) continue;
+
       var createState = statefulWidget.members
           .whereType<MethodDeclaration>()
           .firstWhereOrNull((e) => e.name2.lexeme == 'createState');
--- a/lib/src/rules/unnecessary_stateful_widgets.dart
+++ b/lib/src/rules/unnecessary_stateful_widgets.dart
@@ -8,6 +8,7 @@ import 'package:analyzer/dart/element/element.dart';
 import 'package:collection/collection.dart';
 
 import '../analyzer.dart';
+import '../extensions.dart';
 import '../util/flutter_utils.dart';
 
 const _desc = r'Unnecessary StatefulWidget.';
@@ -80,13 +81,13 @@ class _Visitor extends SimpleAstVisitor<void> {
     var classes = node.declarations.whereType<ClassDeclaration>().toList();
     widgetLoop:
     for (var statefulWidget in classes
-        .where((e) => isExactStatefulWidget(e.declaredElement2?.supertype))) {
+        .where((e) => isExactStatefulWidget(e.declaredElement?.supertype))) {
       // if Stateful is used in name, we ignore it
-      if (statefulWidget.name2.lexeme.contains('Stateful')) continue;
+      if (statefulWidget.name.lexeme.contains('Stateful')) continue;
 
       var createState = statefulWidget.members
           .whereType<MethodDeclaration>()
-          .firstWhereOrNull((e) => e.name2.lexeme == 'createState');
+          .firstWhereOrNull((e) => e.name.lexeme == 'createState');
       if (createState == null) continue;
       var createStateBody = createState.body;
       if (createStateBody is! ExpressionFunctionBody) continue;
@@ -95,11 +96,11 @@ class _Visitor extends SimpleAstVisitor<void> {
       var stateName = stateElement.name;
 
       var stateDeclaration =
-          classes.where((e) => e.name2.lexeme == stateName).singleOrNull;
+          classes.where((e) => e.name.lexeme == stateName).singleOrNull;
       if (stateDeclaration == null) continue;
       if (stateDeclaration.withClause != null) continue;
       // check state is private
-      if (stateDeclaration.declaredElement2?.isPublic ?? true) continue;
+      if (stateDeclaration.declaredElement?.isPublic ?? true) continue;
       // check `extends State`
       var extendsClause = stateDeclaration.extendsClause;
       if (extendsClause == null) continue;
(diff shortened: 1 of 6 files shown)


test_data/rules/unnecessary_stateful_widgets.dart
```diff
@@ -0,0 +1,113 @@
...
+  final int i = 2;
+
+  @override
+  Widget build(BuildContext context) {
+    return Container();
+  }
+}
+
+// State public
+class MyWidget4 extends StatefulWidget // OK
```
goderbauer, line 70: What's the reasoning behind marking this as OK? I think, naively I would have marked this with the lint as well...

test_data/rules/unnecessary_stateful_widgets.dart
```diff
@@ -0,0 +1,113 @@
...
+// State with field
+class MyWidget3 extends StatefulWidget // OK
+{
+  const MyWidget3({super.key});
+
+  @override
+  State<MyWidget3> createState() => _MyWidget3State();
+}
+
+class _MyWidget3State extends State<MyWidget3> {
```
goderbauer, line 60: This should really be linted as well since the state is not really holding on to any state...

goderbauer: But, of course, if a member `i` holds on to anything stateful (say a FocusNode or a ScrollController) it shouldn't lint. I guess, we probably can't differentiate between these cases?

goderbauer (commented):
How does this do when run over the flutter repository? Does it find anything? Any false alarms?

test_data/rules/unnecessary_stateful_widgets.dart
a14n: Removing the state would be a breaking change as someone could have referenced this type outside. For instance in flutter package `PopupMenuButtonState` is one of these cases (and the reason why I added this exception).

test_data/rules/unnecessary_stateful_widgets.dart
a14n: I started with only a rule to catch usage of `setState` but on the flutter codebase I face several false positives with State having fields. As you mentioned it's not easy to catch what field is stateful.

a14n:
> How does this do when run over the flutter repository? Does it find anything?

On packages/flutter : https://github.com/flutter/flutter/pull/112296

> Any false alarms?

The exceptions I added in this linter PR is the result of a too broaden application.

test_data/rules/unnecessary_stateful_widgets.dart
a14n: If i remove the absence of fields in state condition, here's the list of diagnotics

```
   info • Unnecessary StatefulWidget • packages/flutter/lib/src/material/about.dart:530:7 • unnecessary_stateful_widgets
   info • Unnecessary StatefulWidget • packages/flutter/lib/src/material/bottom_sheet.dart:67:7 • unnecessary_stateful_widgets
   info • Unnecessary StatefulWidget • packages/flutter/lib/src/material/bottom_sheet.dart:350:7 • unnecessary_stateful_widgets
   info • Unnecessary StatefulWidget • packages/flutter/lib/src/material/dropdown.dart:95:7 • unnecessary_stateful_widgets
   info • Unnecessary StatefulWidget • packages/flutter/lib/src/widgets/fade_in_image.dart:61:7 • unnecessary_stateful_widgets
   info • Unnecessary StatefulWidget • packages/flutter/test/material/reorderable_list_test.dart:1782:7 • unnecessary_stateful_widgets
   info • Unnecessary StatefulWidget • packages/flutter/test/widgets/animated_list_test.dart:494:7 • unnecessary_stateful_widgets
   info • Unnecessary StatefulWidget • packages/flutter/test/widgets/heroes_test.dart:162:7 • unnecessary_stateful_widgets
   info • Unnecessary StatefulWidget • packages/flutter/test/widgets/inherited_model_test.dart:55:7 • unnecessary_stateful_widgets
   info • Unnecessary StatefulWidget • packages/flutter/test/widgets/slivers_block_global_key_test.dart:11:7 • unnecessary_stateful_widgets
```

A lot of them are not simple/impossible to make statelessWidget

asashour:
How about checking if an override is a default one, so it is not preventing the lint from being triggered?

E.g. 

```dart
@override
void dispose() => super.dispose();
```

The relevant check is in [FlutterConvertToStatelessWidget](https://github.com/dart-lang/sdk/blob/cfddeee68e0cf7451a2eee1a9fbbff8a8d9c0b92/pkg/analysis_server/lib/src/services/correction/dart/flutter_convert_to_stateless_widget.dart#L228) assist.

a14n:
> How about checking if an override is a default one, so it is not preventing the lint from being triggered?

Thanks for the suggestion.

This case should be catch by [unnecessary_overrides](https://dart-lang.github.io/linter/lints/unnecessary_overrides.html). Not sure it is worth adding complexity here to catch things like this.

This `unnecessary_stateful_widgets` could surely be really more complex to catch more cases. But is it worth it regarding the cost of maintenance and how many additionnal diagnotics could arise.

goderbauer:
I added some comment over at: https://github.com/flutter/flutter/pull/112296#discussion_r981553217

I wonder if we should omit the lint if the State implementation is actually using functionality of the `State` class. E.g. if you access `State.context` or `State.mounted` or some such I think it is ok to be a stateful widget.

a14n:
I updated the lint to have an exception if the WidgetState uses a member of its parent.

I think this PR is now ready from a functional point of view.

goderbauer (commented):
LGTM for the cases in which the lint does and doesn't trigger (as documented in the `test_data`).

The lint implementation needs a review from somebody more knowledgable in the linter space... :)

srawlins:
I want to double check that we're "accepting" the new lint request at https://github.com/dart-lang/linter/issues/3683. Then we'll review.

lib/src/rules/unnecessary_stateful_widgets.dart
```diff
@@ -0,0 +1,172 @@
...
+  }
+
+  @override
+  void visitSimpleIdentifier(SimpleIdentifier node) {
+    if (node.name == 'widget' || !_visit(node.canonicalElement)) {
+      super.visitSimpleIdentifier(node);
+    }
+  }
+
+  bool _visit(Element? methodElement) {
```
srawlins, line 155: Can you document what this method returns?

srawlins (approved):
Technically this looks good to me. The proposal has been accepted.

CC @pq @bwilkerson if you care to review from a technical perspective.

devoncarew:

Turskyi:
@devoncarew , @goderbauer , @srawlins I’m struggling to understand the current situation regarding this issue. If the unnecessary_stateful_widgets linter has moved to a new repository, why is the issue (https://github.com/dart-lang/linter/issues/3683) still open? 

If it hasn’t been addressed, why wasn’t the approved pull request merged during the year it remained pending, especially since it passed all tests and was approved?

It’s deeply concerning that contributions from the community—like this PR, which could have improved the framework—are effectively discarded due to delays in merging and the subsequent repository move. This discourages contributors who are genuinely invested in helping Flutter progress.

If the move to a new repository made the merge impossible, shouldn’t the team have proactively migrated this PR or implemented the changes themselves, given that it was already approved? Asking the contributor to recreate their work seems unfair and reflects poorly on the process.

I hope this feedback is taken constructively to avoid similar situations in the future and to better respect the time and efforts of community contributors.

srawlins:
> I’m struggling to understand the current situation regarding this issue. If the unnecessary_stateful_widgets linter has moved to a new repository, why is the issue (https://github.com/dart-lang/linter/issues/3683) still open?

The issues have not yet been moved to the dart-lang/sdk repo. They will be.

> If it hasn’t been addressed, why wasn’t the approved pull request merged during the year it remained pending, especially since it passed all tests and was approved?

We are not always good about triaging this repository's pull request queue. Additionally, I see an un-addressed comment above.

> It’s deeply concerning that contributions from the community—like this PR, which could have improved the framework—are effectively discarded due to delays in merging and the subsequent repository move. This discourages contributors who are genuinely invested in helping Flutter progress.

Apologies for this. The Developer Experience team is actively working to improve our triage processes.

> If the move to a new repository made the merge impossible, shouldn’t the team have proactively migrated this PR or implemented the changes themselves, given that it was already approved? Asking the contributor to recreate their work seems unfair and reflects poorly on the process.

Great question! I don't know of a way to "migrate a PR." And for a languishing PR, it does not seem to be worth the effort. Some PRs are genuinely abandoned.

devoncarew:
@Turskyi, thanks for the feedback.

We're transferring the issues shortly; see the last comment on https://github.com/dart-lang/linter/issues/4729. They won't be closed here and opened elsewhere, they'll just transfer to the sdk repo.

Re: the PR - the source of truth for the code for the linter moved over a year ago (on or around https://github.com/dart-lang/linter/commit/59cd18d1664915df1d1115cda84acda75675e1e2); after that point merging PRs here would no longer makes sense. Existing PRs should have been cleaned up then.

And I can't speak to this specific PR, but PRs can't be transferred like issues; a re-review of already approved content would likely be relatively brief.

Turskyi:
Thank you for your transparency. However, the underlying message here seems to be: "_We’re okay with letting this PR go because fixing it ourselves isn’t worth the effort._" While I understand the challenges of repository migrations and triaging contributions, this approach risks alienating contributors who invest time and effort to improve the framework.
Open-source projects thrive on collaboration, and dismissing already-approved contributions sends the wrong message. I hope this can be a lesson to prioritize processes that respect community efforts and ensure their work doesn’t go to waste due to internal changes.

PR #3726, opened 2022-09-23
unnecessary_overrides: correctly identify the overridden signature

### LINKED ISSUE #1997: unnecessary_overrides does not work for generic methods and null safety mix
The lint is not reported for:
```dart
class A<T> {
  T foo() {
    throw 42;
  }
}

class B extends A<int> {
  @override
  int foo() => super.foo();
}
```

This is a failing test from analysis server.
```dart
  Future<void> test_method_nullSafety_optIn_fromOptOut() async {
    createAnalysisOptionsFile(
      experiments: ['non-nullable'],
      lints: [lintCode],
    );
    newFile('/home/test/lib/a.dart', content: r'''
class A {
  int foo() => 0;
}
''');
    await resolveTestUnit('''
// @dart = 2.7
import 'a.dart';

class B extends A {
  @override
  int foo() => super.foo();
}
''');
    await assertHasFix('''
// @dart = 2.7
import 'a.dart';

class B extends A {
}
''');
  }
```

srawlins:
# Description

In the case of a subclass which instantiates a type parameter of a super class, the signature of the inherited method must also be instantiated.

As the code is before this change, given these classes:

```dart
class F<T> {
  T m1(T p) => throw 42;
}

class G extends F<int> {
  @override
  int m1(int p) => super.m1(p); // LINT
}
```

the signature of `G.m1` is `int Function(int)`, and the rule _thinks_ that the signature of the super method is `T Function(int)`, but it's really `int Function(int)`. This fix correctly calculates the signature of the super method.

Fixes #1997

In fact, I think this is true even if the subclass does not instantiate the type parameter, but rather uses its own type parameter in the `extends` clause (maybe this is still instantiation? to a type variable?). So this doesn't work before this fix either:

```dart
class F<T> {
  T m1(T p) => throw 42;
}

class G<T> extends F<T> {
  @override
  T m1(T p) => super.m1(p); // LINT
}
```


srawlins:
@scheglov if you don't mind, I'd like your review in particular, to know if I've done the best, or the simplest route that can get me the properly instantiated ExecutableElement for a given method's super method.

lib/src/rules/unnecessary_overrides.dart
```diff
@@ -227,13 +235,22 @@ class _UnnecessaryMethodOverrideVisitor
   _UnnecessaryMethodOverrideVisitor(super.rule);
 
   @override
-  ExecutableElement? getInheritedElement(node) => node.lookUpInheritedMethod();
+  ExecutableElement? getInheritedElement(node) {
+    var inheritedMethod = node.lookUpInheritedMethod();
+    if (inheritedMethod == null) return null;
+    var enclosingTypeAsInstanceOfInheritedEnclosingType =
```
srawlins, line 241: I am very open to better names here, both for the local variable and for the extension method on MethodDeclaration :D. I thought for a while though and couldn't think of anything shorter which satisfied me.

scheglov:
Yes, it does not matter whether we provide `int` as a type argument for the supertype, or a type parameter type `T`, or nothing at all - semantically the supertype is always instantiated.

lib/src/rules/unnecessary_overrides.dart
```diff
@@ -227,13 +237,23 @@ class _UnnecessaryMethodOverrideVisitor
   _UnnecessaryMethodOverrideVisitor(super.rule);
 
   @override
-  ExecutableElement? getInheritedElement(node) => node.lookUpInheritedMethod();
```
scheglov, line 230: Do we still use `lookUpInheritedMethod()`?

scheglov: We have in `InterfaceType`:
```dart
  MethodElement? lookUpMethod2(
    String name,
    LibraryElement library, {
    bool concrete = false,
    bool inherited = false,
    bool recoveryStatic = false,
  }) {
```

So, I wonder if `thisType.lookUpMethod2(name, library, concrete: true, inherited: true)` would work.

srawlins:
Flutter CI and google3 CI look green.

lib/src/rules/unnecessary_overrides.dart
srawlins: Thanks that works perfectly!

pq (approved):

PR #3728, opened 2022-09-26
Ensure `BAD` comes before `GOOD` in the documentation

### LINKED ISSUE #2488: Standardize on whether GOOD or BAD comes first in docs
Some lint docs list the GOOD example first (e.g. https://dart-lang.github.io/linter/lints/no_adjacent_strings_in_list.html) others list the BAD example first (e.g. https://dart-lang.github.io/linter/lints/hash_and_equals.html).

Maybe the order could be standardized?

asashour:
Fixes #2488

Please note that the details of `comment_references` was manually changed, as there was a statement before the `GOOD`.

This PR also:
- Ensures keywords are followed by a colon, since we had `**GOOD**:` for example.
- Ensures keywords are in upper-case, e.g. we had `**Note:**`
- In few cases, the new line just after the `**GOOD**` is removed, e.g. `prefer_relative_imports`. No test case added.


lib/src/rules/always_put_required_named_parameters_first.dart
```diff
@@ -12,26 +12,26 @@ const _desc = r'Put required named parameters first.';
...
 ```
 
 **BAD:**
 ```dart
 m({b, c, @required a}) ;
 ```
 
+**GOOD:**
+```dart
+m({@required a, b, c}) ;
```
albertms10, line 32: Shouldn’t we use `required` keyword instead of the deprecated annotation?

asashour: The documentation shows `good` and `bad` for both the `required` and `@required`. I guess it is there because it still can used.

srawlins:
Sorry for the delay, @asashour . I'm reviewing now; it's not easy with the line-by-line diffs, haha. But I do like the BAD before GOOD.

srawlins (changes requested):
A few nits but this looks great.

srawlins:
@pq I imagine we should hold off on PRs which are unrelated to the in-progress roll?

srawlins:
Sorry, @asashour , one more rebase; then I propose we land this!

pq (approved):
💎

pq:
@srawlins: if you want to take another look, I think this should be good to land.  Thanks a million, @asashour!

srawlins (approved):
Huzzah!

PR #3730, opened 2022-09-26
Support unnamed libraries

srawlins:
# Description

Support unnamed libraries across all lint rules. The only ones which use LibraryDirective are:

* `library_names`
* `package_prefixed_library_name` which is just a big comment. So, easy peasy.

### DIFF
--- a/lib/src/rules/library_names.dart
+++ b/lib/src/rules/library_names.dart
@@ -57,8 +57,9 @@ class _Visitor extends SimpleAstVisitor<void> {
 
   @override
   void visitLibraryDirective(LibraryDirective node) {
-    if (!isLowerCaseUnderScoreWithDots(node.name2.toString())) {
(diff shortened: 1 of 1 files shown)


srawlins:
@bwilkerson as far as tests, since each Dart library can only have one `library` directive, I think I need to convert the `library_names` lint rule to use reflective loader tests. Is there a good example PR of that I can follow?

bwilkerson:
I don't know of one, but we haven't been requiring converting older tests to the new style; we have several lints with tests of both forms.

srawlins:
Since each Dart library can only have one library directive, I think I need to convert the library_names lint rule to use reflective loader tests, right?

srawlins:
I switched to unit tests; so very nice.

bwilkerson (approved):

Issue #3565, opened 2022-07-29
Possible memory leak in use_late_for_private_fields_and_variables


bwilkerson:
The visitor for `use_late_for_private_fields_and_variables` defines two static fields. They are used to keep track of information across visitors when a library has multiple compilation units (that is, has parts).

While it makes sense to allow a lint to examine all of the compilation units in a library, the current approach is susceptible to a memory leak because entries are only removed from the map when all of the units in the library have been visited. But (a) something could happen to prevent some of the units from being linted, or (b) an exception could occur when analyzing the last unit in the library that prevents the clean-up code from being reached.

I believe that the right way to support this kind of lint is to make these fields be instance fields and to visit all of the compilation units in the library at the same time. However, I'm not sure that the linting framework is written to allow diagnostics to be reported against anything other than the current compilation unit, so that might not be practical, and we might need a different approach.

And, assuming a safe implementation strategy is found, this should be done in a way that makes it easier to write other cross-library lints in the future.

srawlins:
There are no tests for the `part` support, so those will need to be added when a fix is made.

srawlins:
This support might already be broken. When I analyze these two files in an integration test:

```dart
// library.dart:

part 'part.dart';

int? _i2; // LINT

// part.dart:

part of 'library.dart';

m2() {
  _i2 = 1;
}
```
I get the following output:

```none
/Users/srawlins/code/dart-linter/test_data/integration/use_late_for_private_fields_and_variables/part.dart 6:1 [lint] Use late for private members with a non-nullable type.

^

2 files analyzed, 1 issue found, in 3819 ms.
```

mosuem:
Is this still a P1? Just asking, as it has been a while since the last update.

srawlins:
Oops, we have completely dropped the ball on triaging whether P1 issues in this repo are being worked on. 😦

mosuem:
No worries, I was just checking the status page on https://dart-pr-dashboard.web.app/ and saw this.

PR #3731, opened 2022-09-28
Move use_late_for_private_fields_and_variables tests to reflective

srawlins:
# Description

Before addressing https://github.com/dart-lang/linter/issues/3565, we need to convert the tests to reflective (in order to use part files). This PR does exactly two things:

* A simple translation of the existing tests to unit test cases; I tried to capture the spirit of each existing `// OK` and `// LINT` in the integration test file, but some are probably wrong.
* One new test: a variable is declared in a library file, and referenced in a part file, and should be reported properly.


bwilkerson (approved):
I can't verify that the new tests cover all of the cases covered by the old tests (there are fewer new tests than markers in old tests, and the old tests don't always have `// OK` markers where it should have been) but it's probably complete enough. It looks like some of the old tests might have been testing the same thing and it's coverage, not count, that matters.

srawlins:
Good catch. In making it a one-to-one mapping to unit tests, I think I did catch one or two that seemed just duplicated with no new value.

PR #3732, opened 2022-09-28
Correct Documentation Examples for avoid_equals_and_hash_code_on_mutable_classes

MelbourneDeveloper:
# Description

The three examples for avoid_equals_and_hash_code_on_mutable_classes don't compile because `hashCode` is a getter and not a method. Also, the class `C` references `B` instead of itself.

You can now copy and paste these examples into a Dart app.

Just a note: the second example also violates [const_constructor_with_non_final_field](https://dart.dev/diagnostics/const_constructor_with_non_final_field) which could be a bit confusing.

### DIFF
--- a/lib/src/rules/avoid_equals_and_hash_code_on_mutable_classes.dart
+++ b/lib/src/rules/avoid_equals_and_hash_code_on_mutable_classes.dart
@@ -30,7 +30,7 @@ class A {
   @override
   operator ==(other) => other is A && other.key == key;
   @override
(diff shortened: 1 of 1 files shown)


srawlins (approved):
Thanks for the fix!!

pq:
This is great.  Thanks a million, @MelbourneDeveloper!

When you have a moment, could you add an attribution for yourself to [AUTHORS](https://github.com/dart-lang/linter/blob/main/AUTHORS)?

PR #3740, opened 2022-10-03
New rule: `no_adjacent_strings_in_collection`

### LINKED ISSUE #2343: no_adjacent_strings_in_list should trigger when the collection if is missing a trailing comma

### LINKED ISSUE #2727: Deprecate no_adjacent_strings_in_list, replacing with no_adjacent_strings_in_collections

We should replace no_adjacent_strings_in_list with a new lint rule, no_adjacent_strings_in_collections, which would catch adjacent strings in sets, lists, if-elements, and for-elements.

**Examples**

```dart
[
  if (1 == 2) 'a' // LINT
  'b'
];

[
  for (var v in []) 'a' // LINT
  'b'
];
```

**Additional context**

It would be a breaking change to add if-collection and for-collection support to `no_adjacent_strings_in_list` (https://github.com/dart-lang/linter/pull/2726/). It would not make naming sense to add set support to that existing rule, and also be a breaking change.

I can't conceive of a reason someone would want adjacent strings reported in lists but not sets, or vice versa, so it makes sense to put them in one rule.

srawlins:
# Description

New rule! `no_adjacent_strings_in_collection`, an alternative to https://github.com/dart-lang/linter/pull/2726, and a replacement for `no_adjacent_strings_in_list`. Improvements (differences) include:

* works on set literals
* works on collection for-elements
* works on collection if-elements

Fixes #2343 and fixes #2727

### DIFF
--- a/example/all.yaml
+++ b/example/all.yaml
@@ -88,6 +88,7 @@ linter:
     - list_remove_unrelated_type
     - literal_only_boolean_expressions
     - missing_whitespace_between_adjacent_strings
+    - no_adjacent_strings_in_collection
     - no_adjacent_strings_in_list
     - no_default_cases
     - no_duplicate_case_values
--- a/lib/src/rules.dart
+++ b/lib/src/rules.dart
@@ -91,6 +91,7 @@ import 'rules/lines_longer_than_80_chars.dart';
 import 'rules/list_remove_unrelated_type.dart';
 import 'rules/literal_only_boolean_expressions.dart';
 import 'rules/missing_whitespace_between_adjacent_strings.dart';
+import 'rules/no_adjacent_strings_in_collection.dart';
 import 'rules/no_adjacent_strings_in_list.dart';
 import 'rules/no_default_cases.dart';
 import 'rules/no_duplicate_case_values.dart';
@@ -308,6 +309,7 @@ void registerLintRules({bool inTestMode = false}) {
     ..register(ListRemoveUnrelatedType())
     ..register(LiteralOnlyBooleanExpressions())
     ..register(MissingWhitespaceBetweenAdjacentStrings())
+    ..register(NoAdjacentStringsInCollection())
     ..register(NoAdjacentStringsInList())
     ..register(NoDefaultCases())
     ..register(NoDuplicateCaseValues())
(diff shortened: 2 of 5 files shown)


bwilkerson:
I think that we want to be very conservative about adding new lint rules at this point. Given that there's a discussion about whether we're going to keep adjacent strings, I wonder whether it might not be better to wait on adding more lints around them until a decision has been made. I'd hate to add a new lint and then deprecate it a couple of months later.

Also, given that `no_adjacent_strings_in_list` isn't in the core or recommended lint sets, it might be appropriate to just extend that rule to cover the set case.

WDYT?

srawlins:
> I'd hate to add a new lint and then deprecate it a couple of months later.

I don't think "no adjacent strings" is on the table for Dart 3.0, so it will be at least a year, and more like 5? Right?

> Also, given that no_adjacent_strings_in_list isn't in the core or recommended lint sets, it might be appropriate to just extend that rule to cover the set case.

I love it. We can land https://github.com/dart-lang/linter/pull/2726 instead of this. The lint rule _name_ would stay the same though; it would say `in_list` even though it guarded set literals too.

bwilkerson:
> I don't think "no adjacent strings" is on the table for Dart 3.0 ...

I have no idea what timeframe they're thinking about, even assuming that they decide to remove it (which isn't at all clear to me either). I just thought it might be good to give the discussion a few days to see whether a decision is made or whether they stall out.

> The lint rule _name_ would stay the same though ...

I've wondered a few times whether we might want some kind of 'alias' feature for lints so that we could more easily rename them. We decided a long time ago to not use "do", "don't", "prefer", etc. in lint names, but we don't have a good way to go back and clean up the older lints to follow the newer naming conventions.

Not sure we have time to add such a feature right now, but it would support this renaming as well.

srawlins:
Replaced by #3741

PR #3741, opened 2022-10-03
Check for-elements, if-elements, and set-literals for adjacent strings

### LINKED ISSUE #2343: no_adjacent_strings_in_list should trigger when the collection if is missing a trailing comma

srawlins:
# Description

Adds for-elements, if-elements, and set literals to the `no_adjacent_strings_in_list` rule.

Fixes #2343 and #2727. This is a replacement for https://github.com/dart-lang/linter/pull/2726 as I don't use that source repo any more.

This is an alternative to https://github.com/dart-lang/linter/pull/3740 (creating a new rule).


lib/src/rules/no_adjacent_strings_in_list.dart
```diff
@@ -55,6 +58,20 @@ class _Visitor extends SimpleAstVisitor<void> {
...
+  void visitForElement(ForElement node) {
+    if (node.body is AdjacentStrings) {
+      rule.reportLint(node.body);
+    }
+  }
+
+  @override
+  void visitIfElement(IfElement node) {
+    if (node.thenElement is AdjacentStrings) {
+      rule.reportLint(node.thenElement);
```
bwilkerson, line 71: We probably don't want to report a lint here if there's an `elseElement`, but should report a lint in the `elseElement` if it exists. For example, given
```
var list = [
  if (c) 'a' 'b' else 'c' 'd'
];
```
It wouldn't make any sense to add a comma between `'a'` and `'b'` because `'b' else 'c' 'd'` isn't be a valid expression. But it would name sense to suggest a comma between `'c'` and `'d'`.

test_data/rules/no_adjacent_strings_in_list.dart
```diff
@@ -5,23 +5,45 @@
...
   ];
 
-  List<String> list2 = <String>[
+  var list2 = [
     'a' // LINT
     'b'
     'c'
   ];
+
+  var list3 = [
```
bwilkerson, line 20: Consider using reflective tests for the new tests.

lib/src/rules/no_adjacent_strings_in_list.dart
srawlins: Great catch! Done.

test_data/rules/no_adjacent_strings_in_list.dart
srawlins: I was thinking it would be more confusing (like, really confusing) to have the test cases split between integration tests and reflective tests.

But I am really on board with moving to reflective tests ASAP.

srawlins:
Sorry for the delay.

Testing in flutter at https://dart-review.googlesource.com/c/sdk/+/277046

srawlins:
This passed flutter analysis. 🎉

test_data/rules/no_adjacent_strings_in_list.dart
bwilkerson: I'm fine with you not changing it, but the reason I said something is that we've been writing new tests as reflective tests so that we don't increase the technical debt in the process of writing new tests. I understand that having both styles is confusing, but I think that not increasing the technical debt is more important.

bwilkerson (approved):

PR #3743, opened 2022-10-03
migrate `avoid_returning_this` from `traverseNodesInDFS`

pq:
Minor improvement but one step closer to deprecating `traverseNodesInDFS`.

**BEFORE**

image

**AFTER**

image

/cc @bwilkerson

### DIFF
--- a/lib/src/rules/avoid_returning_this.dart
+++ b/lib/src/rules/avoid_returning_this.dart
@@ -43,8 +43,6 @@ var buffer = StringBuffer()
 
 ''';
 
-bool _isFunctionExpression(AstNode node) => node is FunctionExpression;
-
 bool _returnsThis(ReturnStatement node) => node.expression is ThisExpression;
 
 class AvoidReturningThis extends LintRule {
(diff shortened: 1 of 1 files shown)


lib/src/rules/avoid_returning_this.dart
```diff
@@ -63,6 +61,29 @@ class AvoidReturningThis extends LintRule {
   }
 }
 
+class _BodyVisitor extends UnifyingAstVisitor {
```
bwilkerson, line 64: I believe that this could be a subclass of `RecursiveAstVisitor`. The difference is that a `UnifyingAstVisitor` will invoke `visitNode` on every node in addition to the normal `visit` method, so there's an extra method invocation per node. Probably won't impact performance by a measurable amount, but it will also keep the code a bit cleaner.

bwilkerson (approved):

lib/src/rules/avoid_returning_this.dart
pq: Great catch.  Thanks!

PR #3747, opened 2022-10-04
migrate `prefer_constructors_over_static_methods` from `traverseNodesInDFS`

pq:
A modest gain in the common case.

**BEFORE**

image

**AFTER**

image

/cc @bwilkerson


bwilkerson (approved):
Consider adding blank lines between members in `_BodyVisitor`. Other than that, lgtm.

PR #3750, opened 2022-10-05
have the workflows default to read-only permissions

devoncarew:
- update the workflows to default to read-only permissions
- let issue-labeler.yml write to issues
- and labeler.yml write to pull requests


pq (approved):
Nice!

PR #3763, opened 2022-10-12
Add lint codes for the about half the recommended lints

bwilkerson:


lib/src/rules/library_private_types_in_public_api.dart
```diff
@@ -41,13 +41,22 @@ class _Private {}
 ''';
 
 class LibraryPrivateTypesInPublicAPI extends LintRule {
+  static const LintCode code = LintCode('library_private_types_in_public_api',
+      'Invalid use of a private type in a public API.',
+      correctionMessage:
+          'Try using a public supertype of the private type, or making the '
```
srawlins, line 47: I think the most common fix I've applied, in practice, is to actually change the definition of the type in the return type to be public. For example, a public method might return `List<_Foo>` and the `_Foo` is declared in the same library (obviously), and the fix is to rename `_Foo` to `Foo`.

lib/src/rules/throw_in_finally.dart
```diff
@@ -49,13 +49,20 @@ class BadThrow {
 ''';
 
 class ThrowInFinally extends LintRule {
+  static const LintCode code = LintCode(
+      'throw_in_finally', "Use of '{0}' in finally block.",
+      correctionMessage: "Try moving the '{0}' outside the finally block.");
```
srawlins, line 54: Above, in `control_flow_in_finally`, "finally" is quoted. e.g. "outside the 'finally' block."

srawlins (approved):

pq (approved):

lib/src/rules/library_private_types_in_public_api.dart
bwilkerson: I agree, that's a better suggestion. I've updated the message.

lib/src/rules/throw_in_finally.dart
bwilkerson: Done. Sometimes I'm not as consistent as I should be, so thanks for pointing this out.

PR #3770, opened 2022-10-17
Do not consider super formals as local variables

srawlins:
# Description

When considering the names of local variables, do not consider super formals to be local variables.

Fixes https://github.com/dart-lang/linter/issues/3769.


pq (commented):
Nice! 💎

PR #3773, opened 2022-10-18
No longer document pedantic rule set

parlough:
Pedantic has been discontinued for more than a year now, so I don't see a need to have tooling for it or list it as a rule set a linter rule is in.


pq (approved):

pq:
LGTM.  Is this ready to land as-is?

parlough:
Yeah it's good to go. I'll follow up with a separate PR to gh-pages

pq:
> I'll follow up with a separate PR to the gh-pages branch once this lands.

This isn't necessary, as of https://github.com/dart-lang/linter/commit/1c5e78f5f7d953cba61763f33c38f2b358d040e7 right?

parlough:
Thanks for the review and merging!

I was thinking about deleting the [pedantic svg](https://github.com/dart-lang/linter/blob/gh-pages/lints/style-pedantic.svg), but I think we can keep it for now incase anyone is still using it. Doesn't hurt to leave it there.

PR #3781, opened 2022-10-21
`prefer_const_constructors`: constant check re-ordering

pq:
Delay more expensive node `isConst` check.

A pretty minor bump but a few runs suggest some relative improvement (and objectively this shouldn't **_hurt_** and is no less clear).

**Before**

image

**After**

image


bwilkerson (approved):
Are the before and after screenshots reversed? It actually looks like both the timing and the ranking got worse.

pq:
Ranking went from 28 to 40, the "after" run was just slower all around.

Issue #3774, opened 2022-10-20
`flutter_style_todos` takes expensive time during checks


AlexV525:
**Describe the issue**
After #3618, `flutter_style_todos` will cost more time during checks. As discussed in https://github.com/dart-lang/linter/pull/3618#issuecomment-1231899870.

**Expected behavior**
Rewrite the lint rule to not use a regular expression, or improve the performance when checking with regular expressions.

cc @pq

pq:
@lrhn added some great analysis in https://github.com/dart-lang/linter/pull/3782#issuecomment-1287755497.  👍

AlexV525:
What will be our goal for improvements? I saw the time was reduced by ~3x.

PR #3782, opened 2022-10-21
`flutter_style_todos`: factor out sync*

pq:
Refactoring the regular expressions (#3774) should get us another win but this addresses the known slowness in `sync*` (https://github.com/dart-lang/sdk/issues/32102) and speeds this up ~3x.

**Before**

image

**After**

image

/cc @bwilkerson @srawlins

### DIFF
--- a/lib/src/rules/flutter_style_todos.dart
+++ b/lib/src/rules/flutter_style_todos.dart
@@ -52,25 +52,25 @@ class _Visitor extends SimpleAstVisitor<void> {
 
   _Visitor(this.rule);
 
+  void checkComments(Token token) {
+    Token? comment = token.precedingComments;
+    while (comment != null) {
+      _checkComment(comment);
+      comment = comment.next;
+    }
+  }
+
   @override
   void visitCompilationUnit(CompilationUnit node) {
     Token? token = node.beginToken;
     while (token != null) {
-      _getPrecedingComments(token).forEach(_visitComment);
+      checkComments(token);
       if (token == token.next) break;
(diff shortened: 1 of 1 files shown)


bwilkerson (approved):

lrhn:
Just saw your RegExps:
```dart
static final _todoRegExp = RegExp(r'//+(.* )?TODO\b', caseSensitive: false);

static final _todoExpectedRegExp =
      RegExp(r'// TODO\([a-zA-Z0-9][-a-zA-Z0-9]*\): ');
```

The `+` on `//`, since `.*` matches `/` as well, can cause quadratic backtracking if the line does not contain a `TODO`.
You don't need to capture in the `(.* )`, it could be `(?:. *)`.

Am I correct that you're just trying to check for a `TODO` on the same line as a comment (case-insensitive, even)?

Then I'd do:
```dart
static final _todoRegExp = RegExp(r'//.*?\bTODO\b', caseSensitive: false);
```
Or do we expect a false positive where the TODO is not preceded by a space?
(That is, is the space important in the original RegExp?)

This RegExp uses `.*?` to not consume the entire line first, then backtrack until it sees a TODO. If the TODOs are all valid, that's wasted time. It'll still consume the entire line if there is no TODO.

So, consider:
```dart
  static final _todoRegExp = RegExp(r'//.*?\bTODO\b', caseSensitive: false);
  static final _todoExpectedRegExp = RegExp(r'// TODO\([a-z\d][-a-z\d]*\): ');
```

PR #3785, opened 2022-10-24
migrate from sync* for super iteration

pq:
(WIP)

### DIFF
--- a/lib/src/rules/prefer_const_literals_to_create_immutables.dart
+++ b/lib/src/rules/prefer_const_literals_to_create_immutables.dart
@@ -88,30 +88,6 @@ class _Visitor extends SimpleAstVisitor<void> {
     _visitTypedLiteral(node);
   }
 
-  Iterable<InterfaceType> _getSelfAndInheritedTypes(InterfaceType type) sync* {
-    InterfaceType? current = type;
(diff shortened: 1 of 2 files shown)


lib/src/rules/prefer_const_literals_to_create_immutables.dart
```diff
@@ -135,4 +111,25 @@ class _Visitor extends SimpleAstVisitor<void> {
...
+  // todo(pq): consider making this a utility and sharing w/ `avoid_equals_and_hash_code_on_mutable_classes`
+  static bool _hasImmutableAnnotation(DartType? type) {
+    if (type is! InterfaceType) {
+      // This happens when we find an instance creation expression for a class
+      // that cannot be resolved.
+      return false;
+    }
+
+    InterfaceType? current = type;
+    // TODO(a14n) the is check looks unnecessary but prevents https://github.com/dart-lang/sdk/issues/33210
```
bwilkerson, line 124: I don't understand what `is` check this is referring to. If the comment is stale, consider updating or removing it.

pq: Good catch!  Removed.

pq:
@bwilkerson: could you take a look?

bwilkerson (approved):

PR #3791, opened 2022-10-26
Implement unnecessary_library_directive

kevmoo:
Closes https://github.com/dart-lang/linter/issues/3686


kevmoo:
ZERO clue if I'm holding this right!

Would also like to talk about naming. Should we drop `avoid_`?

CC @srawlins

lib/src/rules/avoid_library_directive.dart
```diff
@@ -0,0 +1,70 @@
...
+
+class _Visitor extends SimpleAstVisitor<void> {
+  final LintRule rule;
+
+  _Visitor(this.rule);
+
+  @override
+  void visitLibraryDirective(LibraryDirective node) {
+    if (node.documentationComment == null &&
+        node.sortedCommentAndAnnotations.isEmpty) {
```
bwilkerson: We should not report a lint for library directives in libraries that contain part directives.

lib/src/rules/avoid_library_directive.dart
```diff
@@ -0,0 +1,70 @@
...
+''';
+
+class AvoidLibraryDirective extends LintRule {
+  static const LintCode code = LintCode('avoid_library_directive',
+      'Library directives without comments or annotations should be avoided.',
+      correctionMessage: 'Try deleting the library directive.');
+
+  AvoidLibraryDirective()
+      : super(
+            name: 'avoid_library_directive',
```
bwilkerson: Yes, we want to not use `avoid` whenever possible. Consider `unnecessary_library_directive` instead.

lib/src/rules/avoid_library_directive.dart
kevmoo: How would one do that?

lib/src/rules/avoid_library_directive.dart
kevmoo: Done!

lib/src/rules/avoid_library_directive.dart
bwilkerson: The parent of the `LibraryDirective` is a `CompilationUnit`, and from there you can get to all the other `directives` to see whether there are any part directives.

If you only want to lint when at least one of the parts uses the library name, that will be a bit harder.

Also, as per the issue, we might want to not lint when the library directive includes a name (though I'm not sure whether we need to care about `dart:mirrors` uses at this point).

kevmoo: I don't care about the mirrors case. Folks can ignore the lint there. Might be worth adding a note to the description, though.

kevmoo: Hrm. We have 2 options.

1. Leave as is and document the part case – maybe push folks to drop named parts.
2. Be safe and just skip this lint for libraries with parts – which are pretty rare.

Otherwise we have to go digging through the analyzer context (or whatever) to find the corresponding part and see if it uses names, etc etc – right?

lib/src/rules/unnecessary_library_directive.dart
```diff
@@ -0,0 +1,76 @@
+// Copyright (c) 2015, the Dart project authors. Please see the AUTHORS file
```
pq, line 1: => 2022

test_data/rules/unnecessary_library_directive.dart
```diff
@@ -0,0 +1,7 @@
+// Copyright (c) 2020, the Dart project authors. Please see the AUTHORS file
```
pq, line 1: => 2022

test/integration/unnecessary_library_directive.dart
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) 2020, the Dart project authors. Please see the AUTHORS file
```
pq, line 1: => 2022

lib/src/rules/unnecessary_library_directive.dart
```diff
@@ -0,0 +1,76 @@
...
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+
+import '../analyzer.dart';
+
+const _name = 'unnecessary_library_directive';
+
+const _desc =
+    'Avoid library directives unless that have documentation comments or '
```
pq, line 13: that

  =>

they

lib/src/rules/unnecessary_library_directive.dart
kevmoo: You can use the suggest change feature, too – very nice!

lib/src/rules/avoid_library_directive.dart
kevmoo: It wouldn't be for all code generators – just code generators that use the old part syntax. Which I think is mostly dead!

example/all.yaml
```diff
@@ -173,6 +173,7 @@ linter:
     - unnecessary_getters_setters
     - unnecessary_lambdas
     - unnecessary_late
+    - unnecessary_library_directive
```
pq, line 176: Could we rename this to `unnecessary_library_directives`?  The plural seems more consistent w/ other similar lints.

pq (changes requested):

example/all.yaml
kevmoo: Meh...we have a lot of singular ones up there to ☝️ – since you can only have one `library` directive in a...`library` going simple seems better to me. Happy to discuss more – or curious about others opinions...

pq (approved):
👍

kevmoo:
I'm going to move this to "draft" and get that exception for parts in per @bwilkerson

lib/src/rules/avoid_library_directive.dart
kevmoo: done!

lib/src/rules/unnecessary_library_directive.dart
```diff
@@ -0,0 +1,86 @@
...
+class _Visitor extends SimpleAstVisitor<void> {
+  final LintRule rule;
+
+  _Visitor(this.rule);
+
+  @override
+  void visitLibraryDirective(LibraryDirective node) {
+    var parent = node.parent! as CompilationUnit;
+    if (parent.directives.any((element) => element is PartDirective)) {
+      // Parts may still use library names. No be safe, we don't lint those
```
bwilkerson, line 77: nit: "No be safe" --> "To be safe"

PR #3796, opened 2022-10-27
New rule: dangling_library_doc_comments

srawlins:
# Description

Fixes https://github.com/dart-lang/linter/issues/3789

The issue states the intent of the rule well, but the devil is, as usual, in the details.

I have implemented this rule specifically to use a heuristic of which doc comments _might_ be trying to document a library. Here are the criteria:

* `///`-style comments only.
* comments above the first directive, unless the first directive is a `library` or `part` directive, so doc comments mixed _within_ directives (imports, exports) are not considered.
* comments above the first declaration, which are not physically attached to the declaration; that is, there exists at least one blank line between the doc comment and the first non-comment token of the declaration. The tests cover this case well, as it is tricky to get right.
* comments at the end of a library, in which there are no declarations. This should cover public libraries whose function is to export private libraries; they may contain no declarations, and may contain a comment intended to be a library doc comment.

### DIFF
--- a/lib/src/rules/dangling_library_doc_comments.dart
+++ b/lib/src/rules/dangling_library_doc_comments.dart
@@ -86,8 +86,9 @@ class _Visitor extends SimpleAstVisitor<void> {
         return;
       }
 
-      if (firstDirective.documentationComment != null) {
-        rule.reportLint(firstDirective.documentationComment);
+      var docComment = firstDirective.documentationComment;
+      if (docComment != null) {
+        rule.reportLintForToken(docComment.beginToken);
         return;
       }
 
--- a/lib/src/rules/dangling_library_doc_comments.dart
+++ b/lib/src/rules/dangling_library_doc_comments.dart
@@ -90,6 +90,8 @@ class _Visitor extends SimpleAstVisitor<void> {
         rule.reportLint(firstDirective.documentationComment);
         return;
       }
+
+      return;
     }
 
     if (node.declarations.isEmpty) {
--- a/lib/src/rules/dangling_library_doc_comments.dart
+++ b/lib/src/rules/dangling_library_doc_comments.dart
@@ -3,6 +3,7 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/token.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
 
 import '../analyzer.dart';
@@ -36,14 +37,15 @@ class C {}
 ```
 
 **NOTE:** An unnamed library, like `library;` above, is only supported in Dart
-2.19 and later. Code which might run in earlier versions of Dart need to provide
-a name in the `library` directive.
+2.19 and later. Code which might run in earlier versions of Dart will need to
+provide a name in the `library` directive.
 ''';
 
 class DanglingLibraryDocComments extends LintRule {
   static const LintCode code = LintCode(
       'dangling_library_doc_comments', 'Dangling library doc comment.',
-      correctionMessage: 'Attach library doc comments to library directives.');
+      correctionMessage:
+          "Add a 'library' directive after the library comment.");
 
   DanglingLibraryDocComments()
       : super(
@@ -93,9 +95,12 @@ class _Visitor extends SimpleAstVisitor<void> {
     if (node.declarations.isEmpty) {
       // Without any declarations, we only need to check for a doc comment as
       // the last thing in a file.
-      var endComment = node.endToken.precedingComments;
-      if (endComment is DocumentationCommentToken) {
-        rule.reportLintForToken(endComment);
+      Token? endComment = node.endToken.precedingComments;
+      while (endComment is CommentToken) {
+        if (endComment is DocumentationCommentToken) {
+          rule.reportLintForToken(endComment);
+        }
+        endComment = endComment.next;
       }
       return;
     }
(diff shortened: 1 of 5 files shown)


srawlins:
Ready for review! :D

kevmoo:
What about dangling annotations? I'd love both! We should split out that issue since it covers both!

srawlins:
Hold your horses. https://github.com/dart-lang/linter/issues/3790

kevmoo:
Bless you, sir!

lib/src/rules/dangling_library_doc_comments.dart
```diff
@@ -0,0 +1,150 @@
...
+/// This is a great library.
+library;
+
+import 'package:math';
+
+class C {}
+```
+
+**NOTE:** An unnamed library, like `library;` above, is only supported in Dart
+2.19 and later. Code which might run in earlier versions of Dart need to provide
```
pq, line 39: => 

    Code which might run in earlier versions of Dart will need to provide

?

pq (changes requested):

lib/src/rules/dangling_library_doc_comments.dart
```diff
@@ -0,0 +1,150 @@
...
+a name in the `library` directive.
+''';
+
+class DanglingLibraryDocComments extends LintRule {
+  static const LintCode code = LintCode(
+      'dangling_library_doc_comments', 'Dangling library doc comment.',
+      correctionMessage: 'Attach library doc comments to library directives.');
```
bwilkerson, line 46: Perhaps "Add a library directive after the library comment.". I'm concerned that some users might be confused about how to attach the comment to a library directive that they don't have.

lib/src/rules/dangling_library_doc_comments.dart
```diff
@@ -0,0 +1,150 @@
...
+        // comments in the library.
+        return;
+      }
+      if (firstDirective is PartOfDirective) {
+        // Don't worry about possible "library doc comments" in a part.
+        return;
+      }
+
+      if (firstDirective.documentationComment != null) {
+        rule.reportLint(firstDirective.documentationComment);
```
bwilkerson, line 88: I wonder whether it makes sense to highlight something smaller than the whole comment. Comments can sometimes be very long.

lib/src/rules/dangling_library_doc_comments.dart
```diff
@@ -0,0 +1,150 @@
...
+        return;
+      }
+    }
+
+    if (node.declarations.isEmpty) {
```
bwilkerson, line 93: I think this needs to be an else-if so that we don't get here if there are directives.

lib/src/rules/dangling_library_doc_comments.dart
```diff
@@ -0,0 +1,150 @@
...
+      // Without any declarations, we only need to check for a doc comment as
+      // the last thing in a file.
+      var endComment = node.endToken.precedingComments;
+      if (endComment is DocumentationCommentToken) {
+        rule.reportLintForToken(endComment);
```
bwilkerson, line 98: Note that this only highlights the first line of the doc comment, which is inconsistent with the above (though possibly better).

It also doesn't handle the case where there's a non-doc comment before the doc comments. Consider
```
// Copyright xxxx

/// Library comment
```

lib/src/rules/dangling_library_doc_comments.dart
```diff
@@ -0,0 +1,150 @@
...
+    if (docComment == null) {
+      return;
+    }
+    var lineInfo = node.lineInfo;
+
+    // We must walk through the comments following the doc comment, tracking
+    // pairs of consecutive comments so as to check whether any two are
+    // separated by a blank line.
+    var commentToken = docComment.endToken;
+    var followingCommentToken = docComment.endToken.next;
```
bwilkerson, line 114: "docComment.endToken" --> "commentToken"

test/rules/dangling_library_doc_comments_test.dart
```diff
@@ -0,0 +1,127 @@
...
+    defineReflectiveTests(DanglingLibraryDocCommentsTest);
+  });
+}
+
+@reflectiveTest
+class DanglingLibraryDocCommentsTest extends LintRuleTest {
+  @override
+  String get lintRule => 'dangling_library_doc_comments';
+
+  test_doc_comment_above_declaration() async {
```
bwilkerson, line 20: These names don't following the convention for test names. See https://github.com/dart-lang/sdk/blob/main/pkg/analyzer/doc/implementation/tests.md#test-names.

test/rules/dangling_library_doc_comments_test.dart
```diff
@@ -0,0 +1,127 @@
...
+import '../rule_test_support.dart';
+
+main() {
+  defineReflectiveSuite(() {
+    defineReflectiveTests(DanglingLibraryDocCommentsTest);
+  });
+}
+
+@reflectiveTest
+class DanglingLibraryDocCommentsTest extends LintRuleTest {
```
bwilkerson, line 16: I'll suggest one additional test (besides the ones that might be implied by other comments):
```
/// Library comment

/// Class comment
class C {}
```

I think there should be a lint here but suspect that there won't be because I don't think we look above the class' documentation comment.

lib/src/rules/dangling_library_doc_comments.dart
srawlins: Hmm, I don't think so. I can see someone writing a supposed library dart doc between the directives and the declarations.

test/rules/dangling_library_doc_comments_test.dart
srawlins: Oops, I copied `discared_futures_test`. Fixed. I kept `docComment_` in each name but let me know if you don't like it. I think it improves readability.

lib/src/rules/dangling_library_doc_comments.dart
bwilkerson: Given

```
import 'dart:async';

/// Comment

class C {}
```

I think that we shouldn't lint the doc comment. Neither analyzer nor dartdoc would have previously interpreted this as the doc comment for the library, so I don't think this lint should either.

lib/src/rules/dangling_library_doc_comments.dart
srawlins: Should I just highlight the first line? I'd be in favor of that. Is there a good practice for getting the offsets of the first line?

lib/src/rules/dangling_library_doc_comments.dart
srawlins: Done.

lib/src/rules/dangling_library_doc_comments.dart
srawlins: Done.

lib/src/rules/dangling_library_doc_comments.dart
srawlins: Good catch. Fixed.

lib/src/rules/dangling_library_doc_comments.dart
srawlins: Done.

test/rules/dangling_library_doc_comments_test.dart
srawlins: Ah ha good catch. Done.

lib/src/rules/dangling_library_doc_comments.dart
srawlins: That's too bad. I'm sure will miss some cases. Done, in any case.

lib/src/rules/dangling_library_doc_comments.dart
bwilkerson: You could use the offset and length of the token for the first line. If you want more, `LineInfo` has `getOffsetOfLine` and `getOffsetOfLineAfter` (though the implementation of the latter looks wrong to me; I might be too tired, though).

srawlins: OK I've updated it to use the beginToken.

srawlins:
@pq ready for re-review

pq (approved):
💎

Issue #3798, opened 2022-10-28
proposal: `sort_public_methods_before_private_methods`


ju851zel:
## Description

Define public methods before private methods.

## Details

**DO** define public methods before private methods.

This makes the code uniform across multiple classes and it makes it faster to find specific methods in a class.
Because public methods are the interface to the class they are (often) more important and should come before private methods. 

This would be my first approach to have a defined order for different parts in a class.
[Flutter style guide ordering of class members](https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#order-other-class-members-in-a-way-that-makes-sense)

## Kind

Style

## Good Examples 

```dart
class A {
  int a() => 0;
  int c() => 0;
  int _b() => 0;
}
```

## Bad Examples

```dart
class A {
  int a() => 0;
  int _b() => 0;
  int c() => 0;
}
```

[Writing Lints]: https://github.com/dart-lang/linter/blob/main/doc/writing-lints.md
[Effective Dart]: https://dart.dev/guides/language/effective-dart
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[SDK Tracker]: https://github.com/dart-lang/sdk/issues

bwilkerson:
I made a couple of comments on the PR, but we should really discuss the semantics (and desirability) of the lint before getting too far into the implementation issues.

Do we want to have a rule that enforces part of the sort order, or would it be better to have a rule that enforces a more complete sort order (such as the one produced by the "Sort Members" feature)? We do already have a couple of rules about the order of constructors, but we need to consider interactions between all such rules, and the more rules we have the harder it will be to manage the interactions. We could, for example, have a rule to enforce that "Sort Members" has been run over the code.

If we do want a more limited rule like this, how does this relate to different kinds of members? The "Sort Members" feature, for example produces the following order:
- public static fields
- private static fields
- public instance fields
- private instance fields
- public constructors
- private constructors
- public instance methods
- private instance methods
- public static methods
- private static methods

Public members of a given kind are always before private members of the same kind, but the kinds are kept separate. Does this rule propose that private members of any kind always be after public members of every kind?

srawlins:
Possible duplicate of https://github.com/dart-lang/linter/issues/2367

srawlins:
[...]

PR #3799, opened 2022-10-28
Sort public methods before private methods

ju851zel:
# New linter rule

I created a new linter rule which, when enabled, makes public methods appear before private methods.

My intention was to have a first step in more concise code.
Public methods should be at the top because they define the interface to the class.

Private methods are (most of the times) less important and therefore they should come below public methods.

[Link to Issue](https://github.com/dart-lang/linter/issues/3798)


lib/src/rules/sort_public_methods_before_private_methods.dart
```diff
@@ -0,0 +1,82 @@
...
+  _Visitor(this.rule);
+
+  void check(NodeList<ClassMember> members) {
+    bool privateMethodIsBeforePublicMethod = false;
+    // Members are sorted by source position in the AST.
+    for (var member in members) {
+      if (member is MethodDeclaration) {
+        if (privateMethodIsBeforePublicMethod &&
+            !_isPrivateName(member: member)) {
+          rule.reportLint(member.returnType);
```
bwilkerson, line 66: Do we want to report this lint for _every_ public member after a private member, or just the first?

Personally, I suspect that it's sufficient to report it for a single member.

Do we want to report this lint for public members or for private members? (That is, which does this lint consider to be in the wrong place?) The original name implies that the private members are in the right place and that it's the public members that need to be moved. The alternative I suggested (which might not be right) implies that the public members are right and that it's the private members that need to be moved.

lib/src/rules/sort_public_methods_before_private_methods.dart
```diff
@@ -0,0 +1,82 @@
...
+            name: 'sort_public_methods_before_private_methods',
+            description: _desc,
+            details: _details,
+            group: Group.style);
+
+  @override
+  void registerNodeProcessors(
+      NodeLintRegistry registry, LinterContext context) {
+    var visitor = _Visitor(this);
+    registry.addClassDeclaration(this, visitor);
```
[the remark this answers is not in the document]
ju851zel: Yes I think this is a good idea.
I will add this for enums, mixins and extensions.

lib/src/rules/sort_public_methods_before_private_methods.dart
```diff
@@ -0,0 +1,82 @@
...
+  }
+}
+
+class _Visitor extends SimpleAstVisitor {
+  final LintRule rule;
+
+  _Visitor(this.rule);
+
+  void check(NodeList<ClassMember> members) {
+    bool privateMethodIsBeforePublicMethod = false;
```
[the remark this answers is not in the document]
ju851zel: You are right. The name `foundPrivateMember` reflect that better, so I'm changing it to that one.

lib/src/rules/sort_public_methods_before_private_methods.dart
ju851zel: As far as I know other rules report a lint suggesting for every occurrence.
E.g. the rule `require_trailing_commas` tells me all the places where a comma is missing.
So for being consistent over multiple rules, I would change it to display all the occurrences simultaneously.
Also e.g. in Android Studio/IntelliJ the shortcut that moves the cursor to the next error would be faster if all the occurrences are mentioned at the same time, in comparison to requiring the analyzer to run again and then mention the next occurrence.

So if you are fine with that, I would try to implement to show all the occurrences at simultaneously.

lib/src/rules/sort_public_methods_before_private_methods.dart
```diff
@@ -0,0 +1,82 @@
...
+  int _b() => 0;
+}
+```
+
+''';
+
+class SortPublicMethodsBeforePrivateMethods extends LintRule {
+  SortPublicMethodsBeforePrivateMethods()
+      : super(
+            name: 'sort_public_methods_before_private_methods',
```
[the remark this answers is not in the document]
ju851zel: Could you please elaborate why this also applies for other members?
As I understand it in line 66 `if (member is MethodDeclaration) {` this only applies for method declarations and not for other declarations, right?.
But I may have also missed something, because this is my first attempt to writing a lint rule.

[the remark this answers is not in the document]
ju851zel: For example this piece gets not complained about.
```
class A {
  final _y = 0;
  final x = 0;
}
```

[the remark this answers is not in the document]
ju851zel: Regarding the naming I like the idea to shorten the name.
If this only applies to methods I would suggest `sort_private_methods_last`, when this applies to all members your suggestion looks good.

I will change the name when we decided if this only applies to methods or all members.

lib/src/rules/sort_public_methods_before_private_methods.dart
ju851zel: I'm open for suggestions if this lint should report for private or public members.
Personally I would not prefer one over the other, because when the lint is resolved, the result is the same.
What would you suggest?

ju851zel: Regarding the report of multiple occurrences at one, this should already work.
I added a test to confirm that `test_multiple_cases_are_reported_simultaneously`.

lib/src/rules/sort_public_methods_before_private_methods.dart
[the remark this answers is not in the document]
bwilkerson: You're right, that comment was outdated. I hadn't noticed the test for method declarations on the first read through, so when I went back to start commenting I had the wrong understanding, and forgot to update the comment when I did see the test. Sorry for the noise.

lib/src/rules/sort_public_methods_before_private_methods.dart
bwilkerson: > As far as I know other rules report a lint suggesting for every occurrence.

Not quite, though I can see why it would seem that way.

The criteria we use is that every issue should have a single diagnostic. If there's a missing comma in one invocation, that's one issue and should have one diagnostic. If there's a missing comma in a second invocation, then that's a separate issue and should also have a diagnostic. Two separate issues, so two diagnostics.

But in this case it seems to me that the fact that the methods aren't sorted correctly is a single issue. For example, the code
```
class C {
  void m1() {}
  void _m2() {}
  void m3() {}
  void m4() {}
  void m5() {}
  void _m6() {}
}
```
Looks to me like someone added a private method (`m2`) in the wrong place. But if the lint reports that `m3`, `m4`, and `m5` are in the wrong place, that seems like noise to me. I'd much rather be told once that the methods aren't sorted correctly and then run "Sort Members" (or some other fix) to have it fix all the problems in this container.

It's possible that I'm the only one that feels that way.

> I'm open for suggestions if this lint should report for private or public members.

I don't know which is better. I believe that it depends on the failure mode that most users experience, which is data I don't have. I wondered whether you had an informed decision, and it sounds like you don't. It might not matter too much, but I tend to think about such things.

lib/src/rules/sort_public_methods_before_private_methods.dart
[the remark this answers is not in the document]
ju851zel: I renamed the rule to `sort_private_methods_last`.

lib/src/rules/sort_public_methods_before_private_methods.dart
ju851zel: I know understand what you mean and I think this is a better solution than linting every occurrence.
The linter should complain once per file to move the private methods below the public ones and a quick fix that does that would be a good solution.
I will try to implement that.

ju851zel: A quick fix for that is not implemented in the linter package right?
I guess this is implemented in the dart sdk which cannot simply be extended by a merge request?
Or am I wrong?

bwilkerson: I'd hold off on doing more implementation work until we have gotten more feedback on the questions in the issue. We're being cautious about adding new stylistic lints and would like to gauge the reactions of the community before merging this PR.

> The linter should complain once per file to move the private methods ...

I suspect we'll end up with one lint per class, enum, etc.

> A quick fix for that is not implemented in the linter package right?
> I guess this is implemented in the dart sdk which cannot simply be extended by a merge request?

Yes, quick fixes are implemented in the `analysis_server` package. There's documentation for how to do that (https://github.com/dart-lang/sdk/blob/main/pkg/analysis_server/doc/implementation/quick_fix.md).

You can create a PR against the sdk repo and it will automatically converted into a Gerrit CL. If you look at the workflow steps in GitHub there should be a link to the CL. You won't have permissions to do much with the CL, but you can push new branches to update your PR and they'll also automatically get copied across.

But I'd also hold off on implementing a fix for a couple of reasons:
1. We already have a "Sort Members" function that could act as a fix (although it does more that might be expected from a fix for this lint). It isn't clear that we want to support yet another sorting feature.
2. Once this PR is merged (assuming it is) it can take a week or more for those changes to roll into the sdk, and you won't be able to test the fix until the lint is in the sdk.

ju851zel:
@bwilkerson I also added a check so that static methods are ignored in this rule.
I think there should be a separate rule that handles the case where static methods in a class should be defined.
Like mentioned in the [Flutter style guide](https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#order-other-class-members-in-a-way-that-makes-sense).

lib/src/rules/sort_public_methods_before_private_methods.dart
ju851zel: Alright, thanks for the information.

If the community likes this rule I would also try to create a PR in the sdk (assuming this one is merged).
I'm also looking forward for adding more rules if this is something the community would appreciate.

I would really like rules that sort members etc. like described in the [Flutter style guide](https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#order-other-class-members-in-a-way-that-makes-sense).
So that in the end all of the mentioned rules for ordering can be automatically applied to files (if wished).

But I will wait some time and come back to this PR when more from the community have seen my proposal.
Thanks a lot for the help and feedback.

bwilkerson:
> Like mentioned in the Flutter style guide.

I'll just mention that with very few exceptions, we don't support lints related to the Flutter style guide. Those are styles established for working in the `flutter` repository. Instead we support lints related to the [Dart style guide](https://dart.dev/guides/language/effective-dart). These are the guidelines recommended for Dart and Flutter users (and Dart developers). The Dart style guide specifies the sort order for directives, but doesn't specify any sort order for declarations.

ju851zel:
I will close this as linter rules related to the flutter style guide are not supported.

PR #3803, opened 2022-10-31
New rule: library_annotations

srawlins:
# Description

Report on annotations which should be attached to a library directive.

Basically this is just `@pragma('dart2js:late:trust')` and all `TargetKind.library` directives that are currently allowed by BestPracticesVisitor.

Fixes https://github.com/dart-lang/linter/issues/3790

### DIFF
--- a/lib/src/rules/library_annotations.dart
+++ b/lib/src/rules/library_annotations.dart
@@ -1,0 +1,156 @@
+// Copyright (c) 2022, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/element.dart';
+// ignore: implementation_imports
+import 'package:analyzer/src/dart/element/extensions.dart';
+import 'package:meta/meta_meta.dart';
+
+import '../analyzer.dart';
+
+const _desc = r'Attach library annotations to library directives.';
+
+const _details = r'''
+Attach library annotations to library directives, rather than
+some other library-level element.
+
+**BAD:**
+```dart
+import 'package:test/test.dart';
+
+@TestOn('browser')
+void main() {}
+```
+
+**GOOD:**
+```dart
+@TestOn('browser')
+library;
+import 'package:test/test.dart';
+
+void main() {}
+```
+
+**NOTE:** An unnamed library, like `library;` above, is only supported in Dart
+2.19 and later. Code which might run in earlier versions of Dart will need to
+provide a name in the `library` directive.
+''';
+
+class LibraryAnnotations extends LintRule {
+  static const LintCode code = LintCode('library_annotations',
+      'This annotation must be attached to a library directive.',
+      correctionMessage: 'Attach library annotations to library directives.');
+
+  LibraryAnnotations()
+      : super(
+            name: 'library_annotations',
+            description: _desc,
+            details: _details,
+            group: Group.style);
+
+  @override
+  LintCode get lintCode => code;
+
+  @override
+  void registerNodeProcessors(
+      NodeLintRegistry registry, LinterContext context) {
+    var visitor = _Visitor(this);
+    registry.addClassDeclaration(this, visitor);
+    registry.addClassTypeAlias(this, visitor);
+    registry.addEnumDeclaration(this, visitor);
+    registry.addExportDirective(this, visitor);
+    registry.addExtensionDeclaration(this, visitor);
+    registry.addFunctionTypeAlias(this, visitor);
+    registry.addFunctionDeclaration(this, visitor);
+    registry.addGenericTypeAlias(this, visitor);
+    registry.addImportDirective(this, visitor);
+    registry.addMixinDeclaration(this, visitor);
+    registry.addTopLevelVariableDeclaration(this, visitor);
+  }
+}
+
+class _Visitor extends SimpleAstVisitor<void> {
+  final LibraryAnnotations rule;
+
+  _Visitor(this.rule);
+
+  @override
+  void visitClassDeclaration(ClassDeclaration node) => _check(node);
+
+  @override
+  void visitClassTypeAlias(ClassTypeAlias node) => _check(node);
+
+  @override
+  void visitEnumDeclaration(EnumDeclaration node) => _check(node);
+
+  @override
+  void visitExportDirective(ExportDirective node) => _check(node);
+
+  @override
+  void visitExtensionDeclaration(ExtensionDeclaration node) => _check(node);
+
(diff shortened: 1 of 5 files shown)


lib/src/rules/library_annotations.dart
```diff
@@ -0,0 +1,156 @@
...
+
+**NOTE:** An unnamed library, like `library;` above, is only supported in Dart
+2.19 and later. Code which might run in earlier versions of Dart will need to
+provide a name in the `library` directive.
+''';
+
+class LibraryAnnotations extends LintRule {
+  static const LintCode code = LintCode('library_annotations',
+      'This annotation must be attached to a library directive.',
+      correctionMessage: 'Attach library annotations to library directives.');
```
bwilkerson, line 45: nit: All of our correction messages are phrased as suggestions rather than a command. Consider "Try attaching the annotation to a library directive."

lib/src/rules/library_annotations.dart
```diff
@@ -0,0 +1,156 @@
...
+    registry.addClassTypeAlias(this, visitor);
+    registry.addEnumDeclaration(this, visitor);
+    registry.addExportDirective(this, visitor);
+    registry.addExtensionDeclaration(this, visitor);
+    registry.addFunctionTypeAlias(this, visitor);
+    registry.addFunctionDeclaration(this, visitor);
+    registry.addGenericTypeAlias(this, visitor);
+    registry.addImportDirective(this, visitor);
+    registry.addMixinDeclaration(this, visitor);
+    registry.addTopLevelVariableDeclaration(this, visitor);
```
bwilkerson, line 71: Consider visiting the `CompilationUnit` instead. I believe that there will be other diagnostics if the annotation isn't on the first member of the unit (directive or declaration), so we should only need to check the first member's annotations.

We might also want to bail if the unit is a part because there should also be diagnostics produced elsewhere in those cases.

lib/src/rules/library_annotations.dart
```diff
@@ -0,0 +1,156 @@
...
+      _check(node);
+
+  void _check(AnnotatedNode node) {
+    for (var annotation in node.metadata) {
+      var elementAnnotation = annotation.elementAnnotation;
+      if (elementAnnotation == null) {
+        return;
+      }
+
+      if (elementAnnotation.targetKinds.contains(TargetKind.library) &&
```
bwilkerson, line 121: This should check that the only target kind is `library`. If an annotation were allowed to be on either a library, class, or method, we wouldn't want to flag it here just because it's on the first class.

lib/src/rules/library_annotations.dart
```diff
@@ -0,0 +1,156 @@
...
+          (node.parent as CompilationUnit?)?.directives.first == node) {
+        rule.reportLint(annotation);
+        return;
```
bwilkerson, line 124: I think we want to find all of the annotations that should be on the library directive, not the first one, so we shouldn't return either here or on line 129.

lib/src/rules/library_annotations.dart
srawlins: Thanks, done!

lib/src/rules/library_annotations.dart
srawlins: Good catch! Done

lib/src/rules/library_annotations.dart
srawlins: Done

lib/src/rules/library_annotations.dart
srawlins: Done, and I agree about the part file. Done.

srawlins:
This is ready for re-review, @bwilkerson

lib/src/rules/library_annotations.dart
```diff
@@ -0,0 +1,125 @@
...
+      var elementAnnotation = annotation.elementAnnotation;
+      if (elementAnnotation == null) {
+        return;
+      }
+
+      if (elementAnnotation.targetKinds.length == 1 &&
+          elementAnnotation.targetKinds.contains(TargetKind.library) &&
+          (node.parent as CompilationUnit?)?.directives.first == node) {
+        rule.reportLint(annotation);
+        continue;
```
bwilkerson, line 96: Does the `continue` do anything? I think it can be removed.

test/rules/library_annotations_test.dart
```diff
@@ -0,0 +1,139 @@
...
+    await assertDiagnostics(
+      r'''
+@pragma('dart2js:late:trust')
+typedef Fn = void Function();
+''',
+      [lint(0, 29)],
+    );
+  }
+
+  test_importDirective() async {
```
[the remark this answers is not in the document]
srawlins: I cannot add a test case (I think) where the library_annotations lint is fired. If a `TargetKind.library` annotation is on the first directive, then the `invalid_annotation_target` Hint will not fire, but this Lint will fire.

If such an annotation is on the first declaration, then the `invalid_annotation_target` Hint _will_ fire, but this Lint _will not_ (to avoid double reporting.

So, I've added a test indicating what I wrote above.

lib/src/rules/library_annotations.dart
```diff
@@ -0,0 +1,125 @@
...
+  void _check(AnnotatedNode node) {
+    for (var annotation in node.metadata) {
+      var elementAnnotation = annotation.elementAnnotation;
+      if (elementAnnotation == null) {
+        return;
+      }
+
+      if (elementAnnotation.targetKinds.length == 1 &&
+          elementAnnotation.targetKinds.contains(TargetKind.library) &&
+          (node.parent as CompilationUnit?)?.directives.first == node) {
```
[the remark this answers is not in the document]
srawlins: True, but if there are no directives, then no annotation which has `TargetKind.library` can exist in the file. In any case, I've fixed to be defensive.

> It looks like the code is ignoring library directives everywhere except on the first directive. Should it also check the first declaration in the case where there are no directives?

Sorry, I don't follow. Can you explain more?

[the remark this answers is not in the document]
bwilkerson: > ... if there are no directives, then no annotation which has `TargetKind.library` can exist in the file.

Oh, I always thought that we allowed library annotations / comments on the first declaration when there were no directives. Of course, we're heading away from allowing them anywhere other than on a `library` directive, so it's probably not a big deal.

> Sorry, I don't follow. Can you explain more?

That might be because of a typo. It should have read

> It looks like the code is ignoring library-specific annotations everywhere except on the first directive.

That comment was also predicated on the assumption that they were accepted on the first declaration, so if that's not the case then the comment is moot.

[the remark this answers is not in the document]
srawlins: The exception in best_practices_verifier only allows a `TargetKind.library` annotation to be placed on the first directive, not on the first declaration:

https://github.com/dart-lang/sdk/blob/main/pkg/analyzer/lib/src/error/best_practices_verifier.dart#L1749

So I mirrored that here.

PR #3824, opened 2022-11-08
Properly match the types in the good versus bad prefer_collection_literals examples

parlough:


pq (approved):
💎

pq:
LGTM.  Holding merges briefly as we see if the last release sticks internally but this is low risk so we can land shortly I'm sure. Thanks!

PR #3831, opened 2022-11-10
Define a lint code for package_names

bwilkerson:
This code was commented out pending the publishing of a later version of the analyzer. That version has been published and moved to, so this code can now be uncommented.

There's no rush to merge this.


pq (approved):
👍

Issue #3825, opened 2022-11-09
Clarify text for unnecessary_final lint


timsneath:
The [`unnecessary_final` lint](https://dart-lang.github.io/linter/lints/unnecessary_final.html) contains a controversial explanation of its rationale.

Specifically:

> var is shorter, and final does not change the meaning of the code.

The historical context is here: https://github.com/dart-lang/linter/pull/1827

While there are clearly divergent preferences around the use of `final`, it _does_ change the meaning of the code (as [the `prefer_final_locals` lint](https://dart-lang.github.io/linter/lints/prefer_final_locals.html) explains). And while `var` is shorter by two characters, that seems like a poor rationale for a lint. Readability is preferred over terseness, anyway. 

Lastly, Effective Dart has [settled case law on the subject](https://dart.dev/guides/language/effective-dart/design#prefer-making-fields-and-top-level-variables-final): 

> State that is not mutable—that does not change over time—is easier for programmers to reason about. Classes and libraries that minimize the amount of mutable state they work with tend to be easier to maintain. Of course, it is often useful to have mutable data. But, if you don’t need it, your default should be to make fields and top-level variables final when you can.

While we don't need to remove the rule, we should have explanatory text that is more accurate. For example, "some people prefer the simplicity of using `var` for all variables."

MelbourneDeveloper:
Thanks @timsneath. 

There is nothing wrong with the lint itself. I understand why some teams may want to embrace this. The issue is the documentation, as you say. 

> var is shorter, and final does not change the meaning of the code.

I looked at this lint and was very confused. This isn't a reason to prefer `var` over `final` but it's also just plain wrong.  

Some teams may want to reassign variables as a matter of preference. For the record, I think that something like this would be better.

> You should make assignments to existing variables where possible to avoid creating multiple variables for the same logical value

and perhaps 

> `var` allows for reassignments while final means your code may end up with multiple variable names for the same logical value.

davidmorgan:
Hi Tim,

Thanks :) this is a fun one.

The Dart team--particularly Bob and Lasse, and they persuaded me, which is why I added this lint--prefer and recommend using `var`for locals. Nobody else on the team has a strong opinion as far as I'm aware.

The `prefer_final_locals` lint description is wrong: the compilers definitely know whether local variables are reassigned, `final` does not help.

[...]

PR #3832, opened 2022-11-10
Update unnecessary_final.dart

mit-mit:
Clarify this this is an opinionated lint, that there is a semantic difference between `final` and `var`, and what the alternatives are.

Fixes https://github.com/dart-lang/linter/issues/3825

### DIFF
--- a/lib/src/rules/unnecessary_final.dart
+++ b/lib/src/rules/unnecessary_final.dart
@@ -10,9 +10,13 @@ import '../analyzer.dart';
 const _desc = "Don't use `final` for local variables.";
 
 const _details = r'''
-**DON'T** use `final` for local variables.
+Use `var`, not `final`, when declaring local variables.
 
-`var` is shorter, and `final` does not change the meaning of the code.
+Per Effective Dart, there are two styles in wide use. This rule
+enforces the `var` style. For the alternative style that prefers `final`,
+enable `prefer_final_locals` and `prefer_final_in_for_each` instead.
+
+For fields, final is always recommended; see the rule `prefer_final_fields`.
 
 **BAD:**
 ```dart
--- a/lib/src/rules/unnecessary_final.dart
+++ b/lib/src/rules/unnecessary_final.dart
@@ -12,9 +12,10 @@ const _desc = "Don't use `final` for local variables.";
 const _details = r'''
 Use `var`, not `final`, when declaring local variables.
 
-Per Effective Dart, there are two styles in wide use. This rule
-enforces the `var` style. For the alternative style that prefers `final`,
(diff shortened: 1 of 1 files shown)


lib/src/rules/unnecessary_final.dart
```diff
@@ -10,9 +10,13 @@ import '../analyzer.dart';
 const _desc = "Don't use `final` for local variables.";
 
 const _details = r'''
-**DON'T** use `final` for local variables.
+Use `var`, not `final`, when declaring local variables.
 
-`var` is shorter, and `final` does not change the meaning of the code.
+Per Effective Dart, there are two styles in wide use. This rule
```
davidmorgan, line 15: Link to the Effective Dart rule?

https://dart.dev/guides/language/effective-dart/usage#do-follow-a-consistent-rule-for-var-and-final-on-local-variables

Not sure if/how linking works here though.

lib/src/rules/unnecessary_final.dart
```diff
@@ -10,9 +10,13 @@ import '../analyzer.dart';
...
+enforces the `var` style. For the alternative style that prefers `final`,
+enable `prefer_final_locals` and `prefer_final_in_for_each` instead.
+
+For fields, final is always recommended; see the rule `prefer_final_fields`.
```
davidmorgan, line 19: Maybe backticks for final here too? `final`

davidmorgan (commented):
Much better, thanks :)

lib/src/rules/unnecessary_final.dart
davidmorgan: Found an example in the `comment_references` lint

https://dart-lang.github.io/linter/lints/comment_references.html

looks like standard markdown, square brackets for the link text then rounded brackets for the link

```dart
const _details = r'''
**DO** reference only in scope identifiers in doc comments.

If you surround things like variable, method, or type names in square brackets,
then [`dart doc`](https://dart.dev/tools/dart-doc) will look
up the name and link to its docs.  For this all to work, ensure that all
identifiers in docs wrapped in brackets are in scope.
```

mit-mit: Nice, added the link!

MelbourneDeveloper: @mit-mit this is good. It's what I had in mind, and it's not towards what I think all the documentation should have: a reference to effective Dart

pq (approved):
💎

pq:
LGTM.  Thanks all!

munificent (approved):

pq (approved):
🏁

PR #3841, opened 2022-11-13
Add explicit lint codes to more lints

bwilkerson:
Also fixes a typo in a comment that I noticed while adding the codes.

### DIFF
--- a/lib/src/rules/avoid_field_initializers_in_const_classes.dart
+++ b/lib/src/rules/avoid_field_initializers_in_const_classes.dart
@@ -36,12 +36,22 @@ class A {
 ''';
 
 class AvoidFieldInitializersInConstClasses extends LintRule {
+  static const LintCode code = LintCode(
+      'avoid_field_initializers_in_const_classes',
+      "Fields in 'const' classes should not have initializers.",
+      correctionMessage:
+          'Try removing converting the field to a getter or initialize the '
+          'field in the constructors.');
+
   AvoidFieldInitializersInConstClasses()
       : super(
             name: 'avoid_field_initializers_in_const_classes',
             description: _desc,
             details: _details,
             group: Group.style);
+
+  @override
+  LintCode get lintCode => code;
 
   @override
   void registerNodeProcessors(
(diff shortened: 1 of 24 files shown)


lib/src/rules/unsafe_html.dart
```diff
@@ -33,20 +33,36 @@ var script = ScriptElement()..src = 'foo.js';
...
 ''';
 
-extension on DartType? {
-  /// Returns whether this type extends [className] from the dart:html library.
-  bool extendsDartHtmlClass(String className) =>
-      extendsClass(className, 'dart.dom.html');
-}
-
 class UnsafeHtml extends LintRule {
+  // static const LintCode attributeCode = LintCode(
```
srawlins, line 37: Should this be... removed?

srawlins (approved):

lib/src/rules/unsafe_html.dart
bwilkerson: See the TODO on line 77. We want to clean up the lint codes, and I'm proposing that these declarations be used when we do so. (In all of the other lints the codes are in the linter rather than the visitor, and I think we should follow the same convention here.)

lib/src/rules/avoid_field_initializers_in_const_classes.dart
```diff
@@ -36,13 +36,23 @@ class A {
 ''';
 
 class AvoidFieldInitializersInConstClasses extends LintRule {
+  static const LintCode code = LintCode(
+      'avoid_field_initializers_in_const_classes',
+      "Fields in 'const' classes should not have initializers.",
+      correctionMessage:
+          'Try removing converting the field to a getter or initialize the '
```
pq, line 43: Try removing **or**?

lib/src/rules/avoid_returning_null.dart
```diff
@@ -48,13 +48,22 @@ bool _isPrimitiveType(DartType type) =>
         type.isDartCoreNum);
 
 class AvoidReturningNull extends LintRule {
+  static const LintCode code = LintCode(
+      'avoid_returning_null',
+      "Don't return 'null' when the return type is 'bool', 'double', 'int', "
+          "or 'num'.",
+      correctionMessage: "Try returning a sentinal value other than 'null'.");
```
pq, line 55: => sentinel

pq (approved):

lib/src/rules/avoid_returning_null.dart
bwilkerson: Done.

PR #3845, opened 2022-11-14
add no response workflow to close stale issues w/ requests for info

pq:
Cribbed from https://github.com/dart-lang/dart-pad/blob/master/.github/workflows/no-response.yml.

/cc @bwilkerson @srawlins 

/fyi @devoncarew

### DIFF
--- a/.github/workflows/no-response.yml
+++ b/.github/workflows/no-response.yml
@@ -1,0 +1,33 @@
+# A workflow to close issues where the author hasn't responded to a request for
+# more information; see https://github.com/godofredoc/no-response for docs.
+
+name: response check
+
+# Both `issue_comment` and `scheduled` event types are required.
+on:
+  issue_comment:
+    types: [created]
+  schedule:
+    # Schedule for five minutes after the hour, every hour
+    - cron: '5 * * * *'
+
+# All permissions not specified are set to 'none'.
+permissions:
+  issues: write
+
+jobs:
+  noResponse:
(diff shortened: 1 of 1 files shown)


devoncarew:
👍 

FYI, this will immediately try and act on existing 'needs-info' issues (https://github.com/dart-lang/linter/labels/needs-info). I'd clear the label off them if it no longer applies - if the people have responded.

pq:
Thanks!  I did a run through of the existing issues w/ that tag and removed one that wasn't stale.  Happy for the others to get culled.

.github/workflows/no-response.yml
```diff
@@ -0,0 +1,33 @@
...
+
+name: response check
+
+# Both `issue_comment` and `scheduled` event types are required.
+on:
+  issue_comment:
+    types: [created]
+  schedule:
+    # Schedule for five minutes after the hour, every hour
+    - cron: '5 * * * *'
```
srawlins, line 12: This seems excessive. Once per day? I dunno what it costs us...

srawlins (approved):

.github/workflows/no-response.yml
pq: Good point/  I cribbed this entirely but on reflection, I agree.  I expect it's a cheap workflow but still.

@devoncarew:  curious what you think?

pq: Here's daily:
   
    '0 0 * * *'

PR #3849, opened 2022-11-15
look at unique names when checking fix status

pq:
I'll think about adding some proper tests (`doc_test` is a smoke test) but I did some manual verification too.

/cc @bwilkerson

### DIFF
--- a/tool/doc.dart
+++ b/tool/doc.dart
@@ -197,11 +197,12 @@ Future<void> generateDocs(String? dir,
   var fixStatusMap = await fetchFixStatusMap();
 
   // Generate rule files.
-  for (var l in rules) {
-    var fixStatus = fixStatusMap[l.name] ?? 'unregistered';
-    RuleHtmlGenerator(l, fixStatus).generate(outDir);
+  for (var rule in rules) {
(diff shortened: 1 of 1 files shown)


tool/doc.dart
```diff
@@ -245,6 +246,16 @@ String getBadges(String rule, [String? fixStatus]) {
   return sb.toString();
 }
 
+String getFixStatus(LintRule rule, Map<String, String> fixStatusMap) {
+  for (var code in rule.lintCodes) {
+    var status = fixStatusMap[code.uniqueName.substring(9)];
+    if (status != null) {
+      return status;
```
bwilkerson, line 253: If there are multiple codes, and the first has a status of 'needsFix', but the second has a status of 'hasFix', which should we show? What if the order is reversed?

bwilkerson (approved):

tool/doc.dart
pq: Hrmmm.  Good point.  I guess we probably want to emphasize the positive here 😉 

Probably best to look for a `hasFix` in any order.  I'll update.

PR #3854, opened 2022-11-17
Fix broken markdown caused by unclosed code block

parlough:
Also puts `@immutable` in code style.


pq (approved):
Thanks!

It would be great to add validation to the CLI... I looked at our markdown parser and commonmark but they don't seem bothered by un-closed code blocks.  Maybe it's just not worth it?

parlough:
The problem arises when there is other text after it, such as when I include the descriptions on dart.dev. Not sure verification is worth it, but I'll look and see if there are any good Dart based tools we could easily integrate.

pq:
Cool.  Thanks!

PR #3862, opened 2022-11-28
don't overreport `prefer_equal_for_default_values` when it's a warning

pq:
See also: https://github.com/dart-lang/linter/pull/3855.

/cc @bwilkerson


parlough:
I like the new direction! Makes sense to just not do anything in this case. Sorry I didn't think to make this change :)

bwilkerson (approved):

PR #3869, opened 2022-11-30
update unrelated_type_equality_checks for fixnum restructuring

### LINKED ISSUE #3868: Lint `unrelated_type_equality_checks` no longer recognizes `Int64`, `Int32`.
The lint `unrelated_type_equality_checks` has special-casing recognizing `Int64` and `Int32` from `package:fixnum`, to allow them to be compared to integers (which are otherwise an unrelated type).

This recognition fails after a refactoring of `package:fixnum`, which moved the classes into individual libraries, instead of using `part` files.

**To Reproduce**
Use tip-of-tree code from https://github.com/dart-lang/fixnum

Try to do `if (FixNum(1) == 1) ...` with the lint enabled. This used to not give a warning. Now it does.

**Expected behavior**

The `Int64` and `Int32` classes should still be recognized and special cased.

**Additional context**

The relevant code is:
```dart
bool _isFixNumIntX(DartType type) {
  if (type is! InterfaceType) {
    return false;
  }
  Element element = type.element;
  return (element.name == 'Int32' || element.name == 'Int64') &&
      element.library?.name == 'fixnum';
}
```
This approach depends on the package not changing, which is a fragile assumption to have for a separate package.

Consider instead checking whether the the classes are named `Int32` or `Int64`, and the *path* to the library starts with `package:fixnum/`. That should cover most potential future refactorings of the package as well, because all it checks for is a class with one of those names provided by `package:fixnum`, without assuming anything about the internal structure of the package.

(That will make the lint harder to test, since you cannot just create a library with name `fixnum` and an `Int64` type. Maybe retain the library name check as well, or change it to `.name == 'fixnum_for_testing'`, so you can still trigger the lint without depending on `package:fixnum`. On the other hand, depending on the real `package:fixnum` for testing, as a dev-dependency, might help detecting if the lint breaks again.)

pq:
Fixes #3868 

(Note that we'll want to hustle this into a cherry-pick release for 2.19.)

/cc @bwilkerson

### DIFF
--- a/lib/src/rules/unrelated_type_equality_checks.dart
+++ b/lib/src/rules/unrelated_type_equality_checks.dart
@@ -7,6 +7,7 @@ import 'package:analyzer/dart/ast/token.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
 import 'package:analyzer/dart/element/element.dart';
 import 'package:analyzer/dart/element/type.dart';
+import 'package:collection/collection.dart';
 
 import '../analyzer.dart';
 import '../extensions.dart';
@@ -153,7 +154,9 @@ bool _isFixNumIntX(DartType type) {
   if (type is! InterfaceType) return false;
   InterfaceElement element = type.element;
   if (element.name != 'Int32' && element.name != 'Int64') return false;
-  return element.library.source.uri.path.startsWith('fixnum/');
+  var uri = element.library.source.uri;
+  if (!uri.isScheme('package')) return false;
+  return uri.pathSegments.firstOrNull == 'fixnum';
 }
 
 class UnrelatedTypeEqualityChecks extends LintRule {
--- a/lib/src/rules/unrelated_type_equality_checks.dart
+++ b/lib/src/rules/unrelated_type_equality_checks.dart
@@ -7,8 +7,6 @@ import 'package:analyzer/dart/ast/token.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
 import 'package:analyzer/dart/element/element.dart';
 import 'package:analyzer/dart/element/type.dart';
-// ignore: implementation_imports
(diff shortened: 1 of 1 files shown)


lib/src/rules/unrelated_type_equality_checks.dart
```diff
@@ -148,12 +150,16 @@ bool _hasNonComparableOperands(TypeSystem typeSystem, BinaryExpression node) {
...
+  if (type is! InterfaceType) return false;
   Element element = type.element;
-  return (element.name == 'Int32' || element.name == 'Int64') &&
-      element.library?.name == 'fixnum';
+  if (element.name != 'Int32' && element.name != 'Int64') return false;
+  var libraryElement = element.library;
+  if (libraryElement is! LibraryElementImpl) return false;
+  var reference = libraryElement.reference;
+  if (reference == null) return false;
+  return reference.name.startsWith('package:fixnum/');
```
bwilkerson, line 162: There aren't any tests. How confident are you that the logic here is correct?

In particular, `Reference.name` is documented to return "The simple name of the reference in its [parent].", which doesn't sound like it's the URI of the library, but maybe.

bwilkerson (approved):
Assuming that we'll figure out how to test this and that you're very confident that the code is correct, then it's probably OK to roll it without tests.

lib/src/rules/unrelated_type_equality_checks.dart
pq: Right.  I've tested this locally with `fixnum` checked out from head.  And verified that it works for the previous `fixnum`.  It's just tricky to test both package structures.

As for digging into the reference, you're 100% right.  Chatting with Konstantin, I was reminded that the source URI is the ticket.

PR #3887, opened 2022-12-07
Detect doubled unnecessary_parenthesis, but allow ternary and equality

### LINKED ISSUE #3885: unnecessary_parenthesis sometimes not reported for doubled parentheses
Some examples of what it currently misses but should lint:

```dart
Int64((12 * 60 * 60))   // ArgumentList(BinaryExpression)

`a.compareTo((b.bar))`  // ArgumentList(MethodInvocation)

foo.addAll((await fut)) // ArgumentList(AwaitExpression)

`'${(foo == bar.baz)}'  // InterpolationExpression(BinaryExpression)
```

So I would like to introduce a rule that
* **`${(  )}` and `((  ))` are never OK, drop the inner ones.**

It's just that it's a bit tricky to find in the language what exactly constitutes these doubled parentheses. For example:
* `ArgumentList` with length of 1 is always an extra set of parentheses
   (which already covers almost all bad cases for this issue).
* So is an `if` statement, but somehow I can't trigger any false negative with it.

I will immediately open a pull request that would address this issue.

### LINKED ISSUE #3886: unnecessary_parenthesis should forgive conditional operator and equality
This came up in a Google-internal discussion as something that some people like to write and would hate the linter to drop the parentheses for them.  
As such, it's a blocker for internal adoption.

I would like to introduce these rules:
* **`(a ? b : c)` is always OK**
* **`(a == b)` and `(a != b)` are OK in assignment-like and return-like statements.**

---

Examples of replacements that I wish didn't happen (copying my own comments from that discussion):

* ```dart
    Foo(String? name) : name = (name == null || name.isEmpty ? _generateNodeName() : name),
    Foo(String? name) : name = name == null || name.isEmpty ? _generateNodeName() : name,
  ```
  
  This is a downgrade because my mind jumps to read it as `(name = name == null || the rest)`.

* ```dart
        node.hasValue = (value == null);
        node.hasValue = value == null;
  ```
  
  I dislike the fact that `Value = value` jumps out at me so strongly. Part of it might just be an issue of *distance* because `==` is just *longer* and makes the precedence appear lower than `=`. The other part is that I want to use the parentheses to stress the fact that I'm assigning the result of a boolean expression, not assigning "value with some other code trailing behind it".

---

I will immediately open a pull request that would address this issue.

oprypin:
* Fixes #3885
* Fixes #3886

---

This implements new rules for `unnecessary_parenthesis` lint:
* `${(  )}` and `((  ))` are never OK, drop the inner ones

else,
* `(a ? b : c)` is always OK
* `(a == b)` and `(a != b)` are OK in assignment-like and return-like statements.

---

This pull request addresses two seemingly completely separate changes, but I think it should be done all at once. That is because:

* Currently some cases where doubled parentheses *happen to be* dropped are cases of `((a == b))` etc. So I'd be introducing more false negatives if I remove the old reason to drop these but not add the new one.

* Removing doubled parentheses just misses one case of `ArgumentList` so it's unclear why I am adding all the other numerous cases here. But more rules are necessary *only because* they will become the only reason for some cases of `((a ? b : c))` to still be dropped.

### DIFF
--- a/lib/src/rules/unnecessary_parenthesis.dart
+++ b/lib/src/rules/unnecessary_parenthesis.dart
@@ -3,6 +3,7 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/token.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
 
 import '../analyzer.dart';
@@ -100,8 +101,19 @@ class _Visitor extends SimpleAstVisitor<void> {
       }
     }
 
-    if (parent is ParenthesizedExpression) {
+    // Directly wrapped into parentheses already - always report.
+    if (parent is ParenthesizedExpression ||
+        parent is InterpolationExpression ||
+        (parent is ArgumentList && parent.arguments.length == 1) ||
+        (parent is IfStatement && node == parent.condition) ||
+        (parent is WhileStatement && node == parent.condition) ||
+        (parent is IfElement && node == parent.condition)) {
       rule.reportLint(node);
+      return;
+    }
+
+    // `(foo ? bar : baz)` is OK.
+    if (expression is ConditionalExpression) {
       return;
     }
 
@@ -119,6 +131,18 @@ class _Visitor extends SimpleAstVisitor<void> {
     if (parent is ConstructorFieldInitializer &&
         _containsFunctionExpression(node)) {
       return;
+    }
+
+    // `foo = (a == b)` is OK, `return (count != 0)` is OK.
+    if (expression is BinaryExpression &&
+        (expression.operator.type == TokenType.EQ_EQ ||
+            expression.operator.type == TokenType.BANG_EQ)) {
+      if (parent is AssignmentExpression ||
+          parent is VariableDeclaration ||
+          parent is ReturnStatement ||
+          parent is YieldStatement) {
+        return;
+      }
     }
 
     if (parent is Expression) {
--- a/lib/src/rules/unnecessary_parenthesis.dart
+++ b/lib/src/rules/unnecessary_parenthesis.dart
@@ -136,8 +136,11 @@ class _Visitor extends SimpleAstVisitor<void> {
         parent is InterpolationExpression ||
         (parent is ArgumentList && parent.arguments.length == 1) ||
         (parent is IfStatement && node == parent.condition) ||
+        (parent is IfElement && node == parent.condition) ||
         (parent is WhileStatement && node == parent.condition) ||
-        (parent is IfElement && node == parent.condition)) {
+        (parent is DoStatement && node == parent.condition) ||
+        (parent is SwitchStatement && node == parent.expression) ||
+        (parent is SwitchExpression && node == parent.expression)) {
       rule.reportLint(node);
       return;
     }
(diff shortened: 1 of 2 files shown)


lib/src/rules/unnecessary_parenthesis.dart
```diff
@@ -100,11 +101,22 @@ class _Visitor extends SimpleAstVisitor<void> {
       }
     }
 
-    if (parent is ParenthesizedExpression) {
+    // Directly wrapped into parentheses already - always report.
+    if (parent is ParenthesizedExpression ||
+        parent is InterpolationExpression ||
+        (parent is ArgumentList && parent.arguments.length == 1) ||
```
bwilkerson, line 107: It isn't clear to me why this is only true when there's a single argument. Are there valid reasons for allowing `o.m(a, (b), c);`?

Also, do we want to handle the case of named arguments (`0.m(a: (b));`)?

lib/src/rules/unnecessary_parenthesis.dart
```diff
@@ -100,11 +101,22 @@ class _Visitor extends SimpleAstVisitor<void> {
...
+        (parent is IfStatement && node == parent.condition) ||
+        (parent is WhileStatement && node == parent.condition) ||
```
bwilkerson, line 109: This might want to also check for the condition in a do-while statement and the expression in a switch statement or switch expression.

lib/src/rules/unnecessary_parenthesis.dart
```diff
@@ -121,6 +133,18 @@ class _Visitor extends SimpleAstVisitor<void> {
...
     }
 
+    // `foo = (a == b)` is OK, `return (count != 0)` is OK.
+    if (expression is BinaryExpression &&
+        (expression.operator.type == TokenType.EQ_EQ ||
+            expression.operator.type == TokenType.BANG_EQ)) {
+      if (parent is AssignmentExpression ||
+          parent is VariableDeclaration ||
+          parent is ReturnStatement ||
+          parent is YieldStatement) {
```
bwilkerson, line 143: One of the examples in the issue is in the context of a constructor initializer, but this doesn't cover that case (the parent will be a `ConstructorFieldInitializer`.

bwilkerson (commented):
No need to address these comments until we've decided whether we want to make all of the changes.

lib/src/rules/unnecessary_parenthesis.dart
oprypin: Yes, `o.m(a, (b), c);` are currently not handled and will still not be handled.

**But** this is intentionally written like this because this whole set of conditions aims to detect **specifically** `((b))` and nothing else, and that's also the only rule I'm arguing for within this proposal.

pq: Keeping this scoped to single args sounds reasonable to me. (Though I do think we should consider cases like `o.m(a, (b), c)` in the future.)

lib/src/rules/unnecessary_parenthesis.dart
pq: @oprypin: do you want to add these cases or add a todo?

lib/src/rules/unnecessary_parenthesis.dart
pq: @oprypin: could you add this case (and a corresponding test)?

pq (changes requested):

lib/src/rules/unnecessary_parenthesis.dart
oprypin: Added DoStatement (+test), SwitchStatement (+test), SwitchExpression

lib/src/rules/unnecessary_parenthesis.dart
oprypin: Added ConstructorFieldInitializer (+test)

pq (approved):

Issue #3848, opened 2022-11-15
linter 1.32.0 plan


pq:
# 1.32.0

- update `avoid_types_as_parameter_names` to handle type variables
- update `avoid_positional_boolean_parameters` to handle typedefs
- improve `unnecessary_parenthesis` support for property accesses and method invocations 
- update `avoid_redundant_argument_values` to check parameters of redirecting constructors
- performance improvements for `prefer_const_literals_to_create_immutables`
- update `use_build_context_synchronously` to check context properties
- fix false positive for `avoid_private_typedef_functions` with generalized type aliases

---

## release testing

**Try:** https://dart-review.googlesource.com/c/sdk/+/275044

### downstream fixes

- [x] fix new `avoid_redundant_argument_values` diagnostics in `flutter` 
    - (https://logs.chromium.org/logs/dart/buildbucket/cr-buildbucket/8797403260884127729/+/u/analyze_flutter_flutter/stdout)
- [x] fix new `unnecessary_parenthesis` diagnostics in the SDK (https://dart-review.googlesource.com/c/sdk/+/275044)
- [x] fix new `unnecessary_parenthesis` diagnostics in `flutter` 
   - [x] https://github.com/flutter/flutter/pull/117081
   - [x] fix false positive with nullable property accesses (🍻 @asashour -- #3902)
   - [x] fix false positive with a postfix `!` (🎉 @asashour -- #3904)
- [x] fix new `unnecessary_parenthesis` diagnostics in `flutter plugins`
  - [x]  https://github.com/flutter/plugins/pull/6841
- [ ]  fix new `unnecessary_parenthesis` diagnostics in `flutter engine` 
  - [x] https://github.com/flutter/engine/pull/38291

---

/fyi @srawlins

asashour:
The link is private, I guess it should better be the public https://dart-review.googlesource.com/c/sdk/+/269747

(there were other places where the the dart-review SDK link didn't contain `googlesource`, but this probably seems important).

pq:
Thanks for catching this!

pq:
Reviving this as 1.31.0 has landed and will (soon 🤞) be cherry-picked to beta.

I'll update the issue description with a changelog.  There's a lot queued up so I expect this could be a pretty fun one to land (but good to do before we start on some plumbing for removals, etc (#3880)).  🥳 

/fyi @srawlins @bwilkerson @oprypin

pq:
[...]

PR #3902, opened 2022-12-13
Fix `unnecessary_parenthesis` with null aware expressions.

asashour:
As reported in #3848

This fixes the situation where a nullable expression is surrounded by parentheses.

Please note, I couldn't create a test case for `SpreadElement` or `ExtensionOverride` (which possibly have also `isNullAware`).


srawlins (approved):
Thanks!

pq (approved):
🦅

pq:
Thank you!!!

PR #3904, opened 2022-12-13
Fix `unnecessary_parenthesis` with postfix bang operator.

asashour:
#3848

Add a fix for the `PostfixExpression` with `!`.

Please also note the Flutter code can be simplified as proposed in https://github.com/flutter/flutter/pull/117026


pq (approved):

pq:
Thanks!

PR #3907, opened 2022-12-14
Fix examples in avoid_null_checks_in_equality_operators

srawlins:
There are a few issues with the text, which I fix here:

* the example used a class with a non-nullable field with no initializer,
* the example did not specify a parameter type for `operator ==`; in Dart versions before null safety, this parameter type was `dynamic` (I think), but after null safety, it is `Object`, so even aside from this lint rule, comparing to null wouldn't make sense.
* incorrect text in the first sentence summary.

### DIFF
--- a/lib/src/rules/avoid_null_checks_in_equality_operators.dart
+++ b/lib/src/rules/avoid_null_checks_in_equality_operators.dart
@@ -17,16 +17,17 @@ const _desc = r"Don't check for null in custom == operators.";
 const _details = r'''
 **DON'T** check for null in custom == operators.
 
-As null is a special type, no class can be equivalent to it.  Thus, it is
-redundant to check whether the other instance is null. 
+As null is a special value, no instance of any class (other than `Null`) can be
+equivalent to it.  Thus, it is redundant to check whether the other instance is
+null.
 
 **BAD:**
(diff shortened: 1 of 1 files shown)


lib/src/rules/avoid_null_checks_in_equality_operators.dart
```diff
@@ -17,27 +17,28 @@ const _desc = r"Don't check for null in custom == operators.";
...
 
 **GOOD:**
 ```dart
 class Person {
-  final String name;
+  final String? name;
 
   @override
-  operator ==(other) => other is Person && name == other.name;
+  operator ==(Object? other) => other is Person && name == other.name;
```
bwilkerson, line 41: Best practice is for the parameter to be `Object` because a `null` can't be passed in. (I haven't checked to see whether we have another warning for that.)

bwilkerson (approved):

pq (approved):

lib/src/rules/avoid_null_checks_in_equality_operators.dart
srawlins: Haha I know. But the lint won't fire under that practice.

PR #3916, opened 2022-12-14
1.32.0

pq:
# 1.32.0

- update `avoid_types_as_parameter_names` to handle type variables
- update `avoid_positional_boolean_parameters` to handle typedefs
- improve `unnecessary_parenthesis` support for property accesses and method invocations
- update `avoid_redundant_argument_values` to check parameters of redirecting constructors
- performance improvements for `prefer_const_literals_to_create_immutables`
- update `use_build_context_synchronously` to check context properties
- fix false positive for `avoid_private_typedef_functions` with generalized type aliases

---

Closes: #3848 

RC being tested on https://dart-review.googlesource.com/c/sdk/+/275044.

/cc @bwilkerson @srawlins

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,18 @@
+# 1.32.0
+
+- update `avoid_types_as_parameter_names` to handle type variables
+- update `avoid_positional_boolean_parameters` to handle typedefs
+- improve `unnecessary_parenthesis` support for property accesses and method invocations
+- update `avoid_redundant_argument_values` to check parameters of redirecting constructors
+- performance improvements for `prefer_const_literals_to_create_immutables`
+- update `use_build_context_synchronously` to check context properties
+- fix false positive for `avoid_private_typedef_functions` with generalized type aliases
+
 # 1.31.0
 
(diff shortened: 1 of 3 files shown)


bwilkerson (approved):

CHANGELOG.md
```diff
@@ -1,8 +1,18 @@
+# 1.32.0
```
devoncarew, line 1: FYI - as a drive by comment - it looks like `1.31.0` has not been published; the latest published version is `1.30.0`.

srawlins: @pq has been experimenting not publishing linter, as there is no known use case for publishing it.

devoncarew: OK, that makes a certain amount if sense :)

Perhaps change the listing on pub.dev to 'unlisted'? That'll mean it won't show up in search results.

If we know were never going to publish it again (which it sounds like we're not yet certain of?) it can instead be marked 'discontinued' in pub.

pq: Unlisted is an interesting option.  We've got a few dependent packages so we'll want to sort that out before we go too far.  Maybe we can discuss in person?

PR #3918, opened 2022-12-15
Tidy up null_closures for null safety

srawlins:
# Description

`null_closures` was written before null safety for guarding against sending `null` where it was not expected, in Dart SDK APIs. For _most cases_, this has been solved by null safety 🎉 🎉 🎉 🎉 🎉 . In fact, every single positional closure parameter was made non-nullable, so this rule no longer applies to any positional parameters.

The named parameters however are all still relevant, and were the primary motivation for the rule (before the rule, I cleaned up many cases of `list.firstWhere(..., orElse: null)`).

### DIFF
--- a/lib/src/rules/null_closures.dart
+++ b/lib/src/rules/null_closures.dart
@@ -8,7 +8,6 @@ import 'package:analyzer/dart/element/element.dart';
 import 'package:analyzer/dart/element/type.dart';
 
 import '../analyzer.dart';
-import '../extensions.dart';
 
 const _desc = r'Do not pass `null` as an argument where a closure is expected.';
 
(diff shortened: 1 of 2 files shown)


srawlins:
Oh my there is a NullClosuresPreNNBD test. Does that mean we should put this PR in the ice box until we're done with pre-null safe code?

bwilkerson:
> Does that mean we should put this PR in the ice box ...

Here's my understanding. If the rule is being used internally, then we want to enable or disable based on the flag in `analyzer` (which isn't in a published version). Otherwise, it likely isn't possible to pull a version of `linter` into an SDK that isn't on the 3.0 track (it would have to be done as a CP, and is probably too big to be considered), so we can make ay 3.0 changes to linter that we're ready to make.

srawlins:
> Here's my understanding. If the rule is being used internally, then we want to enable or disable based on the flag in `analyzer` (which isn't in a published version).

I don't think I really understand this.

So, in a pre-null safe library, the first parameter of `Future.then` is nullable, so this rule guards against passing `null` there, since that couldn't be specified at the API level. In a null safe library, the first parameter of `Future.then` is non-nullable, so the rule will never fire on an argument passed to `Future.then`.

How would this rule use the `noSoundNullSafety` flag? I can see how I could update the rule to use `LibraryElement. isNonNullableByDefault`, but not how I could update the rule to use `noSoundNullSafety`.

bwilkerson:
Yes, if it's gated on 2.12 (null safety), then you don't need the flag from analyzer. (I'm assuming that in the 3.0 world we will set the language level to 2.12 even if a user erroneously attempts to opt back to an earlier version. If that's not the case then there might be a difference.)

srawlins:
Closing; this won't land before the linter code moves to the SDK, and I have this PR for my history :D

Issue #3441, opened 2022-06-02
proposal: `avoid_nullable_equals_parameter`


natebosch:
## Description

"The null value will never be passed to `operator ==`. Use a non-nullable `Object` parameter"

## Details

A definition like `bool operator ==(Object? other)` causes other classes which override the interface to _also_ accept a nullable argument, even though `null` will never be passed.

## Kind

Style

## Good Examples 

```dart
  @override
  bool operator ==(Object other) =>
```

## Bad Examples

```dart
  @override
  bool operator ==(Object? other) =>
```

## Discussion

Mockito codegen currently fails to generate fakes for classes with a nullable argument to `operator ==`.

The main motivator is that the code is misleading - `null` can't show up so it's not worth making it look like it can.

natebosch:
This could perhaps supersede `avoid_null_checks_in_equality_operators`

If I understand correctly that lint only fires when the argument is nullable.

natebosch:
There is one place where using `Object?` is useful, it allows promotion to a generic type when the bound on the generic is also `Object?`. If the argument is `Object` then we can't promote to a generic type with no bound.

Example:

```dart
class C<T> {
  bool checkStuff(T other) => true;

  @override
  bool operator ==(Object other) => other is T && checkStuff(other);
  //                                                         ^^^^^
  // The argument type 'Object' can't be assigned to the parameter type 'T'.
}
```

I think it's probably better to find a different pattern for the body than to widen the argument type to get promotion.

natebosch:
Could also flag no argument `dynamic` as an unnecessary widening as well.

srawlins:
+100. One question:

> "The null value will never be passed to `operator ==`. Use a non-nullable `Object` parameter"

Does this fact come from spec? Or just existing behavior of today's backends?

bwilkerson:
It's from the spec (the third bullet item):

> 17.27 Equality equality
> Equality expressions test objects for equality.
...
> Evaluation of an equality expression ee of the form e1 == e2 proceeds as follows:
> • The expression e1 is evaluated to an object o1.
> • The expression e2 is evaluated to an object o2.
> • If either o1 or o2 is the null object (17.4), then ee evaluates to true if both o1 and o2 are the null object and to false otherwise. Otherwise,
> • evaluation of ee is equivalent to the method invocation o1.==(o2).

srawlins:
[...]

PR #3923, opened 2022-12-17
Add non_nullable_equals_parameter rule

srawlins:
# Description

This rule checks that a parameter to an `operator ==` implementation has a non-nullable type.

I intentionally did not enforce, in this rule, that the parameter is exactly `Object`. It is legal to narrow the parameter type to a different non-nullable type, like `int`. I can't imagine doing it, but it seems to be unrelated to whether the type should be nullable or not.

Fixes https://github.com/dart-lang/linter/issues/3441

### DIFF
--- a/lib/src/rules/non_nullable_equals_parameter.dart
+++ b/lib/src/rules/non_nullable_equals_parameter.dart
@@ -1,0 +1,104 @@
+// Copyright (c) 2022, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/token.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+
+import '../analyzer.dart';
+
+const _desc = r'The parameter type of `==` operators should be non-nullable.';
+
+const _details = r'''
+The parameter type of `==` operators should be non-nullable.
+
+`null` is never passed to `operator ==`, so the parameter type of an `==`
+operator override should always be non-nullable.
+
+**BAD:**
+```dart
+class C {
+  String key;
+  B(this.key);
+  @override
+  operator ==(Object? other) => other is C && other.key == key;
+}
+```
+
+**GOOD:**
+```dart
+class C {
+  String key;
+  B(this.key);
+  @override
+  operator ==(Object other) => other is C && other.key == key;
+}
+```
+''';
+
+class NonNullableEqualsParameter extends LintRule {
+  static const LintCode code = LintCode('non_nullable_equals_parameter',
+      'The parameter should have a non-nullable type.',
+      correctionMessage: 'Try using a non-nullable type.');
+
+  NonNullableEqualsParameter()
+      : super(
+            name: 'non_nullable_equals_parameter',
+            description: _desc,
+            details: _details,
+            group: Group.style);
+
+  @override
+  LintCode get lintCode => code;
+
+  @override
+  void registerNodeProcessors(
+      NodeLintRegistry registry, LinterContext context) {
+    var visitor = _Visitor(this, context);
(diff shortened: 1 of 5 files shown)


lib/src/rules/non_nullable_equals_parameter.dart
```diff
@@ -0,0 +1,104 @@
...
+    if (node.name.type != TokenType.EQ_EQ) {
+      return;
+    }
+
+    var parameters = node.parameters;
+    if (parameters == null) {
+      return;
+    }
+
+    if (parameters.parameters.isEmpty) {
```
bwilkerson, line 81: Alternately, we could test for `length != 1`, because if there's more than one there will also be a different diagnostic.

lib/src/rules/non_nullable_equals_parameter.dart
```diff
@@ -0,0 +1,104 @@
...
+
+    if (parameterElement == null) {
+      return;
+    }
+
+    var type = parameterElement.type;
+
+    if (!type.isDartCoreObject && !type.isDynamic) {
+      // There is no legal way to define a nullable parameter type, which is not
+      // `dynamic` or `Object?`.
```
bwilkerson, line 96: Actually, there's no way to define a valid override where the type isn't a supertype of `Object`; narrowing the parameter type also isn't allowed, contrary to your comment on the PR.

bwilkerson (approved):
I'm not sure that this shouldn't be a warning rather than a lint, but other than that it looks good.

lib/src/rules/non_nullable_equals_parameter.dart
srawlins: But you could make a legal override where the parameter tyoe is a subtype of `Object`.

```dart
class C {
  @override
  bool operator ==(covariant C other) => false;
}
```

bwilkerson: Oh, right. I didn't think of `covariant`. But I agree that we shouldn't flag that case.

lib/src/rules/non_nullable_equals_parameter.dart
pq: Nice idea.

lib/src/rules/non_nullable_equals_parameter.dart
pq: Nice.  Maybe consider adding a `covariant` test case?

lib/src/rules/non_nullable_equals_parameter.dart
srawlins: Love it! Done.

lib/src/rules/non_nullable_equals_parameter.dart
srawlins: I've added a test case.

srawlins:
> I'm not sure that this shouldn't be a warning rather than a lint, but other than that it looks good.

Yeah I think it's pretty grey. I would immediately recommend this go to the core lints set, so shouldn't I just make it a Warning? Or should it take a natural path from Lint to Warning, as that does get more people involved in the decision / more vetting? A puzzle...

Do you have a preference, @pq

bwilkerson:
I don't think that all lints in the core set should be warnings. I think a check should only be a warning if it finds a real bug, or code that's never what the user intended, with no false positives. This seems to meet that criteria, but I might be missing something.

srawlins:
> I don't think that all lints in the core set should be warnings. I think a check should only be a warning if it finds a real bug, or code that's never what the user intended, with no false positives. This seems to meet that criteria, but I might be missing something.

Agreed. I'll wait for @pq's response.

pq:
Sorry for the slow response!

This one seems like a great candidate for an analyzer warning proper for all the reasons Brian enumerated.  I can't think of a downside.

srawlins:
Replaced by a new warning: https://dart-review.googlesource.com/c/sdk/+/277700

Issue #3134, opened 2022-01-08
Analyzer incorrectly reports "prefer_initializing_formals"


obeobe:
Analyzer incorrectly reports "prefer_initializing_formals" in this code:

Code:
```dart
class Hello {
  bool isAdult = true;
  late int age;

  Hello({
    this.isAdult = true,
    int? age
  }) {
    age ??= this.isAdult ? 30 : 12;
    this.age = age; // <-- "prefer_initializing_formals" reported for this line
  }
}
```

Screenshot: 
![image](https://user-images.githubusercontent.com/1394039/148644071-a3703d83-7750-4758-bde1-29e1cd646c59.png)

Using:
 * Linux.
 * Dart SDK Version: `Dart SDK version: 2.15.1 (stable) (Tue Dec 14 13:32:21 2021 +0100) on "linux_x64"`
 * IntelliJ IDEA 2021.2.1
 * JetBrains Dart plugin: 212.5080.8
 * Flutter plugin: 60.1.4

bwilkerson:
I believe that the problem here is that an initializing formal doesn't introduce a parameter into the name scope. Converting the original code to use an initializing formal therefore change the reference to `age` from being a reference to a parameter of type `int?` into a reference to a field of type `int`, which introduces errors in the body.

The most complete solution here would be to analyze the initializers and body to see whether the parameter is being referenced in such a way that making that conversion would be breaking. That might be difficult to get right. A simpler approach would be to see whether the parameter is referenced at all, and to not lint if it is, but that would likely introduce some false negatives.

oprypin:
> see whether the parameter is referenced at all

That approach seems great to me. In fact, maybe that's even *the* important part.
If the order of occurrences is [parameter definition], [unrelated reference], [actual assignment],
the lint shouldn't change it to [parameter definition+actual assignment], [unrelated reference]

PR #3926, opened 2022-12-20
Loosen prefer_initializing_formals to not include different params vs fields types

oprypin:
The code below triggers the lint:
```dart
class CaseWithDifferentType {
  Iterable<int> ids;
  CaseWithDifferentType(List<int> ids) : ids = ids { // LINT
    ids.add(1);
  }
}
```
The fixed code doesn't work:
```dart
class CaseWithDifferentType {
  Iterable<int> ids;
  CaseWithDifferentType(List<int> this.ids) {
    ids.add(1);  // Error: The method 'add' isn't defined for the class 'Iterable<int>'.
  }
}
```
This pull request will loosen this edge case to also match the type before linting so that this won't be an issue.

Closes #1605

I have revived that pull request from @dramos07 and adapted it to null-safe world. The tests all pass but I'm not so sure that I didn't do something wrong.

### DIFF
--- a/lib/src/rules/prefer_initializing_formals.dart
+++ b/lib/src/rules/prefer_initializing_formals.dart
@@ -134,15 +134,16 @@ class PreferInitializingFormals extends LintRule {
   @override
   void registerNodeProcessors(
       NodeLintRegistry registry, LinterContext context) {
-    var visitor = _Visitor(this);
+    var visitor = _Visitor(this, context);
     registry.addConstructorDeclaration(this, visitor);
   }
 }
 
 class _Visitor extends SimpleAstVisitor<void> {
   final LintRule rule;
+  final LinterContext context;
 
-  _Visitor(this.rule);
+  _Visitor(this.rule, this.context);
 
   @override
   void visitConstructorDeclaration(ConstructorDeclaration node) {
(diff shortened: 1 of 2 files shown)


bwilkerson:
I think the comment I added to https://github.com/dart-lang/linter/issues/3134 applies here as well (field name and types changed to match the example above):

> I believe that the problem here is that an initializing formal doesn't introduce a parameter into the name scope. Converting the original code to use an initializing formal therefore change the reference to `ids` from being a reference to a parameter of type `List<int>` into a reference to a field of type `Iterable<int>`, which introduces errors in the body.
>
> The most complete solution here would be to analyze the initializers and body to see whether the parameter is being referenced in such a way that making that conversion would be breaking. That might be difficult to get right. A simpler approach would be to see whether the parameter is referenced at all, and to not lint if it is, but that would likely introduce some false negatives.

In other words, the problem here isn't related to the types of the field and parameter. If it weren't for the invocation of `add` in the body, it would be perfectly reasonable to rewrite the code to the following:

```dart
class CaseWithDifferentType {
  Iterable<int> ids;
  CaseWithDifferentType(List<int> this.ids);
}
```

srawlins:
@oprypin gentle ping for Brian's comment above

PR #3929, opened 2022-12-20
ensure `one_member_abstracts` doesn't lint sealed classes

### LINKED ISSUE #3928: ensure `one_member_abstracts` does not lint on sealed classes

pq:
Fixes #3928

**NOTE:** this needs a new version of analyzer including https://github.com/dart-lang/sdk/commit/7ff20a67ca1a378f4e11df687693e7fc7c77c73e to land. 👍 

**UPDATE:** landed 🦅 

/cc @bwilkerson


test/rules/one_member_abstracts_test.dart
```diff
@@ -0,0 +1,31 @@
...
+  @override
+  String get lintRule => 'one_member_abstracts';
+
+  @override
+  List<String> get experiments => ['sealed-class'];
+
+  test_sealed_noDiagnostic() async {
+    await assertNoDiagnostics(r'''
+sealed class C {
+  void f() { }
```
bwilkerson, line 26: The rule only fires when there's an abstract class (which sealed classes are by default) with a single abstract member. It's possible that the test is only passing because the method isn't abstract.

pq: Thanks!

```dart
sealed class C {
  void f();
}
```

isn't valid and triggers a `CONCRETE_CLASS_WITH_ABSTRACT_MEMBER` compile time error.

and

```dart
abstract sealed class C {
  void f();
}
```

also isn't valid, triggering a `ABSTRACT_SEALED_CLASS` parse error.

So I think the above is the only test case we need to consider...  Or do you see anything else?

bwilkerson: > ... triggers a `CONCRETE_CLASS_WITH_ABSTRACT_MEMBER` compile time error.

Unless the spec is out of date, and assuming that you're using the latest version of the analyzer (which hasn't been published yet), that's a bug in the analyzer's support. The spec currently reads:

> A class marked `sealed` is implicitly an _abstract class_ with all of the existing restrictions and capabilities that implies. _It may contain abstract member declarations, it is a compile-time error to directly invoke its constructors, etc._

pq: I'm pretty sure I was testing with tip of tree but will cycle back and re-try after break.

/fyi @kallentu (who I think is working on this?)

kallentu: [This commit](https://github.com/dart-lang/sdk/commit/7ff20a67ca1a378f4e11df687693e7fc7c77c73e) should be the point where sealed classes are flagged to be abstract. I'm not sure how the linter works, but assuming the version includes this commit, it should work as intended.

pq: OK, great. Thanks!  I just resynced and all looks good.  (I must have been behind your commit.)  Sorry for the noise!

pq:
@bwilkerson : could you take a look?  Thanks!

bwilkerson (approved):

PR #3933, opened 2022-12-21
[flutter_style_todos] Only allow blank characters before `TODO`

### LINKED ISSUE #3919: analyzer rule flutter_style_todos has false positives
This tracker is for issues related to:

* Analyzer

* Dart SDK Version: Dart SDK version: 3.0.0-21.0.dev (dev) (Mon Dec 12 06:46:26 2022 -0800) on "linux_x64" (no issue on stable)
* Whether you are using Windows, MacOSX, or Linux (if applicable): had issue on mac and linux

I was using analyzer w flutter_style_todos rule on dev version and saw it flag some code that wasn't an actual TODO comment, just a comment that happened to have the word TODO in it.

Comment looked this this:
```dart
  /// Generates a request for a single instance of the `modelType` by `id`.
  ///
  /// ```dart
  /// final request = ModelQueries.get(Todo.classType, 'some-todo-id-123');
  /// ```
```

I would not expect it to consider this a TODO comment.

AlexV525:
# Description

The `flutter_style_todos` rule is used to allow any single character before `TODO`, this PR disallows it.

Fixes #3919

### DIFF
--- a/lib/src/rules/flutter_style_todos.dart
+++ b/lib/src/rules/flutter_style_todos.dart
@@ -50,7 +50,7 @@ class FlutterStyleTodos extends LintRule {
 }
 
 class _Visitor extends SimpleAstVisitor<void> {
-  static final _todoRegExp = RegExp(r'//+(.* )?TODO\b', caseSensitive: false);
+  static final _todoRegExp = RegExp(r'//+\s?TODO\b', caseSensitive: false);
 
   static final _todoExpectedRegExp =
       RegExp(r'// TODO\([a-zA-Z0-9][-a-zA-Z0-9]*\): ');
--- a/test/rules/flutter_style_todos_test.dart
+++ b/test/rules/flutter_style_todos_test.dart
@@ -20,6 +20,7 @@ class FlutterStyleTodosTest extends LintRuleTest {
   // TODO(srawlins): This test is called, "bad patterns", contains 10 TODO-like
   // comment lines, but then only expects 9 lints. Why?
   test_badPatterns() async {
+    // Test with comments that will be recognized as TODOs.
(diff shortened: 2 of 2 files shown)


lrhn:
The description here was confusing to me. It's technically correct in that this PR changes (restricts) the definition of what is considered a TODO, and therefore what can trigger being an *invalid* TODO.

That means that the lint *now* allows characters to occur before the word "TODO" that would previously trigger the lint warning.

Is the following description a correct interpretation of the change?
> Changes what counts as a TODO to include fewer occurrences of "todo".
> 
> Previously anything of the form `//<anything> TODO` (any casing) would count as a 
> TODO, and could then be considered a malformed TODO by the lint.
> With the change, only two or more `/`s, followed by zero or one whitespace, and the "TODO" letters
> will be considered as a TODO, and other occurrences of "//" followed by "TODO" will not trigger the lint.
> That allows code like `/// if (test) log(Severity.low, ToDo("Warning"));` (fx as part of a DartDoc code
> sample) to not be linted against.

AlexV525:
> Maybe it's the word "disallows" that throws me off, because the change _allows_ (removes a lint warning from) a line like:

@lrhn The request was long ago. If I understand correctly I'm trying to say that pattern will not be recognized as TODOs. Does that sound good to you?

lrhn:
Ack, I accidentally fell over this CL and didn't notice how old the request was. 
But yes, the charge is to not recognize some lines as being intended as todos, which were previously considered (possibly invalid) todos.
It restricts the lines that are considered todos.

AlexV525:
I'd probably reboot this in recent weeks. The reason why I didn't work on it anymore is I didn't receive responses.

PR #3944, opened 2022-12-23
Move avoid_equals_and_hash_code_on_mutable_classes tests to reflective tests

srawlins:


pq:
Awesome.  FWIW, I don't think it's super necessary to migrate simple tests but don't feel strongly if folks have a preference.

What I **would** like to see happen is all the integration tests migrated over since these are more complex and harder to maintain.

pq (approved):

srawlins:
I don't have a strong preference to move the simple tests in general. But I find it confusing sometimes when a rule is covered by simple tests _and_ reflective tests. It's hard to understand coverage, and adds choices to the question of where to write new tests when a bug is fixed. So I'm trying to at least clean up some set of rules which have both simple and reflective tests. Unfortunately I'm realizing there are many tests in that boat 🤣 . I'm still in the "A"s.

Issue #3943, opened 2022-12-22
use_build_context_synchronously example code and description are wrong


fourman-alex:
The [documentation](https://dart-lang.github.io/linter/lints/use_build_context_synchronously.html) states `When a BuildContext is used, its mounted property must be checked after an asynchronous gap` While `mounted` is a property of `State` and NOT `BuildContext`

Also this example (in the description) is wrong for the reason mentioned above:
```
void onButtonTapped() async {
  await Future.delayed(const Duration(seconds: 1));

  if (!context.mounted) return;
  Navigator.of(context).pop();
}
```
This statement will not compile: `if (!context.mounted) return;`

**Expected behavior**

- A better example that takes into account a StatelessWidget that has an async gap in one of its callbacks
- change ` if (!context.mounted) return;` to `if (!mounted) return;` but emphasize that this is only for StatefulWidgets

pq:
Ah, yes.  Looks like we have lots of room for improvement here.  Thanks!  (And sorry for any confusion caused.)

If you want to take a crack at improving the docs, a PR is most welcome.  Otherwise any ideas for better examples would be appreciated!

fourman-alex:
@pq I've made an initial attempt at this. Though this lint seems to touch upon a problematic limitation of `BuildContext` and its lifecycle. Any way, if someone can have a look at my PR and comment, that would be great.

pq:
Thanks a million!

srawlins:
Closing as `mounted` is now a property on `BuildContext`. Please re-open if the issue remains.

PR #3946, opened 2022-12-25
fix use_build_context_synchronously description error

fourman-alex:
This PR is trying to address this issue: https://github.com/dart-lang/linter/issues/3943 use_build_context_synchronously lint is pointing out the issue of using a `BuildContext` after an sync gap when its state can become defunct.  The first problem was that the description and the example were incorrect. Because calling `if (!context.mounted) return` is not possible as `mounted` is not a property of `BuildContext` rather of `State`.  So fixing that is one thing by emphasizing that you can only check `mounted` on a `StatefulWidget`.  The more general advice I feel we can give the devs at this point is (without getting to technical in the lint description) is to store the "state" that would be part of `BuildContext`'s tree before the async gap. Acting on that state would be ok even after an async gap, even if the `BuildContext` is `defunct` at that point because the state that we expect to affect after an async gap is still that state that was referenced before the async gap. In other words we expect that we would still affect the `Navigator`/`ScaffoldMessenger`/`Provider`/etc. that we got before the async gap.  To look at it from another direction. Even if we assume the risk of the `BuildContext` being defunct after an async gap, and call `Navigator.of(context)`, in the case where the context is now part of a different location in the tree we actually don't want the "new" `Navigator` but rather the "old" one which we would have gotten before the async gap.

IMO this whole thing is a caused by a limitation of the framework. What I mean by that is, if we know that `BuildContext` can have a state where it is not supposed to be used for certain calls it would be best to have a way to check that on the `BuildContext` itself (like `BuildContext.mounted` would have done if it existed)

Fixes # (issue)
https://github.com/dart-lang/linter/issues/3943

### DIFF
--- a/lib/src/rules/use_build_context_synchronously.dart
+++ b/lib/src/rules/use_build_context_synchronously.dart
@@ -17,11 +17,13 @@ const _details = r'''
 **DON'T** use BuildContext across asynchronous gaps.
 
 Storing `BuildContext` for later usage can easily lead to difficult to diagnose
-crashes. Asynchronous gaps are implicitly storing `BuildContext` and are some of
-the easiest to overlook when writing code.
+crashes, because it can change its state during the async gap and no longer be 
+asociated to an `Element`. Asynchronous gaps are implicitly storing 
+`BuildContext` and are some of the easiest to overlook when writing code.
 
-When a `BuildContext` is used, its `mounted` property must be checked after an
-asynchronous gap.
+
+In any `Widget` it is possible to store the data related to the `BuildContext`
+before the async gap.
 
 **BAD:**
 ```dart
@@ -37,6 +39,20 @@ void onButtonTapped(BuildContext context) {
   Navigator.of(context).pop();
 }
 ```
+
+**GOOD:**
+```dart
+void onButtonTapped() async {
+  final navigator = Navigator.of(context);
+  await Future.delayed(const Duration(seconds: 1));
+
+  navigator.pop();
+}
+```
+
+**EXCEPTION:**
+In a `StatefulWidget` it is possible to check the `State`'s `mounted` property, after an
+asynchronous gap and before using `BuildContext`.
 
 **GOOD:**
 ```dart


bwilkerson:
@pq @goderbauer

pq (approved):

pq:
Thanks @fourman-alex!

Could you add yourself to [AUTHORS](https://github.com/dart-lang/linter/blob/main/AUTHORS) while you're at it?

Content looks good to me but will defer to @goderbauer.

Again, thanks!

lib/src/rules/use_build_context_synchronously.dart
```diff
@@ -38,6 +40,20 @@ void onButtonTapped(BuildContext context) {
...
+  final navigator = Navigator.of(context);
+  await Future.delayed(const Duration(seconds: 1));
+
+  navigator.pop();
+}
+```
+
+**EXCEPTION:**
+In a `StatefulWidget` it is possible to check the `State`'s `mounted` property, after an
+asynchronous gap and before using `BuildContext`.
```
goderbauer, line 55: This does not match the example below where `mounted` is called on the `BuildContext` and not the `State`.

This statement is also confusing. With the latest version of Flutter, `mounted` is a property of `BuildContext` and that one should be used over the `State` property as it is also available in StatelessWidgets.

Please remove this and replace this with a note that starting with Flutter version xxx, one can also check the mounted property on the BuildContext. We could then also have a section that talks about older versions of Flutter where this was only available on the State. But we should make clear that going forward `BuildContext.mounted` is the right way to do these checks.

lib/src/rules/use_build_context_synchronously.dart
```diff
@@ -38,6 +40,20 @@ void onButtonTapped(BuildContext context) {
...
 
+**GOOD:**
+```dart
+void onButtonTapped() async {
+  final navigator = Navigator.of(context);
+  await Future.delayed(const Duration(seconds: 1));
+
+  navigator.pop();
+}
+```
```
goderbauer, line 51: This is NOT good practice. The navigator could now become invalid during the async gap (the `await`) and your code crashes. You should not hold on to the navigator like that and instead look it up after the async gap and after checking that the context is still valid.

lib/src/rules/use_build_context_synchronously.dart
```diff
@@ -17,11 +17,13 @@ const _details = r'''
 **DON'T** use BuildContext across asynchronous gaps.
 
 Storing `BuildContext` for later usage can easily lead to difficult to diagnose
-crashes. Asynchronous gaps are implicitly storing `BuildContext` and are some of
-the easiest to overlook when writing code.
+crashes, because it can change its state during the async gap and no longer be 
+asociated to an `Element`. Asynchronous gaps are implicitly storing
```
goderbauer, line 21: typo: associated

lib/src/rules/use_build_context_synchronously.dart
```diff
@@ -17,11 +17,13 @@ const _details = r'''
 **DON'T** use BuildContext across asynchronous gaps.
 
 Storing `BuildContext` for later usage can easily lead to difficult to diagnose
-crashes. Asynchronous gaps are implicitly storing `BuildContext` and are some of
-the easiest to overlook when writing code.
+crashes, because it can change its state during the async gap and no longer be
```
goderbauer, line 20: instead of talking about changing state, let's just say that the BuildContext becomes invalid because it may no longer be associated with an Element. A `BuildContext` doesn't have a state per-se.

lib/src/rules/use_build_context_synchronously.dart
```diff
@@ -17,11 +17,13 @@ const _details = r'''
...
-the easiest to overlook when writing code.
+crashes, because it can change its state during the async gap and no longer be 
+asociated to an `Element`. Asynchronous gaps are implicitly storing 
+`BuildContext` and are some of the easiest to overlook when writing code.
 
-When a `BuildContext` is used, its `mounted` property must be checked after an
-asynchronous gap.
+
+In any `Widget` it is possible to store the data related to the `BuildContext`
+before the async gap.
```
goderbauer, line 26: This sentence is unclear to me. What data is this talking about?

lib/src/rules/use_build_context_synchronously.dart
```diff
@@ -17,11 +17,13 @@ const _details = r'''
...
 
 Storing `BuildContext` for later usage can easily lead to difficult to diagnose
-crashes. Asynchronous gaps are implicitly storing `BuildContext` and are some of
-the easiest to overlook when writing code.
+crashes, because it can change its state during the async gap and no longer be 
+asociated to an `Element`. Asynchronous gaps are implicitly storing 
+`BuildContext` and are some of the easiest to overlook when writing code.
 
-When a `BuildContext` is used, its `mounted` property must be checked after an
-asynchronous gap.
```
goderbauer, line 24: Please keep this advice. It is the number one solution for fixing this lint. 

(Like I said in another comment, it only applies to newer versions of flutter, older versions only provide the `mounted` property on the `State` and it should be checked there instead if one uses an older version of flutter.)

goderbauer (changes requested):

goderbauer:
For context, `BuildContext.mounted` was added in https://github.com/flutter/flutter/pull/111619. Looks like that change has not been included yet in a stable release, but it will be available in the next stable release and going forward it will be the right way to fix this lint. Please do not remove that advice from the lint description here. However, adding a note that this only applies to Flutter versions after 3.4.0-26.0.pre is probably a good idea. For older versions, `State.mounted` should be used instead.

fourman-alex:
Thanks @goderbauer for the review. 
I think that just knowing that `BuildContext.mounted` is about to be available in the stable release makes this lint correct for these versions. I think that in this case I will just add a clarification on the lack of `BuildContext.mounted` in older versions and the existence of `State.mounted` as a workaround

goderbauer:
Closing this since the feature is now released on stable.

PR #3951, opened 2022-12-29
Consolidate avoid_private_typedef tests as reflective tests

srawlins:
This consolidates both the simple tests and the integration tests.

The integration tests had good coverage of various 'part' situations, which required an additional helper function for the reflective tests.


pq (approved):
💎

PR #3955, opened 2023-01-03
Move overridden_fields tests to reflective tests

srawlins:
There are still overridden_fields tests in `test_data/rules`, but this at least moves the test from `test_data/integration`.


pq (approved):
w00t!

PR #3958, opened 2023-01-04
remove implicit casts option use

pq:
/cc @srawlins @bwilkerson

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -1,8 +1,6 @@
 include: package:lints/recommended.yaml
 
 analyzer:
-  strong-mode:
-    implicit-casts: false
   exclude:
     # ignore vscode .history files
     - .history/**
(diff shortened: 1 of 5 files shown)


analysis_options.yaml
```diff
@@ -1,8 +1,6 @@
 include: package:lints/recommended.yaml
 
 analyzer:
-  strong-mode:
-    implicit-casts: false
```
srawlins, line 5: if you want you can switch to:

```yaml
analyzer:
  language:
    strict-casts: true
```

Should be nearly the same

srawlins (approved):

analysis_options.yaml
pq: Oh!  I do.  And fancy that, there's even a quick fix!  😎

PR #3962, opened 2023-01-05
[wip] document lint lifecycle

### LINKED ISSUE #3012: 📝 "Life of a lint" lifecycle doc (Ideation → Proposal → Acceptance → Implementation)

### LINKED ISSUE #58539: 📝 "Life of a lint" lifecycle doc (Ideation → Proposal → Acceptance → Implementation)

pq:
Part and parcel to dart-lang/sdk#58947 and dart-lang/sdk#58948, the start of some lifecycle docs.

Feedback welcome!

Fixes dart-lang/sdk#58539.

/cc @bwilkerson @srawlins @MaryaBelanger

### DIFF
--- a/doc/lint-lifecycle.md
+++ b/doc/lint-lifecycle.md
@@ -10,6 +10,8 @@ A lint begins its life as a **proposal** which is _pending_ until _accepted_.
 
 After discussion and sufficient agreement within the team, a proposed lint becomes **accepted**.
 
+An accepted lint is ready for implementation.
+
 ### Experimental
 
 Lints can be marked experimental. 
--- a/doc/lint-lifecycle.md
+++ b/doc/lint-lifecycle.md
@@ -1,0 +1,64 @@
+# The Life of a Lint
+
+## Lint States
+
+### Proposed
+
+A lint begins its life as a **proposal** which is _pending_ until _accepted_.
+
+### Accepted
+
+After discussion and sufficient agreement within the team, a proposed lint becomes **accepted**.
+
+### Experimental
+
+Lints can be marked experimental. 
+
+Reasons a lint might be marked experimental include a lint
+
+* being tentatively introduced
+* have unknown value
+* have incomplete semantics
+* have outstanding (fixable) false positives or
+* might be known to be temporary.
+
+### Stable
+
+Stable lints are lints whose semantics and implementation are considered complete. False
+positives (that are not known, accepted and documentted limitations of lint semantics) are
+considered bugs and fixing them should be prioritized. False negatives may be bugs or
+enhancements (depending on semantics).
+
+### Deprecated
+
+Deprecated lints are lints that we plan to remove.
+
+Reasons for deprecation include:
+
+* semantics that don't make sense with current language semantics (e.g., null-safety)
+* stale advice
+* poor performance or
+* poor user experience (e.g., too many false positives).
+
+### Removed
+
+Lints we no longer support are removed.
+
+In general removal is preceded by a period of deprecation.
+
+## State Transitions
+
+### Deprecation
+
+Implemented lints can be deprecated.
+
+Deprecating lints that are in common lint sets (e.g., in [package:lints](https://github.com/dart-lang/lints)
+can be impactful so should be done with care.
+
+### Marking Stable
+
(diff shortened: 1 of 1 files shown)


doc/lint-lifecycle.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Lint States
+
+### Proposed
+
+A lint begins its life as a **proposal** which is _pending_ until _accepted_.
+
+### Accepted
+
+After discussion and sufficient agreement within the team, a proposed lint becomes **accepted**.
```
srawlins, line 11: Do we want to say something here about details being hashed out? A problem with some of the older lint rules (and maybe some newer 🤣 ) is that they're really underspecified when they're implemented. So then people start using them, and raise new issues, which require whole new cycles of discussion, which might result in certain behavior being called false positives or false negatives, which are then again expensive to fix.

Alternatively, this could be a requirement for a lint moving to **stable**. WDYT?

doc/lint-lifecycle.md
```diff
@@ -0,0 +1,66 @@
...
+### Deprecated
+
+Deprecated lints are lints that we plan to remove.
+
+Reasons for deprecation include:
+
+* semantics that don't make sense with current language semantics (e.g., null-safety)
+* stale advice
+* poor performance or
+* poor user experience (e.g., too many false positives).
```
srawlins, line 43: You really put a lot of thought into this. I love these lists.

doc/lint-lifecycle.md
```diff
@@ -0,0 +1,66 @@
...
+Deprecating lints that are in common lint sets (e.g., in [package:lints](https://github.com/dart-lang/lints)
+can be impactful so should be done with care.
+
+### Marking Stable
+
+Experimental lints should aspire to be stable. An experimental lint is a candidate for stable when it has
+
+* complete semantics
+* complete implementation (with no false-positives)
+* established long-term value (e.g., inclusion in a recommended lint set)
```
srawlins, line 66: Well... I don't know about how the recommended lint sets work. But it seems that the requirements would go the other way: if someone wanted an experimental lint rule to be included in a recommended lint set, the response would be that it needs to meet the qualifications of being stable before it can be recommended.

srawlins (approved):

doc/lint-lifecycle.md
pq: Sorry, I meant _desire_ to be included.  That said, folks in general don't seem to pay much attention to these states and so they're if anything soft signals.  (But maybe we can change that?)

doc/lint-lifecycle.md
pq: Yeah.  The truth is this is pretty ad hoc.  I'm not sure how formal we want to make it but I agree that getting more crisp about defining semantics *before* landing/accepting a lint would be good in general.  In practice, maybe this means prototyping and we might encourage that as legitimate too.

bwilkerson: I'd like to see a fully specified design be part of the criteria for moving to accepted.

But even if we do that, there will still be questions that arise after people start using the lint, so it probably needs to also be a requirement for moving to stable.

doc/lint-lifecycle.md
```diff
@@ -0,0 +1,66 @@
...
+
+### Removed
+
+Lints we no longer support are removed.
+
+In general removal is preceded by a period of deprecation.
+
+## State Transitions
+
+### Deprecation
```
bwilkerson, line 53: Should this be moved under "Deprecated"? It feels like a discussion about how lints get moved to that state, so it might be something we can combine.

Also, do we want to move lints from experimental to deprecated before removing them, or is the ability to remove them without notice part of what makes them experimental?

doc/lint-lifecycle.md
bwilkerson: I don't think that a desire to get a lint into the core or recommended lint sets should be a reason to make it stable. But I do think that a commitment on our part to support the lint is part of the criteria.

Also, similar question as above: should this section be moved under "Stable"?

bwilkerson (approved):
Overall, this looks good!

pq:
Thanks for all the feedback!  There are still a few unresolved issues but I think this is a good start.

Tracking issue for refinements: https://github.com/dart-lang/linter/issues/4100.

Issue #3908, opened 2022-12-14
`use_if_null_to_convert_nulls_to_bools` points to outdated guidance


MaryaBelanger:
[This lint](https://github.com/dart-lang/linter/blob/main/lib/src/rules/use_if_null_to_convert_nulls_to_bools.dart) refers to [a section in Effect Dart](https://dart.dev/guides/language/effective-dart/usage#prefer-using--to-convert-null-to-a-boolean-value) that is [changing](https://github.com/dart-lang/site-www/pull/4423) quite a bit (based on [this issue](https://github.com/dart-lang/site-www/issues/4388)), to the point where the lint using it as a reference doesn't really make sense.  

To summarize, the guidance will no longer be "PREFER using `??` to convert `null` to a boolean value", but just "DON'T use `true` or `false` in equality operations". 
- `if (nullableBool ?? false)` isn't necessarily "good" (like the lint promotes) because "Using a null-aware operator on a variable inside a condition doesn’t promote the variable to a non-nullable type". 
- Is "converting" null to boolean really even a use-case? The new guidance is more precise and applicable

The lint should change its name, description, or just not point to Effective Dart anymore.

bwilkerson:
I'm confused. The advice appears to be to use `??` when a `bool?` variable is used as a condition, which is what I think this lint is encouraging. So I think the reference back to Effective Dart is still valid.

But if the name of the advice has changed then we might want to rename the lint to make it easier for users to find.

@pq

MaryaBelanger:
@bwilkerson My mistake, I didn't explain that well. Yes, the name of the advice is changing and renaming the lint would be ideal. 

`use_if_null` is no longer the main point, because while it's still "good", the better option is an explicit `!= null` check

And `to_convert_nulls_to_bools` is no longer the main point. It's now just the more general idea that using boolean literals in equality expressions is never a good idea (null check or not). It encompasses the old lint but includes more cases.

Here's [the staged new section](https://dart-dev--pr4423-effective-operators-pi52gp0c.web.app/guides/language/effective-dart/usage#dont-use-true-or-false-in-equality-operations) so it's easier to  get the full picture. 

I don't know what the lint name should change to, though.

pq:
> I'm confused. 

[...]

PR #3973, opened 2023-01-09
Lint about comparing booleans to boolean literals

### LINKED ISSUE #3005: Lint that warns about comparing boolean expressions against literals
Comparing boolean expressions against a boolean literal is almost always a code smell. Less experienced developers are more likely to perform such comparisons. Therefore, I think it will be beneficial to have a lint that warns against such comparisons.

In my opinion, the lint should be applied inside all conditional expressions and maybe even all comparisons.
```dart
if (isDigit() == true) {
   ...
}
```
```dart
while (isDigit() == true) {
   ...
}
```
```dart
final foo = isDigit()== true ? a : b;
```

The lint would also need to take into account if the boolean literal is on the LHS of the comparison.

oprypin:
This lints against any expression like `x == true` or `false == x`.
It kicks in only if the variable is of type `bool`, non-nullable.

Fixes #3005
Ref https://github.com/dart-lang/linter/issues/3005#issuecomment-941293191
Ref https://github.com/dart-lang/linter/issues/3908#issuecomment-1352445021

### DIFF
--- a/lib/src/rules/no_literal_bool_comparisons.dart
+++ b/lib/src/rules/no_literal_bool_comparisons.dart
@@ -1,0 +1,76 @@
+// Copyright (c) 2023, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/analysis/features.dart';
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/type.dart';
+
+import '../analyzer.dart';
+
+const _desc = r"Don't compare booleans to boolean literals.";
+
+const _details = r'''
+From [Effective Dart](https://dart.dev/guides/language/effective-dart/usage#dont-use-true-or-false-in-equality-operations):
+
+**DON'T** use `true` or `false` in equality operations.
+
+**BAD:**
+```dart
+if (someBool == true) {
+}
+while (someBool == false) {
+}
+```
+
+**GOOD:**
+```dart
+if (someBool) {
+}
+while (!someBool) {
+}
+```
+''';
+
+class NoLiteralBoolComparisons extends LintRule {
+  NoLiteralBoolComparisons()
+      : super(
+          name: 'no_literal_bool_comparisons',
+          description: _desc,
+          details: _details,
+          group: Group.style,
+        );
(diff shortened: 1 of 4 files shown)


lib/src/rules/no_literal_bool_comparisons.dart
```diff
@@ -0,0 +1,76 @@
...
+import 'package:analyzer/dart/analysis/features.dart';
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/type.dart';
+
+import '../analyzer.dart';
+
+const _desc = r"Don't compare booleans to boolean literals.";
+
+const _details = r'''
```
bwilkerson, line 14: We should make it explicit in the docs that the lint doesn't catch cases where `someBool` is nullable.

lib/src/rules/no_literal_bool_comparisons.dart
```diff
@@ -0,0 +1,76 @@
...
+}
+
+class _Visitor extends SimpleAstVisitor<void> {
+  final LintRule rule;
+  final LinterContext context;
+
+  _Visitor(this.rule, this.context);
+
+  @override
+  void visitBinaryExpression(BinaryExpression node) {
```
bwilkerson, line 62: This also needs to check that the operator is either `==` or `!=`. This node is used for all binary operators, including `&&` and `||` (which arguably are just as bad, but I don't think this lint is intended to catch those cases).

There should also be at least one test to show that we don't report where we shouldn't.

lib/src/rules/no_literal_bool_comparisons.dart
oprypin: Done

lib/src/rules/no_literal_bool_comparisons.dart
oprypin: Oops, what a huge omission.
Done.

bwilkerson: It's easy to miss. `AssignmentExpression` has a similar issue; it's easy to forget about compound assignments.

bwilkerson (approved):

pq (approved):
💎

oprypin:
Ah I will now just add a test that this works for any expression, not just a variable.

pq:
@oprypin: is this ready to land?

oprypin:
Yes

PR #3980, opened 2023-01-09
enable `dangling_library_doc_comments`

pq:
/cc @bwilkerson @srawlins


lib/src/ast.dart
```diff
@@ -3,6 +3,8 @@
 // BSD-style license that can be found in the LICENSE file.
 
 /// Common AST helpers.
+library;
```
srawlins, line 6: Hey it caught one! 🎉

srawlins (approved):

PR #3981, opened 2023-01-09
+ unnecessary_breaks

### LINKED ISSUE #3671: add new lint: `unnecessary_breaks`
From the [spec](https://github.com/dart-lang/language/blob/master/accepted/future-releases/0546-patterns/feature-specification.md#implicit-break):

> this proposal removes the requirement that each non-empty case body definitely exit. Instead, a non-empty case body implicitly jumps to the end of the switch after completion.

We might consider linting on unnecessary `break`s.

pq:
First crack at a new `unnecessary_breaks` lint.

Feedback and other cases (hah) welcome.

Fixes #3671

/cc @bwilkerson 

/fyi @MaryaBelanger this is what docs typically start as

### DIFF
--- a/lib/src/rules/unnecessary_breaks.dart
+++ b/lib/src/rules/unnecessary_breaks.dart
@@ -2,6 +2,7 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'package:analyzer/dart/analysis/features.dart';
 import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
 
@@ -73,6 +74,7 @@ class UnnecessaryBreaks extends LintRule {
   @override
   void registerNodeProcessors(
       NodeLintRegistry registry, LinterContext context) {
+    if (!context.isEnabled(Feature.patterns)) return;
     var visitor = _Visitor(this);
     registry.addBreakStatement(this, visitor);
   }
--- a/lib/src/rules/unnecessary_breaks.dart
+++ b/lib/src/rules/unnecessary_breaks.dart
@@ -13,7 +13,8 @@ const _desc = r"Don't use explicit `breaks`s when a break is implied.";
 const _details = r'''
 Only use a `break` in a non-empty switch case statement if you need to break
 before the end of the case body.  Dart does not support fallthrough execution
-for non-empty cases, so these break statements are unnecessary.
+for non-empty cases, so `break`s at the end of non-empty switch case statements
+are unnecessary.
 
 **BAD:**
 ```dart
--- a/lib/src/rules/unnecessary_breaks.dart
+++ b/lib/src/rules/unnecessary_breaks.dart
@@ -1,0 +1,90 @@
+// Copyright (c) 2023, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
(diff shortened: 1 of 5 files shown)


lib/src/rules/unnecessary_breaks.dart
```diff
@@ -0,0 +1,90 @@
...
+switch (1) {
+  case 1:
+    break;
+  case 2:
+    print("just two");
+}
+```
+''';
+
+class UnnecessaryBreaks extends LintRule {
```
bwilkerson, line 58: Consider defining a lint code from the outset that has both a problem and correction message.

lib/src/rules/unnecessary_breaks.dart
```diff
@@ -0,0 +1,90 @@
...
+  }
+}
+
+class _Visitor extends SimpleAstVisitor {
+  final LintRule rule;
+
+  _Visitor(this.rule);
+
+  @override
+  visitBreakStatement(BreakStatement node) {
```
bwilkerson, line 80: We probably want to ensure that the break either
- has no label, or
- has a label associated with the enclosing switch statement.

Obviously, these cases need tests.

test/rules/unnecessary_breaks_test.dart
```diff
@@ -0,0 +1,65 @@
...
+    case 1:
+      break;
+    case 2:
+      f();
+  }
+}
+''');
+  }
+
+  test_switchPatternCase_notLast_ok() async {
```
bwilkerson, line 52: I think the case being tested here is not that the break isn't the last statement, but that the break isn't a direct child of the switch pattern case clause.

A test that the break isn't the last statement would look like this:
```
case 1:
  break;
  f(true);
```

which would also have a dead code warning. Both would be good tests.

pq: Good catch.  Updated.

lib/src/rules/unnecessary_breaks.dart
```diff
@@ -0,0 +1,98 @@
...
+            details: _details,
+            group: Group.style);
+
+  @override
+  LintCode get lintCode => code;
+
+  @override
+  void registerNodeProcessors(
+      NodeLintRegistry registry, LinterContext context) {
+    var visitor = _Visitor(this);
```
bwilkerson, line 76: I think we want to return here if 'patterns' isn't enabled.

test/rules/unnecessary_breaks_test.dart
```diff
@@ -0,0 +1,98 @@
...
+import '../rule_test_support.dart';
+
+main() {
+  defineReflectiveSuite(() {
+    defineReflectiveTests(UnnecessaryBreaksTest);
+  });
+}
+
+@reflectiveTest
+class UnnecessaryBreaksTest extends LintRuleTest {
```
bwilkerson, line 16: We should add a test for a library opted back to 2.19 to ensure it doesn't fire for older code.

bwilkerson (approved):

lib/src/rules/unnecessary_breaks.dart
pq: Good call.  Since we'll never encounter a `SwitchPatternCase` w/o patterns, this shouldn't over-report as-is but it's doing needless work.  (And a check here would make it more clear for sure.)

test/rules/unnecessary_breaks_test.dart
pq: Agreed.  Done.

PR #3984, opened 2023-01-10
Add lint codes to a few more lints

bwilkerson:
After this, there are only 7 lints remaining that don't have a `LintCode` defined.

### DIFF
--- a/lib/src/rules/diagnostic_describe_all_properties.dart
+++ b/lib/src/rules/diagnostic_describe_all_properties.dart
@@ -46,7 +46,7 @@ class Absorber extends Widget {
     properties.add(DiagnosticsProperty<bool>('absorbing', absorbing));
     // Missing reference to ignoringSemantics
   }
-}  
+}
 ```
 
 **GOOD:**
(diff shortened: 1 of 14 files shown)


lib/src/rules/diagnostic_describe_all_properties.dart
```diff
@@ -62,11 +62,17 @@ class Absorber extends Widget {
...
+}
 ```
 ''';
 
 class DiagnosticsDescribeAllProperties extends LintRule {
+      'diagnostic_describe_all_properties',
+      "The public property isn't described by either 'debugFillProperties' or "
+          "'debugDescribeChildren'.",
+      correctionMessage: 'Try described the property.');
```
srawlins, line 74: "describing"?

lib/src/rules/discarded_futures.dart
```diff
@@ -44,13 +44,22 @@ Future<void> createDir(String path) async {}
 ''';
 
 class DiscardedFutures extends LintRule {
+  static const LintCode code = LintCode('discarded_futures',
+      "Asynchronous function invoked in a non-'async' function.",
+      correctionMessage:
+          "Try converting the enclosing function to be 'async' and then "
+          "'await' the future");
```
srawlins, line 51: Should Future be capitalized?

srawlins (approved):

lib/src/rules/prefer_final_locals.dart
```diff
@@ -46,6 +46,10 @@ void mutableCase() {
 ''';
 
 class PreferFinalLocals extends LintRule {
+  static const LintCode code = LintCode(
+      'prefer_final_locals', 'Local variables should be final.',
+      correctionMessage: 'Try making the variables final.');
```
pq, line 51: I think this is reported on each variable, in which case maybe 

    'Try making the variable final.'

pq (approved):

lib/src/rules/diagnostic_describe_all_properties.dart
bwilkerson: Done.

lib/src/rules/discarded_futures.dart
bwilkerson: In other docs I've written (not sure whether there's a standard), I've capitalized "Future" when referring to the class, but not when referring to the concept. This is a reference to the concept, hence the lowercase letter.

lib/src/rules/prefer_final_locals.dart
bwilkerson: Done.

PR #3986, opened 2023-01-11
update rule template stub to define a LintCode

### LINKED ISSUE #3985: update `rule.dart` to add a `LintCode` field to rule stubs
We might also consider generating a unit test stub (rather than the old `test_data` approach).

pq:
Fixes #3985

/cc @bwilkerson

### DIFF
--- a/tool/rule.dart
+++ b/tool/rule.dart
@@ -128,12 +128,19 @@ const _details = r'''
 ''';
 
 class $className extends LintRule {
+  static const LintCode code = LintCode(
+      '$ruleName', '<add problem message here>',
+      correctionMessage: '<add correction message here>');
+
   $className()
       : super(
             name: '$ruleName',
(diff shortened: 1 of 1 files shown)


tool/rule.dart
```diff
@@ -128,13 +128,20 @@ const _details = r'''
 ''';
 
 class $className extends LintRule {
+  static const LintCode code = LintCode(
```
bwilkerson, line 131: We might want to add documentation somewhere for how to add more than one code, but this looks good.

bwilkerson (approved):

tool/rule.dart
pq: In fact, we might inline some docs here in general.  (Like the Flutter templates.)  Maybe a link to your diagnostic writing doc?

/fyi @MaryaBelanger

PR #3991, opened 2023-01-16
Refactor leak detector

srawlins:
* Add some doc comments.
* In `_buildVariableReporter`, in the first loop over DartTypePredicates, extract out the null check for `variable`'s `declaredElement`, which is unrelated to the loop.
* In `_buildVariableReporter`, `validators` was a List of Iterables, whose elements were never used; instead, the List was added to in various ways, and then we asked if all of the Iterables were empty. This is a long-winded way of saying `List<bool>`, so I make each of the functions which added to `validators` instead return `bool`. Also, we don't need to carry out _all_ of the checks. If any of these checks returns false, we can just exit `_buildVariableReporter`, without reporting any lint.
  * Some of the functions should now be renamed, but I didn't want to confuse in the PR by adding a lot of re-ordering, so I thought I'd do that in a follow-up PR.
* Many functions take a `VariableDeclaration`, query whether the `declaredElement` is non-null, and then only concern themselves with the `declaredElement`. To reduce the number of `declaredElement` calls and null checks, I change the following functions to accept a `VariableElement`: all `_PredicateBuilder` functions, `_findMethodInvocationWithVariableAsArgument`, `_isElementEqualToVariable`, `_isInvocationThroughCascadeExpression`, `_isPropertyAccessThroughThis`,
`_isSimpleIdentifierElementEqualToVariable`, and `_isPostfixExpressionOperandEqualToVariable`.
* `_hasMatch` used `fold` as a complicated implementation of `any`, so I switch to use `any`.

### DIFF
--- a/lib/src/util/leak_detector_visitor.dart
+++ b/lib/src/util/leak_detector_visitor.dart
@@ -12,45 +12,52 @@ import '../analyzer.dart';
 import '../ast.dart';
 import '../extensions.dart';
 
+/// Returns a predicate that returns true for a node `n` if `n` is a
+/// [ConstructorFieldInitializer] initializing [v].
 _Predicate _hasConstructorFieldInitializers(
         VariableDeclaration v) =>
     (AstNode n) =>
         n is ConstructorFieldInitializer &&
         n.fieldName.staticElement == v.declaredElement;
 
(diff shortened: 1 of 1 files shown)


pq (approved):
Awesome.  Thanks!

PR #3994, opened 2023-01-19
Rename methods in leak detector and sort

srawlins:
If we're not into the sorting, I can un-sort.


pq (approved):
I'm good with sorting (as long as it's consistent w/ the DAS "sort members" action).

srawlins:
Yep, that's the command I ran. 👍

PR #3996, opened 2023-01-19
flip the `noSoundNullSafety` flag for tests

pq:
/cc @bwilkerson

### DIFF
--- a/test/integration/always_require_non_null_named_parameters.dart
+++ b/test/integration/always_require_non_null_named_parameters.dart
@@ -5,6 +5,7 @@
 import 'dart:io';
 
 import 'package:analyzer/src/lint/io.dart';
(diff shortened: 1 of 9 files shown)


bwilkerson (approved):
This all looks fine to me, but in some cases it isn't clear to me why the tests would need to opt out of null safety in order to be valid. If we're going to have to convert the tests at some point, maybe we should consider doing so now. But if we don't want to do that, then we might need to consider running the tests in both modes for some of these lints.

But I'm probably second-guessing you, and I don't need to do that.

pq:
Great points! I did look at doing conversions now and found that a number of tests will just go away entirely and figured it would be easier to just convert what was left once that work was completed.

We might also do a bit of proactive conversion as we continue to convert integration tests to a unit test style (fyi @srawlins).

PR #3997, opened 2023-01-19
Move directives_ordering tests from integration to reflective

srawlins:


test/rules/directives_ordering_test.dart
```diff
@@ -0,0 +1,277 @@
+// Copyright (c) 2022, the Dart project authors. Please see the AUTHORS file
```
pq, line 1: => 2023?

pq (approved):
🎉🎉🎉

test/rules/directives_ordering_test.dart
srawlins: Done!

PR #3998, opened 2023-01-19
Fix documentation for `omit_local_variable_types`

Rexios80:
# Description

Not sure what was up with the old documentation, but that's definitely not how this lint works in its current state

### DIFF
--- a/lib/src/rules/omit_local_variable_types.dart
+++ b/lib/src/rules/omit_local_variable_types.dart
@@ -54,7 +54,7 @@ annotate the variable with the type you want.
 **GOOD:**
 ```dart
 Widget build(BuildContext context) {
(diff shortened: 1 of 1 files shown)


parlough:
We use this syntax to highlight specific code in analyzer diagnostics, but we don't support it for linter rules anywhere, so removing this makes sense for now. Nice catch :)

bwilkerson (approved):

pq (approved):

pq:
Thanks!

Rexios80:
Is seems like the action to deploy this change failed for some reason: https://github.com/dart-lang/linter/actions/runs/3963114079

Should it be re-ran?

pq:
Thanks.  It was a transient error and seems to have cleared itself up.

Issue #3880, opened 2022-12-06
remove deprecated lints in 3.0


pq:
3.0 would be a good time to remove deprecated lints.

Candidates:

- [x] #4005
- [x] #4009
- [x] #4012
- [x] #4013
- [x] #4014
- [x] #4020
  * (replaced by https://github.com/dart-lang/sdk/issues/50694)

srawlins:
I think [enable_null_safety](https://dart-lang.github.io/linter/lints/enable_null_safety.html) should also be removed in 3.0. This rule is being replaced by https://github.com/dart-lang/sdk/issues/50694.

srawlins:
I'd also like to remove [`avoid_null_checks_in_equality_operators`](https://dart-lang.github.io/linter/lints/avoid_null_checks_in_equality_operators.html), which mostly doesn't make sense under null safety, with a proposed rule: [`equality_operator_override_with_nullable_parameter `](https://github.com/dart-lang/linter/issues/3906)

srawlins:
None of these have to do with null safety, so they can be deleted even before Google internal is null safe.

(The last one, `enable_null_safety`, _does_ obviously pertain to null safety, but is not used internally, so it can also be deleted.)

mit-mit:
Can we also remove `default_list_constructor` ?

mit-mit:
Also, how is the present issue different from https://github.com/dart-lang/lints/issues/95 ?

pq:
> Can we also remove default_list_constructor ?

I think so?  @bwilkerson?

> Also, how is the present issue different from https://github.com/dart-lang/lints/issues/95 ?

This issue tracks the removal of the lint implementations from the linter.  https://github.com/dart-lang/lints/issues/95 is about removing newly removed/deprecated lints from the published lint sets.

bwilkerson:
The `default_list_constructor` diagnostic is a compile-time error, not a lint, but I believe that we can remove it. It won't be reported in 3.0 because it looks for invocations of the default constructor, but the default constructor is gone so there can't be an invocation of it (the error instead is "The class 'List' doesn't have an unnamed constructor.").

pq:
Closing linter side as complete.

PR #4000, opened 2023-01-19
"removed" rule doc codegen WIP

pq:
Room for improvement but a start.  

Adds "removed" state context to generated rule docs.

**Example:**

image

@parlough: I'm curious if you have some thoughts here.

We'd like to keep generating docs for removed lints (to keep backlinks working) but would like it to be clear at a glance that a rule has been removed.

Any ideas most welcome!

See: #3880 

/cc @bwilkerson @parlough

### DIFF
--- a/tool/doc.dart
+++ b/tool/doc.dart
@@ -583,8 +583,8 @@ class RuleHtmlGenerator {
   String get detailsHeader {
     if (state.isRemoved) {
       var since = state.since;
-      var sinceDetail = since != null ? 'in Dart $since.' : '';
-      return 'Removed $sinceDetail';
+      var sinceDetail = since != null ? ' in Dart $since.' : '';
+      return '<p style="font-size:30px"><strong>Removed$sinceDetail</strong></p>';
     }
     return '';
   }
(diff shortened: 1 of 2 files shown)


parlough:
Makes sense and looks good to me :)

Two potential suggestions:

- Rename "Maturity" to "State"
- Strikethrough the lint name on the index to make it more clear or even consider removing it from that page. I feel the main value of the page is to find new lints but they shouldn't add this one.

tool/doc.dart
```diff
@@ -628,6 +629,21 @@ class RuleHtmlGenerator {
     return 'Dart SDK: $sdkVersion • <small>(Linter $linterVersion)</small>';
   }
 
+  State get state => rule.state;
+
+  String get stateString {
+    if (state.isDeprecated) {
+      return '<span style="color:orangered;font-weight:bold;" >${state.label}</span>';
+    }
+    if (state.isRemoved) {
```
bwilkerson, line 638: Totally your call, but I prefer using an `else if` structure when only one branch can be taken.

bwilkerson (approved):

PR #4030, opened 2023-01-30
Fix `unnecessary_parenthesis` with null aware expressions.

### LINKED ISSUE #4028: `unnecessary_parenthesis` false positive with null-aware `toString`
* This fix covers most cases: #3902 by @asashour 

But seems that this case still slips through:

* ```dart
  String foo() {
    int? a;
    return (a?.abs().sign).toString(); // Unnecessary use of parentheses.
  }
  ```

* ```dart
  String foo() {
    int? a;
    return a?.abs().sign.toString(); // 'String?' can't be returned from a function with return type 'String'
  }
  ```

Probably the complicating factor is that `toString` is available on `Object?` **and** the null-aware expression has another dot after it

asashour:
Fixes #4028

# Description

`unnecessary_parenthesis` should check for an expression being `null aware` recursively, through the `target`.

### DIFF
--- a/lib/src/rules/unnecessary_parenthesis.dart
+++ b/lib/src/rules/unnecessary_parenthesis.dart
@@ -88,11 +88,7 @@ class _Visitor extends SimpleAstVisitor<void> {
   void visitParenthesizedExpression(ParenthesizedExpression node) {
     var parent = node.parent;
     var expression = node.expression;
-    if (expression is SimpleIdentifier ||
-        (expression is CascadeExpression && expression.isNullAware) ||
-        (expression is PropertyAccess && expression.isNullAware) ||
-        (expression is MethodInvocation && expression.isNullAware) ||
-        (expression is IndexExpression && expression.isNullAware)) {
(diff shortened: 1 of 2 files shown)


bwilkerson (approved):

oprypin:
Thanks. This definitely fixes the false positive.

But I noticed this also adds a new caught case. Which definitely seems to also be an improvement, but just thought I'd point it out because I'm not sure if it was expected.

Wasn't reported before, now is:
```dart
var x = (foo?.a.b.c.length) ?? 0;
```

asashour:
I think the new behavior is correct, the parenthesis are unnecessary, and they can be safely removed, or I am missing something.

```dart
class A {
  A a = A();
  A b = A();
  String c = '';
}

void f(A? foo) {
  int x = (foo?.a.b.c.length) ?? 0;
}
```

oprypin:
Sure, can be removed.

> definitely seems to also be an improvement, but just thought I'd point it out

PR #4032, opened 2023-01-31
fix lint code

pq:
/cc @bwilkerson


bwilkerson (approved):
I wonder how many of these I missed. :-(

oprypin:
I think this was the last one :)
I was seeing mismatches on the internal lint dashboard and wondering what was going on. I only now realized that this is the cause.
Just this one and invariant_booleans were the mismatches

PR #4039, opened 2023-02-01
Don't report `avoid_function_literals_in_foreach_calls` with null-aware call

### LINKED ISSUE #3354: avoid_function_literals_in_foreach_calls is unfriendly to ?.
**Describe the issue**

`forEach` can be handy when looping over a null iterator using conditional member access (`?.`):

```dart
List<MapEntry<String, String>>? list;

list?.forEach((e) {
  // use e
});
```

The quick fix tries to convert this to:

```dart
for (var e in list) {
  // use e
}
```

which is invalid code since `list` cannot be `null` in a for loop.

The alternatives that do work are verbose:

Option 1:

```dart
if (list != null) {
  for (var e in list) {
    // use e
  }
}
```

Option 2 (with `implicit_dynamic`)

```dart
for (var e in list ?? <MapEntry<String, String>>[]) {
  // use e
}
```

implicit_dynamic is itself a good idea, but it would have been better if the implicit type of `[]` were not dynamic in the first place. It seems there is enough information here to infer that.

**To Reproduce**

The previous section contains the example.

**Expected behavior**

It would be nice to have a valid quick fix here, but I think the recommended lint is also not ideal in the first place as it applies to null lists, unless we have an alternative way to write this code that is concise.

**Additional context**
None.

oprypin:
The current behavior is fully correct in null-safe mode, but in legacy mode the check for nullable type is inverted.

Not adding a test because there is no legacy-mode test file.

Closes #3354


bwilkerson:
> Not adding a test because there is no legacy-mode test file.

The preferred method for testing, going forward, is to write a test class in `test/rules` and use the reflective test loader. It's better for several reasons, but one of them is that it makes it easy to write tests against different language versions.

oprypin:
Thanks. In any case, I am not taking up the task to add a completely new test focused on legacy mode. So if this is a blocker, it will remain a blocker.

oprypin:
PTAL

pq (approved):
Thanks!

PR #4042, opened 2023-02-02
Fix `unnecessary_parenthesis`

### LINKED ISSUE #4041: `unnecessary_parenthesis` false positive with null-aware + cascade + assignment
False positive for this case:
```dart
class Foo {
  var bar = false;
  void method() {}
}

void testLint(Foo? foo) {
  (foo?..bar = true)?.method();    // Unnecessary use of parentheses.
}
```

Wrong fix that is being suggested:
```dart
void testLint(Foo? foo) {
  foo?.bar = true.method();
}
```

asashour:
Fixes #4041

`CascadeExpression` should be simply ignored in `_isNullAware` (which is done in a later part of the code).

### DIFF
--- a/test_data/rules/unnecessary_parenthesis.dart
+++ b/test_data/rules/unnecessary_parenthesis.dart
@@ -131,6 +131,10 @@ main() async {
   (a?..abs())!;
   (a?[0])!;
 
+  (a?..toString())?.abs();
+  (a?..field = true)?.setter;
+  (a?..[0] = 1)?.length;
+
   print(!({"a": "b"}["a"]!.isEmpty)); // LINT
 
   print((1 + 2)); // LINT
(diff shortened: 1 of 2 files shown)


bwilkerson:
See also https://github.com/dart-lang/linter/pull/4043

asashour:
Closed in favor of the other one.

test_data/rules/unnecessary_parenthesis.dart
```diff
@@ -131,6 +131,10 @@ main() async {
   (a?..abs())!;
   (a?[0])!;
 
+  (a?..toString())?.abs();
```
pq, line 134: @asashour: if you want to pick this up, that'd be awesome and it would be great to get these cases added to `UnnecessaryParenthesisTest`.

Sorry I didn't look at this PR before pushing ahead on #4042.  A lot of plates in the air...

pq:
Sorry to miss this one @asashour.  And thanks, as always, for the continued contributions!

asashour:
No worries, it was 3 minutes difference. Two persons were doing the same thing :)

Please go ahead with your PR, there are many other plates :)

PR #4043, opened 2023-02-02
don't report `unnecessary_parenthesis` for null-aware cascades

### LINKED ISSUE #4041: `unnecessary_parenthesis` false positive with null-aware + cascade + assignment

pq:
Fixes #4041

/cc @bwilkerson

### DIFF
--- a/lib/src/rules/unnecessary_parenthesis.dart
+++ b/lib/src/rules/unnecessary_parenthesis.dart
@@ -96,6 +96,7 @@ class _Visitor extends SimpleAstVisitor<void> {
           return;
         }
       } else if (parent is MethodInvocation) {
+        if (expression is CascadeExpression) return;
         var name = parent.methodName.name;
         if (name == 'noSuchMethod' || name == 'toString') {
           // Code like `(String).noSuchMethod()` is allowed.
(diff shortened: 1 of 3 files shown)


bwilkerson (approved):

bwilkerson:
See also https://github.com/dart-lang/linter/pull/4042

lib/src/rules/unnecessary_parenthesis.dart
```diff
@@ -96,6 +96,7 @@ class _Visitor extends SimpleAstVisitor<void> {
           return;
         }
       } else if (parent is MethodInvocation) {
+        if (expression is CascadeExpression) return;
```
asashour, line 99: This condition can be moved to the root of the method, in this case, there is a second check (in line 150 now), which should be removed.

Another solution is to use the implementation of the other PR.

pq: Ah!  This check here is scoped to just the null aware and simple identifier cases so I'm not sure it's identical to the check in 150 but I'm sure you're right that this could be tidied up.  (Lots of special cases here...)

Feel free to follow-up w/ any refinements?  It'd be great to flesh out the tests too.  (FYI @oprypin who may have more cases to consider.)

PR #4045, opened 2023-02-03
Fix `UnnecessaryParenthesis`

asashour:
This PR cover more cases for #4041

The fix in #4043 covers only the case of `parent is MethodInvocation`, but it should be generalized to be not based on the `parent` but rather that `expression is CascadeExpression`.

This PR removes the condition from `_isNullAware` and leaves the logic to reach line 150 to return without linting.


pq (approved):
💎 

Thank you!

Issue #4044, opened 2023-02-02
implement: `invalid_case_patterns`


munificent:
## ☂️ Tasks

- [x] initial implementation (for testing and quick-fix hook up) 
  * https://github.com/dart-lang/linter/pull/4047
- [ ] #4172
- [ ] #4055

---

## Description

The new patterns language feature in Dart is [a breaking change to a very small number of switch cases][break]. To help users migrate and mitigate the breakage, I think it would help to have a lint that warns on switch cases in libraries whose language version is 2.19 or lower if the case will become an error or have its semantics change when the library is upgraded to 3.0.

[break]: https://github.com/dart-lang/language/blob/master/accepted/future-releases/0546-patterns/feature-specification.md#breaking-existing-switches

Almost all existing switch cases are fine as far as I can tell. If the case is one of these constant expressions:

* Identifier (qualified or not), not named `_`
* Number literal
* String literal (possibly with interpolations)
* `null`
* `as` expression containing any of the above

Then it behaves exactly the same with patterns as it did before. There’s no breaking and nothing to migrate. Looking at a corpus of open source code, that covers >99.9% of all switch cases.

Some valid switch cases today will become compile errors in Dart 3.0:

* Set literals
* Parenthesized expressions
* Calls to `identical()`.
* Unary operator expressions `!`, `-`, or `~` (except for `-` before an integer literal, which is a valid pattern and is fine)
* Binary operator expressions `!=`, `==`, `&`, `|`, `^`, `~/`, `>>`, `>>>`, `<<`, `+`, `-`, `*`, `/`, `%`, `<`, `<=`, `>`, `>=`, `??`.
* Conditional operator `?:`
* `.length` calls on strings
* `is` and `is!` expressions

Examples of all of them:

```dart
switch (obj) {
  case {1}: // Set literal.
  case (1): // Parenthesized expression.
  case identical(1, 2): // `identical()` call.
  case -pi: // Unary operator.
  case 1 + 2: // Binary operator.
  case true ? 1 : 2: // Conditional operator.
  case 'hi'.length: // .length call.
  case i is int: // is expression.
}
```

Some valid switch cases today are also syntactically valid patterns, but the pattern matching behavior may be different from the current constant equality behavior. They are:

* **List and map literals.** A list or map literal can appear as a constant in a case:

    ```dart
    switch (obj) {
      case [1, 2]: ...
      case {'k': 'v'}: ...
    }
    ```

[...]

PR #4047, opened 2023-02-04
new lint: `invalid_case_patterns`

pq:
See: #4044.

WIP but feedback very welcome. **Especially for docs.**

Because there's a lot to say here, perhaps we could link out to something in the core dev docs?

@MaryaBelanger: thoughts?

Minimally here though, I think the correction message and description could use tightening up. (@bwilkerson? @munificent?)

As for the code, it's a WIP and possibly over-reporting.  Needless to say, input welcome there!


lib/src/rules/invalid_case_patterns.dart
```diff
@@ -0,0 +1,92 @@
...
+const _details = r'''
+Some case expressions that are valid in Dart 2.19 and below will become an error
+or have changed semantics when a library is upgraded to 3.0. This lint flags
+those expressions in order to ease migration to Dart 3.0.
+''';
+
+class InvalidCasePatterns extends LintRule {
+  static const LintCode code = LintCode(
+      'invalid_case_patterns', 'Invalid case pattern.',
+      // todo(pq): can we have a doc link here?
```
bwilkerson, line 22: I'm guessing that you mean "can we link directly to some doc other than the one generated from the `_details`. In which case the answer is that it isn't currently supported.

I think the link to the other doc should be in the `_details`. There will automatically be a link in the IDE to the doc page generated from the `_details`. That might seem like a bit of an indirection, but it's also more consistent with the way the rest of the IDE links work, and I think that consistency trumps the extra indirection.

lib/src/rules/invalid_case_patterns.dart
```diff
@@ -0,0 +1,92 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+
+import '../analyzer.dart';
+
+const _desc = r'Use case expressions that are valid in Dart 3.0.';
+
+/// todo(pq): add details or link out to a doc?
```
bwilkerson, line 12: First, that's a false dichotomy; we could have both.

But I think the answer to the question of whether to have more details here depends on how well the doc we should link to (so "yes" to the second question) does of explaining the behavior of this lint. If the doc provides important details about switch statements, but doesn't directly address the breaking changes, then we need details here. If the doc is dedicated to the breaking changes, then we might not need any extra details here.

But either way, we need to provide the kind of before and after examples that we have in the other diagnostic documentation in order to help users figure out how to migrate their code.

lib/src/rules/invalid_case_patterns.dart
```diff
@@ -0,0 +1,92 @@
...
+const _details = r'''
+Some case expressions that are valid in Dart 2.19 and below will become an error
+or have changed semantics when a library is upgraded to 3.0. This lint flags
+those expressions in order to ease migration to Dart 3.0.
+''';
+
+class InvalidCasePatterns extends LintRule {
+  static const LintCode code = LintCode(
+      'invalid_case_patterns', 'Invalid case pattern.',
```
bwilkerson, line 21: I think it's reasonable to assume that users of this lint might not be familiar with the patterns feature, in which case I think this doesn't really help much. Minimally I think we want to say something like "This expression is not valid in a 'case' clause in Dart 3.0." Maybe with a correction message that suggests reading the lint's documentation for help migrating the code.

test/rules/invalid_case_patterns_test.dart
```diff
@@ -0,0 +1,182 @@
...
+    case true ? 1 : 2: 
+  }
+}
+''', [
+      lint(43, 12),
+      error(HintCode.DEAD_CODE, 54, 1),
+    ]);
+  }
+
+  test_constConstructorCall() async {
```
bwilkerson, line 57: We also need a test in which there's an explicit `const` keyword, because I think that one is valid in 3.0.

test/rules/invalid_case_patterns_test.dart
```diff
@@ -0,0 +1,182 @@
...
+  switch (o) {
+    case ''.length: 
+  }
+}
+''', [
+      lint(43, 9),
+    ]);
+  }
+
+  test_listLiteral() async {
```
bwilkerson, line 110: It would be good to add tests for list, map, and set literals that have type arguments.

test/rules/invalid_case_patterns_test.dart
```diff
@@ -0,0 +1,182 @@
...
+''', [
+      lint(43, 3),
+    ]);
+  }
+
+  test_prefixedExpression() async {
+    await assertDiagnostics(r'''
+void f(Object o) {
+  switch (o) {
+    case -1:
```
bwilkerson, line 150: I believe that there should be no lint here, but we should have a lint for other valid constant expressions involving unary operators. Maybe also one with `-2.0` that doesn't lint.

test/rules/invalid_case_patterns_test.dart
pq: Good idea. Done.

test/rules/invalid_case_patterns_test.dart
pq: Ah, I see.  A better case might be something like:

```dart
void f() {
  const o = 1;
  switch (1) {
    case -o:
  }
}
```

with the int literal case producing no diagnostic.

As for doubles, I'm seeing [this diagnostic](https://dart.dev/tools/diagnostic-messages?utm_source=dartdev&utm_medium=redir&utm_id=diagcode&utm_content=case_expression_type_implements_equals#case_expression_type_implements_equals) so is that case something we need to lint for?

```dart
void f() {
  switch (1.0) {
    case -1.0:
  }
}
```

pq: **EDIT:** rebuilding my SDK since @bwilkerson isn't seeing this diagnostic locally.

test/rules/invalid_case_patterns_test.dart
pq: 👍  Added.  Thanks.

lib/src/rules/invalid_case_patterns.dart
```diff
@@ -59,7 +59,9 @@ class _Visitor extends SimpleAstVisitor {
         rule.reportLint(expression);
       }
     } else if (expression is PrefixExpression) {
-      rule.reportLint(expression);
+      if (expression.operand is! IntegerLiteral) {
```
oprypin, line 62: I'm guessing DoubleLiteral would also be affected?

pq: Double literals are not valid const expressions in Dart `<= 2.19` and will produce a (different) [warning](https://dart.dev/tools/diagnostic-messages?utm_source=dartdev&utm_medium=redir&utm_id=diagcode&utm_content=case_expression_type_implements_equals#case_expression_type_implements_equals).

oprypin: Oh! Sorry.

lib/src/rules/invalid_case_patterns.dart
MaryaBelanger: I'm not sure if this is the kind of input you're looking for @pq, but hopefully it helps.

I'm imagining that within the umbrella of patterns documentation in dart.dev, there will be a reference on switch cases, what works, what doesn't work, the relationship/significance of switch for patterns, etc. 

Or, that the [language tour reference on switch](https://dart.dev/guides/language/language-tour#switch-and-case) will document the new valid and invalid case patterns, agnostic of patterns, but that points to a page about the relationship between switch and patterns.

>  If the doc provides important details about switch statements, but doesn't directly address the breaking changes, then we need details here.

@bwilkerson I'm not sure what "directly addressing the breaking changes" on dart.dev would mean, like explicitly calling out that this is effective from 3.0+? Or that the result of these invalid cases is a compile time error?

> ...depends on how well the doc we should link to...does of explaining the behavior of this lint. 

If I understand correctly, "explaining the behavior of the lint" is a combination of:
- explaining the relationship between patterns and switch (why)
- the advent of that change, i.e. Dart 3 (when)
- showing everything that's invalid (what)
- alongside what _is_ valid/what should be done instead (how). 

Basically documenting #4044 in dart.dev. Would that satisfy documenting the lint "well"?

I'm not against more detail in the lint message, but I am biased towards more complete dart.dev content. So, if fleshing it out on dart.dev solves the problem of making this lint message really long, then I'm happy to make sure the doc it links to instead is a sufficient resource.

_Also important to consider...._

Patterns documentation will definitely be up _by Dart 3_, but I currently don't have a timeline more granular than that, AND it won't actually be published to the website before release day. So that's something to consider, if you need a working link before then. I could prioritize that work, or at the very least create the page and section to link to if the cutoff is very soon.

bwilkerson: > I'm not sure what "directly addressing the breaking changes" on dart.dev would mean ...

The lint is designed to catch expressions in pre-3.0 switch case clauses whose semantics will change, without a diagnostic, when the code is migrated to 3.0. Those are the breaking changes I'm referring to.

> If I understand correctly, "explaining the behavior of the lint" is a combination of: ...

That's exactly what I meant by "directly addressing" the breaking changes.

> Would that satisfy documenting the lint "well"?

Yes, that would be perfect.

> Patterns documentation will definitely be up by _Dart 3_ ...

The lint will, hopefully, ship with the next beta, so we'll want to have documentation explaining what the lint is for before we ship 3.0. We can always change the docs later to be reduced and to point to the `dart.dev` page, but I think it's important for users to understand how this helps them prepare for 3.0 sooner than the release of 3.0.

pq: Thanks @MaryaBelanger!  In the short term, I agree w/ @bwilkerson that we should make sure that the lint has docs that are complete enough that folks who try it out aren't totally lost.  In the medium term, it would be great to link out to a more complete doc on dart.dev (and what you describe sounds perfect).  The good news is the linter docs are auto-generated so it's easy for us to update and refresh them once the doc we want to link to goes live.

MaryaBelanger: Aha, with that timeline it seems like the lint docs are the best place to hold all that info (I think you both already knew that, I'm catching up haha). I'm not sure if you were looking for input on the lint itself, but now that I'm thinking about it...

I might be missing something, but since the correction message says "Try refactoring" you'd want to provide them with a solution or explanation for each potential case right? Unless you don't typically explain lint topics that come with a quick fix (but there's a few expressions that don't have a quick fix so the question still stands for those).

To avoid having to write out every expression's GOOD/BAD example, would it be possible to add a 
string token to the correction message that's populated based on the expression that the linter flagged? Please excuse my horrifying dart/pseudocode mashup, but something like: 

```dart
class InvalidCasePatterns extends LintRule {
  static const LintCode code = LintCode('invalid_case_patterns',
      "This expression is not valid in a 'case' clause in Dart 3.0.",
      correctionMessage: 'Try refactoring the expression to be valid in 3.0:' token);
...
    } else if (expression is ParenthesizedExpression) {
      token = "discard parenthesis";
      rule.reportLint(expression);
    }
```

Otherwise I'd imagine the lint details could just list out all of the affected expressions, maybe splitting them up by "compile time error/syntactically breaking" and "behavior/semantically breaking". Again, a messy mashup but just for reference, something like:

BAD: 
```dart
switch (obj) {
// syntactic error
  case {1}: // Set literal.
  case (1): // Parenthesized expression.
  case identical(1, 2): // `identical()` call.
  case -pi: // Unary operator.
  case 1 + 2: // Binary operator.
  case true ? 1 : 2: // Conditional operator.
  case 'hi'.length: // .length call.
  case i is int: // is expression.

// semantic error
  case [1, 2]: // list literal
  case {'k': 'v'}:  // map literal
  case Point(1, 2): // constant constructor call
  case _: // wildcard
  case true && false: // logic operators
}
```
GOOD:
```dart
switch (obj) {

// List literals, map literals, set literals, and constant constructor calls:
// Put const before the literal or call:

  case const [1, 2]:
  case const {'k': 'v'}:
  case const {1, 2}:
  case const Point(1, 2):

// ... etc. for the rest
```

Maybe you've already considered all of this 😅 I'm still not 100% sure how to read a lint rule's source. Anyway, I'll let you know when the Dart 3 docs start coming together and where you can link to once the content is available.

pq: > I might be missing something, but since the correction message says "Try refactoring" you'd want to provide them with a solution or explanation for each potential case right? 

You're not missing anything.  Where we can, I think we *should* provide a more specific message.  (Often we do and here we should for sure.)  TL;DR you're 💯 spot on.

> would it be possible to add a string token to the correction message that's populated based on the expression that the linter flagged?

Yep. Possible!

>  I'm not sure if you were looking for input on the lint itself, but now that I'm thinking about it...

Always.  So thank you!

test/rules/invalid_case_patterns_test.dart
```diff
@@ -0,0 +1,298 @@
...
+''', [
+      lint(42, 25),
+    ]);
+  }
+
+  test_parenthesizedExpression() async {
+    await assertDiagnostics(r'''
+void f(Object o) {
+  switch (o) {
+    case (1):
```
munificent, line 208: On further though, this should be OK. We do have parenthesized patterns too, so as long as the inner expression is a valid constant/pattern, the parentheses aren't a problem. You can remove this from the lint.

munificent (approved):
One thing then it looks to me like the tests are covering everything. Thank you for the quick turnaround!

test/rules/invalid_case_patterns_test.dart
pq: Got it.  Thanks!

PR #4051, opened 2023-02-06
fix pr labeler yaml indent

pq:
Testing.

(Trying to understand failures like: https://github.com/dart-lang/linter/actions/runs/4108840402/jobs/7089992781)

Indent is now correct but that could be a red herring.


srawlins (approved):
Love it

PR #4057, opened 2023-02-08
[only_throw_errors] don't report on values of type Never

### LINKED ISSUE #4053: [only_throw_errors] prevents explicit control flow
Consider the `only_throw_errors` lint rule and the following example:

```dart
//               vvvvv Only throw instances of classes extending either Exception or Error.
void aa() {throw foo();}
Never foo() {throw Exception();}
```

In this example only_throw_errors kicks in when a value of type Never is being thrown.

I think this goes against the general unspoken convention that explicit control flow should be preferred in Dart (_as opposed to Rust where among others, allowing the last expression in a list of statements to be returned implicitly is fine_)

I think `only_throw_errors` should not report on expressions that are known to be of type Never to allow users that would like to prefer explicit control flow (_i.e. to throw values of type Never instead of implicitly letting it jump_) to do that.

Related (in the sense that this also prefers implicit control flow): https://github.com/dart-lang/linter/issues/2352

modulovalue:
Fixes #4053

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -46,3 +46,4 @@ Nils Reichardt <hi@nils.re>
 Alex Li <alexv.525.li@gmail.com>
 Michail Smirnov <miva200267@gmail.com>
 Peter Leibiger <kuhnroyal@gmail.com>
+Modestas Valauskas <valauskasmodestas@gmail.com>
--- a/lib/src/rules/only_throw_errors.dart
+++ b/lib/src/rules/only_throw_errors.dart
@@ -56,6 +56,7 @@ bool _isThrowable(DartType? type) {
   var typeForInterfaceCheck = type?.typeForInterfaceCheck;
   return typeForInterfaceCheck == null ||
       typeForInterfaceCheck.isDynamic ||
+      type is NeverType ||
       typeForInterfaceCheck.implementsAnyInterface(_interfaceDefinitions);
 }
 
--- a/test/rules/only_throw_errors_test.dart
+++ b/test/rules/only_throw_errors_test.dart
@@ -120,4 +120,14 @@ void f() {
       lint(19, 7),
     ]);
   }
+
+  test_never() async {
+    await assertNoDiagnostics(r'''
+void f() {
+  throw e();
+}
+
+Never e() => throw Exception();
+''');
+  }
 }


pq:
I'm not hugely worried since this addresses a false positive but to be extra-safe we should test internally.

srawlins:
This should not be landed until we decide the attached issue is a bug. That should be informed by whatever guidance informed `only_throw_errors`.

pq:
I'm with @lrhn (see https://github.com/dart-lang/linter/issues/4053#issuecomment-1423314728) and am OK with this change (even if analyzer decides to report the throw as dead code down the road).

modulovalue:
> That should be informed by whatever guidance informed only_throw_errors.

Here is what I think has happened here: because this lint has existed pre null-safety, I think that this is just an edge case. Null-safety changed the bottom type from Null to Never. The fact that a value that is known to be of type Null is thrown couldn't be statically proven pre nnbd, but now, throwing a value that is known to be of type Null is a compile-time error. `Never`, as Lasse said, introduced a type that is a subtype of Error and Exception, but that does not directly implement any of their interfaces. This lint checks for whether the interface of Error or Exception are being implemented, and not for whether the thrown expression is a subtype of Error or Exception, so this edge case occurs.

> This should not be landed until we decide the attached issue is a bug

I don't think that this is a bug. I think that, given the discussion in the attached issue, it would make sense to change the implicit specification of this lint from `implements the interface of Error or Exception` to `is a subtype of Error or Exception`. 

Since the new specification causes fewer expressions to be reported (_anything that implements something is a subtype of it, but the converse is not necessarily true_), I expect this to not be a breaking change (e.g. there should be no new warnings that could break someones CI).

pq (changes requested):
Thanks for this!  Could you add yourself to `AUTHORS`?

modulovalue:
> Thanks for this! Could you add yourself to AUTHORS?

@pq Done.

pq (approved):
👍

pq:
Whoops, sorry @modulovalue: could you fix up the merge conflict?

Thanks!

modulovalue:
@pq Done.

pq:
Thanks!

PR #4059, opened 2023-02-08
convert mixed in abstract class to a proper mixin

### LINKED ISSUE #4058: Migrate mixins to be mixin declarations
With the "class modifiers" feature coming, we need to migrate all of the `abstract class` declarations that are used as mixins to be either `mixin` declarations, or possibly `mixin class` declarations.

This will likely be a blocker for enabling Dart 3 internally.

pq:
Fixes #4058

/cc @jakemac53  @bwilkerson

### DIFF
--- a/lib/src/rules/control_flow_in_finally.dart
+++ b/lib/src/rules/control_flow_in_finally.dart
@@ -112,7 +112,7 @@ class ControlFlowInFinally extends LintRule {
 /// [ControlFlowInFinally] which is in a separate rule to allow a more granular
 /// configurability given that reporting throw statements in a finally clause is
 /// controversial.
-abstract class ControlFlowInFinallyBlockReporterMixin {
+mixin ControlFlowInFinallyBlockReporterMixin {
   LintRule get rule;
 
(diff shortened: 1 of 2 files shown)


lib/src/rules/control_flow_in_finally.dart
```diff
@@ -112,7 +112,7 @@ class ControlFlowInFinally extends LintRule {
 /// [ControlFlowInFinally] which is in a separate rule to allow a more granular
 /// configurability given that reporting throw statements in a finally clause is
 /// controversial.
-abstract class ControlFlowInFinallyBlockReporterMixin {
+mixin ControlFlowInFinallyBlockReporterMixin {
```
bwilkerson, line 115: Consider, at some point, dropping `Mixin` from the name. I don't think it adds anything.

bwilkerson (approved):

lib/src/rules/control_flow_in_finally.dart
pq: Works for me.  Done!

Issue #1882, opened 2019-12-10
Lint when return type `void` is overridden by `Future<T>` for some `T`?


eernstg:
Cf. https://github.com/dart-lang/language/issues/735.

It is error-prone and not likely to be useful when the return type `void` is overridden by the return type `Future<T>` or `FutureOr<T>`, for any `T`. For example:

```dart
abstract class Bar {
  void test();
}

class Baz extends Bar {
  Future<void> test() async {
    return Future.delayed(Duration(seconds: 3), () => print('Baz'));
  }
}
```

A lint on such overrides could supplement the support for not accidentally discarding futures that we already have with `unawaited_futures`.

bwilkerson:
@MichaelRFairhurst @jcollins-g (Because this came up in a CL this morning.)

bwilkerson:
I have to ask whether this should be extended to prohibit any override of a method with a return type of `void` with a non-`void` return type. It seems similarly error-prone to ignore any returned value from an overriding method, and of marginal utility to allow an overriding method to return a value (and the latter case could always be handled with an `ignore` comment as a way of documenting that the behavior is intentional).

(That seems especially true for overrides that are marked with the `useResult` annotation, but that might be better addressed by expanding the checking associated with the annotation.)

eernstg:
The argument that usually comes up in favor of allowing a method to have return type `void`, and then be overridden by a method with some other return type, is that the overridden method may not have anything meaningful to return.

A particularly tricky case occurs when the returned result should be ignored because it cannot be interpreted correctly at the more abstract level.

Here's an extreme example where the interpretations are nearly opposite. We would probably never see anything this blatant in real life, but we could still have cases where two different overrides of a method return a result that should be interpreted _somewhat_ differently, and hence it could be useful to use the return type `void` in the case where it isn't known which interpretation is appropriate:

```dart
abstract class Action {
  // We could make it `bool doit();`, but we wouldn't know what that bool means.
  void doit();
}

class ActionThatShouldSucceed implements Action {
  bool doit() {
    ... // Do the thing.
    ... // Return true if everything was done successfully.
  }
}

class ActionThatShouldFail implements Action {
  bool doit() {
    ... // Attempt to do the thing.
    ... // Return true if various checks detected this attempt and stopped it.
  }
}
```

[...]

PR #4064, opened 2023-02-12
Add rule 'override_void_return'

srawlins:
# Description

Fixes https://github.com/dart-lang/linter/issues/1882

This rule prevents dangerous overrides, and helps to stay safe with [`unawaited_futures`](https://dart-lang.github.io/linter/lints/unawaited_futures.html).

Two things to maybe discuss:

* [ ] the name? I welcome suggestions :)
* [x] Are there edge cases to special-case? Overriding with `Future?`, `Future<Never>`, `Future<Null>`, FutureOr?`, ...? CC @eernstg

### DIFF
--- a/example/all.yaml
+++ b/example/all.yaml
@@ -74,6 +74,7 @@ linter:
     - exhaustive_cases
     - file_names
     - flutter_style_todos
+    - future_return_overriding_void_return
     - hash_and_equals
     - implementation_imports
     - implicit_call_tearoffs
--- a/example/all.yaml
+++ b/example/all.yaml
@@ -74,7 +74,6 @@ linter:
     - exhaustive_cases
     - file_names
     - flutter_style_todos
-    - future_return_overriding_void_return
     - hash_and_equals
     - implementation_imports
     - implicit_call_tearoffs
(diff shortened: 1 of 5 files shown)


srawlins:
This is ready for review 👍

eernstg:
Do we need a lifted version for futures?:

```dart
abstract class A {
  Future<void> foo();
}

class B implements A {
  Future<int> foo() async => 1;
}
```

A more general lifted version would probably not add much (because `void` isn't used very often as a type argument, except for `Future<void>`):

```dart
abstract class A {
  List<void> foo();
}

class B implements A {
  List<int> foo() => [];
}
```

srawlins:
Hmm, would that serve a very different purpose? Overriding `void`-return with non-`void`-return hampers `unawaited_futures`. But it seems like a check on overriding `Future<void>`-return with `Future<non-void>`-return would be something else, like void_checks or strict-void thing? I'm not against it, but likely a different rule?

eernstg:
I was just thinking that if we want to lint

```dart
abstract class A {
  void foo();
}

class B implements A {
  int foo() => 0;
}
```

then we might also want to lint

```dart
abstract class A {
  Future<void> foo();
}

class B implements A {
  Future<int> foo() async => 0;
}
```

"Same objects, different timing."

srawlins:
The proposal has changed dramatically. Closing.

PR #4067, opened 2023-02-14
Lint against assigning a variable to itself

### LINKED ISSUE #1597: Lint request: flag useless assignments

Just spent 10 minutes debugging a problem that got introduced during a refactor:

```dart
  void _showIfHidden(Rect globalBounds) {
    if (!show) {
      setState(() {
          globalBounds = globalBounds;
          show = true;
        });
    }
  }
```

The `globalBounds = globalBounds` is a no-op. It's supposed to be `this.globalBounds = globalBounds`.

It would be nice to flag these useless assignments because they're an easy mistake to make, are never intentional, and can waste a bunch of developer time.

**Examples**

See above.

oprypin:
This lints against any assignment like `x = x` or `foo.bar = foo.bar` -- only for simple identifiers.

For setters with custom update logic, there *are* valid use cases for this, but they should be refactored anyway. There is an example about that included.

Fixes #1597

I don't insist on the current name of the lint, all suggestions welcome.


lib/src/rules/no_self_assign.dart
```diff
@@ -0,0 +1,150 @@
...
+}
+
+class _Visitor extends SimpleAstVisitor<void> {
+  final LintRule rule;
+
+  _Visitor(this.rule);
+
+  @override
+  void visitAssignmentExpression(AssignmentExpression node) {
+    if (node.operator.type != TokenType.EQ) return;
```
srawlins, line 141: Is there any motivation to catch things like `??=` or `+=`?

test/rules/no_self_assign_test.dart
```diff
@@ -0,0 +1,110 @@
...
+  test_classMemberAssignment() async {
+    await assertDiagnostics(r'''
+class C {
+  static String foo = "foo";
+}
+
+void main() {
+  C.foo = C.foo;
+  String foo;
+  foo = C.foo; // OK
```
srawlins, line 105: Would you mind splitting out the lint case and the OK case into two tests? It hurts me to count 58 characters to see what it is reported :)

test/rules/no_self_assign_test.dart
```diff
@@ -0,0 +1,110 @@
...
+''');
+  }
+
+  test_fieldAssignmentExplicitSameVar() async {
+    await assertDiagnostics(r'''
+class C {
+  int x = 5;
+
+  void update(C other) {
+    other.x = other.x;
```
srawlins, line 62: Could you add another case like `this.x = other.x` or `one.x = two.x`? I don't think I see a case like that (left side and right side are each PrefixedIdentifier).

srawlins (changes requested):

albertms10:
This new lint also suggests the usage of `??` with the same variable, as might sometimes happen in the body of `copyWith` methods, missing `this.`—which I’ve already faced in the past:

```diff
  class A {
    final String? label;
    final int? value;

    const A({this.label, this.value});

    A copyWith({String? label, int? value}) => A(
-         label: label ?? label,
+         label: label ?? this.label,
          value: value ?? this.value,
        );
  }
```

WDYT?

srawlins:
I think that would generalize to include equality as well (`foo == foo` and `foo != foo` when the left and the right are the same identifier).

lib/src/rules/no_self_assign.dart
```diff
@@ -0,0 +1,150 @@
+import 'package:analyzer/dart/ast/ast.dart';
```
pq, line 1: Copyright header?

lib/src/rules.dart
```diff
@@ -103,6 +103,7 @@ import 'rules/no_leading_underscores_for_library_prefixes.dart';
 import 'rules/no_leading_underscores_for_local_identifiers.dart';
 import 'rules/no_logic_in_create_state.dart';
 import 'rules/no_runtimeType_toString.dart';
+import 'rules/no_self_assign.dart';
```
pq, line 106: We haven't been totally consistent but I think we've standardized more on pluralized names.  In which csae, how about:

    no_self_assignments

Or, maybe:

    self_assignments

?

lib/src/rules/no_self_assign.dart
oprypin: `+=`: no, because `a += a` is a perfectly valid way of multiplying something by 2 or something like that.

`??=`: haven't investigated how it looks

lib/src/rules/no_self_assign.dart
oprypin: Done

test/rules/no_self_assign_test.dart
oprypin: Added both

test/rules/no_self_assign_test.dart
oprypin: Done

lib/src/rules.dart
oprypin: Done `no_self_assignments`

test/rules/no_self_assignments_test.dart
```diff
@@ -0,0 +1,145 @@
...
+  }
+
+  test_propertyAssignment() async {
+    await assertDiagnostics(r'''
+class C {
+  int _x = 5;
+
+  int get x => _x;
+
+  set x(int x) {
```
pq, line 109: I feel like this came up in discussion but I can't recall and maybe this is why you call out "direct" effects above but I'm guessing we don't mind reporting in a case where a setter is side-effecting?  That's an anti-pattern too but slightly different than variable self-assignment.

oprypin: I don't think I've seen a discussion on this myself. It's just coming from simple considerations:

* I didn't distinguish between normal fields and setters in the initial implementation. And after seeing the results, I thought that they were both worth flagging and refactoring.
* The "caller site" looks exactly the same (just `x = x`) and lint rules should be simple to understand and resistant to code edits in distant locations.

lib/src/rules.dart
pq: Thank you!

pq (approved):

srawlins (approved):
Sorry for the delay. Love it. Just need to rebase.

PR #4068, opened 2023-02-14
WIP: avoid_future_catcherror rule

christopherfujino:
TODO description

### DIFF
--- a/lib/src/rules.dart
+++ b/lib/src/rules.dart
@@ -24,6 +24,7 @@ import 'rules/avoid_escaping_inner_quotes.dart';
 import 'rules/avoid_field_initializers_in_const_classes.dart';
 import 'rules/avoid_final_parameters.dart';
 import 'rules/avoid_function_literals_in_foreach_calls.dart';
+import 'rules/avoid_future_catcherror.dart';
 import 'rules/avoid_implementing_value_types.dart';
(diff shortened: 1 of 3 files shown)


srawlins:
Quick drive by as this would require a lot of rename work: We typically convert camelCase to snake_case, like in `avoid_type_to_string` or `no_logic_in_create_state` (ignore `avoid_function_literals_in_foreach_calls` and `no_runtimeType_toString`! 🤣  ).

christopherfujino:
> Quick drive by as this would require a lot of rename work: We typically convert camelCase to snake_case, like in `avoid_type_to_string` or `no_logic_in_create_state` (ignore `avoid_function_literals_in_foreach_calls` and `no_runtimeType_toString`! rofl ).

so s/avoid_future_catcherror/avoid_future_catch_error, or other updates too?

srawlins:
Oh, also, we I think are trying to avoid the word `avoid`. @pq would this be just `future_catch_error`?

In any case, you don't have to put more work into this until we settle on what kind of lint rule we want.

christopherfujino:
> Oh, also, we I think are trying to avoid the word `avoid`. @pq would this be just `future_catch_error`?
> 
> In any case, you don't have to put more work into this until we settle on what kind of lint rule we want.

ack on both points.

devoncarew:

Issue #4062, opened 2023-02-09
update `unnecessary_parenthesis` to support parenthesized patterns


pq:

pq:
This could prove to be a lot of (finicky) work.  Bumping.

PR #4075, opened 2023-02-15
unnecessary_parenthesis: parenthesized pattern precedence (WIP)

pq:
See: https://github.com/dart-lang/linter/issues/4062

/cc @bwilkerson


lib/src/rules/unnecessary_parenthesis.dart
```diff
@@ -84,6 +85,14 @@ class _Visitor extends SimpleAstVisitor<void> {
 
   _Visitor(this.rule);
 
+  @override
+  void visitParenthesizedPattern(ParenthesizedPattern node) {
+    var parent = node.parent;
+    if (parent is DartPattern && parent.precedence < node.precedence) {
```
bwilkerson, line 91: This looks good.

We should explore whether we want to flag any use of a parenthesized pattern where the parent isn't another pattern. Seems like that would be a good case, but I'm not familiar enough with the patterns feature to know for sure.

pq: Nor am I.  I'll do a little digging.  Thanks!

pq:
This needs a more thorough revisit. Closing for now.

PR #4078, opened 2023-02-16
Move cancel_subscriptions tests from integration to reflective

srawlins:


test/rules/cancel_subscriptions_test.dart
```diff
@@ -0,0 +1,128 @@
+// Copyright (c) 2023, the Dart project authors. Please see the AUTHORS file
```
pq, line 1: The new test needs to get added to `all.dart` but otherwise, LGTM.

pq (changes requested):

pq (approved):

PR #4089, opened 2023-02-24
unnecessary_lambdas: Do not report when expression has late final variables

srawlins:
Fixes https://github.com/dart-lang/linter/issues/2182

Final properties and variables are understood by this rule to be "safe for tearing off" as they shouldn't change. While this is not true for final fields, it is the understanding. This change also requires the property or variable to not be late, so that it is definitely initialized when it is torn off.


pq (approved):
Nice!

Issue #2301, opened 2020-10-20
gh-pages / generated docs updates


devoncarew:
We might consider:

- changing from hosting the docs on a separate branch (https://github.com/dart-lang/linter/tree/gh-pages) to using a specific directory in the main branch; github added the ability to do this a while ago - it's just a setting for the repo

- changing from generating html to generating markdown, and then having a jekyll template serve that as html. Jekyll is well supported by github pages; we'd be using a process slightly more similar to other github repos, and, it would give us more options in terms of doc styles and presentation

devoncarew:
@pq - if you generate the docs into `doc/lints/` (or similar) in this repo, and then update the gh-pages setting, you can switch to serving from the master branch independent of any switch to generating to markdown / using jekyll.

parlough:
Perhaps we want to use https://github.com/dart-lang/linter/pull/2362 along with [Jekyll Data Files](https://jekyllrb.com/docs/datafiles/) to surface lints on `dart.dev` similar to analyzer diagnostics?

bwilkerson:
There's one consideration that I know of. We have shipped versions of the SDK that produce links to the lint documentation. Those links won't update to the new location if we change where the documentation is published, which means that unless we also redirect from the old location to the new any previously published IDEs will have broken links.

Redirecting probably isn't a problem, but that needs to be part of the work.

pq:
Thanks Brian. In addition to published SDKs there are lots of open issues w/ links to the existing docs.  Having a redirect to keep them alive would be great.

parlough:
Thanks, both those situations definitely make sense! Once the site-www version is finalized, released, and perhaps given some time to settle, I can definitely figure something out to redirect the current site links to redirect to the updated entries.

pq:
I think we can close this as there are a few fresher issues tracking current work on doc generation.

PR #4106, opened 2023-03-04
refactor doc generation

devoncarew:
This is currently a speculative refactor of the linter's doc generation.

This is generally in the spirit of https://github.com/dart-lang/linter/issues/2301 (though I should probably open a more specific issue). The main motivation for me here is that the current linter docs (https://dart-lang.github.io/linter/lints/index.html) don't have either a dart.dev style or a github style; they have their own separate branding.

Some potential changes:
- no longer generate html docs from this repo
- have the analysis server link to the newer dart.dev docs for lint codes instead of the above linter ones
- have the current linter doc pages all redirect to the dart.dev page (to avoid breaking any existing search results / pointers to those docs)
- update the generation here to write (simpler) markdown pages; these could be committed to the repo, and would then be auto-rendered as std github markdown
- update the generation to be more deterministic (to not rely on downloading from other repos at generation time, but instead just render w/ the info available in the current repo).

Some sample generated markdown files included w/ this PR.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,7 +4,6 @@ Contributing to the Linter
 [![Build Status](https://github.com/dart-lang/linter/workflows/linter/badge.svg)](https://github.com/dart-lang/linter/actions)
 ![GitHub contributors](https://img.shields.io/github/contributors/dart-lang/linter.svg)
 
-
 Want to contribute? Great! First, read this page (including the small print at
 the end).
 
--- a/doc/index.md
+++ b/doc/index.md
@@ -1,0 +1,304 @@
+# Linter for Dart
+
+## Lint Rules
+
+Welcome! For general information on using lint rules, see 
+[Using the Linter](https://dart.dev/guides/language/analysis-options#enabling-linter-rules).
+For information about configuring which lints are used, see the
+[analysis options file](https://dart.dev/guides/language/analysis-options#the-analysis-options-file).
+documentation.
+
+Lints can also be used via predefined rulesets; common ones include:
+
+* [core](https://github.com/dart-lang/lints) for official "core" Dart team lint
+  rules.
+* [recommended](https://github.com/dart-lang/lints) for additional lint rules
+  "recommended" by the Dart team.
+* [flutter](https://github.com/flutter/packages/blob/main/packages/flutter_lints/lib/flutter.yaml)
+  for rules recommended for Flutter projects (`flutter create` enables these by
+  default).
+
+## package:lints Core Rules
+
+The following rules are included in the "core" Dart team lint rules (see
+[core](https://github.com/dart-lang/lints/blob/main/lib/core.yaml)). To use
+these rules, add a pubspec dependency on `package:lints` and create an
+analysis_options.yaml file with the line:
+
+```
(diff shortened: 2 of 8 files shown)


devoncarew:
@parlough - the dart.dev site now includes lint codes? Just codes+docs for stable lints?

parlough:
Yep, dart.dev includes linter rules, their descriptions, and version support at https://dart.dev/tools/linter-rules. This includes both stable and experimental rules.

The/my current thinking is to split linter documentation by two separate use cases.

- The diagnostics themselves and how to fix them, in the same style as the [diagnostic messages](https://dart.dev/tools/diagnostic-messages). They would be combined with other diagnostics (https://github.com/dart-lang/site-www/issues/4498), likely with each diagnostic on its own page rather one big one like now. These would be the pages tools would link to, and likely wouldn't be in the sidenav of the site.
- Newly written "Why" documentation for each lint, for users determining lints to enable. This would also include things like version support, lint status, what sets the lint is in, links to Effective Dart, etc. This could have a link back to the diagnostic information as well. This page would be on the sidenav of the site.

Some extra context on this idea: https://github.com/dart-lang/site-www/issues/4498#issuecomment-137605037

devoncarew:
Thanks for the info! I see from one of the linked issues that one of the goals is `deprecate github.io/linter`, which resonates w/ me.

I think generated markdown files here still provide some value, but am not sure that I can really describe why. Perhaps the value is marginal enough to not be worth the maintenance cost.

It looks like we have stable `'go'` links for diagnostic codes (https://github.com/dart-lang/site-www/blob/main/firebase.json#L97). Are these what IDEs would point to? Do we do that today? If no, we may want to update things, so that we could have links we can update as and if the lint docs locations change.

parlough:
Yep, that's what the analysis server [uses today for diagnostics](https://cs.opensource.google/dart/sdk/+/master:pkg/_fe_analyzer_shared/lib/src/base/errors.dart;l=112).

With the discussed changes this would be still be intended access point for old diagnostics, and now the new access point for lint diagnostics.

parlough:
I thought about updating the analysis server to link to the dart.dev linter rule docs rather than `github.io/linter`, but I thought it might be best to wait until we've consolidated them with the other diagnostics.

bwilkerson:
For what it's worth, I already have a significant number of the linter docs rewritten to use the same format as the analyzer diagnostic messages.

devoncarew:
> I think it's a good idea to wait at least until we're feel like we've converged on a link format that feels stable.

From the comments above, it sounds like the link format is `dart.dev/diagnostics/always_use_package_imports`. That's working now for diagnostics, but doesn't yet do the redirection (to https://dart.dev/tools/linter-rules?) for lint codes.

bwilkerson:
Not necessarily. There's an ongoing discussion about moving all of the lint code documentation into https://dart.dev/tools/diagnostic-messages. The plans are still being discussed (temporarily on hold until the 3.0 work is done), so changes at this point are probably premature.

devoncarew:
Gotcha, makes sense to me. It sounds like doc simplification + consolidation here is already planned, and might even happen in the near term. I'll close this PR.

PR #4113, opened 2023-03-06
dogfood the new version of setup-dart

devoncarew:
Related to https://github.com/dart-lang/setup-dart/issues/79; this updates the CI here to use the new, unreleased version of the `setup-dart` action in order to ensure it's working as expected.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -32,7 +32,7 @@ jobs:
 
     steps:
       - uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c
-      - uses: dart-lang/setup-dart@a57a6c04cf7d4840e88432aad6281d1e125f0d46
+      - uses: dart-lang/setup-dart@929ed5f8bae55086c6fac456ba5acb9763c9e3e2
         with:
           sdk: dev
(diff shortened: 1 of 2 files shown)


pq:
I'll want the benchmark bot to finish but so far it looks good.  The only thing I would test is maybe break a test and make sure the bots still fail?  (I don't imagine exit code stuff has changed but this *is* something that has regressed in the past -- e.g., failed tests going unnoticed and not breaking the build.)

devoncarew:
> I'll want the benchmark bot to finish but so far it looks good. The only thing I would test is maybe break a test and make sure the bots still fail? (I don't imagine exit code stuff has changed but this _is_ something that has regressed in the past -- e.g., failed tests going unnoticed and not breaking the build.)

+1, makes sense. FYI, the full test suite has run successfully at setup-dart, and `dart test` should be unaffected; this is more about `dart` being available at all.

I mostly want to convert 4-5 repos over to using this version, and have their normal repo stuff work for ~a week before we call the new version good.

pq:
Cool.  In that case, I'm all for landing this and just keeping an eye on the bots.

Sound good?

pq (approved):

PR #4115, opened 2023-03-07
Align public_member_api_docs expectations with '// LINT' comments

### LINKED ISSUE #4114: `public_member_api_docs` integration test expectations are not aligned with `// LINT` comments
**Describe the issue**

The integration test specifies line numbers where it expects lints to be reported, for example lines 7 and 9. But there are lines in the test data, for example on line 8, with a comment, `// LINT`, which has no matching expectation. This makes it completely unclear what to expect.

Also, `expect(result.stdout.trim(), stringContainsInOrder(...));` does not enforce anything about lines which _shouldn't_ have lint reported. Maybe a lint _is_ reported on line 8, maybe not. Maybe a line with `// OK` has lint reported at it, and maybe not.

srawlins:
After manually reviewing the full expectations and comparing with the `// LINT` and `// OK` comments, this is the total fix for the misalignment.

Fixes #4114

### DIFF
--- a/test/integration/public_member_api_docs.dart
+++ b/test/integration/public_member_api_docs.dart
@@ -28,6 +28,7 @@ void main() {
           result.stdout.trim(),
           stringContainsInOrder([
             'a.dart:7:1',
+            'a.dart:8:16',
             'a.dart:9:11',
             'a.dart:10:9',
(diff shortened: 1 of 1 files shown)


pq (approved):
Nice!

PR #4117, opened 2023-03-07
Fix reported range for undocumented mixin; re-implement public docs integration test

### LINKED ISSUE #3987: `public_member_api_docs` causes body of mixin to be underlined
This issue is related to #3525. Similar to extensions, the mixin's entire body is underlined. It'll be great if only the mixin's name is underlined.

mixins

### LINKED ISSUE #4116: `public_member_api_docs` integration tests seem to use the system dart binary?
**Describe the issue**

The `public_member_api_docs` integration tests [call](https://github.com/dart-lang/linter/blob/main/test/integration/public_member_api_docs.dart#L23) `Process.runSync('dart', ['analyze', '$integrationTestDir/public_member_api_docs'])` which should have nothing to do with the code-under-test. Other integration tests use a different mechanism, `cli.dart`, which presumably tests the code-under-test.

srawlins:
Fixes #3987 and fixes #4116.

I went to fix #3987 but could not see any change in the results for `test/integration/public_member_api_docs.dart`. It turns out this test does not make any assertions on the code you're writing. It makes assertions on whatever `dart` binary exists in your `PATH`. So I filed #4116, and fix it here.

* Simplify `getNodeToAnnotate` and `_getNodeToAnnotate` by combining them into one method. It was split for nullability purposes, but that logic can be made more local in the ConstructorDeclaration and ExtensionDeclaration cases.
* Add MixinDeclaration case to `getNodeToAnnotate`.
* Sort the cases in `getNodeToAnnotate`.
* Re-implement `test/integration/public_member_api_docs.dart` with the `cli.dart` features that the other integration tests use.
  * Normally this would obfuscate the changed test expectations required for the fix to #3987, but GitHub's diff actually highlights nicely the few lines whose content really changes.
* Remove an outdated comment about performance. A [recent benchmark](https://github.com/dart-lang/linter/actions/runs/4355704004/jobs/7612683974) shows this rule to be one of the fastest.

### DIFF
--- a/lib/src/ast.dart
+++ b/lib/src/ast.dart
@@ -59,8 +59,55 @@ int? getIntValue(Expression expression, LinterContext? context) {
 
 /// Returns the most specific AST node appropriate for associating errors.
 SyntacticEntity getNodeToAnnotate(Declaration node) {
-  var mostSpecific = _getNodeToAnnotate(node);
-  return mostSpecific ?? node;
+  // TODO(srawlins): Convert to a switch expression over `Declaration` subtypes,
+  // assuming `Declaration` becomes an exhaustive type.
+  if (node is ClassDeclaration) {
+    return node.name;
+  }
+  if (node is ClassTypeAlias) {
+    return node.name;
+  }
+  if (node is ConstructorDeclaration) {
+    return node.name ?? node;
+  }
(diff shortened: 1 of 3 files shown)


srawlins:
This is ready for review.

PR #4132, opened 2023-03-09
support record and list patterns in `prefer_final_locals`

pq:
Fixes: https://github.com/dart-lang/linter/issues/4120

/cc @bwilkerson

### DIFF
--- a/lib/src/rules/prefer_final_locals.dart
+++ b/lib/src/rules/prefer_final_locals.dart
@@ -3,6 +3,7 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/token.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
 
 import '../analyzer.dart';
@@ -68,6 +69,8 @@ class PreferFinalLocals extends LintRule {
       NodeLintRegistry registry, LinterContext context) {
     var visitor = _Visitor(this);
     registry.addVariableDeclarationList(this, visitor);
+    registry.addRecordPattern(this, visitor);
+    registry.addListPattern(this, visitor);
   }
 }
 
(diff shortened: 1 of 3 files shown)


pq:
Ah, I should probably handle object patterns too :/

pq:
And Maps!

pq:
OK. A bunch more cases handled.

@bwilkerson: feedback welcome.

srawlins (approved):
❤️

lib/src/rules/prefer_final_locals.dart
```diff
@@ -76,6 +81,101 @@ class _Visitor extends SimpleAstVisitor<void> {
...
+    }
+  }
+
+  bool isDeclaredFinal(AstNode node) {
+    var declaration = node.thisOrAncestorOfType<PatternVariableDeclaration>();
+    if (declaration == null) return false; // To be safe.
+    return declaration.keyword.isFinal;
+  }
+
+  bool isPotentiallyMutated(AstNode pattern, FunctionBody function) {
```
bwilkerson, line 129: This is attempting to return `true` if any variable in the `pattern` is potentially mutated, but it doesn't appear to account for variables nested inside other patterns (such as a logical-and or list pattern).

lib/src/rules/prefer_final_locals.dart
```diff
@@ -76,6 +81,101 @@ class _Visitor extends SimpleAstVisitor<void> {
...
 
+  void checkPatternFields(DartPattern node) {
+    var function = node.thisOrAncestorOfType<FunctionBody>();
+    if (function == null) return;
+
+    late NodeList<PatternField> fields;
+    if (node is RecordPattern) fields = node.fields;
+    if (node is ObjectPattern) fields = node.fields;
+
+    var parent = node.unParenthesized.parent;
```
bwilkerson, line 92: I don't think this is doing what you expect it to do. If the `node` is a `ParenthesizedPattern`, then `unParenthesized` will return the expression inside the parentheses, and `parent` will return the `ParenthesizedPattern`. I think you either want `node.parent` or a loop that will get you to the parent of any enclosing parenthesized patterns.

lib/src/rules/prefer_final_locals.dart
```diff
@@ -100,3 +200,23 @@ class _Visitor extends SimpleAstVisitor<void> {
...
+extension on Token? {
+  bool get isFinal {
+    var self = this;
+    if (self == null) return false;
+    return self.keyword == Keyword.FINAL;
+  }
+}
+
+extension on DartPattern {
+  DeclaredVariablePattern? get declaredVariablePattern {
```
bwilkerson, line 213: I don't understand the semantics you're trying to implement here.

test/rules/prefer_final_locals_test.dart
```diff
@@ -0,0 +1,252 @@
...
+  test_destructured_listPattern_mutated_ok() async {
+    await assertNoDiagnostics(r'''
+f() {
+  var [a, b, ...rest] = [1, 2, 3, 4, 5, 6, 7];
+  print('${++a}$b$rest');
+}
+''');
+  }
+
+  test_destructured_listPattern_ok() async {
```
bwilkerson, line 32: I'd drop the `_ok` (I think the `assertNoDiagnostics` says enough), especially here where we're expecting a lint.

test/rules/prefer_final_locals_test.dart
```diff
@@ -0,0 +1,252 @@
...
+  test_switch_objectPattern() async {
+    await assertDiagnostics(r'''
+class A {
+  int a;
+  A(this.a);
+}
+
+f() {
+  switch (A(1)) {
+    case A(a: >0 && var b): print('$b');
```
bwilkerson, line 180: What happens when you reverse the operands of the `&&`?

lib/src/rules/prefer_final_locals.dart
```diff
@@ -76,6 +81,101 @@ class _Visitor extends SimpleAstVisitor<void> {
...
+      if (element == null || function.isPotentiallyMutatedInScope(element)) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  @override
+  void visitListPattern(ListPattern node) {
+    if (isDeclaredFinal(node)) return;
```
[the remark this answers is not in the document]
srawlins: Addressed in https://github.com/dart-lang/linter/pull/4225

lib/src/rules/prefer_final_locals.dart
srawlins: Addressed in https://github.com/dart-lang/linter/pull/4225

lib/src/rules/prefer_final_locals.dart
srawlins: Addressed in https://github.com/dart-lang/linter/pull/4225

lib/src/rules/prefer_final_locals.dart
srawlins: Removed in https://github.com/dart-lang/linter/pull/4225

test/rules/prefer_final_locals_test.dart
srawlins: Done in https://github.com/dart-lang/linter/pull/4225

PR #4138, opened 2023-03-13
migrate `prefer_mixin` to reflective tests

pq:
Migrate `prefer_mixin` to reflective tests and break out 2.19-specific tests.

/cc @scheglov


test/rules/prefer_mixin_test.dart
```diff
@@ -8,17 +8,76 @@ import '../rule_test_support.dart';
...
   defineReflectiveSuite(() {
-    defineReflectiveTests(PreferMixinTest);
+    defineReflectiveTests(PreferMixin30Test);
+    defineReflectiveTests(PreferMixin219Test);
   });
 }
 
-@reflectiveTest
-class PreferMixinTest extends LintRuleTest {
+abstract class BaseTest extends LintRuleTest {
```
scheglov, line 16: In the analyzer this would be `PreferMixinTestCases`, but it is a mixin. I don't remember why we decided to make these mixins. A base class like this would work too.

But naming it just `BaseTest` seems not ideal, you would get many of these. OTOH, this is a short nice name, easy to find once you open a file.

scheglov (approved):

test/rules/prefer_mixin_test.dart
pq: renamed.  Thanks!

test/rules/prefer_mixin_test.dart
```diff
@@ -8,17 +8,76 @@ import '../rule_test_support.dart';
 
 main() {
   defineReflectiveSuite(() {
-    defineReflectiveTests(PreferMixinTest);
+    defineReflectiveTests(PreferMixin30Test);
+    defineReflectiveTests(PreferMixin219Test);
```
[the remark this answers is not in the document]
pq: Happy to take the analyzer's lead here.  Thanks!

PR #4139, opened 2023-03-13
migrate `no_duplicate_case_values` to reflective tests

pq:
More migrations and segregation of 2.19 and 3.0 testing.

/cc @scheglov


test/rules/no_duplicate_case_values_test.dart
```diff
@@ -0,0 +1,202 @@
...
+  });
+}
+
+abstract class BaseNoDuplicateCaseValuesTestTest extends LintRuleTest {
+  @override
+  String get lintRule => 'no_duplicate_case_values';
+}
+
+@reflectiveTest
+class NoDuplicateCaseValuesTestLanguage219
```
pq, line 22: FWIW, we can't use underscores here because it breaks a lint we have enabled.

scheglov (approved):

PR #4144, opened 2023-03-13
stop indexing removed lints

### LINKED ISSUE #4135: Auto-generated list of rules should not contain "removed" ones
(reporting it as a bug, but not sure this is right )

**Describe the issue**
The documentation hosted here: https://dart-lang.github.io/linter/lints/options/options.html contains rules such as `avoid_as` (https://dart-lang.github.io/linter/lints/avoid_as.html) which are marked as **_removed_**.

**To Reproduce**
Look at https://dart-lang.github.io/linter/lints/options/options.html

**Expected behavior**
https://dart-lang.github.io/linter/lints/options/options.html should only contain rules that are considered valid.

**Additional context**
Not sure if fixing it would be considered a breaking change, or if maybe there should be a separate page with that filtering applied, or maybe there should be a page for each version of the Dart SDK, showing only the list of options that make sense "as of" that version ?

pq:
Fixes #4135

/cc @bwilkerson

### DIFF
--- a/tool/doc.dart
+++ b/tool/doc.dart
@@ -515,7 +515,7 @@ linter:
 ''');
 
     var sortedRules = rules
-        .where((r) => !r.state.isDeprecated)
+        .where((r) => !r.state.isDeprecated && !r.state.isRemoved)
         .map((r) => r.name)
         .toList()
       ..sort();


parlough:
I think this is probably fine, but also should consider that some of these were removed in Dart 3 and are still useful as of Dart 2.19, which people referencing the documentation are possibly using.

pq:
Yeah.  It's a good point @parlough.  That said, the generated docs won't go away, this will just elide removed options from the sample at https://dart-lang.github.io/linter/lints/options/options.html. Maybe OK?

parlough:
Yeah I think it's fine since it's excluding deprecated ones already and the use of this file is likely limited.

Could potentially be expanded so it only excludes these if they've been deprecated or removed before the current stable version, but probably not worth it.

bwilkerson (approved):

PR #4145, opened 2023-03-13
switch pattern support for `no_adjacent_strings_in_list`

### LINKED ISSUE #4119: `no_adjacent_strings_in_list` tests should include list patterns
Child issue of https://github.com/dart-lang/linter/issues/3624 I suppose.

```dart
List<String?> row = ...

// If the first column is 'one', and the second is 'two', we expect to have a name after it.
switch (row) {
  case ['one' 'two', var name!]:
    // name is a non-nullable string here.
}
```

pq:
Fixes #4119

/cc @srawlins @bwilkerson


srawlins (approved):
Cool!

PR #4147, opened 2023-03-14
add relational pattern support to `unrelated_type_equality_checks`

### LINKED ISSUE #4123: `unrelated_type_equality_checks` tests should include relational patterns

For example:

```dart
String asciiCharType(int char) {
  const space = 32;
  const zero = 48;
  const nine = 57;

  return switch (char) {
    == 'space' => 'space', // LINT
    // Etc...
  };
}
```

pq:
Fixes #4123

/cc @bwilkerson

### DIFF
--- a/test/rules/unrelated_type_equality_checks_test.dart
+++ b/test/rules/unrelated_type_equality_checks_test.dart
@@ -50,6 +50,21 @@ String f(String char) {
     ]);
   }
 
+  test_switchExpression_notEq() async {
+    await assertDiagnostics(r'''    
+const space = 32;
+
+String f(int char) {
+  return switch (char) {
+    != 'space' => 'space',
+  };
+}
+''', [
+      error(CompileTimeErrorCode.NON_EXHAUSTIVE_SWITCH, 49, 6),
+      lint(69, 10),
+    ]);
+  }
+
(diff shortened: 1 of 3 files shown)


test/rules/unrelated_type_equality_checks_test.dart
```diff
@@ -0,0 +1,80 @@
...
+    await assertDiagnostics(r'''    
+const space = 32;
+
+String f(int char) {
+  return switch (char) {
+    == 'space' => 'space',
+  };
+}
+''', [
+      error(CompileTimeErrorCode.NON_EXHAUSTIVE_SWITCH, 49, 6),
```
bwilkerson, line 34: You can, and really should, get rid of extraneous errors. While we make an effort for the lints to work well in the face of errors, that isn't really what we're trying to test here. In most cases it's as simple as converting to use a switch statement rather than an expression. For `test_switchExpression_lessEq_ok` you can probably just reverse the types of the param and value.

bwilkerson (approved):

test/rules/unrelated_type_equality_checks_test.dart
pq: Thanks!  I'll do a round of cleanup.

PR #4148, opened 2023-03-14
null assert pattern support for `null_check_on_nullable_type_parameter`

### LINKED ISSUE #4130: `null_check_on_nullable_type_parameter` tests should include null-assert patterns

```dart
(T?, T?) position = ...

// We know if we get here that the coordinates should be present:
var (x!, y!) = position;
```

in which the lint rule would report that casting to `T` is preferable to null-asserting.

pq:
Fixes #4130

/cc @bwilkerson @srawlins 

/fyi @a14n


srawlins (approved):
Very cool, thanks!

PR #4153, opened 2023-03-15
Add new rule: deprecated_member_use_from_same_package

srawlins:
# Description

Fixes https://github.com/dart-lang/linter/issues/4118

This rule implements (more or less*) the same feature currently implemented in the analyzer `HintCode.DEPRECATED_MEMBER_USE_FROM_SAME_PACKAGE`.

See [this comment on the issue](https://github.com/dart-lang/linter/issues/4118#issuecomment-1460627011) for the steps to roll this to analyzer, which involve simultaneously removing `HintCode.DEPRECATED_MEMBER_USE_FROM_SAME_PACKAGE`.

*: The changes are improvements to the rule, which are similarly proposed and in-progress for `HintCode.DEPRECATED_MEMBER_USE`: https://github.com/dart-lang/sdk/issues/51664.

### DIFF
--- a/test/rules/deprecated_member_use_from_same_package_test.dart
+++ b/test/rules/deprecated_member_use_from_same_package_test.dart
@@ -1,0 +1,786 @@
+// Copyright (c) 2023, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:collection/collection.dart';
+import 'package:test_reflective_loader/test_reflective_loader.dart';
+
+import '../rule_test_support.dart';
+
+main() {
+  defineReflectiveSuite(() {
+    defineReflectiveTests(DeprecatedMemberUseFromSamePackageTest);
+  });
+}
+
+@reflectiveTest
+class DeprecatedMemberUseFromSamePackageTest extends LintRuleTest {
+  @override
+  String get lintRule => 'deprecated_member_use_from_same_package';
+
+  @override
+  Future<void> assertDiagnostics(
+      String code, List<ExpectedDiagnostic> expectedDiagnostics) async {
+    addTestFile(code);
+    await resolveTestFile();
+    var filteredErrors = errors
+        .whereNot((e) =>
+            e.errorCode == HintCode.DEPRECATED_MEMBER_USE_FROM_SAME_PACKAGE ||
+            e.errorCode ==
+                HintCode.DEPRECATED_MEMBER_USE_FROM_SAME_PACKAGE_WITH_MESSAGE)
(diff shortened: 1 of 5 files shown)


test/rules/deprecated_member_use_from_same_package_test.dart
```diff
@@ -0,0 +1,786 @@
...
+  @override
+  String get lintRule => 'deprecated_member_use_from_same_package';
+
+  @override
+  Future<void> assertDiagnostics(
+      String code, List<ExpectedDiagnostic> expectedDiagnostics) async {
+    addTestFile(code);
+    await resolveTestFile();
+    var filteredErrors = errors
+        .whereNot((e) =>
```
pq, line 27: I'm curious what errors you're filtering out here? Could we re-write this to be more targeted? (My concern is that we not ignore/mask surprising errors in the test data.) FWIW I'm all for enumerating a handful of diagnostics that might be ignored in a lot of tests.  (I was thinking just today how we might want to ignore `unused_local_variable` at least.)

srawlins: These are targeted at the two error codes listed. What are you thinking for "more targeted"? I'm open to ideas here, cuz it's a little weird. I can just add a comment too.

bwilkerson: The filter removed everything except the listed codes. I think the question is: what codes are being removed? In general we try to write code that doesn't produce any diagnostics other than the ones we're explicitly testing for. That helps us catch code that needs to be updated as the language changes, but ignoring everything negates that benefit.

bwilkerson (approved):

test/rules/deprecated_member_use_from_same_package_test.dart
srawlins: I might be thinking about this wrong, but normally we take _all_ reported errors and compare them against the short list of expected errors.

> I think the question is: what codes are being removed?

In this bit, I am taking all errors which are not `HintCode.DEPRECATED_MEMBER_USE_FROM_SAME_PACKAGE` or `HintCode.DEPRECATED_MEMBER_USE_FROM_SAME_PACKAGE_WITH_MESSAGE`. So it's the same set of reported errors as usual, minus these two.

> ignoring everything negates that benefit.

I certainly don't intend to ignore everything. Just two error codes.

These two error codes are produced today, but will not be produced soon, with an SDK commit. The only way to get the tests to pass in google3 in both states is to filter out the HintCodes.

bwilkerson: I read the code backwards and thought you were filtering out everything except those two. Filtering those two is fine.

PR #4162, opened 2023-03-17
unreachable_from_main: Report unreachable constructors

srawlins:
# Description

This adds support for reporting unreachable constructors in `unreachable_from_main`.

* Constructors (named and unnamed) can be reachable through direct call, tear-off, explicit super call, implicit super call (default constructor), transitive implicit super call, redirecting constructor initializer, redirected factory constructor.
* Unused public enum constructors are already reported as `UNUSED_ELEMENT`.

### DIFF
--- a/lib/src/rules/unreachable_from_main.dart
+++ b/lib/src/rules/unreachable_from_main.dart
@@ -43,8 +43,8 @@ void f() {}
 ''';
 
 class UnreachableFromMain extends LintRule {
-  static const LintCode code = LintCode(
-      'unreachable_from_main', 'Unreachable member in an executable library.',
+  static const LintCode code = LintCode('unreachable_from_main',
+      "Unreachable member '{0}' in an executable library.",
       correctionMessage: 'Try referencing the member or removing it.');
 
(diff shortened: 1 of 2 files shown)


srawlins:
This is ready for review.

lib/src/rules/unreachable_from_main.dart
```diff
@@ -42,8 +43,8 @@ void f() {}
 ''';
 
 class UnreachableFromMain extends LintRule {
-  static const LintCode code = LintCode('unreachable_from_main',
-      'Unreachable top-level member in an executable library.',
+  static const LintCode code = LintCode(
+      'unreachable_from_main', 'Unreachable member in an executable library.',
```
bwilkerson, line 47: Consider adding the name of the member to the message for command-line users.

bwilkerson (approved):

lib/src/rules/unreachable_from_main.dart
srawlins: Good call, done!

PR #4187, opened 2023-03-21
Do not require API docs on public enum constructors

### LINKED ISSUE #3655: proposal: `public_member_api_docs` should not lint on enum constructors
```dart
enum Example {
  one(1),
  two(2),
  three(3);
  
  // LINT: public_member_api_docs
  const Example(int value);
}
```

Enum constructors are unique where the constructor is (essentially) class-private by default.

My recommendation is to just change the lint not to fire on
- Methods that are constructors
- Where that constructor's class type is an `Enum`

srawlins:
As described in the issue, enum constructors are really class-private, and this lint rule should not flag them.

Fixes #3655


pq (approved):
Nice!

PR #4189, opened 2023-03-22
declared variable pattern support for `unnecessary_nullable_for_final_variable_declarations`

### LINKED ISSUE #4181: `unnecessary_nullable_for_final_variable_declarations` should have tests with patterns

```dart
final (List<int>? a, List<int?> b, num? c) = ([], [], 1);
```

in which the lint rule would report that all three of the types are unnecessarily nullable.

pq:
Fixes #4181

/cc @bwilkerson

### DIFF
--- a/lib/src/rules/unnecessary_nullable_for_final_variable_declarations.dart
+++ b/lib/src/rules/unnecessary_nullable_for_final_variable_declarations.dart
@@ -4,6 +4,7 @@
 
 import 'package:analyzer/dart/analysis/features.dart';
 import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/token.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
 
 import '../analyzer.dart';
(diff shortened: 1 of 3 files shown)


lib/src/rules/unnecessary_nullable_for_final_variable_declarations.dart
```diff
@@ -63,6 +65,19 @@ class _Visitor extends SimpleAstVisitor<void> {
   final LinterContext context;
   _Visitor(this.rule, this.context);
 
+  void check(AstNode? element) {
```
srawlins, line 68: I think this is only called with non-nullable `DartPattern`s.

test/rules/unnecessary_nullable_for_final_variable_declarations_test.dart
```diff
@@ -0,0 +1,52 @@
+// Copyright (c) 2022, the Dart project authors. Please see the AUTHORS file
```
srawlins, line 1: 2023 😄

srawlins (approved):
❤️

lib/src/rules/unnecessary_nullable_for_final_variable_declarations.dart
pq: Good catch!

PR #4192, opened 2023-03-22
null assert pattern support for `unnecessary_null_checks`

### LINKED ISSUE #4183: `unnecessary_null_checks` should have tests with object patterns

```dart
void f(int? a, int? b) {
  (b!, ) = (a, );
}
```

in which the lint rule would report that null-asserting at `b!` is unnecessary because the variable type is `int?`.

pq:
Fixes #4183

/cc @bwilkerson @srawlins

### DIFF
--- a/lib/src/rules/unnecessary_null_checks.dart
+++ b/lib/src/rules/unnecessary_null_checks.dart
@@ -175,6 +175,7 @@ class UnnecessaryNullChecks extends LintRule {
   void registerNodeProcessors(
       NodeLintRegistry registry, LinterContext context) {
     var visitor = _Visitor(this, context);
+    registry.addNullAssertPattern(this, visitor);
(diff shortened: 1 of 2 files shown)


lib/src/rules/unnecessary_null_checks.dart
```diff
@@ -185,6 +186,16 @@ class _Visitor extends SimpleAstVisitor<void> {
   final LinterContext context;
   _Visitor(this.rule, this.context);
 
+  @override
+  void visitNullAssertPattern(NullAssertPattern node) {
+    if (node.operator.type != TokenType.BANG) return;
```
bwilkerson, line 191: There aren't any other operators that could appear there, so we don't need to test that it's the right one.

bwilkerson (approved):

PR #4198, opened 2023-03-22
Add a test for a documented annotated enum value

### LINKED ISSUE #1396: `public_member_api_docs` on Enum values having an annotation
_From @cedx on December 31, 2018 10:45_

* Dart SDK: 2.1.0 (Tue Nov 13 18:22:02 2018 +0100) on "windows_x64"
* OS: Windows 10

I'm using the rule `public_member_api_docs` from the Dart linter.

If I use an annotation on a documented member of a class, everything is OK. But if I do the same thing with a enumerated value, the analyzer triggers an error: `Document all public members`.

```dart
// This class does not trigger the analyser warning.
class Foo {
  /// A comment.
  @JsonKey(defaultValue: '')
  String bar;
}

// This enum triggers the analyzer warning.
enum Baz {
   /// A comment.
  @JsonValue('contact')
  qux
}
```

Removing the annotation on the enumerated value also removes the analyzer warning.

_Copied from original issue: dart-lang/sdk#35527_

srawlins:
Closes #1396 (an analyzer bump, long ago, likely fixed the issue)

### DIFF
--- a/test/rules/public_member_api_docs_test.dart
+++ b/test/rules/public_member_api_docs_test.dart
@@ -49,6 +49,20 @@ enum A {
 ''');
   }
 
+  test_annotatedEnumValue() async {
+    await assertNoDiagnostics(r'''
+/// Documented.
+enum A {
+  /// This represents 'a'.
+  @Deprecated("Use 'b'")
+  a,
+
+  /// This represents 'b'.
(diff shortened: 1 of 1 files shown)


pq (approved):
Nice!

PR #4203, opened 2023-03-23
`prefer_void_to_null` cast pattern tests

### LINKED ISSUE #4201: `prefer_void_to_null` should not be reported in a Cast Pattern or an As Expression
`void` is not allowed in either:

```dart
void f(int a) {
  switch (a) {
    case var _ as void: // compile-time error
  }
  a as void; // compile-time error
}
```

So `prefer_void_to_null` should not be reported in those positions.

pq:
Fixes #4201

/cc @bwilkerson @srawlins


bwilkerson (approved):

srawlins:
`prefer_void_to_null` would never fire on this test case, as it has no instances of `Null`.

Issue #4195, opened 2023-03-22
Lint for using a type literal in a switch case


munificent:
Consider:

```dart
switch (obj) {
  case int: print('int');
}
```

In Dart both before 3.0 and in 3.0 with patterns, this case does not match integers, it matches only the *type object `int` itself*. Prior to 3.0, there was no way to do a type test in a switch, so users would rarely ever think to write the name of a type and expect it to type test.

With patterns, we expect type tests to be common in switches. There are two idiomatic ways to write them:

```dart
switch (obj) {
  case int(): print('int');

  case int _: print('int');
}
```

(The first, an object pattern, is useful when you want to do further destructuring on the value after testing its type. It's likely the idiomatic style to use in algebraic datatype style code. The latter, a wildcard variable declaration, allows you to match any possible type annotation. Object patterns only support named types, but not more complex type annotations like records, functions, nullable types, etc.)

Because type tests are common in patterns, we expect that users will sometimes write a type literal when they meant to do a type test. This is *valid* code (though in some cases it will lead to non-exhaustive switch errors), but it probably doesn't do what the user wants.

We tried hard to tweak the syntax to avoid this footgun, but every other approach we came up with seemed to be worse overall. We also considered making it an error to have a constant in a pattern that resolves to a type literal. This would force users away from the footgun... but it's also a breaking change. There *is* code in the wild today that deliberately uses type literals in switch cases to test for type objects.

Given that, I think it would help the user experience if we added a lint that reports a type literal being used as a constant in a switch case (in a Dart 3.0 library). Instead, it should suggest that users use either an object pattern or variable pattern if they meant to do a type test, or an explicit `==` pattern if they really did mean to compare for equality to a type literal.

The heuristic I'd suggest is:

1.  If the matched value type is `Type`, then suggest they change the case from `case SomeType` to `case == SomeType`. That has the same semantics, but makes it clearer to the reader that a type equality test is intended (and silences the lint).

2.  Otherwise, if the matched value type is anything else, suggest that they use `case SomeType _` if they indended to do a type test and `== SomeType` otherwise.

This lint will be useful whenever it arrives, but nothing is blocked in 3.0 around having it. It should just help users avoid what's likely a mistake.

pq:
I'm all for this and agree with Brian that a lint is the right place to implement.

munificent:
[...]

PR #4207, opened 2023-03-23
Issue 4195. Add type_literal_in_constant_pattern

scheglov:
Obviously, cannot be landed with dependency overrides, will clean up after initial review.

Fixes https://github.com/dart-lang/linter/issues/4195

### DIFF
--- a/example/all.yaml
+++ b/example/all.yaml
@@ -166,6 +166,7 @@ linter:
     - tighten_type_of_initializing_formals
     - type_annotate_public_apis
     - type_init_formals
+    - type_literal_in_constant_pattern
     - unawaited_futures
     - unnecessary_await_in_return
     - unnecessary_brace_in_string_interps
--- a/lib/src/rules.dart
+++ b/lib/src/rules.dart
@@ -176,6 +176,7 @@ import 'rules/throw_in_finally.dart';
 import 'rules/tighten_type_of_initializing_formals.dart';
 import 'rules/type_annotate_public_apis.dart';
 import 'rules/type_init_formals.dart';
+import 'rules/type_literal_in_constant_pattern.dart';
 import 'rules/unawaited_futures.dart';
 import 'rules/unnecessary_await_in_return.dart';
 import 'rules/unnecessary_brace_in_string_interps.dart';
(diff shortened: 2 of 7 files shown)


lib/src/rules/no_constant_pattern_type_literal.dart
```diff
@@ -0,0 +1,108 @@
...
+void f(Object? x) {
+  if (x case int _) {
+    print('int');
+  }
+}
+```
+
+''';
+
+class NoConstantPatternTypeLiteral extends LintRule {
```
pq, line 53: I think I'd prefer `NoConstantPatternTypeLiterals` --- on balance we have more plural thank singular names.

lib/src/rules/no_constant_pattern_type_literal.dart
```diff
@@ -0,0 +1,108 @@
...
+  static const String lintName = 'no_constant_pattern_type_literal';
```
bwilkerson, line 54: Or `type_literal_in_constant_pattern`?

There's always been some inconsistency between lints names describing what you ought to do and describing what you did wrong. All of the non-lint diagnostics describing what's wrong with the code, and I'm leaning in that direction for this lint, but it's a judgement call.

pq (approved):
Looks great!  I'll defer to @eernstg in case he has any questions about semantics but this handles the cases I'm aware of (and most concerned about), so cool!

lib/src/rules/no_constant_pattern_type_literal.dart
```diff
@@ -0,0 +1,108 @@
...
+
+  static const LintCode matchType = LintCode(
+    lintName,
+    "Use '== TypeName ' instead of a type literal.",
```
bwilkerson, line 58: nit: extra space after 'TypeName'

bwilkerson (approved):

lib/src/rules/no_constant_pattern_type_literal.dart
scheglov: With the comment from Brian below, I renamed it to `TypeLiteralInConstantPattern`, to be consistent with the `type_literal_in_constant_pattern` name. Is this a correct way to name lint rules?

lib/src/rules/no_constant_pattern_type_literal.dart
scheglov: Done.

lib/src/rules/no_constant_pattern_type_literal.dart
scheglov: Fixed.

lib/src/rules/no_constant_pattern_type_literal.dart
bwilkerson: Yes. It's easier to find the implementation if the class name and lint name match.

PR #4208, opened 2023-03-23
don't overreport `prefer_void_to_null` on cast patterns and as expressions

pq:
Fixes https://github.com/dart-lang/linter/issues/4201 **(actually)** 😊 

/cc @srawlins


srawlins (approved):
❤️

PR #4209, opened 2023-03-23
`use_late_for_private_fields_and_variables` pattern assignment test

### LINKED ISSUE #4180: `use_late_for_private_fields_and_variables` should have tests with patterns

I think there should be a test case with an assignment pattern like:

```dart
int? _i;
m() {
  _i?.abs();
  (_i, ) = (null, );
}
```

in which the lint rule would allow that `_i` can be nullable, as it is assigned `null`, in a pattern.

pq:
It looks like you can't do a pattern assignment of a field or top level member so the diagnostics are expected.  Verifying no lints.

@srawlins: let me know if this doesn't get the case you had in mind?

Fixes #4180


srawlins:
Ah good catch; if the lint rule doesn't apply to local variables, then there's nothing to do. Thanks!

srawlins (approved):

PR #4211, opened 2023-03-23
`use_named_constants` constant pattern tests

### LINKED ISSUE #4197: `use_named_constants` should have tests with constant patterns

I think there should be a test case with a constant pattern like:

```dart
void f(Duration d) {
  switch (d) {
    case const Duration(seconds: 0):
  }
}
```

in which the lint rule would recommend using `Duration.zero` instead of `Duration(seconds: 0)`.

pq:
Fixes #4197

/cc @bwilkerson @srawlins


srawlins (approved):
🎉  I'm so happy this was already working, haha.

Thanks!

PR #4213, opened 2023-03-23
don't wildcard patterns w/ `non_constant_identifier_names`

pq:
/cc @bwilkerson @srawlins

### DIFF
--- a/test/rules/non_constant_identifier_names_test.dart
+++ b/test/rules/non_constant_identifier_names_test.dart
@@ -19,16 +19,6 @@ class NonConstantIdentifierNamesPatternsTest extends LintRuleTest {
   @override
   String get lintRule => 'non_constant_identifier_names';
 
-  test_identifierPattern() async {
-    await assertNoDiagnostics(r'''
-f() {
-  var a = 0;
-  [_, a] = [1,2];
-  print(a);
-}
-''');
-  }
-
   test_patternForStatement() async {
     await assertDiagnostics(r'''
 void f() {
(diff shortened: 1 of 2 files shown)


test/rules/non_constant_identifier_names_test.dart
```diff
@@ -19,6 +19,16 @@ class NonConstantIdentifierNamesPatternsTest extends LintRuleTest {
   @override
   String get lintRule => 'non_constant_identifier_names';
 
+  test_identifierPattern() async {
+    await assertNoDiagnostics(r'''
+f() {
+  var a = 0;
```
srawlins, line 25: Hmm, this is not new syntax, so I wouldn't expect anything to be surprising here. But a variable declaration pattern might not get caught: `var [_, a] = [1, 2];`

pq: > Hmm, this is not new syntax

I believe this is a "wildcard pattern" 

https://github.com/dart-lang/language/blob/master/accepted/future-releases/0546-patterns/feature-specification.md#wildcards

And this is a different case 

> var [_, snake_case] = [1, 2];

(where `_` is being declared as a local variable).

The question is whether all-underscores should be ok in this case.

I'll look into this today.

srawlins: As for the wildcard, I was just avoiding an `unused_local_variable` issue, we could remove that; `snake_case` is the important part.

srawlins (approved):
Awesome

PR #4216, opened 2023-03-24
filter UNUSED_LOCAL_VARIABLE diagnostics in test expectations

pq:
These diagnostics are just noise and we need to do hokey things to ignore them.

In case we ever care a test can always override `ignoredErrorCodes`.

I'll take a pass to remove all the gross `print(..)` work-arounds next.

/cc @bwilkerson @srawlins

### DIFF
--- a/test/rule_test_support.dart
+++ b/test/rule_test_support.dart
@@ -476,7 +476,7 @@ abstract class _ContextResolutionTest with ResourceProviderMixin {
       .toList();
 
   /// Error codes that by default should be ignored in test expectations.
-  List<HintCode> get ignoredErrorCodes => [HintCode.UNUSED_LOCAL_VARIABLE];
+  List<HintCode> get ignoredErrorCodes => [WarningCode.UNUSED_LOCAL_VARIABLE];
 
   Folder get sdkRoot => newFolder('/sdk');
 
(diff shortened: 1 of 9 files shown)


test/rule_test_support.dart
```diff
@@ -472,7 +471,12 @@ abstract class _ContextResolutionTest with ResourceProviderMixin {
   List<String> get collectionIncludedPaths;
 
   /// The analysis errors that were computed during analysis.
-  List<AnalysisError> get errors => result.errors;
+  List<AnalysisError> get errors => result.errors
+      .whereNot((e) => ignoredErrorCodes.any((c) => e.errorCode == c))
+      .toList();
+
+  /// Error codes that by default should be ignored in test expectations.
+  List<HintCode> get ignoredErrorCodes => [HintCode.UNUSED_LOCAL_VARIABLE];
```
srawlins, line 479: Please use `WarningCode.UNUSED_LOCAL_VARIABLE` as this HintCode will (🤞 🤞 🤞 ) be removed soon. https://github.com/dart-lang/sdk/blob/main/pkg/analyzer/lib/src/error/codes.g.dart#L7112

srawlins (approved):

test/rule_test_support.dart
pq: Oh, yeah!  Totally.

bwilkerson (approved):
This is great for going forward.

It's too bad we won't be able to easily find code that could be removed from the tests because it's no longer needed in order to suppress the warning, but I can't think of any reasonable way to find it.

pq:
> It's too bad we won't be able to easily find code that could be removed from the tests because it's no longer needed in order to suppress the warning,

I wracked my brain for this too. The good news is I've been pretty consistent in my approach so a `grep` will go a long way...

PR #4217, opened 2023-03-24
remove unneeded test `print` calls

pq:
Follow-up from https://github.com/dart-lang/linter/pull/4216.

/cc @bwilkerson


bwilkerson (approved):
Nice.

PR #4225, opened 2023-03-27
Improve robustness of prefer_final_locals with patterns

srawlins:
This addresses some of the comments at https://github.com/dart-lang/linter/pull/4132/files

* I reworked the visitors to only visit `DeclaredVariablePattern` (for record and object cases) and `PatternVariableDeclaration` (for variable declaration cases), and added an additional visitor for visiting (potentially highly nested) declared variables.
* I renamed test cases, removing `_ok`. A few renames would result in a new sort order, but I intentionally did _not_ re-sort, so as to make the review easier. I can re-sort next.
* I removed unnecessary statements (`print` mostly).
* I added new test, all of which would not pass with the previous code.


test/rules/prefer_final_locals_test.dart
```diff
@@ -112,74 +107,88 @@ f() {
...
+
+  test_destructured_listPattern() async {
     await assertDiagnostics(r'''
 f() {
   var [a, b] = ['a', 'b'];
-  print('$a$b');
 }
 ''', [
-      lint(12, 6),
+      lint(8, 3),
```
pq, line 140: I'm having a hard time understanding this diff...  Are the test expectations changing here?

pq (commented):
I *love* the cleanup!

The renames are making the diffs hard to understand though. On the surface, it looks like a bunch of offsets have changed... Is that right? Were we flagging the wrong places before?

test/rules/prefer_final_locals_test.dart
srawlins: `[12, 6]` is the range for `[a, b]`, the ListPattern. `[8, 3]` is the range for `var`. This change keeps `PatternVariableDeclaration`s in sync with `VariableDeclarationList`s. I think this is a positive change, but I haven't thought too much of more complicated examples...

srawlins:
> The renames are making the diffs hard to understand though.

Sorry about that. How do the renames affect the assertion/code diffs? I did not re-sort, so the renames should not affect the test code.

> On the surface, it looks like a bunch of offsets have changed... Is that right? Were we flagging the wrong places before?

Yes, the offsets changed. I don't know if there is a right range to highlight and a wrong one. The change is mostly to highlight the word `var` instead of the internal patterns. This is what we do for other variable declarations, and I think highlights what text should change to comply with the rule. For example, in `var {'first': a, 'second': b} = ...`, we used to report the map pattern, but with this PR we report the keyword `var`. This would keep the highlighted range short (a map could be long and span multiple lines).

test/rules/prefer_final_locals_test.dart
pq: I think this is the right direction. I'm doing some spot checking and will confirm. Thanks!

pq:
Let's land and iterate. Thanks!

pq (approved):

pq:
Sorry, no, I'm not. Just want to do more testing. Looks good to me!

PR #4235, opened 2023-03-29
Simplify logic in implicit_reopen

srawlins:
* Many of the extension getters on `InterfaceElement?` helped account for a `null` receiver, but it turns out we don't want to report any lint when the annotated element is `null`, or when it's supertype's element is `null` either (we we also null-checking `supertype` in a few positions already).
* We also don't want to report any lint when the annotated element is not a `ClassElement`, or when it's supertype is not a `ClassElement`, so when we promote those variables, a lot of the benefits of the extension getters fall away.

### DIFF
--- a/lib/src/rules/implicit_reopen.dart
+++ b/lib/src/rules/implicit_reopen.dart
@@ -87,35 +87,30 @@ class _Visitor extends SimpleAstVisitor {
 
   void checkElement(InterfaceElement? element, NamedCompilationUnitMember node,
       {required String type}) {
-    if (element == null) return;
+    if (element is! ClassElement) return;
     if (element.hasReopen) return;
     if (element.isSealed) return;
     if (element.isMixinClass) return;
 
     var library = element.library;
-    var supertype = element.superElement;
+    var supertype = element.supertype?.element;
+    if (supertype is! ClassElement) return;
     if (supertype.library != library) return;
 
     if (element.isBase) {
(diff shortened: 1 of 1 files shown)


lib/src/rules/implicit_reopen.dart
```diff
@@ -87,35 +87,30 @@ class _Visitor extends SimpleAstVisitor {
 
   void checkElement(InterfaceElement? element, NamedCompilationUnitMember node,
       {required String type}) {
-    if (element == null) return;
+    if (element is! ClassElement) return;
```
bwilkerson, line 90: Is it impossible for a mixin to re-open another mixin (or a `mixin class`)? (That might be true, I'm just not sure.)

bwilkerson (approved):

lib/src/rules/implicit_reopen.dart
srawlins: I'm not sure either, but this lint rule only registers `ClassDeclaration`, so `element` _should_ only be a ClassDeclaration. I'd like to investigate this mixin question though.

eernstg: Drive-by comment: It is not possible to reopen a mixin or a mixin class: It has the modifier `base` or no modifier, and no subtype of a `base` declaration can exist without a modifier which is at least as strong as `base`.

PR #4238, opened 2023-03-30
unreachable: Do not count references in as- or is-expressions, or various patterns

srawlins:
This is the last planned work for this rule. None of the following references count as a "reachable use." Explanations are given in the comments for each case.

* a reference to a type in the right side of an as-expression, an is-expression, or a cast pattern
* a reference to a constructor as the constructed expression in a constant pattern
* a reference to a type as the type in an object pattern


lib/src/rules/unreachable_from_main.dart
```diff
@@ -142,12 +141,30 @@ class _ReferenceVisitor extends RecursiveAstVisitor {
     super.visitAnnotation(node);
   }
 
+  @override
+  void visitAsExpression(AsExpression node) {
+    node.expression.accept(this);
+    // Intentionally do not visit `node.type`, as a reference in the type of an
+    // as-expression is not good enough to count as "reachable". Marking a type
+    // as reachable only because it was seen in an as-expression would be a
+    // miscategorization if the type is never instantiated or subtyped.
```
bwilkerson, line 150: Just curious, but does the same reasoning apply for type annotations?
```
class _C {}
void f() {
  _C? c; // Is this a use? It's certainly dead code.
}
```

lib/src/rules/unreachable_from_main.dart
```diff
@@ -163,6 +180,22 @@ class _ReferenceVisitor extends RecursiveAstVisitor {
...
+  @override
+  void visitConstantPattern(ConstantPattern node) {
+    var expression = node.expression;
+    if (expression is InstanceCreationExpression) {
+      expression.argumentList.accept(this);
+      // Intentionally do not visit `expression.constructorName`, as a reference
+      // in a constant pattern is not good enough to count as "reachable".
+      // Marking a type as reachable only because it was seen in a constant
+      // pattern would be a miscategorization if the type is never instantiated
+      // or subtyped.
```
bwilkerson, line 192: Technically, this will instantiate the class, and the object being matched might even be equal to the instance even if it isn't an instance of the same class. I'm not sure that this should be skipped.

bwilkerson (approved):

lib/src/rules/unreachable_from_main.dart
srawlins: I _think_ that type annotations should also not count as a "use". I don't think we have that logic for `unused_element` so I decided not to include it in this PR. But I think a type annotation would be the same as an as-expression, etc (in fact, its typically the same AST node, TypeAnnotation). I want to think about parameter type annotations, local, type arguments, type parameter bounds etc.... (maybe none of them count as a use...)

lib/src/rules/unreachable_from_main.dart
srawlins: Hmm wait, can you explain that more? Can the constant being instantiated be equal to an object that wasn't instantiated the same way? Maybe I just need tests with factory constructors?

bwilkerson: I might be wrong. Does the constant have to use the implementation of `==` from `Object` (that is, `identical`)? I was thinking that it would be possible for the `==` to return `true` for an object other than the same constant, but maybe that's not true.

But in any case, an instance of the class is actually created so that it can be compared to the value being switched over.

pq:
I had an initial worry about the performance cost of this change but while this does bump the lint up in the ranks it's still in the pack of middle-expensive lints.

**before**

image

**after**

image


It'd be interesting to brainstorm a bit on speeding this up.  (Definitely something to prioritize if we see a lot of adoption.)

pq:
But anyway, cool beans!

pq (approved):
💎

natebosch:
@srawlins - should we land this?

srawlins:
Brian asked a question above that I haven't answered yet. I'm still game too though. Just need to get back to this.

lib/src/rules/unreachable_from_main.dart
srawlins: Good catch!! A constant pattern which is a const invocation is totally valid code there, and does instantiate an object. I've reverted this bit.

PR #4249, opened 2023-04-03
prefer_const_contsructors_in_immutables: Use @immutable class in test

srawlins:
The linter never had a chance to report lint in this test case.


pq (approved):
Good catch!

PR #4250, opened 2023-04-03
1.35.0

pq:
1.35.0

- add new lints: 
  - `explicit_reopen`
  - `unnecessary_breaks`
  - `type_literal_in_constant_pattern`
- updates to existing lints to support patterns and class modifiers
- remove support for:
  - `enable_null_safety`
  - `invariant_booleans`
  - `prefer_bool_in_asserts`
  - `prefer_equal_for_default_values`
  - `super_goes_last`
- fix `unnecessary_parenthesis` false-positives with null-aware expressions
- fix `void_checks` to allow assignments of `Future<dynamic>?` to parameters
  typed `FutureOr<void>?`
- fix `use_build_context_synchronously` in if conditions
- fix a false positive for `avoid_private_typedef_functions` with generalized
  type aliases
- update `unnecessary_parenthesis` to detect some doubled parens
- update `void_checks` to allow returning `Never` as void
- update `no_adjacent_strings_in_list` to support set literals and for- and
  if-elements
- update `avoid_types_as_parameter_names` to handle type variables
- update `avoid_positional_boolean_parameters` to handle typedefs
- update `avoid_redundant_argument_values` to check parameters of redirecting
  constructors
- improve performance for `prefer_const_literals_to_create_immutables`
- update `use_build_context_synchronously` to check context properties
- improve `unnecessary_parenthesis` support for property accesses and method
  invocations

---

/cc @bwilkerson @srawlins

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,36 @@
+1.35.0
+
+- add new lints: 
+  - `explicit_reopen`
+  - `unnecessary_breaks`
+  - `type_literal_in_constant_pattern`
+- updates to existing lints to support patterns and class modifiers
+- remove support for:
+  - `enable_null_safety`
+  - `invariant_booleans`
+  - `prefer_bool_in_asserts`
+  - `prefer_equal_for_default_values`
+  - `super_goes_last`
+- fix `unnecessary_parenthesis` false-positives with null-aware expressions
+- fix `void_checks` to allow assignments of `Future<dynamic>?` to parameters
+  typed `FutureOr<void>?`
+- fix `use_build_context_synchronously` in if conditions
+- fix a false positive for `avoid_private_typedef_functions` with generalized
+  type aliases
+- update `unnecessary_parenthesis` to detect some doubled parens
+- update `void_checks` to allow returning `Never` as void
+- update `no_adjacent_strings_in_list` to support set literals and for- and
+  if-elements
+- update `avoid_types_as_parameter_names` to handle type variables
+- update `avoid_positional_boolean_parameters` to handle typedefs
+- update `avoid_redundant_argument_values` to check parameters of redirecting
+  constructors
+- improve performance for `prefer_const_literals_to_create_immutables`
+- update `use_build_context_synchronously` to check context properties
+- improve `unnecessary_parenthesis` support for property accesses and method
+  invocations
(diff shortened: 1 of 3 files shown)


bwilkerson (approved):

mono0926:
@pq 

> explicit_reopen

There is no such rule: https://dart-lang.github.io/linter/lints/explicit_reopen.html

Is [implicit_reopen](https://dart-lang.github.io/linter/lints/implicit_reopen.html) correct?

parlough:
You're correct, it was fixed in https://github.com/dart-lang/linter/commit/4c9ad1ad0d87532bec8e4d654f5f4de1f2ae9368 :)

mono0926:
@parlough 

> You're correct, it was fixed in https://github.com/dart-lang/linter/commit/4c9ad1ad0d87532bec8e4d654f5f4de1f2ae9368 :)

Thanks, this is also should be fixed, isn't it?

> - explicit_reopen
> 
> https://github.com/dart-lang/linter/releases/tag/1.35.0

parlough:
I'm not sure Git supports updating tag messages without recreating them. It's probably fine as it's not the main reference for learning about new lints.

mono0926:
@parlough @pq 

https://github.com/dart-lang/linter/releases/tag/1.35.0 is a release note, not "tag messages" 👌 

So @pq seems to be able to update it.

https://github.com/dart-lang/linter/releases/tag/1.35.0 is important information for me, so I was confused with `explicit_reopen` mistake, and asked as https://github.com/dart-lang/linter/pull/4250#issuecomment-1543288916.

pq:
Done. Thanks @mono0926! And sorry for the confusion!

natebosch:
@pq - why are we repeatedly changing the version number without publishing to pub?

Should this package have a new publish?

What does it accomplish to change the version number and _not_ publish? Is the version number meaningful when this package is used from the SDK?

natebosch:
IMO no - we should not have a version number which doesn't have meaning (and is never validated). We should also rethink how we handle the changelog since that isn't meaningful without a publish.

pq:
SGTM. Do you want to open an issue? We should probably pull a few more folks into the conversation but needless to say I'm happy to help drive.

natebosch:
Opened https://github.com/dart-lang/linter/issues/4389 for discussion.

PR #4254, opened 2023-04-03
`prefer_const_constructors_in_immutables`: simplify visitConstructorDeclaration override

srawlins:
While migrating the test cases, I had trouble reading the `visitConstructorDeclaration` override, so I tweaked it to be simpler, IMO. Let me know what you think.

I also moved the `prefer_const_constructors_in_immutables` tests to reflective. They are the same test cases.


pq (approved):
💎

PR #4255, opened 2023-04-04
Move several if-concerned use_build_context_synchronously tests

srawlins:
These tests are quite specific in their coverage, so I'm only migrating one subset. I hope the names on the reflective tests are meaningful, but I might add comments at the top of each...


srawlins:
FYI this is ready for review

pq (approved):
👍

PR #4263, opened 2023-04-07
New rule: Use matching super parameter names

srawlins:
Implements the proposal at https://github.com/dart-lang/linter/issues/4259

### DIFF
--- a/lib/src/rules/matching_super_parameters.dart
+++ b/lib/src/rules/matching_super_parameters.dart
@@ -92,10 +92,12 @@ class _Visitor extends SimpleAstVisitor<void> {
 
   @override
   void visitConstructorDeclaration(ConstructorDeclaration node) {
-    var positionalSuperParameters = node.parameters.parameters
-        .whereType<SuperFormalParameter>()
-        .where((p) => p.isPositional)
-        .toList();
+    var positionalSuperParameters = <SuperFormalParameter>[];
+    for (var parameter in node.parameters.parameters) {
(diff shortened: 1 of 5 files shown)


bwilkerson (approved):

pq:
I'd like to get a little more conversation on the proposal (https://github.com/dart-lang/linter/issues/4259) before landing.

srawlins:
Sounds good. Anything specific?

pq:
As per https://github.com/dart-lang/linter/blob/main/doc/lint-lifecycle.md we generally try and get a bunch of input before "accepting" a proposed lint. I'll go ahead and cc some folks who are reliably opinionated.

lib/src/rules/matching_super_parameters.dart
```diff
@@ -0,0 +1,133 @@
...
+}
+
+class _Visitor extends SimpleAstVisitor<void> {
+  final LintRule rule;
+
+  const _Visitor(this.rule);
+
+  @override
+  void visitConstructorDeclaration(ConstructorDeclaration node) {
+    var positionalSuperParameters = node.parameters.parameters
```
pq, line 95: I'd be inclined to do this less idiomatically in preference of just one iteration and without all the temporary objects. (I did a series of changes that dumbed down code like this and made it measurably faster.)

pq: I mean, parameter lists are short but I guess I think every bit counts. 🤔

srawlins: Done

PR #4264, opened 2023-04-08
Move more use_build_context_synchronously tests

srawlins:
This also makes some of the existing tests more compact, allowing for the reported range to have a smaller offset (like `lint(115, 21)` vs `lint(145, 21)`).


pq (approved):
💎

Issue #4265, opened 2023-04-08
"unrelated type" checking does not account for record types


srawlins:
**Describe the issue**

"Unrelated type checking" is used by a few rules:

* [`unrelated_type_equality_checks`](https://dart-lang.github.io/linter/lints/unrelated_type_equality_checks.html)
* [`collection_methods_unrelated_type`](https://dart-lang.github.io/linter/lints/collection_methods_unrelated_type.html)
* [`iterable_contains_unrelated_type`](https://dart-lang.github.io/linter/lints/iterable_contains_unrelated_type.html) (soft-deprecated)
* [`list_remove_unrelated_type`](https://dart-lang.github.io/linter/lints/list_remove_unrelated_type.html)

It currently decides that any a record type is "related" to any other type ☹️ . I discovered this when converting use of `Tuple2` map keys to record map keys. The analyzer did not complain when I asked a `Map<(int, int), String>` if it `containsKey` with a `Tuple2` ☹️ .

**To Reproduce**

Use any of the above linter rules with:

* a record type and a non-Null, non-Object, non-record type (considered related, but definitely unrelated)
* two record types with different shape (considered related, but definitely unrelated)
* two record types which are mutually non-assignable (considered related, but definitely unrelated)

**Expected behavior**

Each of the above should be considered unrelated. The only types that a given record type, `R1` should be considered "related to", modulo nullability, are:
* Null (maybe only if the record type is nullable? Match existing behavior for other types)
* Object
* another record type, `R2` such that `R2` is assignable to `R1`, or `R1` is assignable to `R2`. Assignability is the only consideration, and that is provided by `TypeSystem.isAssignableTo`.

(Maybe also `FutureOr<T>` where `T` is "related" to the record type.)

**Additional context**

The notion of "related types" in these lint rules has only grown organically, and has not been super principled. But reviewing the implementation, it is _very_ close to the Dart language's idea of "assignability". I think there are only a few main differences:

* "relatedness" is a symmetric notion. `T` is related to `U` iff `U` is related to `T`.
* In deciding whether two types are unrelated, each is considered modulo nullability (and I think the same for each component type). This is partially to support the symmetric notion. `Future<int>?` is related to `Future<num?>` even though neither is assignable to the other.
* In deciding whether two types are related, each of their component types (for example `T` in `Future<T>`, or in `Future<U extends T>`) are considered "related" independently. So `Map<num, int>` and `Map<int, num>` are related, even though neither is assignable to the other.
[...]

PR #4266, opened 2023-04-08
Account for records in type relatedness checks

srawlins:
Fixes https://github.com/dart-lang/linter/issues/4265

Uses `isAssignableTo` to handle relatedness, accounting for shape and subtypes.

### DIFF
--- a/lib/src/util/dart_type_utilities.dart
+++ b/lib/src/util/dart_type_utilities.dart
@@ -167,7 +167,7 @@ bool typesAreUnrelated(
     return false;
   }
   if (promotedLeftType is InterfaceType && promotedRightType is InterfaceType) {
-    return typeSystem._interfaceTypesAreUnrelated(
+    return typeSystem.interfaceTypesAreUnrelated(
         promotedLeftType, promotedRightType);
   } else if (promotedLeftType is TypeParameterType &&
       promotedRightType is TypeParameterType) {
@@ -262,7 +262,7 @@ class InterfaceTypeDefinition {
 }
 
 extension on TypeSystem {
-  bool _interfaceTypesAreUnrelated(InterfaceType type1, InterfaceType type2) {
+  bool interfaceTypesAreUnrelated(InterfaceType type1, InterfaceType type2) {
     // In this case, [leftElement] and [rightElement] each represent
     // the same class, like `int`, or `Iterable<String>`.
     var element1 = type1.element;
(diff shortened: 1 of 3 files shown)


lib/src/util/dart_type_utilities.dart
```diff
@@ -282,6 +261,39 @@ class InterfaceTypeDefinition {
   }
 }
 
+extension on TypeSystem {
+  bool _interfaceTypesAreUnrelated(InterfaceType type1, InterfaceType type2) {
```
pq, line 265: I might renamed this to `interfaceTypesAreUnrelated` but no biggie.

pq (approved):
Super awesome!

Might be good to have @bwilkerson take a quick look but LGTM!

lib/src/util/dart_type_utilities.dart
srawlins: Oops, I think I normally would have too 🤣  copy-paste error.

lib/src/util/dart_type_utilities.dart
```diff
@@ -282,6 +261,39 @@ class InterfaceTypeDefinition {
   }
 }
 
+extension on TypeSystem {
+  bool interfaceTypesAreUnrelated(InterfaceType type1, InterfaceType type2) {
+    // In this case, [leftElement] and [rightElement] each represent
```
bwilkerson, line 266: Consider renaming the parameters to `leftElement` and `rightElement`. I think they read better. Otherwise, consider updating the comments to match the current names.

(And I don't understand what "this case" is referring to here, but that's not your problem because you didn't write the original code.)

test/rules/unrelated_type_equality_checks_test.dart
```diff
@@ -18,6 +18,28 @@ class UnrelatedTypeEqualityChecksTestLanguage300 extends LintRuleTest
   @override
   String get lintRule => 'unrelated_type_equality_checks';
 
+  test_recordAndInterfaceType_unrelated() async {
```
bwilkerson, line 21: It would be good to have some tests involving records with named fields, and maybe some that have both positional and named fields.

bwilkerson (approved):

lib/src/util/dart_type_utilities.dart
srawlins: Good note; renamed.

I also really didn't understand what that comment referred to. After looking for a few minutes, I decided to just remove it.

test/rules/unrelated_type_equality_checks_test.dart
srawlins: Done!

PR #4271, opened 2023-04-11
Add comments to use_build_context_synchronously tests with fuller explanations

srawlins:
I've found myself writing linguistic gymnastics in the test case names for `use_build_context_synchronously`. This is an experiment where I use slightly shorter names, and appending suffixes like `2`, `3`, etc. and then using simpler English with punctuation in a comment at the top.

WDYT?


pq (approved):
Nice!

PR #4267, opened 2023-04-08
[WIP] dart 3.0!

pq:
Not ready for landing until we can bump the SDK constraint.

PR #4272, opened 2023-04-12
inverted token type extension

pq:
Follow-up from: https://github.com/dart-lang/linter/pull/4267

/cc @bwilkerson @kevmoo

### DIFF
--- a/lib/src/extensions.dart
+++ b/lib/src/extensions.dart
@@ -3,6 +3,7 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/token.dart';
 import 'package:analyzer/dart/constant/value.dart';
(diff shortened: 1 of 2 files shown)


lib/src/extensions.dart
```diff
@@ -254,6 +255,28 @@ extension ExpressionExtension on Expression? {
   bool get isNullLiteral => this?.unParenthesized is NullLiteral;
 }
 
+extension InhertanceManager3Extension on InheritanceManager3 {
```
kevmoo, line 258: Uh...tiny bit confusing w/ the code move around in this PR.

kevmoo (approved):

bwilkerson (approved):

lib/src/extensions.dart
pq: Yeah. I did a sort. (Which we generally try to do but are not currently enforcing.)

https://github.com/dart-lang/linter/issues/4273

PR #4279, opened 2023-04-16
[use_build_context_synchronously] Fix for unprotected 'last' invocation.

scheglov:
Also enables `propagate-linter-exceptions`, so that we get the crash when resolving test code.

This issue started to happen quite often internally, there is a spike of crashes.


srawlins (approved):

pq:
Fanastic. Thanks!

PR #4282, opened 2023-04-18
Fix unrelated_type_equality_checks when comparing enums

davidmartos96:
# Description

Currently the `unrelated_type_equality_checks` doesn't work when comparing different enums. This can be a gotcha when comparing similar enums in terms of context.

```dart
enum EnumOpKind1 { insert, update, delete }
enum EnumOpKind2 { upsert, delete }

void fun() {
  var x = EnumOpKind1.delete;
  if (x == EnumOpKind1.delete) print('delete'); // OK
  if (x == EnumOpKind2.delete) print('delete'); // This should lint
}
```

Same thing happens when comparing different Mixins. Although this I've noticed when I looked into the source code of the lint.

### DIFF
--- a/lib/src/util/dart_type_utilities.dart
+++ b/lib/src/util/dart_type_utilities.dart
@@ -290,10 +290,9 @@ extension on TypeSystem {
     } else {
       var sameSupertypes = leftElement.supertype == rightElement.supertype;
 
-      // Unrelated Enums or Mixins have the same supertype, but they are not the same element, so
+      // Unrelated Enums have the same supertype, but they are not the same element, so
       // they are unrelated.
-      if (sameSupertypes &&
-          (leftElement is EnumElement || leftElement is MixinElement)) {
+      if (sameSupertypes && leftElement is EnumElement) {
         return true;
       }
 
--- a/test_data/rules/unrelated_type_equality_checks.dart
+++ b/test_data/rules/unrelated_type_equality_checks.dart
@@ -196,18 +196,11 @@ void function32() {
   if (y == EnumImplements.a) print('a'); // LINT
 }
 
-void someFunction33(Mixin m1, Mixin2 m2) {
-  if (m1 == m1) print('someFunction32'); // OK
-  if (m1 == m2) print('someFunction32'); // LINT
-}
-
 class ClassBase {}
 
 class DerivedClass1 extends ClassBase {}
 
 mixin Mixin {}
-
-mixin Mixin2 {}
 
 class DerivedClass2 extends ClassBase with Mixin {}
 


lib/src/util/dart_type_utilities.dart
```diff
@@ -288,8 +288,16 @@ extension on TypeSystem {
       // Otherwise, they might be related.
       return false;
     } else {
+      var sameSupertypes = leftElement.supertype == rightElement.supertype;
+
+      // Unrelated Enums or Mixins have the same supertype, but they are not the same element, so
+      // they are unrelated.
+      if (true && (leftElement is EnumElement || leftElement is MixinElement)) {
```
srawlins, line 295: `true &&`? That's curious.

davidmartos96: Oh, nice catch. It remained while debugging. It used to be `sameSupertypes && (leftElement ...)` but it doesn't seem to be required. Should we add it anyway? I don't have enough background with the linter, so there may be edge cases I'm not aware of.

srawlins:
I'd actually prefer we address mixins separately. Mixins can have superconstraints (`mixin M on C`), and we might want to consider that. I think it is less cut-and-dry than enums. I think two mixins `M1` and `M2` could be considered related if they each have the same superconstraint... but I could be wrong. I think it is exceedingly rare for an element's static type to be a mixin...

test_data/rules/unrelated_type_equality_checks.dart
```diff
@@ -178,12 +178,26 @@ void function30() {
   if (c.determinant == callable2) print('someFunction30'); // OK
 }
 
+void function31() {
+  var x = EnumOpKind1.delete;
+  if (x == EnumOpKind1.delete) print('delete'); // OK
```
srawlins, line 183: Let's add another test comparing two values of the same enum type. `x == EnumOpKind1.update`.

test_data/rules/unrelated_type_equality_checks.dart
```diff
@@ -203,3 +217,6 @@ class ClassWCall {
 }
 
 class SubClassWCall extends ClassWCall {}
+
+enum EnumOpKind1 { insert, update, delete }
```
srawlins, line 221: It shouldn't affect the results, but I'd like a few more enum test cases. For example, the following are legal, and should still be considered unrelated to other enums: `enum E implements C` and `enum E with M`.

lib/src/util/dart_type_utilities.dart
srawlins: Yeah I guess you can revert the `sameSupertypes` calculation, and keep that at the end.

srawlins (changes requested):

davidmartos96:
@srawlins  Should I leave the "basic" Mixins handling or do you prefer them to be handled in a separate commit?

srawlins:
I think even the `|| leftElement is MixinElement` code would be better done in a separate PR.

srawlins (approved):
❤️  Love it. I need to check internal code for new violations, then we can land.

davidmartos96:
Great, thank you!

srawlins:
Internal looks clean. Will check flutter at the linter roll to the Dart SDK.

Issue #3296, opened 2022-03-17
add mock flutter support for rule unit tests


pq:
See: https://github.com/dart-lang/linter/pull/3295 for a work-around.

pq:
Update: the work-around doesn;t fly in Google3 where for some reason the mocked Flutter package isn't properly resolved:

https://dart-in-g3-qa-prod.corp.google.com/dg3/Home#/cbuild/find/9e515e899aaa7923aa5775e612a8fb87ba696a2c

PR #4283, opened 2023-04-20
Move use_key_in_widget_constructors tests to be reflective

srawlins:
All very clean tests, migrated cleanly with additional flutter mock code. :)

I think we can also say this closes https://github.com/dart-lang/linter/issues/3296.


pq (approved):
Nice!

PR #4288, opened 2023-04-20
`unnecessary_final` support for `ForEachPartsWithPattern`s

pq:
/cc @bwilkerson

### DIFF
--- a/lib/src/extensions.dart
+++ b/lib/src/extensions.dart
@@ -428,11 +428,7 @@ extension NullableAstNodeExtension on AstNode? {
 }
 
 extension TokenExtension on Token? {
-  bool get isFinal {
-    var self = this;
-    if (self == null) return false;
-    return self.keyword == Keyword.FINAL;
-  }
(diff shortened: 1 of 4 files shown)


lib/src/extensions.dart
```diff
@@ -427,6 +427,14 @@ extension NullableAstNodeExtension on AstNode? {
   }
 }
 
+extension TokenExtension on Token? {
+  bool get isFinal {
+    var self = this;
+    if (self == null) return false;
+    return self.keyword == Keyword.FINAL;
```
bwilkerson, line 434: Not that it matters, but I think you could just write `return this?.keyword == Keyword.FINAL;` (and drop the two lines above).

bwilkerson (approved):

lib/src/extensions.dart
pq: Ah, but even better still...

```dart
extension TokenExtension on Token? {
  bool get isFinal => this?.keyword == Keyword.FINAL;
}
```

Thanks!

PR #4294, opened 2023-04-21
prefer_final_in_for_each_test for final records

TzviPM:
# Description

Adds ok variant to tests added in #4291 to avoid regressions


pq:
Thanks!

pq (approved):
💎

PR #4296, opened 2023-04-21
fix `no_leading_underscores_for_local_identifiers` pattern field over-reporting

### LINKED ISSUE #4295: no_leading_underscores_for_local_identifiers shouldn’t fire if the variable is defined in an object pattern field with an implied name
**Describe the issue**
When a variable name is declared implicitly using `ObjectPattern(:var name)` syntax, the user doesn't have free choice in the variable name; the name must match the field of the object they are destructuring.

The lint `no_leading_underscores_for_local_identifiers` shouldn't fire in this case, because it can't be fixed by a simple rename.

**To Reproduce**
Given this Dart 3.0 code:
```dart
class Foo {
  var _bar;
}
test(Object o) {
  switch (o) {
    case Foo(:var _bar): // (1)
      print(_bar);
  }
}
```
The lint `no_leading_underscores_for_local_identifiers` is firing at (1).  However the suggested quick fix ('Remove leading underscore') produces this:
```dart
  switch (o) {
    case Foo(:var bar): // (1)
      print(bar);
  }
```
(the omitted lines are unchanged).  This is a compile-time error because there is no field `bar` in the class `Foo`.

It's tempting to suggest also renaming `Foo._bar`, but that would be bad because it would expose private details of the class `Foo` to consumers of the library.

In principle, one *could* avoid the lint by rewriting the object pattern using an explicit name, i.e.:
```dart
  switch (o) {
    case Foo(_bar: var bar): // (1)
      print(bar);
  }
```
But I would argue that this is worse than the original because of the repetition of `bar`.  I think it would be better to simply make an exception so that the lint rule doesn't fire in cases like these.

**Expected behavior**
The lint shouldn't fire if the variable is defined by a variable pattern, and that variable pattern is either:
- An immediate subpattern of an object pattern, where the corresponding field name is omitted,
- or an immediate subpattern of a null-assert or null-check pattern, where the null-assert or null-check pattern is an immediate subpattern of an object pattern, where the corresponding field name is omitted.

Reproduced in Dart SDK https://github.com/dart-lang/sdk/commit/742612ddaca82b51387a129ed50738256ae2e243, which includes linter version 2212d95ec16f8c243e3fb9e8cd7d2328c914a0f4.

pq:
Fixes #4295

/cc @bwilkerson

### DIFF
--- a/test/rules/no_leading_underscores_for_local_identifiers_test.dart
+++ b/test/rules/no_leading_underscores_for_local_identifiers_test.dart
@@ -97,6 +97,18 @@ f() {
     ]);
   }
 
+  test_objectPattern_destructured_field() async {
+    await assertNoDiagnostics(r'''
+class A {
+  int _a;
+  A(this._a);
+}
+f() {
+  var A(:_a) = A(1);
+}
+''');
+  }
+
   test_objectPattern_ifCase() async {
     await assertDiagnostics(r'''
 class C {
@@ -129,6 +141,20 @@ f() {
     ]);
   }
 
+  test_objectPattern_switch_field() async {
+    await assertNoDiagnostics(r'''
+class A {
+  var _a;
+}
+
+f(A a) {
+  switch (a) {
+    case A(:var _a):
+  }
+}
+''');
+  }
+
   test_recordPattern_destructured() async {
     await assertDiagnostics(r'''
 f() {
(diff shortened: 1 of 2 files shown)


test/rules/no_leading_underscores_for_local_identifiers_test.dart
```diff
@@ -129,6 +141,20 @@ f() {
...
 
+  test_objectPattern_switch_field() async {
+    await assertNoDiagnostics(r'''
+class A {
+  var _a;
+}
+
+f(A a) {
+  switch (a) {
+    case A(:var _a):
```
stereotype441, line 152: Also consider testing:
```dart
case A(:var _a?):
case A(:var _a!):
```
(IMHO the lint should also be suppressed in these cases)

pq: Ah, great catch. I'd have thought these case would "just work" but I need to handle `NullCheckPattern`s and `NullAssertPattern`s explicitly. Thank you!

stereotype441 (approved):
lgtm!

PR #4305, opened 2023-04-24
Don't report avoid_function_literals_in_foreach_calls with null-aware chain

oprypin:
`foo.bar?.forEach` already worked correctly (is not reported) but `foo?.bar.forEach` was reported and even auto-fixed to non-working code.

Also extract a common util `containsNullAwareInvocationInChain` because it is now used in 3 places.

* Yet another followup for #3354


lib/src/util/nullability_utils.dart
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) 2023, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/ast/ast.dart';
+
+/// Return `true` if the expression is null aware, or if one of its recursive
+/// targets is null aware.
+bool containsNullAwareInvocationInChain(AstNode? node) {
```
pq, line 9: Consider implementing this as an extension? (And possibly putting it in `extensions.dart`?)

oprypin: Applied the change

pq (approved):
Nice! 👍

oprypin:
Could we merge these PRs in the next days? I wanna go enforcing the lint :)

PR #4313, opened 2023-04-27
Don't report avoid_function_literals_in_foreach_calls inside a cascade

oprypin:


lib/src/rules/avoid_function_literals_in_foreach_calls.dart
```diff
@@ -94,7 +94,8 @@ class _Visitor extends SimpleAstVisitor<void> {
         node.argumentList.arguments.isNotEmpty &&
         node.argumentList.arguments.first is FunctionExpression &&
         _isNonNullableIterable(target.staticType) &&
-        !_hasMethodChaining(node)) {
+        !_hasMethodChaining(node) &&
+        node.thisOrAncestorOfType<CascadeExpression>() == null) {
```
srawlins, line 98: This would reach too far, right? This would escape outside expressions and functions, etc.

```dart
foo
  ..one = 1
  ..two = lists.first.forEach((list) => print('$list!'));
```

The lint should first on this case, but there is a CascadeExpression in the syntax tree above the `forEach`.

oprypin: Done

srawlins: You are correct, haha. I guess you really need to even escape the function body to get a nice test case of that.

srawlins (approved):
Nice tests. Love it.

pq (approved):
💎

PR #4358, opened 2023-05-15
Report type_literal_in_constant_pattern more often.

scheglov:
As requested in https://github.com/dart-lang/linter/issues/4195#issuecomment-1523954808

### DIFF
--- a/lib/src/rules/type_literal_in_constant_pattern.dart
+++ b/lib/src/rules/type_literal_in_constant_pattern.dart
@@ -10,25 +10,29 @@ import '../analyzer.dart';
 const _desc = r"Don't use constant patterns with type literals.";
 
 const _details = r'''
-Use `== TypeName` or `TypeName _` instead of type literals in patterns.
+If you meant to test if the object has type `Foo`, instead write `Foo _`.
 
 **BAD:**
 ```dart
-void f(Type x) {
-  if (x case int) {
-    print('int');
+void f(Object? x) {
+  if (x case num) {
+    print('int or double');
   }
 }
 ```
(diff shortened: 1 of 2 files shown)


pq (approved):
🎉

bwilkerson (approved):

mit-mit:
Is this something we might be able to cherry pick?

scheglov:
The past experience left me with a feeling that we don't to cherry picks for linter.
@pq might confirm, or claim that I'm exaggerating :-)

bwilkerson:
We are able to do cherry picks from the linter, but it requires creating a new linter branch based on the last branch pulled into the SDK, applying the changes on that branch, and then pulling the new branch into the target SDK branch (bypassing the main SDK branch).

pq:
Yep. What Brian said. The most careful way to cherry-pick a linter fix is a bit painful and time-consuming. Alternatively we could pull in more changes and shoulder a bit of risk.

If this is high priority, I'm happy to look into it.

mit-mit:
We discussed this in the language team, and we think that it's high value enough to pursue as a cherry pick. We'll let you folks decide if the best approach is a linter roll or just patching this single change in.

mit-mit:
@pq did you have a chance to look into a cherry pick?

pq:
Yeah. I'm working on it. The trick is we'd like to get the change landed into main first so that the cherry-pick isn't out of sync with what's at tip of tree. That roll has been stalled for other reasons (https://dart-review.googlesource.com/c/sdk/+/304360) but we're getting there!

PR #4364, opened 2023-05-17
`unnecessary_null_checks` stops checking assignments to local variables

### LINKED ISSUE #2608: False positive from unnecessary_null_checks
Create a file containing the following code:
```
  void f() {
    var s = getS();
    if (s == null) {
      return;
    }
    if (s.length > 2) {
      s = getS()!;
    }
    s.substring(0);
  }
  String? getS() => null;
```
and enable the `unnecessary_null_checks` lint. It will flag the null check, but removing the null check causes a diagnostic on the last line of `f` because the assignment to `s` prevents `s` from being promoted.

The lint should not flag uses of the null check operator when it changes promotion.

oprypin:
- Fixes #2608

### DIFF
--- a/lib/src/rules/unnecessary_null_checks.dart
+++ b/lib/src/rules/unnecessary_null_checks.dart
@@ -5,7 +5,9 @@
 import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/token.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/element.dart';
 import 'package:analyzer/dart/element/type.dart';
+import 'package:analyzer/src/dart/ast/ast.dart'; // ignore: implementation_imports
 
 import '../analyzer.dart';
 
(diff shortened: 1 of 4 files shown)


lib/src/rules/unnecessary_null_checks.dart
```diff
@@ -5,7 +5,9 @@
 import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/token.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/element.dart';
 import 'package:analyzer/dart/element/type.dart';
+import 'package:analyzer/src/dart/ast/ast.dart'; // ignore: implementation_imports
```
scheglov, line 10: Please don't use internal libraries.

lib/src/rules/unnecessary_null_checks.dart
```diff
@@ -79,12 +81,18 @@ DartType? getExpectedType(PostfixExpression node) {
...
-          (parent.leftHandSide as Identifier).name !=
-              (node.operand as Identifier).name)) {
+  if (parent is AssignmentExpression && parent.operator.type == TokenType.EQ) {
+    var lhs = parent.leftHandSide;
+    var rhs = node.operand;
+    if (lhs is Identifier && rhs is Identifier && lhs.name == rhs.name) {
+      return null;
+    }
+    if (lhs is SimpleIdentifierImpl) {
+      var scope = lhs.scopeLookupResult?.getter;
```
scheglov, line 91: @bwilkerson See why I want to put `resolution` into identifiers? He does not know that the `writeElement` from the parent `AssignmentExpression` should be used here.

bwilkerson: It's an interesting assumption that `resolution` would have made the API easier to learn. :-)

PR #4369, opened 2023-05-19
revert null literal checking for `avoid_redundant_argument_values`

pq:
See: #4368

/cc @srawlins @bwilkerson


srawlins (approved):
I strongly prefer not to check in commented out code. Just noting.

Issue #4351, opened 2023-05-14
Lint docs should include installation instructions / example of usage


eseidel:
I always forget the syntax to include lints.  Would love to have something copy-paste in the docs, e.g.:

https://dart-lang.github.io/linter/lints/unawaited_futures.html

Should have a small section:

## Usage

```yaml
linter:
  rules:
    - unawaited_futures
```

Similar to what we do in pub:
https://pub.dev/packages/openapi_generator/install

Although honestly maybe a command line would be easier. :)  But even just adding the usage docs I think would be an improvement.

pq:
Great suggestion!

We've talked about a CLI for enabling lints but have never taken it anywhere (fyi @srawlins) but worth revisiting.

/fyi @MaryaBelanger @parlough @bwilkerson re: doc improvements. It'd be super-easy to add a usage footer to the generated docs. I'd be curious to get your 2 cents.

parlough:
I created an SDK issue to track a CLI command: https://github.com/dart-lang/sdk/issues/52394. It has made explaining adding pub dependencies so much easier and would be a big win for documentation purposes.

PR #4370, opened 2023-05-19
Include usage information in generated docs

parlough:
Eventually moving away from the site, but since it's still used and referenced, doesn't hurt to add some usage information.

Staged example: https://parlough-linter.web.app/lints/unnecessary_late.html

Example of usage information for unnecessary_late

Closes https://github.com/dart-lang/linter/issues/4351

### DIFF
--- a/tool/doc.dart
+++ b/tool/doc.dart
@@ -594,8 +594,10 @@ abstract class RuleGenerator {
   String get usageMarkdown => '''
 ## Usage
 
-Add `$name` under **linter > rules** in your
-[`analysis_options.yaml`](https://dart.dev/guides/language/analysis-options):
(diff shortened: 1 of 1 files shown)


pq (approved):
Very nice!

MaryaBelanger (approved):
Nice! I think I'm a little confused, though. For someone to end up seeing this new "Usage" section, wouldn't it be because the lint popped up for them (so, because they already are using it)? I'm guessing there is another use case for these docs; like; are users browsing the linter docs to identify lints they might want to enable?

parlough:
Yeah I find that many users are still using this site to find lints as well.

I'm not going to worry about moving use cases of this site to dart.dev until our improvements later on.

PR #4371, opened 2023-05-19
don't overreport `cast_nullable_to_non_nullable` on undefined classes

pq:
See also: https://logs.chromium.org/logs/dart/buildbucket/cr-buildbucket/8780744067138629361/+/u/analyze_flutter_flutter/stdout

/cc @scheglov

### DIFF
--- a/lib/src/rules/cast_nullable_to_non_nullable.dart
+++ b/lib/src/rules/cast_nullable_to_non_nullable.dart
@@ -80,6 +80,7 @@ class _Visitor extends SimpleAstVisitor<void> {
     if (expressionType != null &&
         type != null &&
         expressionType is! DynamicType &&
+        expressionType is! InvalidType &&
         context.typeSystem.isNullable(expressionType) &&
         context.typeSystem.isNonNullable(type)) {
       rule.reportLint(node);
--- a/lib/src/rules/cast_nullable_to_non_nullable.dart
+++ b/lib/src/rules/cast_nullable_to_non_nullable.dart
@@ -77,11 +77,11 @@ class _Visitor extends SimpleAstVisitor<void> {
   void visitAsExpression(AsExpression node) {
     var expressionType = node.expression.staticType;
     var type = node.type.type;
-    if (expressionType != null &&
-        type != null &&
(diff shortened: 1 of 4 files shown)


test/rules/cast_nullable_to_non_nullable_test.dart
```diff
@@ -0,0 +1,39 @@
...
+  test_castNullable() async {
+    await assertDiagnostics(r'''
+String? s;
+var a = s as String; 
+''', [
+      lint(19, 11),
+    ]);
+  }
+
+  @soloTest
```
scheglov, line 29: Remove it?

test/rules/cast_nullable_to_non_nullable_test.dart
```diff
@@ -0,0 +1,39 @@
...
+  test_castNullable_unresolvedType() async {
+    await assertDiagnostics(r'''
+Undefined? s;
```
scheglov, line 32: I think it does not matter whether you put `?` after invalid type.
But this is probably closer to what the actual code has, so makes sense.

lib/src/rules/cast_nullable_to_non_nullable.dart
```diff
@@ -80,6 +80,7 @@ class _Visitor extends SimpleAstVisitor<void> {
     if (expressionType != null &&
         type != null &&
         expressionType is! DynamicType &&
```
scheglov, line 82: It would have been more logical, if we checked all for `expressionType` first, and then `type`.
From performance point of view, neither type will ever be `null`.

scheglov (approved):

test/rules/cast_nullable_to_non_nullable_test.dart
pq: Oh wow. I'm going to write a check for this.  Sorry.

lib/src/rules/cast_nullable_to_non_nullable.dart
pq: Updated. Thanks!

PR #4372, opened 2023-05-19
check for soloed tests

pq:
Make `@soloTest`s break the build.

Follow-up from https://github.com/dart-lang/linter/pull/4371#discussion_r1199420285.

/cc @scheglov @bwilkerson

### DIFF
--- a/tool/checks/rules/no_solo_tests.dart
+++ b/tool/checks/rules/no_solo_tests.dart
@@ -1,0 +1,62 @@
+// Copyright (c) 2022, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/element.dart';
+import 'package:analyzer/dart/element/type.dart';
+import 'package:linter/src/analyzer.dart';
+
+const _desc = r"Don't commit soloed tests.";
+
+const _details = r'''
+**DON'T** commit a soloed test.
+
+**BAD:**
+```dart
+@soloTest
+test_myTest() async {
+  ...
+}
+```
+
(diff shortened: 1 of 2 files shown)


tool/checks/rules/no_solo_tests.dart
```diff
@@ -0,0 +1,60 @@
...
+}
+
+
+class _Visitor extends SimpleAstVisitor<void> {
+  final LintRule rule;
+
+  _Visitor(this.rule);
+
+  @override
+  void visitMethodDeclaration(MethodDeclaration node) {
```
scheglov, line 53: `test_reflective_loader` also considers methods names `solo_test_` to be, well, solo tests.
So, maybe check for these too.

pq: Thanks!

tool/checks/rules/no_solo_tests.dart
```diff
@@ -0,0 +1,60 @@
...
+    for (var annotation in node.metadata) {
+      if (annotation.name.name == 'soloTest') {
```
scheglov, line 60: In general, a better test would be something like we have in `ElementAnnotationImpl`:
```dart
  @override
  bool get isInternal => _isPackageMetaGetter(_internalVariableName);
```
Not for `package:meta` of course, but kind of like this.
Otherwise, `soloTest` could be from some other package.
And I simply don't like using unresolved data to check anything.
That's why we put this resolution information into AST!

scheglov (approved):
This change is useful, with using resolution, or not.

tool/checks/rules/no_solo_tests.dart
pq: Ah, cool. I'll update. I figured the chances of another soloTest to be remote but it's better practice for sure. Stand by!

bwilkerson (approved):

tool/checks/rules/no_solo_tests.dart
pq: Done.

Issue #4311, opened 2023-04-26
unrelated_type_equality_checks should tell us what the types are


Hixie:
I got this error today:
```
   info • Neither type of the operands of '==' is a subtype of the other • lib/src/builder.dart:87:9 • unrelated_type_equality_checks
```
...however, as far as I can tell, they _should_ be the same type, so it'd be really useful if the message told me what the types actually are (or at least, what the analyzer thinks they are).

Hixie:
(As an aside, the actual code was `newAdjustedStatus != adjustStatus` -- it would be nice if the message used `!=` rather than `==` when that was the actual operator used.)

PR #4380, opened 2023-05-23
Print the type names in unrelated_type_equality_checks

srawlins:
# Description

This PR mainly splits the lint message into two, one appropriate for expressions (BinaryExpression with `==` or `!=`) and one appropriate for patterns. In each case, a conscious decision was made to write the "argument" type before the "target" type (approximate terms for the two cases). In particular for patterns, this produces messages like:

> The operand type, 'String', is not a subtype nor a supertype of the value type, 'int'.

I think this is much more readable than printing the target type (value type) first, as the problem likely does not lie with the value type, but with the operand type. But I'm still very open to word-smithing; this one is wordy 🙁 .

Also some tidying:

* Move some utility functions into the `_Visitor`, so as not to pass around the `TypeSystem`.
* Move some utility functions to extensions, for readability.
* Rename `fixNum` to `fixnum`.

Fixes https://github.com/dart-lang/linter/issues/4311

### DIFF
--- a/lib/src/rules/unrelated_type_equality_checks.dart
+++ b/lib/src/rules/unrelated_type_equality_checks.dart
@@ -5,12 +5,10 @@
 import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/token.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
-import 'package:analyzer/dart/element/element.dart';
 import 'package:analyzer/dart/element/type.dart';
 import 'package:collection/collection.dart';
 
 import '../analyzer.dart';
-import '../extensions.dart';
 import '../util/dart_type_utilities.dart';
 
 const _desc =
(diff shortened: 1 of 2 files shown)


lib/src/rules/unrelated_type_equality_checks.dart
```diff
@@ -196,26 +173,72 @@ class _Visitor extends SimpleAstVisitor<void> {
...
+    var rightOperand = node.rightOperand;
+    if (rightOperand is NullLiteral) return;
+    var leftType = leftOperand.staticType;
+    if (leftType == null) return;
+    var rightType = rightOperand.staticType;
+    if (rightType == null) return;
+
+    if (_nonComparable(leftType, rightType)) {
+      rule.reportLint(
+        node,
```
bwilkerson, line 191: Would it be better to highlight the operator, to reduce the highlight range, or would that be confusing?

lib/src/rules/unrelated_type_equality_checks.dart
```diff
@@ -132,40 +130,19 @@ class DerivedClass2 extends ClassBase with Mixin {}
...
-    !(_isFixNumIntX(leftType) && _isCoreInt(rightType));
-
 class UnrelatedTypeEqualityChecks extends LintRule {
-  static const LintCode code = LintCode('unrelated_type_equality_checks',
-      "Neither type of the operands of '==' is a subtype of the other.",
+  static const LintCode expressionCode = LintCode(
+      'unrelated_type_equality_checks',
+      uniqueName: 'LintCode.unrelated_type_equality_checks_expression',
+      "The right operand type, '{0}', of the equality comparison is not a "
+          "subtype nor a supertype of the left operand type, '{1}'.",
```
bwilkerson, line 138: The message seems a little long and a little formal, so I took a stab at rewriting it. Not sure whether it's better.

The type of the right operand ('{0}') isn't a subtype or a supertype of the left operand ('{1}').

lib/src/rules/unrelated_type_equality_checks.dart
```diff
@@ -132,40 +130,19 @@ class DerivedClass2 extends ClassBase with Mixin {}
...
+      correctionMessage: 'Try changing one or both of the operands.');
+
+  static const LintCode patternCode = LintCode(
+      'unrelated_type_equality_checks',
+      uniqueName: 'LintCode.unrelated_type_equality_checks_pattern',
+      "The operand type, '{0}', is not a subtype nor a supertype of the value "
+          "type, '{1}'.",
```
bwilkerson, line 145: I did the same thing here:

The type of the operand ('{0}') isn't a subtype or a supertype of the value being matched ('{1}').

bwilkerson (approved):
A couple of thoughts, but nothing that requires changes.

lib/src/rules/unrelated_type_equality_checks.dart
srawlins: It would be nice to be smaller. At first I thought the right operand would be a good choice but in `a == b`, the "wrong" value (that the dev will likely change to fix the issue) could be either `a` or `b` (or both, or scrap the whole expression). The operator works, to keep the highlight range small.

PR #4383, opened 2023-05-23
fix `prefer_final_in_for_each` in "collection fors"

### LINKED ISSUE #4353: `prefer_final_in_for_each` doesn't look inside *collection for*
I think this should trigger at least one of the `prefer_final_*` lints, but it doesn't:

```dart
List<int> f() => [
      for (var i in [1, 2]) i + 3
    ];

### PR DESCRIPTION
**Pull request by pq**
Fixes #4353

/cc @bwilkerson

### DIFF
--- a/lib/src/rules/prefer_final_in_for_each.dart
+++ b/lib/src/rules/prefer_final_in_for_each.dart
@@ -61,7 +61,7 @@ class PreferFinalInForEach extends LintRule {
   void registerNodeProcessors(
       NodeLintRegistry registry, LinterContext context) {
     var visitor = _Visitor(this);
-    registry.addForStatement(this, visitor);
+    registry.addForEachPartsWithDeclaration(this, visitor);
   }
 }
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by bwilkerson — APPROVED** [review]
We'll want to make sure that the fix for this (assuming there is one) isn't broken before we pull this into the SDK.

**Comment by pq** [comment]
Agreed. Thanks!

**Comment by srawlins** [comment]
Seeing lots of failures in flutter for this roll: https://logs.chromium.org/logs/dart/buildbucket/cr-buildbucket/8780270934930079521/+/u/analyze_flutter_flutter/stdout?format=raw

**Comment by pq** [comment]
Hmmmm. Thanks for checking this. I didn't realize they had this lint enabled (thinking it was just a Google3 thing).

I guess the question now is whether the new diagnostics are _desired_.  If not, I'll revert the change.

@goderbauer: would you mind weighing in?

**Comment by pq** [comment]
Here's a concrete example.

```dart
        developer.postEvent(
          'Flutter.ImageSizesForFrame',
          <String, Object>{
            for (ImageSizeInfo imageSizeInfo in _pendingImageSizeInfo.values)
              imageSizeInfo.source!: imageSizeInfo.toJson(),
          },
        );
```

should now become:

```dart
        developer.postEvent(
          'Flutter.ImageSizesForFrame',
          <String, Object>{
            for (final ImageSizeInfo imageSizeInfo in _pendingImageSizeInfo.values)
              imageSizeInfo.source!: imageSizeInfo.toJson(),
          },
        );
```

Which *feels* consistent with the intent of this rule.

goderbauer:
The tide has turned a little bit in whether we want to slap `final` on everything. A lot of people find this too verbose for little benefit. However, I do think that the `prefer_final_in_for_each` lint should consistently apply to all kind of `for`s. So, I am in favor of this change. We can have a separate discussion of whether the lint as a whole is too verbose for Flutter's taste and disable it than (as a whole). I'll LGTM the required changes in Flutter as well.

pq:
Thanks!

PR #4386, opened 2023-05-23
Use more direct Effective Dart links

parlough:
This PR primary updates to the more concise `dart.dev/effective-dart` path we've [updated to](https://github.com/dart-lang/site-www/commit/176f6c58c204690fb5f421617b14f2eb99696f58) but also worked historically. It also updates multiple mentions of Effective Dart directly to the related rule for more relevant information. Also some linked to the wrong page completely, so updated those too.


pq (approved):
Nice!

PR #4387, opened 2023-05-23
unreachable_from_main: Do not consider type annotations when computing reachability

srawlins:
# Description

A reference in a type annotation is not good enough to count as "reachable". Marking a type as reachable only because it was seen in a type annotation would be a miscategorization if the type is never instantiated or subtyped.

This was inspired by a comment @bwilkerson made in an earlier review.

CC @goderbauer This is the last change I plan for this rule 😁

### DIFF
--- a/test/rules/unreachable_from_main_test.dart
+++ b/test/rules/unreachable_from_main_test.dart
@@ -83,18 +83,6 @@ void f<T extends C>() {}
 ''');
   }
 
-  test_class_reachableViaTypeInFunction() async {
-    await assertNoDiagnostics(r'''
-void main() {
-  f();
-}
-
-class C {}
-
-void Function(C)? f() => null;
-''');
-  }
-
   test_class_unreachable() async {
     await assertDiagnostics(r'''
 void main() {}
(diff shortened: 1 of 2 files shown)


bwilkerson (approved):
I don't remember the comment, but it sounds reasonable that type annotations don't count. It would be interesting to know whether there are any new violations, especially false positives, as a result of this change.

srawlins:
Yeah I'll try and get some numbers.

srawlins:
It's good you asked and it's good I tested on a few packages! I discovered that this was not sound in the case of type aliases: a type alias may be used in any type annotation, and that is enough to consider it "used." A function type or a record type, in particular, are not instantiated by referring to the alias by name (and interface types don't need to be either).

This PR is quite different now; another review would be appreciated.

test/rules/unreachable_from_main_test.dart
```diff
@@ -1007,6 +1009,109 @@ int x = 1;
     ]);
   }
 
+  test_typedef_reachable_eferencedInObjectPattern() async {
```
bwilkerson, line 1012: nit: typo in "eferenced"

srawlins: Thanks! Fixed

PR #4388, opened 2023-05-24
Expand constraint on `http`

natebosch:
The `1.0.0` release does not have any breaking changes.

Remove the pre-release suffix from the SDK constraint now that the SDK
is published.


srawlins (approved):

pq (approved):
👍

PR #4390, opened 2023-05-24
Update use_build_context_synchronously to track async state in a single visitor.

srawlins:
# Description

This is massive, apologies. I don't think I can make this change smaller though, without regressions. I'll start with the motivating example:

```dart
void foo(BuildContext context) async {
  await f();
  if (1 == 2) {
    if (!mounted) return;
    await f();
  } else {
    if (!mounted) return;
    await f();
  }
  Navigator.of(context);
}
```

(This is one of the new test cases.) The idea is that, today, with the `_MountedCheckVisitor`, we see that both branches inside the `if (1 == 2)` statement contain "exit-if-not-mounted" checks that protect the use of `context` the bottom. But that visitor fails to see that there be more awaits inside, and so does _not_ flag this code.

This PR rewrites the `_MountedCheckVisitor` into the `_AsyncStateVisitor`, which tracks nodes' "async state" relative to a "reference" node which follows any of the nodes in the visitor.

**Review tip:** I think it might be simpler, when looking at the diff, to just look at the right side. I ripped so much apart in converting the `_MountedCheckVisitor` into the `_AsyncStateVisitor`, and the `_MountedCheck` enum into the `_AsyncState` enum, that it doesn't make sense to look at the line-by-line diff, comparing the left and the right.
 
Notes:
* Since the new `_AsyncStateVisitor` must carefully pass "asynchronous gap" code up the tree (up the visitors), it includes overrides of many many many visit methods. Basically all expression-, pattern-, and statement-visiting methods.
* The elegant `_AwaitVisitor` is gone, as this work is now painstakingly done in `_AsyncStateVisitor`.
* I tried very hard to capture the design of the visitor and the enum in doc comments. It's a lot of text that I think is pretty crucial to understanding them.
* For the most part I did not touch the old-style test cases, but there were two that needed updating, so I moved them into the reflective test cases.
* Performance is not hindered. 🎉 . The benchmarks say the rule takes ~37ms before and after.
* I added a lot of tests. They should cover the majority of the new code. A lot of the new visit methods are very simple, forwarding to either `_asynchronousIfAnyIsAsync` or `_inOrderAsyncState` (again, check the docs, should be very helpful), so there is not 100% coverage of each method, but those two helpers are exercised well in the new tests.


srawlins:
This is ready for review. I can also meet for a review as its a big one :)

bwilkerson:
@scheglov Would you mind taking a look at this?

scheglov:
Yes, I will review, given the size, I want to say it before actually looking :-)

srawlins:
I also considered writing a design doc for this change, but haven't. If it would be of great benefit, I can write one up. If the benefit is "meh," I'd prefer to do other things. 😄

lib/src/rules/use_build_context_synchronously.dart
```diff
@@ -85,222 +85,626 @@ class UseBuildContextSynchronously extends LintRule {
...
   static const mountedName = 'mounted';
 
-  final AstNode child;
+  final AstNode reference;
+
+  _AsyncStateVisitor({required this.reference});
+
+  @override
+  _AsyncState? visitAdjacentStrings(AdjacentStrings node) =>
+      _asynchronousIfAnyIsAsync(node.strings);
```
scheglov, line 169: Is it possible?

scheglov (commented):
This looks a lot like reachability during flow analysis, and it is unfortunate if we had to repeat it in the linter. Might be worth to discuss with Paul to see if we can attach tracking an additional feature to the existing flow analysis.

bwilkerson:
I'll note that this PR is clean-up in preparation to moving this to the analyzer package. Not that that invalidates your concern about duplicating effort. I'd love to see us be able to use information from flow analysis to do this computation.

srawlins:
> This looks a lot like reachability during flow analysis, and it is unfortunate if we had to repeat it in the linter. Might be worth to discuss with Paul to see if we can attach tracking an additional feature to the existing flow analysis.

Yes, we discussed with @stereotype441 . It would be very neat to be able to register hooks with flow analysis to keep track of certain properties. He wrote (though I cannot find it, paraphrasing) that the current flow analysis system would not be a good fit for such hooks.

scheglov:
> I'll note that this PR is clean-up in preparation to moving this to the analyzer package. Not that that invalidates your concern about duplicating effort. I'd love to see us be able to use information from flow analysis to do this computation.

Are we moving this lint rule into the analyzer itself?

Hm... It does not looks like just a clean-up, more like a giant fix and expansion :-)

srawlins:
> Are we moving this lint rule into the analyzer itself?

No just the big visitor, maybe + a little more. @keertip needs it in order to implement a data-driven fix, I think for https://github.com/dart-lang/sdk/issues/50698.

So my plan was to land this big visitor, then add termination support (very similar to the ExitDetector), and then move that to analyzer code... somewhere.

> Hm... It does not looks like just a clean-up, more like a giant fix and expansion :-)

From the user's perspective, I think it is a big fix, but not an expansion in the rule being enforced. Looking at the implementation it is a giant expansion of code 😄 .

scheglov (approved):

srawlins:
I'll land this giant thing, and there are some follow-on CLs that need to go out (the rule has 12 TODOs written in it now), but I will try to mail much smaller changes. :)

PR #4395, opened 2023-05-26
Prepare for the breaking change to NamedType.

scheglov:
We will stop visiting `NamedType.name` node (and remove it). So, `NamedType.element` should be used explicitly.


srawlins (approved):

scheglov:
Ah, you updated `UnreachableFromMain` already.
OK, just one change now.

pq (approved):

PR #4396, opened 2023-05-26
+ no_wildcard_variable_uses

### LINKED ISSUE #4392: proposal: `no_wildcard_variable_uses`
## Description

Do not use wildcard parameters or variables.

## Details

There's a proposal [to change the language](https://github.com/dart-lang/language/blob/master/working/wildcards/feature-specification.md) so that (at least) parameters and local variables named `_` (e.g., wildcards) become non-binding. When that happens, existing code that uses wildcard parameters or variables will break. 

Enabling this lint now will make it less breaking to make `_` non-binding in a future Dart release.

## Kind

Error

## Bad Examples

```dart
var _ = 1;
print(_); // LINT
```

```dart
void f(int __) {
  print(__); // LINT multiple underscores too
}
```

## Good Examples 

```dart
for (var _ in [1, 2, 3]) count++;
```

```dart
try {
  ...
} catch (_) {
  ...
}
```

```dart
var [a, _, b, _] = [1, 2, 3, 4];
```

## Discussion

https://github.com/dart-lang/sdk/issues/51221

cc @dart-lang/language-team

### Discussion checklist

- [x] List any existing rules this proposal modifies, complements, overlaps or conflicts with.
- [x] List any relevant issues (reported here, the [SDK Tracker], or elsewhere).
- [x] If there's any prior art (e.g., in other linters), please add references here.
- [x] If this proposal corresponds to [Effective Dart] or [Flutter Style Guide] advice, please call it out. (If there isn't any corresponding advice, should there be?)
- [x] If this proposal is motivated by real-world examples, please provide as many details as you can.  Demonstrating potential impact is _especially valuable_.

[Writing Lints]: https://github.com/dart-lang/linter/blob/main/doc/writing-lints.md
[Effective Dart]: https://dart.dev/effective-dart
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[SDK Tracker]: https://github.com/dart-lang/sdk/issues

pq:
Fixes #4392

/cc @bwilkerson @scheglov @MaryaBelanger

### DIFF
--- a/lib/src/rules/no_wildcard_variable_uses.dart
+++ b/lib/src/rules/no_wildcard_variable_uses.dart
@@ -14,11 +14,12 @@ const _desc = r"Don't use wildcard parameters or variables.";
 const _details = r'''
 **DON'T** use wildcard parameters or variables.
 
-In a future version of the Dart langauge, wildcard parameters and local
-variables (e.g., named with just underscores like `_`, `__`, `___`, etc.) will
-become non-binding. When that happens, existing code that uses wildcard
-parameters or variables will break. In anticipation and to make adoption
-easier, this lint disallows wildcard and variable parameter uses.
+Wildcard parameters and local variables
+(e.g. underscore-only names like `_`, `__`, `___`, etc.) will
+become non-binding in a future version of the Dart language.
+Any existing code that uses wildcard parameters or variables will
+break. In anticipation of this change, and to make adoption easier,
+this lint disallows wildcard and variable parameter uses.
 
 
 **BAD:**
--- a/lib/src/rules/no_wildcard_variable_uses.dart
+++ b/lib/src/rules/no_wildcard_variable_uses.dart
@@ -1,0 +1,79 @@
+// Copyright (c) 2023, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+
+import '../analyzer.dart';
+import '../util/ascii_utils.dart';
+
+const _desc = r"Don't use wildcard parameters or variables.";
+
+const _details = r'''
+**DON'T** use wildcard parameters or variables.
+
+In a future version of the Dart langauge, wildcard parameters and local
+variables (e.g., named with just underscores like `_`, `__`, `___`, etc.) will
+become non-binding. When that happens, existing code that uses wildcard
+parameters or variables will break. In anticipation and to make adoption
+easier, this lint disallows wildcard and variable parameter uses.
+
+
+**BAD:**
+```dart
+var _ = 1;
+print(_); // LINT
+```
+
+```dart
+void f(int __) {
+  print(__); // LINT multiple underscores too
+}
+```
+
+**GOOD:**
+```dart
+for (var _ in [1, 2, 3]) count++;
+```
+
+```dart
+var [a, _, b, _] = [1, 2, 3, 4];
+```
+''';
+
+class NoWildcardVariableUses extends LintRule {
+  static const LintCode code = LintCode(
+      'no_wildcard_variable_uses', 'The referenced identifier is a wildcard.',
+      correctionMessage: 'Use an identifier name that is not a wildcard.');
+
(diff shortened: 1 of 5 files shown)


lib/src/rules/no_wildcard_variable_uses.dart
```diff
@@ -0,0 +1,79 @@
...
+}
+
+class _Visitor extends SimpleAstVisitor {
+  final LintRule rule;
+
+  _Visitor(this.rule);
+
+  @override
+  void visitSimpleIdentifier(SimpleIdentifier node) {
+    if (node.name.isJustUnderscores) {
```
pq, line 75: I *think* it's just this simple... Please tell me if I'm over-simplifying though!

lib/src/rules/no_wildcard_variable_uses.dart
```diff
@@ -0,0 +1,79 @@
...
+import 'package:analyzer/dart/ast/visitor.dart';
+
+import '../analyzer.dart';
+import '../util/ascii_utils.dart';
+
+const _desc = r"Don't use wildcard parameters or variables.";
+
+const _details = r'''
+**DON'T** use wildcard parameters or variables.
+
```
pq, line 15: @MaryaBelanger: any word-smithing appreciated!

lib/src/rules/no_wildcard_variable_uses.dart
scheglov: The lint says: don't *use*.
Why not: don't *declare*?
There might be multiple uses, but they original reason is that something declared it.

Also, this visit method will fire on getters `print(this._)`.
Do we want this?

bwilkerson: > Why not: don't declare?

Read https://github.com/dart-lang/sdk/issues/51221.

> Do we want this?

No, the lint should only fire for uses of local variables (including parameters). We need tests that verify that it doesn't fire for members, including constructors and top-level functions, getters, etc.

pq: Thanks @bwilkerson. I'd forgotten about class members.

scheglov: Ah, of course I should have read the issue discussion.

pq: OK. Added executable element filtering (and tests).  Thanks!

lib/src/rules/no_wildcard_variable_uses.dart
```diff
@@ -0,0 +1,82 @@
...
+}
+
+class _Visitor extends SimpleAstVisitor {
+  final LintRule rule;
+
+  _Visitor(this.rule);
+
+  @override
+  void visitSimpleIdentifier(SimpleIdentifier node) {
+    if (node.staticElement is ExecutableElement) return;
```
scheglov, line 76: This looks too vague in my opinion.
Why to explicitly say what we want to check?
`element is LocalVariableElement || element is ParameterElement`

pq: SGTM. Updated.

scheglov (approved):

lib/src/rules/no_wildcard_variable_uses.dart
```diff
@@ -0,0 +1,85 @@
...
+const _desc = r"Don't use wildcard parameters or variables.";
+
+const _details = r'''
+**DON'T** use wildcard parameters or variables.
+
+In a future version of the Dart langauge, wildcard parameters and local
+variables (e.g., named with just underscores like `_`, `__`, `___`, etc.) will
+become non-binding. When that happens, existing code that uses wildcard
+parameters or variables will break. In anticipation and to make adoption
+easier, this lint disallows wildcard and variable parameter uses.
```
MaryaBelanger, line 21:
```suggestion
Wildcard parameters and local variables
(e.g. underscore-only names like `_`, `__`, `___`, etc.) will
become non-binding in a future version of the Dart language.
Any existing code that uses wildcard parameters or variables will
break. In anticipation of this change, and to make adoption easier,
this lint disallows wildcard and variable parameter uses.
```

MaryaBelanger (approved):
LGTM! Just a slight reword, take it or leave it :)

PR #4415, opened 2023-06-01
remove stale unfixable lint data

pq:
/cc @bwilkerson


tool/canonical/scorecard.dart
```diff
@@ -128,21 +128,9 @@ int _compareRuleSets(List<String> s1, List<String> s2) {
...
-    var name = entry['lint'];
-    var notes = entry['notes'];
-    if (notes != 'TODO') {
-      skipped.add(name as String?);
-    }
-  }
-  return skipped;
-}
+List<String?> _getUnfixableLints() =>
+    // todo(pq): update with data extracted from error_fix_status.yaml
```
bwilkerson, line 132: Or from the `FixProcessor`'s table, which we could access directly after the move. No need for an update to the PR, just a thought.

bwilkerson (approved):

Issue #4389, opened 2023-05-24
Remove the version from the pubspec and find a new process for the CHANGELOG


natebosch:
Since we aren't publishing the package anymore the `version` config in the pubspec does not have meaning.

Since we roll into the SDK by SHA instead of version tag, the headers in the CHANGELOG don't have meaning. Overall without a publish there isn't a meaningful way for end users to find the _relevant_ changelog entries for the SDK version they are using.

We should stop populating these values.

@pq - do we have some kind of process for editing the SDK CHANGELOG when we update the SHA configured in DEPS for the SDK?

srawlins:
I would be in favor of sort of backfilling the linter CHANGELOG with SDK versions. So starting now we'd write an entry for `# Dart 3.1.0-dev` or something, and add to that list. And then when it's released, we probably have to do some fiddling to put the most recent CHANGELOG entries in the right spot. But it's meaningful. Maybe that double entry with the SDK's actual CHANGELOG is too much fuss?

natebosch:
> Maybe that double entry with the SDK's actual CHANGELOG is too much fuss?

That also made me think of another option:

Changelog here is structured like:

```md
## Next

* Some update since the last SDK sync.

## <Date of most recent SDK sync> - <SDK hash>

* Some update which is in the SDK changelog now.

## 2023-01-01 - <SDK hash>

* Some update which is also in the SDK changelog already.
```

We might need to start by having to do and validate it manually, but I think it could be automated. I think we'd get a lot more safety out of it. cc @devoncarew - some potential features we could look at implementing in general?

The process we'd follow is:

- On PR validate that an appropriate CHANGELOG entry is added under `## NEXT`, or the feature is not user facing (validation could be automated)
- On SDK DEPS change, move any CHANGLEOG entries under `## NEXT` to the SDK changelog. (file change could be automated in bump_deps?)
- After merging SDK DEPS change, open a PR here to rename `## NEXT` to `## <Current Date> - SDK merge` (could be automated with github actions?)

It sacrifices some utility of the CHANGELOG as a user facing file, but I think we gain some utility in making the automation easier, and enforcing CHANGELOG from the start. There is a risk for diff conflicts in outstanding PRs after each SDK sync, so we may want to be sure we can mitigate that before we try it.

devoncarew:
[...]

PR #4427, opened 2023-06-03
move to new "versioning" scheme

pq:
Removes the `version` from our pubspec and updates the changelog to a new scheme based on 3.1.0 (the next stable Dart version).

My thinking is we do not need to suffix this with `-dev` since we're not really using semver and don't see the value of a suffix that we'll have to remember to update.

Very open to input though.

See: #4389

/cc @bwilkerson @natebosch @devoncarew

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,10 @@
-1.35.0
+# 3.1.0
+
+- new lint: `no_wildcard_variable_uses`
+
+---
+
+# 1.35.0
 
 - add new lints: 
   - `implicit_reopen`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-# 3.1.0
+# 3.1.0-wip
 
 - new lint: `no_wildcard_variable_uses`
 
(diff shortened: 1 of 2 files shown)


bwilkerson (approved):

CHANGELOG.md
```diff
@@ -1,4 +1,10 @@
-1.35.0
+# 3.1.0
```
devoncarew, line 1: Just for clarity, we plan for this to reflect the sdk version that the change / addition shipped in?

We may want to continue to use `3.1.0-wip` here then, until the sdk does ship, and we can retroactively remove the `-wip` from the latest entry and add a space for a new `3.(N+1).0-wip` entry. I'm not sure how this would all be handled through when the linter moves into the sdk - whether we still want a separate changelog or not.

devoncarew (approved):

CHANGELOG.md
pq: Works for me.  As for when the linter is in the SDK, we might still want this changelog but I'm totally open to deleting it.

pq: > Just for clarity, we plan for this to reflect the sdk version that the change / addition shipped in?

Sorry, yes. This is right.

PR #4429, opened 2023-06-04
generate/update all rule-supporting files

pq:
Updates the rule generation script to:

* (re)generate `all.dart`
* (re)generate `all.yaml`
* (re)generate `rules.dart`

With this, there are **no manual steps** post rule creation besides writing and testing the rule.

You'll notice a few supporting changes too.

1. sorting has been fixed
2. inconsistent names have been fixed
3. a misplaced pub rule has been put in the "pub" rule group directory

/cc @bwilkerson @srawlins


lib/src/rules.dart
```diff
@@ -420,15 +419,16 @@ void registerLintRules({bool inTestMode = false}) {
     ..register(UnnecessaryConst())
     ..register(UnnecessaryConstructorName())
     ..register(UnnecessaryFinal())
-    ..register(UnnecessaryNew())
```
pq, line 423: I won't miss manual sorting.

(Obviously, I'm no good at it! 🙄)

pq:
Hmmmm. I'm seeing version skew with integration-style tests thanks to the test class renames. (Another gotcha I won't miss!)

srawlins (approved):
❤️

bwilkerson (approved):

PR #4430, opened 2023-06-05
+ changelog process notes

pq:
Some notes on `CHANGELOG` updates.

In support of https://github.com/dart-lang/linter/issues/4389.

### DIFF
--- a/doc/lint-lifecycle.md
+++ b/doc/lint-lifecycle.md
@@ -74,6 +74,6 @@ Experimental lints should aspire to be stable. An experimental lint is a candida
 
 ### Removal
 
-Implemented lints can be removed.
+Deprecated or experimental lints can be removed. A stable lint should be deprecated before removal.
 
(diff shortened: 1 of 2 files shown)


doc/lint-lifecycle.md
```diff
@@ -50,17 +50,30 @@ In general removal is preceded by a period of deprecation.
...
 
 Experimental lints should aspire to be stable. An experimental lint is a candidate for stable when it has
 
 * complete semantics
 * complete implementation (with no false-positives)
 * established long-term value (e.g., inclusion in a recommended lint set)
+
+### Removal
+
+Implemented lints can be removed.
```
bwilkerson, line 77: We might want to be more specific about which states can be transitioned to which other states. In particular, "Implemented" isn't very informative because it isn't a state. We might, for example, want to require that only deprecated or experimental lints can be deleted in order to ensure that users have some form of notification.

bwilkerson (approved):
My one comment is just me thinking out loud, and doesn't require any changes.

PR #4436, opened 2023-06-06
use_build_context: Fix a failing test regarding the looping of while loops

srawlins:
# Description

In a while loop, an async gap that comes _after_ a BuildContext reference should be reported!

Statements that precede a BuildContext reference can guard it with a mounted check, but not statements that follow the BuildContextReference.

In order to accomplish this, note that a _lot_ of the complexity in `check()` is moved to the `_AsyncStateVisitor`, mostly in `visitBlock`. I think it is a big benefit to make `check()` simpler, when it comes to special treatment of parent node types.

CC @keertip


srawlins:
This is still ready for review.

keertip (approved):

PR #4441, opened 2023-06-08
[unreachable_from_main] Support for instance members.

scheglov:
When applied to `analyzer/`, `analysis_server`, etc found a handful of violations.
https://dart-review.googlesource.com/c/sdk/+/307975

### DIFF
--- a/lib/src/rules/unreachable_from_main.dart
+++ b/lib/src/rules/unreachable_from_main.dart
@@ -70,12 +70,16 @@ class UnreachableFromMain extends LintRule {
   }
 }
 
-/// This gathers all of the top-level and static declarations which we may wish
-/// to report on.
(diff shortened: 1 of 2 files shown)


lib/src/rules/unreachable_from_main.dart
```diff
@@ -181,6 +229,16 @@ class _ReferenceVisitor extends RecursiveAstVisitor {
     super.visitClassDeclaration(node);
   }
 
+  @override
+  visitConstantPattern(ConstantPattern node) {
+    _patternLevel++;
```
srawlins, line 234: Nice impl

srawlins (approved):
The testing looks pretty complete! I'd love to test this in our packages.

PR #4443, opened 2023-06-08
migrate since SDK data to a new map

pq:
Proposes a new way to associate SDK version info w/ lint rules.  Adopting this will allow us to get rid of a bunch of complex logic elsewhere in `tool/` (starting w/ `dart_sdk.yaml` and all of it's clients) and get us closer to being able to remove our dependency on `package:github` (removing it from rule doc code generation).

See: https://github.com/dart-lang/sdk/issues/59174

See also: dart-lang/sdk#59173

/cc @parlough @bwilkerson


bwilkerson (approved):

parlough:
Nice! I like the test :)

PR #4447, opened 2023-06-08
remove fix entry

pq:
As per new CHANGELOG policy, keep it lean (only lint adds, removes, and deprecations).

See: #4389

/cc @scheglov


scheglov (approved):
Sorry again :-)

pq:
No worries! It's a very recent change!

PR #4455, opened 2023-06-13
Many small improvements to use_build_context_synchronously

srawlins:
# Description

* Properly support `await`s in CatchClauses.
* Other correctness fixes for TryStatements, with new tests.
* Properly support `await`s in ConditionalExpressions and IfElements.
* Support WhenGuards with mounted guards, which can guard statements in a SwitchPatternCase.
* Remove all special casing in `check()`; we just unconditionally move to each parent, and check the async state.
* Move some ForStatement tests from the integration style to the reflective style.

### DIFF
--- a/test/rules/use_build_context_synchronously_test.dart
+++ b/test/rules/use_build_context_synchronously_test.dart
@@ -23,13 +23,13 @@ class AsyncStateTest extends PubPackageResolutionTest {
 
   FindNode get findNode => FindNode(result.content, result.unit);
 
-  Future<void> assertNoErrorsInCode(String code) async {
+  Future<void> resolveCode(String code) async {
     addTestFile(code);
     await resolveTestFile();
   }
 
   test_adjacentStrings_referenceAfter_awaitInString() async {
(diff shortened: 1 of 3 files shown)


srawlins:
CC @keertip

test/rules/use_build_context_synchronously_test.dart
```diff
@@ -353,8 +351,123 @@ void foo(BuildContext context, int i) async {
     expect(forElement.asyncStateFor(reference), AsyncState.mountedCheck);
   }
 
+  test_forStatementWithDeclaration_referenceAfter_awaitInPartCondition() async {
```
keertip, line 354: Exhaustive testing 👍

keertip (approved):

test/rules/use_build_context_synchronously_test.dart
srawlins: Haha yeah, kind of tiresome, but... I guess since these are all encoded in the visitor, its good to cover them.

PR #4463, opened 2023-06-14
stop flagging empty statements in switch pattern cases

### LINKED ISSUE #4410: "empty_statements" doesn't know about case fallthrough
I had this code:

```dart
      switch (customization.touchFeedback) {
        case HapticStyle.none:
          ;
        case HapticStyle.vibrate:
          HapticFeedback.vibrate();
        // ...
```

...and the `;` line got the message "Unnecessary empty statement", but if I remove it, it changes the semantics of the code.

pq:
Fixes #4410

/cc @bwilkerson

### DIFF
--- a/lib/src/rules/empty_statements.dart
+++ b/lib/src/rules/empty_statements.dart
@@ -74,17 +74,13 @@ class _Visitor extends SimpleAstVisitor<void> {
     if (parent is! SwitchPatternCase) return false;
 
     var statements = parent.statements;
-    var justEmpties = true;
+    if (statements.last != node) return false;
 
-    var index = 0;
-    for (var (i, statement) in statements.indexed) {
-      if (statement == node) index = i;
-      if (statement is! EmptyStatement) justEmpties = false;
+    for (var statement in statements) {
+      if (statement is! EmptyStatement) return false;
     }
 
-    var isLast = index == statements.length - 1;
-    if (isLast) return justEmpties;
-    return false;
+    return true;
   }
 
   @override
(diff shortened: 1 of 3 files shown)


lib/src/rules/empty_statements.dart
```diff
@@ -71,6 +71,7 @@ class _Visitor extends SimpleAstVisitor<void> {
 
   @override
   void visitEmptyStatement(EmptyStatement node) {
+    if (node.parent is SwitchPatternCase) return;
```
bwilkerson, line 74: I wonder whether we want to only allow empty statements when the case contains only empty statements and it's the last such statement. For example, consider the following cases
```
case a:
  print('a');
  ; // [LINT]
case b:
  ; // [LINT]
  print('b');
case c:
  ; // [LINT]
  ; // [OK]
case d:
  ; // [OK]
```

pq: Ah, nice. 💯

pq: I updated the implementation and tests. PTAL!

lib/src/rules/empty_statements.dart
```diff
@@ -69,8 +69,27 @@ class _Visitor extends SimpleAstVisitor<void> {
...
 
+  bool definesSemantics(EmptyStatement node) {
+    var parent = node.parent;
+    if (parent is! SwitchPatternCase) return false;
+
+    var statements = parent.statements;
+    var justEmpties = true;
+
+    var index = 0;
+    for (var (i, statement) in statements.indexed) {
```
bwilkerson, line 80: I believe that this works, but it seems a bit more complex than necessary. If the `node` isn't the last statement, then we can return `false`. If it is, then we can iterate over the previous statements and return `false` as soon as we find one that is not an empty statement. Otherwise we return `true`.

Your call whether to change it.

bwilkerson (approved):

PR #4464, opened 2023-06-15
tweak `prefer_conditional_assignment` messaging

### LINKED ISSUE #4398: prefer_conditional_assignment message is confusing
The message for "prefer_conditional_assignment" is "Unnecessary 'null' comparison before assigning a value":

```
Object?	bar() => null;

void main() {
  Object? foo = bar();
  if (foo == null) { // "Unnecessary 'null' comparison before assigning a value"
    foo = bar();
  }
  print(foo);
}  
```

This makes it sound like the fix is to remove the comparison:

```
Object?	bar() => null;

void main() {
  Object? foo = bar();
  foo = bar();
  print(foo);
}  
```

...which of course is semantically quite different.

Something like "Prefer the "??=" operator instead of an "if" statement" might be clearer?

pq:
Here's a crack at a reword.

Feedback welcome.

Fixes #4398

/cc @bwilkerson @hixie

### DIFF
--- a/lib/src/rules/prefer_conditional_assignment.dart
+++ b/lib/src/rules/prefer_conditional_assignment.dart
@@ -68,8 +68,9 @@ Expression? _getExpressionCondition(Expression rawExpression) {
 
 class PreferConditionalAssignment extends LintRule {
   static const LintCode code = LintCode('prefer_conditional_assignment',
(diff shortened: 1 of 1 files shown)


pq:
@hixie: I'll land this as is, thinking it's a definite improvement. Feel free to chime in if you can see other possible tweaks.

PR #4469, opened 2023-06-18
Allow overloading of operators in unnecessary_statements

### LINKED ISSUE #4468: `unnecessary_statements` shouldn't trigger when the operator is overloaded
Say you some code that overloads an operator:

```dart
class Logger {
	void operator<<(String message) {/*...*/}
}

void main() {
	final l = Logger();

	l << "Hello world (づ｡◕‿‿◕｡)づ"; 
	// Unnecessary statement.
	// Try completing the statement or breaking it up. (unnecessary_statements)
}
```
The lint trigger when it shouldn't as the statement has a clear effect here.

gaetschwartz:
# Description

This PR disables the `unnecessary_statements` for overloaded operators, so that code like this is possible:

```dart
class Logger {
	void operator<<(String message) {/*...*/}
}

void main() {
	final l = Logger();

	l << "Hello world (づ｡◕‿‿◕｡)づ"; 
	// Unnecessary statement.
	// Try completing the statement or breaking it up. (unnecessary_statements)
}
```

Fixes #4468

### DIFF
--- a/lib/src/rules/unnecessary_statements.dart
+++ b/lib/src/rules/unnecessary_statements.dart
@@ -106,19 +106,20 @@ class _ReportNoClearEffectVisitor extends UnifyingAstVisitor {
           (c) => c.declaredElement!.extendedType == node.leftOperand.staticType,
         );
 
-    // if the operator in the class definition is overloaded, it has a clear effect
+    // if the operator in the class definition is overloaded, it has a possible effect
     if (clazz != null) {
       // look for the operator in the members of the class
       var op = clazz.members
           .whereType<MethodDeclaration>()
           .firstWhereOrNull((m) => m.name.lexeme == node.operator.lexeme);
       if (op != null) {
-        // Has a clear effect
+        // Has a possible effect. For simplicity sake, we don't check
+        // if the method body has side effects.
         return;
       }
     }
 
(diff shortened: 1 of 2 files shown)


lib/src/rules/unnecessary_statements.dart
```diff
@@ -92,6 +94,48 @@ class _ReportNoClearEffectVisitor extends UnifyingAstVisitor {
 
   @override
   void visitBinaryExpression(BinaryExpression node) {
+    // look for the class of the left operand and its extensions
+    var clazz = context.currentUnit.unit.declarations
+        .whereType<ClassDeclaration>()
```
pq, line 99: Leaving open whether we agree that we *want* to loosen this check (which we should settle in the associated issue), I will note that if we do, we should try and optimized these new bits.

Here's a benchmark run before this change:

image


And after:

image

Either way, it does seem likely that we can establish if an operator is overriding more directly.  Maybe `UnnecessaryOverrides` can provide some ideas?  (In particular, some variation of `enclosingElement.thisType.lookUpMethod2(...)`?)

srawlins:
I don't see us accepting this, as there we are not treating https://github.com/dart-lang/linter/issues/4468 as a valid issue.

PR #4476, opened 2023-06-21
Block github package 9.15.0

srawlins:
# Description

Blocks github package version 9.15.0, which includes a breaking change. See https://github.com/dart-lang/linter/issues/4475


parlough:
Consider https://github.com/dart-lang/linter/pull/4477 instead.

PR #4478, opened 2023-06-21
Mark types of external declarations as "reachable"

### LINKED ISSUE #4474: unreachable_from_main ignores classes used through `external` members
The following has an unexpected lint:

```dart
void main() {
  c.doStuff();
}

class C {}
//    ^ unreachable_from_main

extension on C {
  void doStuff() {}
}

external C get c;
```

The real code this came up for has an `@JS()` on the `class` and the `external` getter, but it doesn't require the annotations for the static reproduction.

https://github.com/dart-lang/test/blob/3d5afedce6fe9643cafbeb89af45ebe9754233c2/pkgs/test/tool/host.dart#L18-L28

srawlins:
# Description

When looking at a NamedType, if it is found in the return type of an external member, call it reachable.

Fixes #4474

### DIFF
--- a/lib/src/rules/unreachable_from_main.dart
+++ b/lib/src/rules/unreachable_from_main.dart
@@ -175,10 +175,6 @@ class _ReferenceVisitor extends RecursiveAstVisitor {
   /// References from patterns should not be counted.
   int _patternLevel = 0;
 
-  /// Whether the visitor is currently in a declaration of an `external`
-  /// declaration.
-  bool _inExternalDeclaration = false;
-
   _ReferenceVisitor(this.declarationMap);
 
   @override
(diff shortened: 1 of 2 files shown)


bwilkerson (approved):

lib/src/rules/unreachable_from_main.dart
```diff
@@ -271,9 +271,13 @@ class _ReferenceVisitor extends RecursiveAstVisitor {
...
-    if (node.type?.alias != null) {
-      // Any reference to a typedef counts as "reachable", since structural
-      // typing is used to match against objects.
+    // Any reference to a typedef is reachable, since structural typing is used
+    // to match against objects.
+    //
+    // The return type of an external variable declaration is reachable, since
+    // the external implementation can instantiate it.
+    if (node.type?.alias != null ||
+        node.isInExternalVariableTypeOrFunctionReturnType) {
```
natebosch, line 280: Do we care to also check whether the external definition is also reachable?

srawlins: The visitor should only be checking the external definition if it was reachable from main.

PR #4481, opened 2023-06-24
New lint: type_test_on_generic_type_variable

### LINKED ISSUE #4480: proposal: `is_check_on_generic_argument` - flag `T is Foo` when `T` is a type variable
```dart
class C<T> {
  void something() {
    if (T is Foo) {} // LINT
    if (T is! Foo) {} // LINT
  }
}
```

I don't think there are any useful is checks on generic type variables. `T is Type` and `T is! AnythingElse` will always be `true`.

It is possible to do `T == AnythingElse`, but this is also risky because it does not handle subtypes like `is` would.

natebosch:
Closes #4480

A generic type variable will always be a `Type` and the check is never
useful. Some authors may expect an `is` to check what type the variable
represents. It's less likely an author would expect `as` to do something
useful, but it is also not useful.

### DIFF
--- a/lib/src/rules/type_test_on_generic_type_variable.dart
+++ b/lib/src/rules/type_test_on_generic_type_variable.dart
@@ -14,9 +14,12 @@ const _desc = r' ';
 const _details = r'''
 **DON'T** perform type tests on generic type variables.
 
-A type variable for a generic class or method will always be a `Type`. A type
-test with `is` or `as` does not check what the type variable is bound to, and
-does not relate to the values which have a type specified by the type variable.
+A type variable, the type parameter of a generic class or method,
+used as an expression always evalautes to a `Type` object.
+The first operand of `is` or `as` type tests is an expression,
+so a type test of a type variable is always type test on a `Type` instance.
+Such a test does not relate to the values which have a type specified by the
+type variable.
 
 **BAD:**
 ```dart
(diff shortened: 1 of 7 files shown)


lib/src/rules/type_test_on_generic_type_variable.dart
```diff
@@ -0,0 +1,90 @@
...
+import 'package:analyzer/dart/element/element.dart';
+
+import '../analyzer.dart';
+
+const _desc = r' ';
+
+const _details = r'''
+**DON'T** perform type tests on generic type variables.
+
+A type variable for a generic class or method will always be a `Type`. A type
```
natebosch, line 17: Copying @lrhn comment from the commit so it's on the PR instead.

https://github.com/dart-lang/linter/commit/b9d99dfddedbe051ac7f9d7eba0ad00b582a0e39#r119914277

Hmm.  A type variable isnt a `Type`, but it evaluates to one when used as an *expression*, not when used as a type.

Also "of that type" does not have another occurrence of "type" to refer back to. It refers to the type bound to the type variable, but a type variable is not a type (it's a name which can be used to refer to a type).

Maybe:

> A type variable, the type parameter of a generic class or method,
> used as an expression always evalautes to a `Type` object.
> The first operand of `is` or `as` type tests is an expression, so
> a type test of a type variable does not check any aspect
> of the type bound to that type variable. 
> It just tests an object of type `Type`.

natebosch: I took some of this wording and changed the last half. WDYT?

natebosch:
Moving over to https://dart-review.googlesource.com/c/sdk/+/311924 for now.

PR #4486, opened 2023-06-26
Move prefer_generic_function_type_aliases tests

srawlins:


bwilkerson (approved):
We should probably have a test for a non-function-type, like `typedef Int = int;`, at some point, but not required for this PR.

PR #4489, opened 2023-06-26
Move empty_catches tests

srawlins:


test/rules/empty_catches_test.dart
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  test_2() async {
+    await assertNoDiagnostics(r'''
+void foo() {
+  try {
+    throw new Exception();
+  } catch (e) {
+    // Nothing.
+  } //OK!
```
scheglov, line 48: This comment is probably not very useful in unit tests.
The same below.

scheglov (approved):

test/rules/empty_catches_test.dart
srawlins: Good catch; done.

Issue #4460, opened 2023-06-14
plan for hosting linter documentation post repo-move


pq:
@parlough: would you mind spitballing some of your ideas for post-repo move doc-hosting?

parlough:
In the short term, before we update and introduce split diagnostic and discovery docs, or whatever that will entail, my plan is loosely as follows:

- [ ] Make dart.dev/lints cover everything currently on https://dart-lang.github.io/linter/
  - [x] Add a page which lists all linter rules. (https://github.com/dart-lang/site-www/pull/4990)
  - [ ] Introduce an index similar to [the old one](https://dart-lang.github.io/linter/lints/index.html) which allows for easy discovery. (https://github.com/dart-lang/site-www/pull/4999)
  - [ ] Move linter rule information to their own pages. (https://github.com/dart-lang/site-www/pull/4999)
- [ ] Update the generator on `dart-lang/linter` to redirect to `dart.dev/lints/<rule>`.
  - [ ] Use [`http-equiv`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta#http-equiv) and specify a [canonical URL](https://developers.google.com/search/docs/crawling-indexing/consolidate-duplicate-urls#rel-canonical-link-method).
  - [ ] Add backup text directing to the new locations.
- [ ] Update links around `dart-lang` and `flutter` repos to point to the new dart.dev equivalents.
  - [ ] Update all links within `dart-lang/linter`, specifically the one in the repository description.
  - [x] Update the analyzer to point to `dart.dev/lints/<rule>` in diagnostic messages (https://github.com/dart-lang/sdk/commit/25c0ae1c423e32e9e539232318ca1c32bb70eee6).
  - [ ] Update other miscellaneous links.

parlough:
At least in the short term, the generated document that remains useful for dart.dev, `package:lints`, and perhaps elsewhere is the [`rules.json`](https://github.com/dart-lang/linter/blob/gh-pages/lints/machine/rules.json). Perhaps we can move this in tree and have a test that makes sure it is up to date?

pq:
@parlough: are we basically done here?  Anything else we need to be tracking?

parlough:
I think this issue can be closed :)

PR #4504, opened 2023-06-28
Add redirects and specify canonical URLs pointing to dart.dev

parlough:
This is step #1 in retiring the old linter site. Using [`http-equiv`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta#http-equiv) it redirects from the generated pages to their dart.dev equivalents. It also specifies a [canonical URL](https://developers.google.com/search/docs/crawling-indexing/consolidate-duplicate-urls#rel-canonical-link-method) to clearly signal to search engines that the content has moved.

Eventually we'll be able to cleanup and eventually disable the automatic generation, but this gets the redirects functional immediately. 

I've staged a version of the site generated with these changes. Visit the below links to see the redirects functional:

**Staged rule index page:** https://parlough-linter.web.app/lints/index.html
**Staged rule page:** https://parlough-linter.web.app/lints/prefer_final_fields.html
**Staged all rule list:** https://parlough-linter.web.app/lints/options/options.html

Contributes to https://github.com/dart-lang/linter/issues/4460 and https://github.com/dart-lang/site-www/issues/4499

### DIFF
--- a/tool/doc.dart
+++ b/tool/doc.dart
@@ -311,6 +311,8 @@ class HtmlIndexer {
       <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0">
       <meta name="mobile-web-app-capable" content="yes">
       <meta name="apple-mobile-web-app-capable" content="yes">
+      <meta http-equiv="refresh" content="0; url=https://dart.dev/lints">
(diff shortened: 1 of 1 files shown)


pq (commented):
💎

PR #4513, opened 2023-06-29
Move unnecessary_new tests

srawlins:
Fixes https://github.com/dart-lang/linter/issues/4512


test/rules/unnecessary_new_test.dart
```diff
@@ -0,0 +1,136 @@
...
+''', [
+      lint(39, 7),
+    ]);
+  }
+
+  test_unnamed_newName_const() async {
+    await assertNoDiagnostics(r'''
+class A {
+  const A();
+  A.c1();
```
bwilkerson, line 102: Not sure this line adds any value to the test.

test/rules/unnecessary_new_test.dart
```diff
@@ -0,0 +1,136 @@
...
+  String get lintRule => 'unnecessary_new';
+
+  test_localFinal_new() async {
+    await assertDiagnostics(r'''
+class A {
+  const A();
+}
+
+void f() {
+  final a = new A();
```
bwilkerson, line 27: Not sure whether it's useful to test all the locations in which an instance creation expression can appear, so we might consider dropping `test_localFinal_new` and `test_localFinal_noNew`.

bwilkerson (approved):

test/rules/unnecessary_new_test.dart
srawlins: I'm certain that it's not useful. :/ I'm just trying to mechanically migrate these to be reflective. But I guess I'm happy to _delete_ uninteresting cases.

bwilkerson: If we don't delete them now then I'm not sure we'll come back to delete them later. That said, I'm not asking for perfection, just pointing out the things I see. But you're right, it more closely reflects the original tests this way, so I'm fine leaving it as-is.

test/rules/unnecessary_new_test.dart
srawlins: Removed

PR #4535, opened 2023-07-05
fix `public_member_api_docs` overreporting on effectively private constructors

### LINKED ISSUE #4526: public_member_api_docs: constructors on sealed classes are not really public
_Similar to https://github.com/dart-lang/linter/issues/284 (/cc @yjbanov)._

Here is a more recent (2023-era) problem, constructors that are not _really_ public.

Consider:

```dart
sealed class Fruit {
  Fruit();
}

final class Apple extends Fruit {}
final class Orange extends Fruit {}
final class Banana extends Fruit {}
```

The `Fruit()` constructor _appears_ public (i.e. it does not start with `_`), but in Dart 3+, it is _functionally_ private (that is, it's only accessible in the same library, similar to methods declared with `_`).

pq:
Fixes #4526

/cc @bwilkerson

### DIFF
--- a/lib/src/rules/public_member_api_docs.dart
+++ b/lib/src/rules/public_member_api_docs.dart
@@ -226,11 +226,21 @@ class _Visitor extends SimpleAstVisitor {
 
   @override
   void visitConstructorDeclaration(ConstructorDeclaration node) {
-    if (!inPrivateMember(node) &&
-        !isPrivate(node.name) &&
-        node.parent is! EnumDeclaration) {
-      check(node);
+    if (inPrivateMember(node) || isPrivate(node.name)) return;
+    var parent = node.parent;
+    if (parent is EnumDeclaration) return;
+    if (parent is ClassDeclaration) {
+      var classElement = parent.declaredElement;
+      if (classElement != null) {
+        if (classElement.isSealed) return;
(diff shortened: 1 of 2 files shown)


lib/src/rules/public_member_api_docs.dart
```diff
@@ -226,11 +226,21 @@ class _Visitor extends SimpleAstVisitor {
 
   @override
   void visitConstructorDeclaration(ConstructorDeclaration node) {
-    if (!inPrivateMember(node) &&
-        !isPrivate(node.name) &&
-        node.parent is! EnumDeclaration) {
-      check(node);
+    if (inPrivateMember(node) || isPrivate(node.name)) return;
```
bwilkerson, line 229: The code duplication with an earlier PR concerns me. I'm wondering whether we want to add the lines below to `inPrivateMember` (or add a new utility method to include the overall semantics), maybe making it an extension method in the process.

bwilkerson (approved):

lib/src/rules/public_member_api_docs.dart
pq: Agreed. I've got another PR in the wings for `@internal`. I'll address this duplication there. Thanks!

PR #4537, opened 2023-07-06
Report unnecessary_parenthesis around Literal and PrefixedIdentifier

oprypin:
Fixes a few false ~~positives~~ negatives that I noticed when mass-editing code and gratuitously sprinkling parentheses everywhere

### DIFF
--- a/test_data/rules/unnecessary_parenthesis.dart
+++ b/test_data/rules/unnecessary_parenthesis.dart
@@ -145,7 +145,24 @@ main() async {
   a2 = (1 == 1); // OK
   a2 = (1 == 1) || "".isEmpty; // OK
   var a3 = (1 + 1); // LINT
+
+  // Tests for Literal and PrefixedIdentifier
+  var a4 = (''); // LINT
+  var a5 = ((a4.isEmpty), 2); // LINT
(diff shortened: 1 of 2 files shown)


bwilkerson (approved):

test_data/rules/unnecessary_parenthesis.dart
```diff
@@ -145,8 +145,25 @@ main() async {
   a2 = (1 == 1); // OK
   a2 = (1 == 1) || "".isEmpty; // OK
   var a3 = (1 + 1); // LINT
+
+  // Tests for Literal and PrefixedIdentifier
+  var a4 = (''); // LINT
```
pq, line 150: You mention only false positives but these tests validate (new) reports. Is that right? Is this actually fixing false negatives?

oprypin: Oh sorry, I mixed up the two. I am only fixing false negatives here. I.e. only adding more complaint cases.

oprypin:
Note: this will cause a few new lint violations in google3, but I'll handle those

pq (approved):

pq:
Good deal. Thanks!

PR #4538, opened 2023-07-06
Fix directives_ordering for a leading dot in path

### LINKED ISSUE #2678: directives_ordering inconsistent with other orderings
The sort order does not seem to be consistent with analyzer/tidy_dart when there are packages with `.` in the name; it looks like we need to split the package name out, sort by then, then sort by name within package.

Other suggestions to check:

```
  import "/src/foo.dart";
  import "./foo.dart";
  import "../foo.dart";  
```

oprypin:
It sorts imports as follows:

```dart
import './foo1.dart';
import '../../foo2.dart';
import '../foo3.dart';
import 'foo4.dart';
```

This change corrects it to the following:

```dart
import '../../foo1.dart';
import '../foo2.dart';
import './foo3.dart';
import 'foo4.dart';
```

Fixes #2678

See https://github.com/dart-lang/linter/issues/2678#issuecomment-1610144987 for an explanation on how this edit relates to the behavior

### DIFF
--- a/lib/src/rules/directives_ordering.dart
+++ b/lib/src/rules/directives_ordering.dart
@@ -116,6 +116,9 @@ const _importKeyword = 'import';
 ///
 /// Package is everything until the first `/`.
 int compareDirectives(String a, String b) {
+  if (!a.startsWith('package:') || !b.startsWith('package:')) {
+    return a.compareTo(b);
+  }
   var indexA = a.indexOf('/');
   var indexB = b.indexOf('/');
   if (indexA == -1 || indexB == -1) return a.compareTo(b);
--- a/test/rules/directives_ordering_test.dart
+++ b/test/rules/directives_ordering_test.dart
@@ -274,4 +274,24 @@ import 'a.dart';
       lint(102, 16),
     ]);
   }
+
+  test_sortDirectiveSectionsAlphabetically_dotInRelativePath_import() async {
+    await assertDiagnostics(r'''
+import './foo1.dart';
+import '../../foo2.dart';
+import '../foo3.dart';
+import 'foo4.dart';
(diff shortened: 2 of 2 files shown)


oprypin:
I see. Yes, I realize my validation was not right. It's not that the fixer was doing the right thing now, it's just that it wasn't kicking in anymore.
Then how can we proceed and fix the code in both places?

bwilkerson:
I can only think of two approaches (though there are probably others).

1. First update both the linter and the server to use the same implementation of this method, as per the comment in the organize imports implementation, then modify the shared implementation.

2. Update the server's implementation at the same time we roll a new version of the linter into the SDK. This will only work if we ensure that the internal version of the linter is always the version in the SDK.

oprypin:
Let's just update sdk/ and rev linter at the same time. Was anyone planning a routine bump of the linter, so perhaps then we could advance it only 1 commit forward when applying this particular change?

If there are some edge cases in their synchronization, I think it's not going to cause much damage, because it's very rare that `"./imports"` are used. Also people are not forced to only rely on the fixer, it's possible to stop the lint complaint even if they are out of sync.

oprypin:
OK, linter got bumped to latest! https://dart.googlesource.com/sdk/+/fab30a7a73cd6946fc0cc340a62c10b69740b02c

I think we're in a great position to proceed with this change. Then I'll be able to make an SDK change that will also bump linter to this change.

cc @pq

srawlins (approved):

srawlins:
That plan sounds good to me. Landing.

srawlins:
Sync CL: https://dart-review.googlesource.com/c/sdk/+/313285

oprypin:
Ah CL as well :)
Thanks much!

PR #4560, opened 2023-07-13
use_build_context_synchronously: Fix not-mounted-or case

srawlins:
This issue was reported internally. The motivating test case was:

```dart
  if (!context.mounted || 1 == 2) {
  } else {
    context /* ref */;
  }
```

The only way to get into the else branch is if `context.mounted` is true.


scheglov (approved):

mernen:
I was about to report this issue, as it is a regression in Flutter 3.13. Is it planned for a hotfix release?

srawlins:
No, I don't believe we will release this in a hotfix.

PR #4568, opened 2023-07-16
Support prefer_final_fields in libs with parts

srawlins:
Fixes https://github.com/dart-lang/linter/issues/4562

This involved a sizeable refactoring, with a few improvements:

* Visit parts in order to find mutations of private fields.
* More idiomatic/readable extension getters.
* Store less fields-which-have-been-mutated. The previous impl stored all fields every mutated in a library,  including fields never declared in the library. This new impl stores only local field elements.
* Do not declare local functions in a for-loop; this should allocate many fewer short-lived objects (`_Closure`s).

### DIFF
--- a/lib/src/rules/prefer_final_fields.dart
+++ b/lib/src/rules/prefer_final_fields.dart
@@ -2,8 +2,6 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-import 'dart:collection';
-
 import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/token.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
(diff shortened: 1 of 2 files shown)


lib/src/rules/prefer_final_fields.dart
```diff
@@ -112,16 +99,39 @@ class PreferFinalFields extends LintRule {
...
+      if (element is FieldElement &&
+          element.isPrivate &&
+          !element.overridesField) {
+        fields[element] = variable;
+      }
+    }
   }
 }
 
 class _MutatedFieldsCollector extends RecursiveAstVisitor<void> {
```
bwilkerson, line 128: nit: This isn't collecting mutated fields anymore (which is a good thing), it's finding fields that have been mutated and updating the list. Perhaps the name could be updated to reflect the change. I'm sure you can come up with something better, but I'm thinking of something like `_FieldMutationFinder`.

bwilkerson (approved):

lib/src/rules/prefer_final_fields.dart
srawlins: That sounds great, and doesn't change the alphabetical ordering 😌

bwilkerson: I'll try to keep that goal in mind when I'm renaming other classes. 🙂

PR #4571, opened 2023-07-17
Add docs for core lints

bwilkerson:
# Description

Fixes # (issue)

### DIFF
--- a/messages.yaml
+++ b/messages.yaml
@@ -1,0 +1,1109 @@
+LintCode:
+  avoid_empty_else:
+    documentation: |-
+      #### Description
+
+      The analyzer produces this diagnostic when the statement after an `else`
+      is an empty statement (a semicolon).
+
+      #### Example
+
+      The following code produces this diagnostic because the statement
+      following the `else` is an empty statement:
+
+      ```dart
+      void f() {
+        if (x > y)
+          print("1");
+        else [!;!]
+          print("2");
+      }
+      ```
+
+      #### Common fixes
+
+      If the statement after the empty statement is intended to be executed only
+      when the condition is `false`, then remove the empty statement:
+
+      ```dart
+      void f() {
+        if (x > y)
+          print("1");
+        else
+          print("2");
+      }
+      ```
+
+      If there is no code that is intended to be executed only when the
+      condition is `false`, then remove the whole `else` clause:
+
+      ```dart
+      void f() {
+        if (x > y)
+          print("1");
+        print("2");
+      }
+      ```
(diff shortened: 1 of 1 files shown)


messages.yaml
```diff
@@ -0,0 +1,1109 @@
...
+      }
+      ```
+    justification: |-
+      It's easy to miss the presence of the semicolon, which can make it hard to
+      spot bugs in the code control flow.
+    seeAlso: |-
+      The `empty_statements` lint catches this case and other similar cases.
+
+      The `curly_braces_in_flow_control_structures` lint also prevents this
+      problem by forcing the statement after the `else` to be a block.
```
parlough, line 54: I really like this. If we do break up the categories, perhaps they should also go under documentation.

messages.yaml
```diff
@@ -0,0 +1,1109 @@
...
+      ```
+  avoid_shadowing_type_parameters:
+    documentation: |-
+      #### Description
+
+      The analyzer produces this diagnostic when a type parameter shadows a type
+      parameter from an enclosing declaration.
+
+      Shadowing a type parameter with a different type parameter can lead to
+      subtle bugs that are difficult to debug.
```
parlough, line 90: I was thinking we would likely still need a short one-liner(or similar) for lint discovery, as the names themselves might not be enough. But perhaps the problem message serves as that?

parlough (commented):
I haven't taken a close look at this, but it's really exciting. I just wanted to leave two quick comments.

messages.yaml
bwilkerson: I think that depends on the question of "discovery for what purpose". In other words, I think we need to think about this in terms of CUJ's. If the user it trying to discover why a diagnostic was generated, the problem message would probably work. If the user is trying to find a lint that would catch a specific problem, then the problem message probably isn't going to work very well.

Discoverability is important, so we should plan on adding whatever we need in order to optimize that experience for the user.

parlough: Yeah I'm thinking about the "find a lint that would catch a specific problem" scenario, so an extra top-level field for that would be great. It would loosely correspond to/replace the current description defined at https://github.com/dart-lang/linter/blob/main/lib/src/rules/avoid_empty_else.dart#L10, but perhaps updated to be more consistent. It's also what could be surfaced in the `analysis_options.yaml` on hover.

parlough: I think it's fine if those don't necessarily land in the first iteration of this, as we still have the inlined `desc`s, but I see that replacing those eventually.

PR #4573, opened 2023-07-18
Move unnecessary_raw_strings tests

srawlins:


test/rules/unnecessary_raw_strings_test.dart
```diff
@@ -0,0 +1,99 @@
...
+    defineReflectiveTests(UnnecessaryRawStringsTest);
+  });
+}
+
+@reflectiveTest
+class UnnecessaryRawStringsTest extends LintRuleTest {
+  @override
+  String get lintRule => 'unnecessary_raw_strings';
+
+  test_doubleQuotes_raw() async {
```
bwilkerson, line 20: I'd remove the '_raw' from every test method name because they're all raw strings (unless we think we should add tests of non-raw strings). But completely up to you.

bwilkerson (approved):

test/rules/unnecessary_raw_strings_test.dart
srawlins: Yeah the non-raw case should be tested.

I've added a test case, and left the other names the same. Thanks for the catch!

PR #4581, opened 2023-07-19
remove unused `invariant_booleans` support bits

pq:
These were used by `invariant_booleans` but no longer!

/cc @bwilkerson @srawlins


srawlins (approved):
🎉

PR #4582, opened 2023-07-20
Leave backup text crumbs pointing to new docs

parlough:
This PR will regenerate each page with textual directions pointing and linking to the new documentation. It should likely never be seen due to the redirects created in https://github.com/dart-lang/linter/pull/4504, however it acts as a backup in case a client doesn't support those for some reason. It also removes unnecessary doc generation code that is no longer used after these changes.

After this lands and the pages on `gh-pages` are updated, this generation logic can be _(mostly)_ removed.

Contributes to https://github.com/dart-lang/linter/issues/4460#issuecomment-1595810116

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -85,4 +85,3 @@ linter:
     - use_string_buffers
     - use_super_parameters
     - use_to_and_as_if_applicable
-    - unreachable_from_main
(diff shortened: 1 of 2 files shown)


pq (approved):
💎

pq:
@parlough: is this ready to land?

parlough:
@pq Yep, this should be good to land!

PR #4585, opened 2023-07-20
Remove most remaining doc generation code

parlough:
This removes most remaining doc generation code, outside of the machine generated file (`rules.json`) which is still used by `package:lints`, dart.dev, a flutter sample, and perhaps elsewhere. 

In the short term, we should perhaps check that file in to the `main` branch instead and have a test to make sure its up to date. Once that is done, we can remove the `gh-pages` workflow.

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -85,4 +85,3 @@ linter:
     - use_string_buffers
     - use_super_parameters
     - use_to_and_as_if_applicable
-    - unreachable_from_main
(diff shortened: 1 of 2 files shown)


pq:
> In the short term, we should perhaps check that file in to the main branch instead and have a test to make sure its up to date. 

I can take care of that! If you have a sec, could you open a tracking issue?

parlough:
That would be appreciated! https://github.com/dart-lang/linter/issues/4587

pq (approved):

PR #4586, opened 2023-07-20
Enable `unreachable_from_main` linter rule

parlough:
It has been quite useful when removing old/stale functionality, especially in `/tool`.

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -79,10 +79,10 @@ linter:
     - unnecessary_null_checks
     - unnecessary_parenthesis
     - unnecessary_statements
+    - unreachable_from_main
     - use_if_null_to_convert_nulls_to_bools
     - use_late_for_private_fields_and_variables
     - use_setters_to_change_properties
     - use_string_buffers
     - use_super_parameters
     - use_to_and_as_if_applicable
-    - unreachable_from_main
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -85,3 +85,4 @@ linter:
     - use_string_buffers
     - use_super_parameters
     - use_to_and_as_if_applicable
+    - unreachable_from_main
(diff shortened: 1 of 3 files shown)


test/rule_test.dart
```diff
@@ -180,15 +180,15 @@ void defineRuleUnitTests() {
...
-void defineSoloRuleTest(String ruleToTest) {
-  for (var entry in Directory(ruleTestDataDir).listSync()) {
-    if (entry is! File || !isDartFile(entry)) continue;
-    var ruleName = p.basenameWithoutExtension(entry.path);
-    if (ruleName == ruleToTest) {
-      testRule(ruleName, entry);
-    }
-  }
-}
+// void defineSoloRuleTest(String ruleToTest) {
```
pq, line 183: Let's just delete this altogether? I think `rule_debug.dart` has superseded this approach.

pq (changes requested):

test/rule_test.dart
pq: Thanks!

pq (approved):
👍

analysis_options.yaml
```diff
@@ -85,3 +85,4 @@ linter:
     - use_string_buffers
     - use_super_parameters
     - use_to_and_as_if_applicable
+    - unreachable_from_main
```
pq, line 88: Whoops! Could you sort this up? (It should go before `use_if_null_to_convert_nulls_to_bools` I think?)

pq (changes requested):

analysis_options.yaml
parlough: Oops. You're super right, thanks!

pq (approved):
Thank you!

test/rule_test.dart
```diff
@@ -179,17 +179,6 @@ void defineRuleUnitTests() {
   });
 }
 
-/// Handy for debugging.
```
srawlins, line 182: CC @pq, we might want to `// ignore` this case.

srawlins: Oh, lol

PR #4590, opened 2023-07-20
write and verify `rules.json`

### LINKED ISSUE #4587: Preserve machine-generated linter rules JSON file
The machine-generated `rules.json` file found on the `gh-pages` branch at https://github.com/dart-lang/linter/blob/gh-pages/lints/machine/rules.json is still used by a few clients. 

For the purposes of moving to the SDK, it should potentially be checked in to the `main` branch with a test to make sure it's up to date.

\cc @pq

pq:
Fixes #4587

@parlough: let me know if this is what you had in mind?

/cc @bwilkerson @srawlins

### DIFF
--- a/test/all.dart
+++ b/test/all.dart
@@ -23,6 +23,7 @@ import 'validate_rule_description_format_test.dart'
     as validate_rule_description_format;
 import 'validate_sdk_version_map.dart' as validate_sdk_version_map;
 import 'verify_checks_test.dart' as verify_checks;
+import 'verify_machine_json_test.dart' as verify_machine_json;
 import 'verify_reflective_test_suites.dart' as verify_reflective_test_suites;
 
 void main() {
@@ -45,5 +46,6 @@ void main() {
   validate_rule_description_format.main();
   validate_sdk_version_map.main();
   verify_checks.main();
+  verify_machine_json.main();
   verify_reflective_test_suites.main();
 }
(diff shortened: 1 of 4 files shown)


parlough:
I'm generally happy with this, but it may not be what you want, since it will break if one of the `lints` packages modifies their list of lints or the error fix status map changes. Perhaps we want to adjust those to the latest stable tag/commit rather than `main`?

pq:
Great points!

My gut is that in practice we won't get bitten here too often. My thinking being:

1. `lints` and `flutter_lints` updates happen very infrequently, and while
2. `error_fix_status.yaml` might change more frequently, it will become less of a hassle when we're co-located in the same repo and can make atomic changes

I guess I'm inclined to take the "simple" approach first and rethink if it becomes a problem down the road.

How does that sound?

parlough:
Sounds good to me :)

pq:
> Another benefit of that is it more closely matches what most users will experience.

Regarding this, won't the version of `rules.json` users generally interact w/ still be on `dart.dev` somewhere? And correspond to the latest stable release?

parlough:
> Regarding this, won't the version of rules.json users generally interact w/ still be on dart.dev somewhere? And correspond to the latest stable release?

It's kind of in an awkward spot for now. Since the analyzer still points to the the linter rule documentation rather than the diagnostic version for lints, I try to keep it up to date for `main`. That way new lints don't have a dead link for users on `beta` and `dev`. So for now I have just been manually updating the fix status to match stable.

bwilkerson (approved):

PR #4620, opened 2023-07-26
Move noop_primitive_operations tests

srawlins:


test/rules/noop_primitive_operations_test.dart
```diff
@@ -0,0 +1,175 @@
...
+}
+''', [
+      lint(21, 8),
+    ]);
+  }
+
+  test_print_stringLiteral() async {
+    await assertNoDiagnostics(r'''
+onPrint() {
+  print(''); // OK
```
albertms10, line 123: May it be:

```suggestion
  print('');
```

srawlins: Thanks!!

pq (approved):

Issue #4643, opened 2023-08-04
remove lint checking workflow


pq:
@parlough, I'm guessing you do something similar on dart.dev and this can safely just be removed?

https://github.com/dart-lang/linter/blob/main/.github/workflows/linkcheck.yml

See: https://github.com/dart-lang/linter/issues/4411

parlough:
Nice catch, I forgot to remove this workflow. Opened https://github.com/dart-lang/linter/pull/4644 to remove it.

PR #4644, opened 2023-08-04
Remove now unnecessary linkchecker

parlough:
The gh-pages link specified here no longer is updated, but I forgot to remove this workflow along with that change. We also check links in the generated docs on site-www anyway.

Closes https://github.com/dart-lang/linter/issues/4643


pq:
Cool. Thanks!

PR #4653, opened 2023-08-09
Enable more lints used in sdk analysis packages

parlough:
To ease transition to the SDK and for consistency, adds a few linter rules enabled on the [analyzer](https://github.com/dart-lang/sdk/blob/main/pkg/analyzer/analysis_options.yaml) and the [analysis_server](https://github.com/dart-lang/sdk/blob/main/pkg/analysis_server/analysis_options.yaml).

The only lint that triggered is `avoid_dynamic_calls`, so this PR fixes those.

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -28,7 +28,6 @@ linter:
     - collection_methods_unrelated_type
     - comment_references
     - dangling_library_doc_comments
-    - depend_on_referenced_packages
     - directives_ordering
     - discarded_futures
     - implicit_call_tearoffs
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -13,6 +13,7 @@ linter:
     - avoid_annotating_with_dynamic
     - avoid_catches_without_on_clauses
     - avoid_catching_errors
+    - avoid_dynamic_calls
     - avoid_positional_boolean_parameters
     - avoid_print
     - avoid_redundant_argument_values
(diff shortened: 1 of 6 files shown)


analysis_options.yaml
```diff
@@ -21,13 +22,18 @@ linter:
     - avoid_returning_this
     - avoid_setters_without_getters
     - avoid_slow_async_io
+    - avoid_unused_constructor_parameters
     - cancel_subscriptions
     - cast_nullable_to_non_nullable
+    - collection_methods_unrelated_type
     - comment_references
     - dangling_library_doc_comments
+    - depend_on_referenced_packages
```
pq, line 31: This one is already included in core so can be safely removed I think.

pq (commented):
What's the analysis you're aligning this with? (Not that I object necessarily.)

Is there a shared set we can include that's a superset of `recommended` that brings these additional analyses in?

parlough:
> What's the analysis you're aligning this with? (Not that I object necessarily.)
> 
> Is there a shared set we can include that's a superset of `recommended` that brings these additional analyses in?

I was just basing this off of the [analyzer](https://github.com/dart-lang/sdk/blob/main/pkg/analyzer/analysis_options.yaml) and the [analysis_server](https://github.com/dart-lang/sdk/blob/main/pkg/analysis_server/analysis_options.yaml).

However, I'd be more than happy to switch to [`dart_flutter_team_lints`](https://github.com/dart-lang/ecosystem/blob/main/pkgs/dart_flutter_team_lints/lib/analysis_options.yaml) which includes many of these. It's available in the DEPS file now: https://github.com/dart-lang/sdk/blob/main/DEPS#L142

analysis_options.yaml
parlough: Nice catch.

pq:
> I was just basing this off of the [analyzer](https://github.com/dart-lang/sdk/blob/main/pkg/analyzer/analysis_options.yaml) and the [analysis_server](https://github.com/dart-lang/sdk/blob/main/pkg/analysis_server/analysis_options.yaml).

OK, this is great. This seems reasonable as we prep for the imminent move to the SDK.  I *do* very much think we should have a shared set of options there and will (re)-kick off that conversation.

Until then let's just land these in-line.

Thanks!

pq (approved):
Thanks!

PR #4660, opened 2023-08-10
extension type support for `library_private_types_in_public_api`

### LINKED ISSUE #4659: `library_private_types_in_public_api` support for extension types

pq:
Fixes #4659 

/cc @bwilkerson

### DIFF
--- a/lib/src/rules/library_private_types_in_public_api.dart
+++ b/lib/src/rules/library_private_types_in_public_api.dart
@@ -81,14 +81,6 @@ class Validator extends SimpleAstVisitor<void> {
   }
 
   @override
-  void visitExtensionTypeDeclaration(ExtensionTypeDeclaration node) {
-    if (Identifier.isPrivateName(node.name.lexeme)) return;
-    node.typeParameters?.accept(this);
-    node.representation.fieldType.accept(this);
-    node.members.accept(this);
-  }
-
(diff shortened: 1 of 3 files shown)


lib/src/rules/library_private_types_in_public_api.dart
```diff
@@ -80,6 +80,14 @@ class Validator extends SimpleAstVisitor<void> {
     node.members.accept(this);
   }
 
+  @override
+  void visitExtensionTypeDeclaration(ExtensionTypeDeclaration node) {
+    if (Identifier.isPrivateName(node.name.lexeme)) return;
+    node.typeParameters?.accept(this);
+    node.representation.fieldType.accept(this);
```
bwilkerson, line 87: I think the representation type only needs to be checked in cases where the name of the representation field is public. That is, I don't think there should be a lint for
```
extension type E(_C _c) {}
class _C {}
```
because the extension field isn't part of the public API in that case.

bwilkerson (approved):
LGTM with the one change made (assuming I'm right about the semantics).

lib/src/rules/library_private_types_in_public_api.dart
pq: Good catch!

PR #4671, opened 2023-08-11
extension type support for `sort_unnamed_constructors_first`

### LINKED ISSUE #4670: `sort_unnamed_constructors_first` support for extension types

pq:
Fixes #4670

/cc @bwilkerson

### DIFF
--- a/test/rules/sort_unnamed_constructors_first_test.dart
+++ b/test/rules/sort_unnamed_constructors_first_test.dart
@@ -15,6 +15,9 @@ main() {
 @reflectiveTest
 class SortUnnamedConstructorsFirstTest extends LintRuleTest {
   @override
+  List<String> get experiments => ['inline-class'];
+
+  @override
   String get lintRule => 'sort_unnamed_constructors_first';
 
   test_class_sorted() async {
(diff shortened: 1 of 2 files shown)


test/rules/sort_unnamed_constructors_first_test.dart
```diff
@@ -62,4 +65,16 @@ enum A {
...
+
+  test_extensionType() async {
+    await assertDiagnostics(r'''
+extension type E(Object o) {
+  void m() { }
+  E(this.o);
+}
+''', [
+      // No lint.
+      // todo(pq): Add `duplicate_constructor` diagnostic when it is reported.
```
bwilkerson, line 77: The lack of an expected lint could either indicate that the lint correctly handles this case, or it could mean that we just aren't visiting constructors in extension types. I think the test needs to produce a lint diagnostic in order to be useful.

Also, rather than add a non-lint diagnostic in the future, you could give the primary constructor a name in order to avoid the diagnostic:
```
extension type E.a(Object o) {
  void m() { }
  E.b(this.o);
  E(this.o);
}
```

pq: Here I'm trying to assert that we do not produce a false positive. The good news is we do not today. The value of the test is to ensure we never do.

But the idea of this

> you could give the primary constructor a name

is bending my brain a bit. I wasn't aware that you could do this!

pq: OK. I've updated it to handle the case where the representation constructor get's a name.

Thanks!

bwilkerson (approved):

PR #4673, opened 2023-08-11
extension type support for `hash_and_equals`

### LINKED ISSUE #4672: `hash_and_equals` support for extension types

pq:
Fixes #4672

/cc @bwilkerson

### DIFF
--- a/test/rules/hash_and_equals_test.dart
+++ b/test/rules/hash_and_equals_test.dart
@@ -15,6 +15,9 @@ main() {
 @reflectiveTest
 class HashAndEqualsTest extends LintRuleTest {
   @override
+  List<String> get experiments => ['inline-class'];
+
+  @override
   String get lintRule => 'hash_and_equals';
 
   test_enum_missingHash() async {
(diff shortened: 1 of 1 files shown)


test/rules/hash_and_equals_test.dart
```diff
@@ -30,4 +33,15 @@ enum A {
...
     ]);
   }
+
+  test_extensionType_missingHash() async {
+    await assertDiagnostics(r'''
+extension type E(Object o) {
+  bool operator ==(Object other) => false;
+}
+''', [
+      // No lint.
```
bwilkerson, line 43: I'm similarly unconvinced that this test is particularly useful, though I guess I don't object too much here because there's nothing else to test.

bwilkerson (approved):

test/rules/hash_and_equals_test.dart
pq: The goal here is to ensure we don't report a lint where there's already an error (since you can't redeclare `hash` or `==` in extension types).

PR #4677, opened 2023-08-11
extension type support for `unnecessary_constructor_name`

### LINKED ISSUE #4676: `unnecessary_constructor_name` support for extension types

pq:
Fixes #4676

/cc @bwilkerson

### DIFF
--- a/lib/src/rules/unnecessary_constructor_name.dart
+++ b/lib/src/rules/unnecessary_constructor_name.dart
@@ -59,7 +59,7 @@ class UnnecessaryConstructorName extends LintRule {
       NodeLintRegistry registry, LinterContext context) {
     var visitor = _Visitor(this);
     registry.addConstructorDeclaration(this, visitor);
-    registry.addExtensionTypeDeclaration(this, visitor);
+    registry.addRepresentationConstructorName(this, visitor);
     registry.addInstanceCreationExpression(this, visitor);
   }
 }
@@ -81,13 +81,13 @@ class _Visitor extends SimpleAstVisitor {
   }
 
   @override
-  void visitExtensionTypeDeclaration(ExtensionTypeDeclaration node) {
-    _check(node.representation.constructorName?.name);
+  void visitInstanceCreationExpression(InstanceCreationExpression node) {
+    _check(node.constructorName.name?.token);
   }
 
   @override
-  void visitInstanceCreationExpression(InstanceCreationExpression node) {
-    _check(node.constructorName.name?.token);
+  void visitRepresentationConstructorName(RepresentationConstructorName node) {
+    _check(node.name);
   }
 
   void _check(Token? name) {
--- a/lib/src/rules/unnecessary_constructor_name.dart
+++ b/lib/src/rules/unnecessary_constructor_name.dart
@@ -70,6 +70,7 @@ class _Visitor extends SimpleAstVisitor {
 
   @override
   void visitConstructorDeclaration(ConstructorDeclaration node) {
+    if (node.parent is ExtensionTypeDeclaration) return;
     _check(node.name);
   }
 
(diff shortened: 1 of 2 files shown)


lib/src/rules/unnecessary_constructor_name.dart
```diff
@@ -70,6 +70,7 @@ class _Visitor extends SimpleAstVisitor {
 
   @override
   void visitConstructorDeclaration(ConstructorDeclaration node) {
+    if (node.parent is ExtensionTypeDeclaration) return;
```
bwilkerson, line 73: I don't understand why we wouldn't check constructors in extension types.

If anything, I think this lint needs to check the primary constructor (which it currently probably doesn't do).

pq: You can't redefine the unnamed constructor in an extension type so my thinking is this lint should never fire.

pq: (OK. I can see I need to update this in case the primary constructor is named. Updating.)

bwilkerson: But you can use `.new` in a constructor declaration (whether it's the primary constructor or not), and I think this lint ought to flag that given that it's flagging the same use in constructors in classes. Examples:

```
extension type E.new(int i) {}
```

and

```
extension type E.name(int i) {
  E.new(this.i);
}
```

pq: Thanks for your help.  I think I've got it right now!

lib/src/rules/unnecessary_constructor_name.dart
```diff
@@ -70,7 +70,12 @@ class _Visitor extends SimpleAstVisitor {
 
   @override
   void visitConstructorDeclaration(ConstructorDeclaration node) {
-    if (node.parent is ExtensionTypeDeclaration) return;
+    var parent = node.parent;
+    if (parent is ExtensionTypeDeclaration &&
+        parent.representation.constructorName == null) {
```
bwilkerson, line 76: I still don't understand why this condition exists. If I have
```
class C {
  C();
  C.new(int x);
}
```

Then I believe we'll still flag the `new` in the second constructor. (In other words, I don't think this lint checks to ensure that this is the first and only unnamed constructor.) I don't know why we'd do that for extension types if we're not doing that for classes.

Also, I think we need to `visitRepresentationDeclaration` in order to catch the use of `.new` there.

bwilkerson: You're effectively checking the primary constructor in `visitExtensionTypeDeclaration`, though using `visitRepresentationDeclaration` might be slightly more direct and more general.

pq: > You're effectively checking the primary constructor in visitExtensionTypeDeclaration, though using visitRepresentationDeclaration might be slightly more direct and more general.

You're right!  Better still might even be `visitRepresentationConstructorName` -- which I hadn't noticed until just now.

> I don't know why we'd do that for extension types if we're not doing that for classes.

No good reason except that I'm paying attention now. I overlooked the class case and think it's a false positive that should be fixed (for the same reasons).

pq: In the end, if you think we shouldn't bother suppressing the lint in case a compilation error is being reported, I'm happy to update. Assuming that is the desire behavior, I added a failing test for the class case so we can swing back and fix it later.

bwilkerson: As long as it gets fixed everywhere I'm fine with not double reporting.

But doing so might require visiting every kind of declaration that can have constructors so that we can set a flag indicating whether we've already seen an unnamed constructor, so that might change the way we want to handle the primary constructor.

pq: Yeah. That was my thinking too. I didn't want to tackle such a rewrite here so I left a test as a breadcrumb.  Happy to rollback the logic for extension types if you'd rather keep the support symmetrical.

bwilkerson: No, as long as we're planning on moving in that direction I'm fine with it being in-progress.

bwilkerson (approved):

PR #4678, opened 2023-08-11
Account for missing lints in changelog

parlough:
Some of this is replicated from the `1.35.0` section, but there were linter rules added after `1.35.0` but before the linter commit included in Dart 3, which were missing from the changelog completely: `matching_super_parameters` and `no_literal_bool_comparisons`.

To clearly indicate to users all of the lints that were added and removed, I added an entry for 3.0.0 including both those already included and missing previously. This has the added benefit of beginning the new style with a full major version :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,21 @@
 - new lint: `no_self_assignments`
 - new lint: `no_wildcard_variable_uses`
 
+# 3.0.0
+
+- new lint: `implicit_reopen`
+- new lint: `unnecessary_breaks`
+- new lint: `type_literal_in_constant_pattern`
+- new lint: `invalid_case_patterns`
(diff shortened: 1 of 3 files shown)


pq (approved):
Excellent. Thank you!

mockturtl:
@parlough @pq Thanks!  Now that Dart SDK 3.1.0 has shipped, can I also assume `linter` version 3.1.0 is stable / released?

parlough:
Yep! [`no_self_assignments`](https://dart.dev/lints/no_wildcard_variable_uses) and [`no_self_assignments`](https://dart.dev/lints/no_wildcard_variable_uses) should both be available in Dart 3.1.0 :) 

I just haven't updated the version here yet since the linter is currently being moved to the `dart-lang/sdk` repository.

Issue #4411, opened 2023-06-01
linter repo move readiness // planning


pq:
🚧  _(This is a living document and a work in progress. Watch this space for emerging details. And do provide feedback!)_ 🚧 

## Background

**Costs.** There are costs to having the linter source living externally and separate from the analyzer. In practice, the two are quite tightly coupled and not being able to make changes atomically across them (when an analyzer API changes or a new lint is added for example) creates extra work (e.g., otherwise needless analyzer package publications) and bookkeeping (e.g., TODOs or issues to track the implementation of quick fixes for new lints that cannot be landed until the new lint is available in the SDK). Lints are also very difficult to evolve as any downstream impact from changed semantics is only detected when a linter roll into the Dart SDK's `DEPS` is evaluated. Having the linter workflow external and unfamiliar to Dart team members has also (arguably) limited Dart team contributions.

**Benefits.** There are benefits too. Particularly in the early days of this project we've greatly benefited from the relative nimbleness of a workflow that favors external contributions, allows easy CI customization and enjoys the relative focus of a dedicated issue tracker. 

**Today.** When the linter was growing rapidly and was a tool that was effectively adjacent to the core Dart SDK offerings, the benefits (IMO clearly) outweighed the costs. The landscape today is different though. The linter is effectively now _itself_ a core offering, with lints a central part of the developer experience and various subsets endorsed by the Dart and Flutter teams and promoted as part of standard best practice (see [package:lints](https://github.com/dart-lang/lints)). 

**Making a move.** In light of this, we'd like to move the linter sources (and issue tracking) into the Dart SDK.

## Plan

_(WIP planning notes. Details to be filled in.)_

**Readiness.** There are a bunch of steps to make the linter ready for a move.

### Sources

**Build.** To make the source buildable in `sdk/pkg`, we'll need to:

- [x] #4412

To not lose the build check automation we enjoy from GitHub actions, we'll want to:

- [x] #4416

**Testing.** To make the source testable in the SDK build, we'll need to:

- [x] https://github.com/dart-lang/linter/issues/4404

_(Nice to have)_ To speed up test runes we might:

- [ ] https://github.com/dart-lang/linter/issues/3535

**Documentation.** We currently auto-generate lint docs here and host them on GitHub pages. We need:

- [x] #4460

We'll also need to:

- [x] update contributing guidelines and docs to integrate with an SDK development workflow

**Process.** We should tidy up stale process bits.

Notably, as we're no longer publishing the package, we should:

[...]

PR #4708, opened 2023-08-23
Call out move in README

parlough:
Contributes to https://github.com/dart-lang/linter/issues/4411
Closes https://github.com/dart-lang/linter/issues/4707 

Redirects contributions to the appropriate repository.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,11 +6,12 @@ The Dart linter is now developed within the [Dart SDK][] repository,
 and can be found at https://github.com/dart-lang/sdk/tree/main/pkg/linter.
 
 For now, please continue to file new linter issues on
-this repository. 
+[this repository][]. 
 To contribute changes, check out the [SDK contribution guide][].
 
 [Repository has moved]: https://github.com/dart-lang/sdk/tree/main/pkg/linter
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,5 +1,20 @@
 # Linter for Dart
 
+## [Repository has moved][]
+
+The Dart linter is now developed within the [Dart SDK][] repository,
+and can be found at https://github.com/dart-lang/sdk/tree/main/pkg/linter.
+
+Please open new linter issues on the [dart-lang/sdk repo][issues].
```
pq, line 8: Actually, let's hold off on the issue redirection for now?  We're not quite ready to migrate the issues and I think having to look in two places will be confusing.

Maybe we can call that out too? (That is, "File issues here for now..." or something similar.) WDYT?

parlough: That's fine for me! 

Either way has a chance of issues placed on the other temporarily. Just consider adding a note to the issue moving _tracking issue_ to update this text when that is complete.

pq: This is great. Thanks!

pq (approved):

PR #4838, opened 2024-01-02
Fix PR labeler

mosuem:
See https://github.com/dart-lang/native/pull/873.


parlough:
Since this repo is [mostly deprecated](https://github.com/dart-lang/linter?tab=readme-ov-file#repository-has-moved), perhaps this config and action can just be removed?

pq:
Yeah. I think @parlough is right. There really shouldn't be many PRs against this repo moving forward. Removing the action sounds right.

pq:
(PS: thanks for the fix @mosuem!)

PR #4894, opened 2024-02-26
typo fixes: its vs. it's

elitree:
# Description

Fixes a few instances where the text should be "its" instead of "it's" or vice versa.


srawlins:
So sorry about this, but this repo is read-only at this point (need to figure out how to not allow a new PR); sorry for the confusion! The linter is moved to https://github.com/dart-lang/sdk/tree/main/pkg/linter and we'd be happy to accept a PR there 😁 . Thanks and sorry again!

PR #4964, opened 2024-05-06
Make descriptions more consistent in using code-quotes.

lrhn:
Use markdown code quotes for keywords and types more consistently.
Rephrase some descriptions as drive-by-fixups.

Removed (deprecated-member-using) case for `==` in extensions, since extensions cannot declare `==` operators.


srawlins:
I love it. I think this is all correct. But the linter source is maintained in the Dart SDK repo now. :/ I think we're not sure how to turn off PRs for this repo.

lrhn:
That was the conclusion I came to too, after failing to make tests run.
So I made: https://dart-review.googlesource.com/c/sdk/+/365460

